PR #256, opened 2014-09-29
New faction system

noudess:
These changes are due to what I believe to be flaws in the existing faction system.

The basic explanation of my proposed system is:
- each char can earn/lose personal faction with each faction witha  range of +1200 to -3000
- racial/class deity adjustments apply on top of that for a total.

From a few discussions I've had online and drawing from personal experience I believe this is much closer to live that what we have today.

The current scheme's personal faction is limited by an over-arching +1500/-1500 range.  That range, when combined with the racial/class/deity bonuses doesn't yield an effective system.    There are also bugs in that range currently.

I started two posts in the community to discuss this.

One in http://www.eqemulator.org/forums/showthread.php?t=38685 which doesn't say much more than this thread and one over at PEQ http://www.peqtgc.com/phpBB3/viewtopic.php?f=1&t=15006 which got some input and discussion.

Read the PEQ one for more details on why I am asking for these changes.

I understand that changing the faction system might create angst to existing servers.  I'd be happy to resubmit changes that onoly use this system if a new Rule is set. 

I do, however feel, that the new system is a step in the right direction.


sebshaik:
I bottomed out my faction Claws of Veeshan. I was getting the, "Your faction standing with Claws of Veeshan could not possibly get any worse."

Now, upon killing Velious giants (which give positive faction hits), I still get the same message.

Unsure if I'm in some incredibly deep hole, where I'll get these messages until my faction goes back to -3000 (which is a bug), if this is just a display bug (and I'm getting the correct +5 faction points or whatever), or if there is something else going on. If Magelo were working, I'd be able to give more information, but it's not so I can't. I'm on the Underfoot client. Same issue on all my characters.

Please fix factions.

noudess:
I will take a look at this. You could be well below 3000 based on race, etc penalties, but positive hits should bring you above 3000 in your personal faction value.
On Jan 5, 2015 1:30 AM, sebshaik <notifications@github.com> wrote:I bottomed out my faction Claws of Veeshan. I was getting the, "Your faction standing with Claws of Veeshan could not possibly get any worse."

Now, upon killing Velious giants (which give positive faction hits), I still get the same message.

Unsure if I'm in some incredibly deep hole, where I'll get these messages until my faction goes back to -3000 (which is a bug), if this is just a display bug (and I'm getting the correct +5 faction points or whatever), or if there is something else going on. If Magelo were working, I'd be able to give more information, but it's not so I can't. I'm on the Underfoot client. Same issue on all my characters.

Please fix factions.

sebshaik:
I have a theory that some of the issues I've had stem from the base faction modifiers. Particularly in Velious (though I suspect Kunark and Classic, also), there are modifiers to your base faction - Necromancers and Paladins are both neutralish to Coldain, by default, but Necromancers will start off lower (perhaps dubious) while Paladins will start higher (perhaps apprehensive or even indifferent). Despite their default faction being different, both classes are able to reach "Max Ally" (after all, the various armor quests required Ally faction, as did the turnins of the various heads of each faction (e.g., Yelinak's head) and you'll never find outcries of folks who looted these but were never able to turn them in despite any amount of factioning they ever did).

Necromancers might /con lower to Coldain at level 1, by default. However, given the proper amount of work, they ought to be able to reach max ally, same as Paladins. Currently, no class is able to reach max Ally with Coldain (or Claws of Veeshan, etc. etc.) and I believe this is due to a default penalty (which is, in some sense, correct - as most classes/races/deities didn't start out indifferent) which should apply at the start, for a given starting faction level, but should not impact the ceiling.

Other times it may be the case that the default penalty is the same as the penalty to the "ceiling" that a character can attain - for example, as an Iksar no matter how many orcs I kill on Live, the Wood Elves will never think of me amiably. I'm not sure my default faction is -2000 points (from indifferent) and my max is -800 points (again, -2000 points from the maximum), but this relationship may be true sometimes.

The system needs a way to say, "Hey, you Iksar Necromancer, these Coldain don't really like you, but only here at the start before you prove yourself," with say a -500 penalty. However, that Iksar Necromancer can still reach +1200 at the end. In Velious, there should be no race/class/deity modifiers that apply beyond a character's specific starting point with a faction. So an Iksar Necromancer of Cazic and a Halfling Ranger of Tunare might each start at, say, -500 and -200 respectively, but that only gives the ranger a head start to the eventually end-point of +1200. The old system we had in place treated all factions this way, which was improper. But certainly a lot of them, particularly in the post-Kunark expansions, act in this fashion. Velious most definitely does.

joligario:
Looking back to the old system and reading your comments, I think the base mod should have actually been just an initial faction "hit" instead of a permanent mod. That way the player could effectively still reach ally, but they have to dig that extra 1000, or whatever amount, out of the hole. I played a Necromancer on live and I did a lot of faction work to max out Coldain and CoV at different stages in the game.

noudess:
The old system did what you're saying, to a certain degree.  The thing is, that "initial hit" affected how low you could go.  If the hit was -600, then the most you could drop was (old-min - 600).  Essentially, the maximum the most hated races could lose was less than the races that were liked.  So, a hated race with an illusion was better off than a liked race without, with the same amount of negative faction hits, because a race illusion could fix the hated race, but could not fix the other.  I gave examples of that.

I believe every character needs a fixed range of personal "action based" faction, and separate racial, class and deity adjustments, which is how it is implemented.  If the width of that range needs to change for certain factions or expansions, then so be it.  It sounds like the messages for being at the limits might need fixing as well.

sebshaik:
I think it's an era thing. Back in Classic, you had a totally different group calling the shots. There was a real ceiling - no matter how many orcs I would kill as an Iksar, the wood elves would never like me. They might tolerate me, they might even sell to me (sometimes the cap allowed for trading, sometimes not), but that was as best I could hope for. Kunark largely held this intact, though it was certainly possible for non-Iksars to achieve high factions with the ones that mattered (e.g., the Greenmist is completable by non-Iksar races, as are the other Cabilis "epics" and I have my doubts this is possible on EQEmu right now).

In later expansions, faction was largely abandoned and/or everyone got the same regardless. The D&D guys got squeezed out and the idea of race-based decisions mattering fell by the wayside as they pushed out big expansions every 4-6 months. There may have been an initial starting "hit" that joligario mentioned, but after Velious that hit was largely the same for all players.

noudess:
sebshaik said:

> I bottomed out my faction Claws of Veeshan. I was getting the, "Your faction standing with Claws of Veeshan could not possibly get any worse."
>
>Now, upon killing Velious giants (which give positive faction hits), I still get the same message.

Yes.  I found this problem.  The code for detecting bottom and top is looking at the total modified faction not the personal faction.  This is a bug in what I wrote.  So lets say your class mod is -100.  The code starts saying you bottomed out at -2900 not -3000.  So you can and do actually lose another 100 even though the message says you are maxed if you take more negative hits.  Then when you start getting positive hits, your at -3100 (with mods) and the code again makes mistakes for 100 positive hits. 

I will fix this and put in a pull request

noudess:
I also see that SetFactionLevel2() does not do any of the checks.  And, the mods have no bearing on the functionality.  That code was wrong as well.  Processing...

noudess:
Lol - faction from quests was not even fixed to use the new system.  Fixing that too.

sebshaik:
Glad to hear you've had some time and identified some of the issues. When this gets pushed to PEQ, I'll do some basic testing.

noudess:
Sounds great.  Sorry it took so long, I got pulled onto a real work contract.  I'm off again now.  That procedure should tell yu exactly where characters (peqdb) stand faction wise, as long as you have at least 1 hit on that faction.

sebshaik:
We also finally got magelo back up today, so that should assist in my data collections. Once I see something on the changelog I'll be sure to poke around.

noudess:
Found a problem which may well be what you were talking about.  Testing the fix now.  Unfortunately, the bug may have hurt existing characters by adjusting their faction up/down incorrectly.  I don't see a way to automatically repair that damage.

joligario:
Server admins should have their players report anomalies and have them manually adjust.

mackal:
If the bug let's them min/max their factions again or w/e the issue is, I'm sure the players won't mind. They'll still curse someone for about a week or two then completely forget about it.

sebshaik:
Existing characters have already had their faction change twice. Basically what mackal said.

noudess:
There was also a longer term bug (back before my code) with SetCharacterFactionLevel. If current_value goes down to 0, the line got deleted from the db but the in memory copy was not updated.  Fixing that to use UPDATE and then only update in memory if that succeeds.  It will mean a zero entry to chars that previously had a value, but I think thats ok.

noudess:
Another interesting thing.  One faction in the db has a starting value of -30000.  If someone wanted to max that, they'd need to earn to 31200.  That barely fits the db field for value which is a 16 bit value.  If we want to handle bigger ranges, wed need to mod that db type.

noudess:
New pull request for fixes in place.

PR #298, opened 2014-11-20
removed unused moremath.cpp/h

addtheice:


KimLS:
Does not auto merge.
Includes spaces in lieu of tabs at points.

PR #331, opened 2015-01-08
Work-around for "strange ADL issues" not needed if using Boost 1.57.0

af4t:
Work-around for "strange ADL issues" not needed if using Boost 1.57.0 (was required through 1.56.0, thus test for less than 1.57.0).  Tested on VS2013 x64 and in-game appears to be working correctly; someone might want to check this on other compilers and/or OS.


mackal:
Just to note, seems to be same patch as here https://github.com/DennisOSRM/luabind/pull/3

PR #336, opened 2015-01-13
Clean up of start zone overrides

RicardoCampos:
Altered the way that start zone overrides work. Previously I think a few bugs crept in as the code changed over time, meaning that it ignored some of the settings provided.

They now no longer use an entry in the variables table, but rather use TitaniumStartZoneID, and SofStartZoneID in the rule_values table.

If this is set, it will look for a match in the start_zones based on race/class/deity and the zone id. If it finds a matching row (by zone_id), it will use that row's details. If not, it will use a default.

If accepted, I'll update the wiki with a howto.


world/worlddb.cpp
```diff
@@ -429,9 +299,10 @@ bool WorldDatabase::GetStartZoneSoF(PlayerProfile_Struct* in_pp, CharCreate_Stru
 	in_pp->x = in_pp->y = in_pp->z = in_pp->heading = in_pp->zone_id = 0;
 	in_pp->binds[0].x = in_pp->binds[0].y = in_pp->binds[0].z = in_pp->binds[0].zoneId = in_pp->binds[0].instance_id = 0;
 
-    std::string query = StringFormat("SELECT x, y, z, heading, bind_id FROM start_zones WHERE zone_id = %i "
-                                    "AND player_class = %i AND player_deity = %i AND player_race = %i",
-                                    in_cc->start_zone, in_cc->class_, in_cc->deity, in_cc->race);
+	//this is wrong. if start_zone is set we should use that id
```
RicardoCampos: This comment is actually wrong :P

mackal:
This is something I noticed when I was fixing some issues but didn't get back around to fixing, still need to look over this code though.

PR #341, opened 2015-01-18
MS Visual Studio vs. position.cpp

af4t:
VS2013 wants <algorithm> included also, to be able to find std::min and std::max.  Enclosed test for MS in #ifdef if is not required by other toolchains.
(Note: came back & edited totally incoherent wee-hours subject line)


joligario:
Done in 52eb287082218e272ec4c79918fdac6a4da9235c

PR #351, opened 2015-01-25
Repair Faction System Messaging

noudess:
The faction messaging system did not match the new personal faction system.  Max and Min faction messages needed to be based on personal faction range, not adjusted range (race/class/etc).

The old code also posted a "you could not get any better"  or a "you could not get any worse" 1 faction hit too soon.  Fixed that.  Changes from min/max need to use the value after the hit to assure the correct message.  Tricky to think through but it is the case. 

Quest faction code had not been updated to match the normal faction code (it did not use the personal faction system).  Factored out common code and called that from both.

Tested both types of faction as well as all the edge cases.  Messages look good.

I also want to donate an SQL StoredProcedure.  Given a character name, it gives you the full rundown on where that char stands on all factions for which they have at least 1 hit.  This should help debug faction issues, be they a DB change that is needed or some new rules as we improve the faction system.

Stored in:  utils/sql/git/optional/2015_01_25_MyFaction_Stored_Procedure.sql


joligario:
We can't tie the database name in SQL. Not everyone uses `peqdb`.

mackal:
The SQL isn't included, he just was posting it as something that could help debugging issues.

noudess:
I was just intending that as a comment that people could use if they wanted.  But, I'm happy to add it to the optional folder.  Did I do it correctly?

joligario:
Oh, I know it wasn't included. That's why I was saying to add it to the optional folder.

sebshaik:
Some quick comments, as I'm not sure how much grinding I'll be doing.

Previously I was getting some other messages, now I seem to be truly bottomed out "Your faction standing cannot get any worse," which is correct.

I'm still getting the "Your faction standing with Coldain/Dain cannot get any better," yet these factions are only warmly to me. Because of this, I fear that BiC and Velious quests are still broken - once I see progress in these, I'll probably do a lot more work investigating the meta-system a bit more.

Could you share what the values are for each faction range? E.g., Scowls -3000 to -750, Ally 1100+, etc. With PEQ Magelo working again, I can easily test (and even count) these ranges to my own experiences and make sure they line up.

Given a quick look, the system looks to be much improved. Just a matter of testing, finding broken quests, and resolving individual factions.

af4t:
A report from the field on ""cannot get any better" vs. "considers you warmly" ... within the past week I did the Signet of Service and Regal Band of Bathezid quests, with a Human/Karana char, on someone else's server where I am just a peon.  These quests require ally or better faction, but I also showed maxed out at "warmly". 
I would suspect that the faction message text display routine might need a look, but have not actually done so yet.

noudess:
I'll take a look.  Can you run my MyFaction("char name") script and send me the result?

joligario:
[06:44] <Kayen> hmm looks like half the perl faction functions are broken quest::faction doesnt seem to work, cant seem to change faction via any of them
[06:44] <Kayen> $client->GetCharacterFactionLevel broken too
[06:53] <Kayen> SetFactionLevel2 broken

noudess:
These work for me.  Do you have all the current changes?  I just tested  $client->GetCharacterFactionLevel() and it worked fine.  I wonder what is different?

joligario:
He removed his name. I'll let him respond if he is still having issues.

noudess:
Ok.  Should I be checking IRC?  I've never used it before. Do I need an account for access to the area devs chat in?

Akkadius:
irc.eqemulator.net/6667
#eqemucoders

noudess:
16:00] [Info] Looking for server irc.eqemulator.net/ (port 6667)...
[16:00] [Error] Connection to server irc.eqemulator.net/ (port 6667) lost: Unknown error.

This is what I always get - thought maybe I needed something else.

noudess:
>Could you share what the values are for each faction range? E.g., Scowls -3000 to -750, Ally 1100+, >etc. With PEQ Magelo working again, I can easily test (and even count) these ranges to my own >experiences and make sure they line up.

I believe the values are:

>=1101 ALLY
701-1100 WARMLY
401-700 KINDLY
101-400 AMIABLE
0-100 INDIFF
-100--1 APPREHENSIVE
-700-0191 DUBIOUS
-999--701 THREAT
-1000 and below KOS

I got this from an old printout but I think its right - will check when I have a few min.

PR #356, opened 2015-01-31
Rof merchantlist

clucksoft:
RoF+ can send 200 items to the client at a merchant. I've tested this with RoF, RoF2, and UF.

### DIFF
--- a/zone/client_process.cpp
+++ b/zone/client_process.cpp
@@ -960,9 +960,9 @@ void Client::BulkSendInventoryItems()
 void Client::BulkSendMerchantInventory(int merchant_id, int npcid) {
 	const Item_Struct* handyitem = nullptr;
 	uint32 numItemSlots = 80; //The max number of items passed in the transaction.
-  if (ClientVersionBit & BIT_RoFAndLater) { // RoF+ can send 200 items
-      numItemSlots = 200;
-  }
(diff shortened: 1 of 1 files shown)


mackal:
Are all other clients 80?

clucksoft:
It's been at 80 for all clients in the server code until now. I can't actually verify that Titanium uses 80, but nothing pre-RoF is above 80.

clucksoft:
If any of the other clients use lower than 80, then its been broken for 10 years and I think we'd have reports of missing merchant items on those clients.
Uleat probably knows best when it comes to items: http://www.eqemulator.org/forums/showpost.php?p=231183&postcount=5
His suspicion about merchant list holds true from what I have seen.

joligario:
Titanium doesn't see more than 80 even if you add this change regardless of client.

clucksoft:
Exactly. This is only supposed to allow RoF+ to see more items, since it can.

mackal:
I was more just wondering if all other clients were verified to be 80 or not.

joligario:
You won't need the version bit check. It didn't crash Titanium or UF.

clucksoft:
Yeah, that is true that it doesn't crash UF if you send it more items than it accepts, but I think you may as well cut it short instead of continuing to send items when you know it won't use them.

joligario:
Good point, thanks!

PR #358, opened 2015-02-01
Repair min/max faction issues.

noudess:
My system was incorrect in that it only allowed characters to max at 1200 and min at 3000.  However, when I removed the race/class/deity aspects that SHOULD not be recoverable, I also left out faction start values, which should NOT.  So, if a faction started you at -800, you could only earn 1200 - leaving you at 400.  Now, you can earn 1200 - starting faction so, 2000 (minusing a minus).  The same fix was made for bottoming out.  I also added code to compensate for horrendous faction start values, as I saw one in the db for -30000.  The code got confused when the starting value was less that -3000.  Now, it understands and allows you to dig out of that hole.  Conversely, if you were to start at 30000 if will let you bottom out.

Race/Class/Deity hits still apply and can not be made up for.

SetCharacterFactionLevel also had a problem when faction came up to or down to zero that would cause the in memory copy to differ from the db.  Fixed that as well.


noudess:
Important note.

The previous patch could have incorrectly adjusted character factions by as much as the entire faction starting value if (value + starting) was bigger than 1200 or less than -3000.  Lets say a faction has a starting faction of -500.  The old system would let you earn 1200, for a net of 700.  If someone had say 1500 in their value, it would have dropped it to 1200.

I hate that I didn't catch this, and I want to apologize to the community for this and I regret any work I have caused GMs.

noudess:
Somehow a rogue change slipped in that was bogus.  Fixed it.

PR #372, opened 2015-02-09
A non magical weapon with an augment tagged as magic now hits as magic

noudess:
When a non magical weapon has an augment placed on it that is tagged as magic, the GUI shows the new adjusted weapon as magical but it still cannot hit a mob that requires a magic weapon.

This change makes the weapon behave as magical after an aug marked as magic is applied in combat.

Ex: It can now hit a willowisp.

I'm not 100% sure of this behavior on live (which augs are marked magical) but this seems quite reasonable.


noudess:
You guys ok with this?  I'd like to close the branch.

PR #376, opened 2015-02-18
Fix for Titanium characters wishing to start in paineel (erudite cazic folk).

noudess:
A recent change ended up with paineel Erudites ending up in erudin.  Added code to Titanium function to start them in paineel.

Also fixed an outdated comment.


noudess:
I think the merge conflict is because I updated changelog without pulling down a new one.  I won't make that mistake again.  Can you guys merge it and I'll make sure I'm current from now on.

PR #381, opened 2015-02-23
Allow server customization of swimming start value.

noudess:
Added a new rule.  If it does not exist, it uses the old value of 100 that was hard coded.

I also moved this to char creation from where it was (was running every time you entered a zone).


noudess:
I see now that I butchered the use of the Rules in that I should be using the macros and there is no need to check for rule existence since each rule has a default.  I;m streamlining and will update this pull request with cleaner code and some more core to handle sense heading as well.

noudess:
The changes are much cleaner now and added Sense Heading code.

noudess:
Any issues with this or can get get this pulled in?

PR #388, opened 2015-03-11
Enable multiple NPC equipment materials

iequalshane:
This change allows #npcedit
armtexture/bracertexture/handtexture/legtexture/feettexture to work
properly and sets individual armor slot materials for NPCs.


iequalshane:
Ah, sorry this was missing a header that somehow didn't get added to the commit.

KimLS:
I see changes to database calls but no SQL changes provided?

iequalshane:
This doesn't require an SQL DB change. The current columns for item material work with this change.

KimLS:
Alright thanks for clarifying

noudess:
The fields for the new textures don't exist in my npc_types and I don't see any sql stuff anywhere in the past optional or required that puts them there.  No biggie, I added em manually, but I'm wondering why noone else has an issue?

iequalshane:
Which ones were missing? I used the same existing fields as the #npcedit texture commands (#npcedit armtexture, legtexture, bracertexture, etc)

noudess:
Just zoning in to a zone after server boot.  The npc creation fails do to the npc_types query from the change to  zone/zonedb.cpp.

iequalshane:
I didn't run into this issue when using the DB generated by EQEmu Akka's PEQ Server Repack. I'll investigate tonight when I get home from work.

noudess:
Could just be my db is out of date for that table, but I;ve run the updates from the sql folder.

iequalshane:
Maybe the updates don't add the new entries because they weren't being consume without gm commands before. I'm not too familiar with this area but I wouldn't mind digging into it and resolving this issue so people upgrading older DBs aren't broken.

PR #395, opened 2015-04-01
Some illusions and some NPC gear not showing up on zone-in & initial spawn if in zone when it occurs

noudess:
Tree/Minor-Illusion-Objects:

These illusions (race 142/143) are problematic in a few ways.  The OP_ZoneSpawns has the correct race sent when you zone in, but the client does not update correctly, usually leaving you with a human appearance.  It is not a gender thing, I tried that along with many other things.  I also noticed that even OP_Illusion sent before OP_ClientReady was rejected.  I also tried moving the OP_ZoneSpawns later in the sequence to no avail.  My solution is to use OP_NewSpawn for any PC with these illusions on them when a new client zones in.  This works fine, as long as these packets wait for CLIENT_CONNECTED.

NPC Gear visibility:

Any gear (other than primary/secondary) worn by NPCs does not show up for clients zoning in.  The data sent by OP_ZoneSpawns is accurate, but again not utilized or somehow broken in the client.  I tried several variants as discussed above.  Nothing worked for these either as far as staying with OP_ZoneSpawn.  I added code to where the BulkZoneSpawn was being done, to send additional WearChange packets only if there are NPCs wearing gear in non equipment slots.  These packets are only sent to the new client zoning in, as again, that is the only need.

These changes won't add many packets and only add packets to someone zoning in so they see NPCs and illusioned PC (tree and obj) correctly.

One problem remains.  Some zones limit these illusions.  We don't implement this on the server and the client reacts somewhat strangely for the zones that limit this.   We perhaps need to disallow these illusions from being cast based on zone flags of some sort (from talking to Uleat).

To duplicate the previous problem without the patch:

NPC gear: Give Sparlus in erudnint some chain arms and you'll see him wear them.  Zone out (nake sure someone else is in zone or it is static) and come back.  They won't be displayed.

Illusion: Cast tree form in an allowed area.  Zone someone else in.  More than half the time the illusion isn't displayed.  For me it was like 90%.

Should both work find now, except in zones that limit these illusions.


noudess:
Hold off on merging for a bit.  I found that NPCs that spawn with gear don't push the appearance to GUI either.  The OP_NewSpawn sends the data to the client, but the client does not perform the wear operations on non primary/secondary.

noudess:
Ok.  Fixed npcs to also properly display worn armor slots on normal spawn when they are wearing something from lootdrops.  The OP_NewSpawn operator to the client was not causing the client to show the armor worn, depsite being filled in, much like the other packet for OP_ZomeSpawns.  Added code to the spawn logic to send clients the WearChange on spawn if its a PC race and they are wearing a lootdrop based armor piece.

noudess:
Any comments or issues?

PR #400, opened 2015-04-30
Now Npcs won't respond to hails if they can't see you.

noudess:
On live, if the NPC can't see you, he does not respond to hails.  Made eqemu match that and tested it.

### DIFF
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -1051,11 +1051,11 @@ void Client::ChannelMessageReceived(uint8 chan_num, uint8 language, uint8 lang_s
 		if(quest_manager.ProximitySayInUse())
 			entity_list.ProcessProximitySay(message, this, language);
 
-		if (GetTarget() != 0 && GetTarget()->IsNPC()) {
+		if (GetTarget() != 0 && GetTarget()->IsNPC() &&
+			!IsInvisible(GetTarget())) {
 			if(!GetTarget()->CastToNPC()->IsEngaged()) {
 				CheckLDoNHail(GetTarget());
 				CheckEmoteHail(GetTarget(), message);
-
 
 				if(DistanceSquaredNoZ(m_Position, GetTarget()->GetPosition()) <= 200) {
 					NPC *tar = GetTarget()->CastToNPC();


sebshaik:
This is giving a bunch of in-game problems.

The issue is NPC's haven't been tagged as "See Invis" or not. The DB needs a lot of work to support this change.

mackal:
That's normal. If the NPC can't see you, there is no hails. Like Madronoa in Natimbi doesn't see invis so you need to drop invis/hide/sneak to port to Qvic. Pixtt Kekken and Pixtt Uxnikk (Taskmistress spawnsers in Riwwi) also don't see invis so you need to drop to request Taskmistress.

sebshaik:
It's normal behavior, but the DB needs to be updated to align with Live in terms of NPC's who /should/ see you while invis.

Planar Projections I know aren't responding/giving flags while invis (I don't recall this being an issue, but would need to be confirmed). I'm also thinking of the various Magi for porting around LDoN and Taeloisia.

mackal:
Report the mobs that are having problems to PEQ or w/e server.

sebshaik:
Will do, thanks @mackal for the prompt replies and dedication.

Also thanks @noudess for all the fixes lately!

AthrogatePEQ:
This was tested on PEQ and Rogue SOS mobs didn't respond to the PC.  However a Bard singing Selos song of travel was responded too.  The was tested with Madronoa in Natimbi.

noudess:
So the message there is we need to make sure song of travel shows as invis then and have the NPC not respomd?  I can take a look if you want.

AthrogatePEQ:
That's what a guildie reported yesterday.  I can confirm the mob ignored my rogues properly, but a guildies Bard was not ignored.

noudess:
One note, Shroud of Steath is must stronger and many mobs that see invis creatures cannot see SoS.

AthrogatePEQ:
I can't see the PEQ DB - I looked in my DB and Madronoa can't see invis or sneak or hide in my database.

AthrogatePEQ:
Further testing I may have been given bad information.  May not be a problem.  Sorry

mackal:
The GL magus at least doesn't see invis.

PR #416, opened 2015-05-25
Making $npc->RemoveFromHateList actually work

hateborne:
$npc->RemoveFromHateList is a mob function, not an NPC function. Castingto Mob to let it work.


mackal:
This isn't needed.

PR #417, opened 2015-05-25
Revert "Making $npc->RemoveFromHateList actually work"

mackal:
Reverts EQEmu/Server#416

### DIFF
--- a/zone/perl_npc.cpp
+++ b/zone/perl_npc.cpp
@@ -823,7 +823,7 @@ XS(XS_NPC_RemoveFromHateList)
 		if(ent == nullptr)
 			Perl_croak(aTHX_ "ent  is nullptr, avoiding crash.");
 
-		THIS->CastToMob()->RemoveFromHateList(ent);
+		THIS->RemoveFromHateList(ent);
 
 	}
 	XSRETURN_EMPTY;


hateborne:
Mackal,

Unless you can show how it works, this does not function (or at least in any normal way). I compiled source, attached VS 2013 Express to all zone processes, set breakpoints at mob.cpp at line 2657. 

Example:
Setting one mob (MobA) to attack a neighbor (MobB), then hitting the second (MobB) creates a hatelist on MoB of "MobA, Hate". 

Original: When I trigger the $npc->RemoveFromHateList() via the snippet below, nothing happens. VS 2013 does not respond, I still exist on the hatelist, and the script stops progressing (does not proceed further from that call).

Commit 41ca23e: When I trigger the $npc->RemoveFromHateList() via the snippet below, my name is removed from the hatelist and the code AFTER it still functions (gmsay appears).


Snippet:
my @hatelist = $npc->GetHateList(); foreach my $h (@hatelist) { my $nh = $h->GetEnt(); if($nh->IsClient()) { $npc->RemoveFromHateList($nh); quest::gmsay("Removing [".$nh->GetCleanName."] from Hate List.",  15); } }

mackal:
My simple test works fine. (just doing $npc->RemoveFromHateList($client); in EVENT_AGGRO)

Issue #418, opened 2015-05-25
Random Character Select Screen Client Crashes In UF


Uleat:
Lead-in:
- http://www.eqemulator.org/forums/showthread.php?t=39684

Error codes seen with client crash:
- 0xC0000005
- 0xC0000096

Cause:
I believe this crash is associated with the prepending of 0x00 to opcodes where (opcode & 0x00FF == 0).

This action occurs both in application and protocol packets during the serialize() method call.

Occurence:
Based on testing, the crash occurs when an out-going packet's opcode is one whose value meets the above criteria (0x4200 in UF's case) and the packet's payload size and protocols are a multiple of eqstream->MaxLen(). (Tested with payload size of 1006 bytes.)

Log dumps:
[UF]
[05-25-2015 :: 01:30:53] [Netcode] 127.0.0.1:49157: Making oversized packet, len 1006
[05-25-2015 :: 01:30:53] [Netcode] 127.0.0.1:49157: First fragment: used 502/1006. Put size 1008 in the packet
[05-25-2015 :: 01:30:53] [Netcode] 127.0.0.1:49157: Subsequent fragment: len 506, used 1008/1006.
[05-25-2015 :: 01:30:53] [Netcode] 127.0.0.1:49157: Subsequent fragment: len 1, used 1009/1006.

[SoD]
[05-25-2015 :: 01:39:15] [Netcode] 127.0.0.1:49159: Making oversized packet, len 1006
[05-25-2015 :: 01:39:15] [Netcode] 127.0.0.1:49159: First fragment: used 502/1006. Put size 1008 in the packet
[05-25-2015 :: 01:39:15] [Netcode] 127.0.0.1:49159: Subsequent fragment: len 506, used 1008/1006.

- Notice the last sub-sequent fragment in [UF].
- When the payload is 1005 or 1007 bytes, the client crash does not occur..only when the length of the last packet is 1 (1006-byte payload in this example.)
- I suspect this is occurring with the affected opcodes where original packets are 502, 1006, 1510, 2014, etc... (+502, +504, +504, +504, etc...)

Fix attempts:
I tried excluding the preceeding 0x00 on this particular opcode and was able to make it to the char select screen. However, there were no characters listed..so, the preceeding 0x00 is required.

The only other fix I can see, which I have not tested as yet, is to check higher in the packet parsing for an affected opcode and adjust the p->size value before serialization.

[...]

PR #419, opened 2015-05-28
Application packet size fix for high byte opcodes

Uleat:
This directly addresses the root cause of issue #418.

I'd like someone to review it and/or have some more testing done on it before merging since it's in an area that I have little experience in.


emagi:
Good catch Uleat this definitely lines up with the EQ2 netcode

PR #424, opened 2015-06-02
Encounter timers

clucksoft:
This allows creation of quest timers that are tied to an encounter instead of a spawn. In particular, an encounter doesn't require a 'status' or 'trigger' spawn to exist in order to manage timers through the whole encounter.

It works by introducing a subclass of Mob called Encounter which gets instantiated any time an encounter is loaded, and timers get tied to that Encounter object. Then timers can be registered like other scripts, with eq.set_timer and that calls back to event_timer as usual.

It is possible I've overlooked something as this system is fairly complex, but my tests are working and I have run a couple existing encounters and they worked fine (but I only had a couple available to run!)
KLS has glanced over the code and I have fixed the only problem he saw.

### DIFF
--- a/zone/CMakeLists.txt
+++ b/zone/CMakeLists.txt
@@ -20,6 +20,7 @@ SET(zone_sources
 	embparser_api.cpp
 	embperl.cpp
 	embxs.cpp
+  encounter.cpp
 	entity.cpp
 	exp.cpp
 	fearpath.cpp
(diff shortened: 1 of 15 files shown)


clucksoft:
Ok. Travis CI isn't my friend today. Its not including encounters.cpp in the build, and I'm not sure how to fix that.

PR #426, opened 2015-06-07
Full group was being excluded from group exp bonus

clucksoft:
Simple fix, but still tested and verified on my server.


mackal:
This is why I hate commits that do a crap ton of unrelated things.

sebshaik:
FYI - this seems working/pushed to PEQ as of today's reset. Thanks again!

PR #434, opened 2015-07-06
Bot saylinks and command aliases.

KinglyKrab:


KimLS:
I see a change to a query but no accompanying SQL changes.

KimLS:
Otherwise I think this is fine.

KinglyKrab:
Well, the SQL change is actually to fix the query itself, it's using old columns 'rezzed' and 'isBurried' when the new columns are 'is_rezzed' and 'is_buried'. Check prntscr.com/7phswk for reference.

PR #438, opened 2015-07-16
Add Optional Enforce Task Level Requirement on AssignTask

hateborne:
Affected:
* quest::assigntask
* $client->AssignTask

This is to allow those task level requirements in DB to be used,
optionally. It defaults to do not enforce (translated: no change for
current servers), but can be set with an additional flag to both
commands (see below). The LUA bit was added thanks to help from demonstar55.

* quest::assigntask(703); # works
* quest::assigntask(703, 1); # denies user
* $client->AssignTask(701, $npc->GetID()); # works
* $client->AssignTask(701, $npc->GetID(), 1); # denies user

### DIFF
--- a/zone/lua_client.cpp
+++ b/zone/lua_client.cpp
@@ -1077,6 +1077,11 @@ void Lua_Client::AssignTask(int task, int npc_id) {
 	self->AssignTask(task, npc_id);
 }
 
+void Lua_Client::AssignTask(int task, int npc_id, bool enforce_level_requirement) {
+	Lua_Safe_Call_Void();
+	self->AssignTask(task, npc_id, enforce_level_requirement);
+}
+
 void Lua_Client::FailTask(int task) {
 	Lua_Safe_Call_Void();
 	self->FailTask(task);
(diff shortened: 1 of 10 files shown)


zone/lua_client.cpp
```diff
@@ -1525,6 +1530,7 @@ luabind::scope lua_register_client() {
 		.def("FindSpellBookSlotBySpellID", (int(Lua_Client::*)(int))&Lua_Client::FindSpellBookSlotBySpellID)
 		.def("UpdateTaskActivity", (void(Lua_Client::*)(int,int,int))&Lua_Client::UpdateTaskActivity)
 		.def("AssignTask", (void(Lua_Client::*)(int,int))&Lua_Client::AssignTask)
+		.def("AssignTask", (void(Lua_Client::*)(int,int,bool))&Lua_Client::AssignTask
```
mackal: Missing closing paran here

hateborne: Did a live edit and resolved. Apologies. :-\

PR #440, opened 2015-07-18
Compile errors with VS2015 Community RC/RTM

af4t:
Very preliminary work with Visual Studio 2015 Community RC/RTM, resolve some version-specific build errors:

1. Standard library functions snprintf & vsnprintf are available thusly named (without the prepended underscores).  Someone may want to check my conditionals against non-MSVC compilers, though.
2. zone/bonuses.cpp throws an internal compiler error with optimize enabled; #pragma disable optimization on VS2015 only. Doing further testing as to whether we have to disable optimization for the WHOLE file ...

### DIFF
--- a/common/tcp_connection.h
+++ b/common/tcp_connection.h
@@ -23,7 +23,9 @@
 */
 
 #ifdef _WINDOWS
-	#define snprintf	_snprintf
+	#if (!defined(_MSC_VER) || (defined(_MSC_VER) && _MSC_VER < 1900))
+		#define snprintf	_snprintf
+	#endif
 	#define strncasecmp	_strnicmp
 	#define strcasecmp	_stricmp
 
--- a/common/types.h
+++ b/common/types.h
@@ -43,7 +43,9 @@ typedef unsigned char		uchar;
 typedef const char Const_char;	//for perl XS
 
 #ifdef _WINDOWS
-	#define snprintf	_snprintf
+	#if (!defined(_MSC_VER) || (defined(_MSC_VER) && _MSC_VER < 1900))
+		#define snprintf	_snprintf
+	#endif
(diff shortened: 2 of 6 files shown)


mackal:
Is number 2 a compiler bug?

number 1 looks good (as long as they're all in windows only bits, which I think they are)

af4t:
With optimizations enabled it throws an internal compiler error at the last line of bonuses.cpp.  I'm not where I can see the exact message, but near as I can recall it refers to "F:\dd\(blah blah)" and suggests simplifying the code.  Which for bonuses.cpp I don't see happening any time soon.  Disable optimizations and it does compile and link.  I don't have an objective way to measure any specific performance hit, though.
I saw that there were some other bug reports for VS2015 around internal compiler errors and switch statements ... so I would say MS will get to this sooner or later.

af4t:
Here's the exact message:
1> ...\eqemu myserver\zone\bonuses.cpp(4686): fatal error C1001: An internal error has occurred in the compiler.
1>  (compiler file 'f:\dd\vctools\compiler\utc\src\p2\main.c', line 247)
1>   To work around this problem, try simplifying or changing the program near the locations listed above.
1>  Please choose the Technical Support command on the Visual C++
1>   Help menu, or open the Technical Support help file for more information

mackal:
My only guess is that in NPC::NegateSpellBonuses the last to cases don't have a break (an error) but still 100% valid code which should compile. Can you try adding breaks and see if it fixes it?

af4t:
Just adding the break;s didn't cure it, BUT this did point me to that turn off optimizations ONLY on Mob:NegateSpellBonuses is enough to get it compiled. Closing this and trying to resubmit

PR #441, opened 2015-07-20
Compile errors with VS2015 Community RC/RTM v2.0

af4t:
Very preliminary work with Visual Studio 2015 Community RC/RTM, resolve some version-specific build errors:
-- Standard library functions snprintf & vsnprintf are available thusly named (without the prepended underscores). Someone may want to check my conditionals against non-MSVC compilers, though
-- zone/bonuses.cpp throws an internal compiler error with optimize enabled; #pragma disable optimization on VS2015 only for the function Mob:NegateSpellBonuses only is enough to work around the internal error.
PS sorry for the multiple commits ...

### DIFF
--- a/common/tcp_connection.h
+++ b/common/tcp_connection.h
@@ -23,7 +23,9 @@
 */
 
 #ifdef _WINDOWS
-	#define snprintf	_snprintf
+	#if (!defined(_MSC_VER) || (defined(_MSC_VER) && _MSC_VER < 1900))
+		#define snprintf	_snprintf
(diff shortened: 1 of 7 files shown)


mackal:
Looks like they released a new version today, is the issue still present?

af4t:
Had to install the new version after I got home from work, so just now finding out ...
Yes, it's fixed now.  Closing this one too, resubmitting with just avoiding the unneeded macro defines for library functions.

PR #447, opened 2015-08-04
Fix snow so it ends correctly.

noudess:
All packets to end weather should have the type.  The code in command.cpp knows this, the automated weather system did not.  Snow was not ending correctly.

### DIFF
--- a/zone/zone.cpp
+++ b/zone/zone.cpp
@@ -1346,7 +1346,6 @@ void Zone::ChangeWeather()
 
 			weathertimer = (duration*60)*1000;
 			Weather_Timer->Start(weathertimer);
-			zone->zone_weather = 0;
 			zone->weather_intensity = 0;
 		}
 		else if(tmpOldWeather == 2)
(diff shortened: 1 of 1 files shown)


mackal:
Can you give me a few steps to reproduce so I can observer? (I know you mentioned them in IRC ...)

noudess:
Sure.  So when stow starts in a zone (stonebrunt is where I noticed the bug and fixed it), you get snow visuals and wind (if you're not in a tunnel).  The wind, never stops.  That's because the code sends up to end-rain packet for everything.  That seems to end the visual correctly, but not the audio (wind).  The code in command.cpp for the #weather command works correctly.  I just fixed the automated weather code to send the correct end packet (type and 0 for intensity).  Stonebrunt is working fine now.

mackal:
I couldn't get it to snow ... I'll trust you that it fixes it :P

noudess:
Takes awhile.  I just plunked someone there with loggin of the packets.

I wonder how they opened Velious..

PR #448, opened 2015-08-08
Fix druid bot casting Share Wolf Form instead of Pack Spirit

N0ctrnl:


mackal:
I guess saying wolf makes more sense for it cast this spell, can you think you can add in another that will cast Pack Spirit still?

N0ctrnl:
I'm actually going with something different on this.

PR #457, opened 2015-09-30
Update to give client access to the proper AAs for the expansions all…

Cilraaz:
…owed

Bit shifting by expansion value minus 1, rather than expansion value


Cilraaz:
Further discussion regarding this change (and the logic behind it) found at the below URL

http://www.eqemulator.org/forums/showpost.php?p=243460&postcount=22

PR #467, opened 2015-11-02
Old Race/Class Experience Mods

Cilraaz:
I wasn't sure if this was something that was wanted in the base code.  This created two rules (false by default) that control the old race and class experience penalties/bonuses, as the class penalties were removed in Velious while the race penalties weren't removed until much later.  The change is put in by increasing the character's required xp to level, as opposed to penalizing xp gains, as this is how it was originally implemented by SOE.  Ideally, they should be used separately from Character:UseRaceClassExpBonuses to avoid doubling up on bonuses.

Let me know if this style change isn't wanted in the base code and I'll refrain from future pull requests of this kind.


noudess:
I'll use it if it's integrated.

mackal:
quick review of the code it looks fine, but touching XP is scary! Don't want to break shit, but it doesn't look like it should :P

PR #469, opened 2015-11-10
Augmentation Feature Patch

Shendare:
RoF+ clients now support the built-in adding, swapping, destroying, and removing of augments in equipment, updating an equipped item's look in case of ornamentation changes. All clients will now verify that the proper distiller (or a perfected distiller for RoF+) is being sent for consumption for safely removing augments. Hard-coded item IDs for distillers have been replaced with checks on item types.

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -2924,7 +2924,6 @@ void Client::Handle_OP_AugmentItem(const EQApplicationPacket *app)
 		return;
 	}
 
-	// Delegate to tradeskill object to perform combine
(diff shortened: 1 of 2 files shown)


KimLS:
Has anyone else looked at this? I kept meaning to; it's so big it's hard to tell if it causes any issues just by reading it quickly and I keep forgetting.

Akkadius:
I've looked at it just not deeply and have not tested

Shendare:
Hopefully someone will get a chance to give it some second-party testing.

Here's a video of tests of each operation in each supported client:

https://www.youtube.com/watch?v=ly0G2NxZp7o

PR #473, opened 2015-12-16
Repair issue with Bind Wounds when you are binding someone that is binding themselves.

noudess:
When binding someone else's wounds, the code tried to send a bind
    would response to client with a type of 2.  The intent (based on
    comments was to get the client to display a "stand still" message
    on the receiving client.

    That reply message was not generating that message, but if the client
    you were binding was also binding his own  wounds, it would cause your
    target to stand up, interrupting his bind.

    I replaced this client reply with a simple, directed client message.  It all
    seems to work fine now.

   I also eliminated a duplicate message.


noudess:
I should add...  the failure I mentioned only occurs if:

- you are binding someone elses wounds
- that person is already binding wounds
- that person is standing (crouching obviously - due to bind anim)

PR #478, opened 2015-12-22
Various cout end line fixes.

stheno:


Akkadius:
What is the point of this?

Also, end line function is not a direct replacement for a \r

stheno:
Everything is fine in a console, This just assists interfaces that may handle the console windows as a sub process.
As it was the cout buffer was not being cleared right and interfaces were not getting line feeds.
Thereby not getting any text at all.

stheno:
Also 90% of the console code has endl; This just fixes some of the missing ones.

PR #486, opened 2016-01-02
Added runspeed to #mystats window

noudess:
Just added runspeed to #mystats window.


mackal:
I know this window has a limited buffer, no idea if we're close to hitting it though ...

PR #496, opened 2016-01-26
Item weights being downcast to uint8

daerath:
Item weight was being downcast to uint8 which impacted any item with weight over 255.  For SoD, SoF, and UF, prior to sending item info to the client we now cap weight at 255 to ensure the item remains heavy instead of being made (in most cases), super light.


Uleat:
Looks ok - I'll work on Titanium once I get it merged.

PR #499, opened 2016-02-08
Simplification of ItemInst IsAugmentable()

daerath:
IsAugmentable uses a for loop to check AugSlotType to see if an item is augmentable.  Non-augmentable items iterate EmuConstants::ITEM_COMMON_SIZE times (6 currently).

Augmentable items only ever iterate once because it is true for every augmentable item that AugSlotType[0] != NO_ITEM.

The change is to just check AugSlotType[0] and return the result.

### DIFF
--- a/common/item.cpp
+++ b/common/item.cpp
@@ -1656,12 +1656,12 @@ bool ItemInst::IsAugmentable() const
 	if (!m_item)
 		return false;
 
-	for (int index = 0; index < EmuConstants::ITEM_COMMON_SIZE; ++index) {
-		if (m_item->AugSlotType[index] != NO_ITEM)
-			return true;
-	}
-
-	return false;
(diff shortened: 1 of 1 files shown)


mackal:
They do not have to be filled in order. RoF+ lets you put them in any slot (that it fits in) and power sources use a specific slot (usually last usable slot) and that can be filled out of order.

daerath:
To make sure I understand, although the six augslot#type columns are all filled in order where if augslot1type is 0 then all of the others are also 0, they don't have to be filled in that way when an item is created?  If that's the case, then yes, this change would be bad.  =)

mackal:
Looking at it again, it probably depends if the client actually cares. If it doesn't care, then this would disagree with the client.

daerath:
I didn't notice any issues, but it may be safer to withdraw this pull request.

PR #506, opened 2016-03-07
Zone crash fix

clucksoft:
This fixes a case where spells with spell target ST_GroupClientAndPet (notably spirit of the panther, etc.. not many spells use this type) crash zone when Mob::IsRaidGrouped() returns true and Mob::GetRaid() returns null. Athrogate has sent me a few PEQ crash reports where this was the cause.

Also fixed a couple spots where the wrong mob was being cast to client when getting the target group id, but I don't think that was causing any real problems.

This doesn't address the underlying mismatch between those functions but that seems to be a pretty deep problem.

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -1732,7 +1732,10 @@ bool Mob::DetermineSpellTargets(uint16 spell_id, Mob *&spell_target, Mob *&ae_ce
 						}
 						else if(IsRaidGrouped())
 						{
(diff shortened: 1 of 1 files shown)


mackal:
Well this is better than crashing, but as you said, why is this happening.

Does anyone know? Like casting on a pet or something? I know someone said it happened every time they left the zone, but the raid should be valid until everyone has left ...

clucksoft:
I'll take some more time to try and track the real issue down.. I didn't know any details like it was happening when zoning out, so I wasn't able to reproduce it. Probably still won't be able to, but that gives me somewhere specific to look. Have you looked into it at all?

PR #517, opened 2016-04-18
VS2015: Fix compile errors after EQDictionary rework

af4t:
VS2015: In zone\client.cpp, fix call(s) to SharedDatabase::GetItem fails if SharedDatabase::GetItem is declared as const.  (This must not be an issue for non-MS compilers? #defined fix for only if MSVC 2015.)  Error appeared after Uleat's "Reworked EQDictionary into namespace EQEmu" commit. 
This doesn't _appear_ to break anything else; then again I am not analytical enough to resolve this other than by brute force.

### DIFF
--- a/common/shareddb.cpp
+++ b/common/shareddb.cpp
@@ -1084,7 +1084,11 @@ void SharedDatabase::LoadItems(void *data, uint32 size, int32 items, uint32 max_
 	}
 }
 
+#if (defined(_MSC_VER) && (_MSC_VER > 1899))
+Item_Struct* SharedDatabase::GetItem(uint32 id) {
+#else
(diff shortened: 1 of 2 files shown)


Uleat:
I'll keep an eye on this one..no idea what changes I made would cause that...

You really don't want to pass around that particular reference as a non-const pointer, though.

af4t:
I am trying to look harder at this myself, too.
Am I correct that it's the _address of the pointer_ to Item_Struct that has to be constant? (I would hate to be working from a faulty premise.)

af4t:
No more linker error after your last "eliminate nested classes" commit, issue resolved.  You rule, sir.

PR #522, opened 2016-05-25
FreeBSD Compile Fixes

9thsector:
Some freebsd compile fixes.


mackal:
Too lazy to verify, I'll trust you

PR #526, opened 2016-05-28
Allow specifying log directory in eqemu_config.xml

9thsector:
Allows specifying the log directory in eqemu_config.xml
Also creates a compile-time define for passing in the path for locating eqemu_config.xml
Defaults to current directory.


Akkadius:
If you can re-pull this I can look at merging this, not sure how this sat for so long

PR #530, opened 2016-06-11
Add luarocks support

mackal:
This allows one to use the lua_modules folder as their luarocks tree
ex (from the server folder with lua_modules using default location)

luarocks --tree=lua_modules install luasql-mysql MYSQL_INCDIR=/usr/include

This will install the luasql mysql module so you can now do queries from lua


mackal:
This does work, I just wanted opinions on "doing it right"

PR #535, opened 2016-07-04
Replacing \\ and / in strings with portable paths

fzzzt:
I'm running into path problems on FreeBSD running the script non-interactively. Maybe interactively too and I just haven't noticed... It was silently not finding any SQL files to execute for the PEQ download because there were path delimiters in strings. Changing these to use a function that should automatically generate \ or / based on the platform fixed it, hopefully in a cross-platform way (I can't test on Windows).

This issue likely exists in various places, but I haven't gone through and changed all of the references yet. This one used \\ in the string, so it's possible that strings with / are already cross-platform somehow, assuming they work on Windows...

P.S.: When I first tried this commit, a few dozen line-endings changed from CRLF to LF. I re-forked, re-cloned and re-edited it with another editor to avoid that. If I save with CRLF, it ends up changing tons more, probably hundreds, so there are mixed endings in the current file. Do I need to continue to preserve the mixed endings or would it be acceptable to commit with all LF endings (fewer changes)?

### DIFF
--- a/utils/scripts/eqemu_update.pl
+++ b/utils/scripts/eqemu_update.pl
@@ -13,9 +13,13 @@ use File::Basename;
 use File::Copy qw(copy);
 use POSIX qw(strftime);
 use File::Path;
+use File::Temp;
 use File::Find;
 use URI::Escape;
 use Time::HiRes qw(usleep);
+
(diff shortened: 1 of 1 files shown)


fzzzt:
Huh, I dunno how the second commit got into this request. I meant to only include the first once, maybe it automatically adds later commits after the request is created, like it's file-based and not commit-based. I'm still a GitHub noob.

PR #536, opened 2016-07-09
Fix Voice Graft to work only when targetting pet.

noudess:
Per what I recall, and a few comments on the spell on Zam, you can still speak yourself if you target yourself.

Fixed to allow this.


noudess:
I need to make a quick fix.  This patch prevents you from speaking to an NPC when you have the spell on your pet regardless of target.

noudess:
Ok it's fixed.

noudess:
Swapped checks for expediency and safety.

PR #547, opened 2016-08-25
Added quest::camerashake(duration,intensity) function. 

TheGrandPackard:
Similar behavior as the #camerashake command that can be called in a perl quest script.


Natedog2012:
Would be nice to see options in this like the CameraEffect.  (global / singleclient ect)

https://github.com/EQEmu/Server/blob/ef2c17748e612959d75328c023b6abb9c364ef1b/zone/perl_mob.cpp#L7268

PR #552, opened 2016-09-05
Implemented some basic commands.

stheno:
broadcast, zone shout, entity say and shout (both npc and pc).
Alpha'd the auth handler list so it is easier to read when/if we put more in.


stheno:
Forgot to mention this:
"Started to try and figure out how to implement repop force."

PR #559, opened 2016-09-29
Fix Random Ground Spawn Z

TheGrandPackard:
This fixes the random ground spawn z location on RoF2 clients, notably the Eastern Wastes skinning rocks, but also other occurrences. Titanium did not have an issue with this, and it is tested to still be functional.

I changed the max_z in the database to be equal to BEST_Z_INVALID (-99,999) to activate this code, and when ground spawns are generated, the Z value is set to the closest Z per the map file. To affect all ground spawns with variance in the X/Y location, use this query: `UPDATE ground_spawns SET max_z = -99999 WHERE max_x != min_x OR max_y != min_y`.

I also had to reorder the zone initialization code for this to work, as the zonemap is loaded after groundspawns, which prevented calculating the closestZ.

### DIFF
--- a/utils/sql/git/optional/2016_09_29_GroundSpawnsMaxZ.sql
+++ b/utils/sql/git/optional/2016_09_29_GroundSpawnsMaxZ.sql
@@ -1,0 +1,1 @@
+UPDATE ground_spawns SET max_z = -99999 WHERE max_x != min_x OR max_y != min_y
--- a/zone/object.cpp
+++ b/zone/object.cpp
@@ -451,6 +451,21 @@ void Object::RandomSpawn(bool send_packet) {
 
 	m_data.x = zone->random.Real(m_min_x, m_max_x);
 	m_data.y = zone->random.Real(m_min_y, m_max_y);
+	
+	if(m_data.z == BEST_Z_INVALID) {
+		glm::vec3 me;
+		me.x = m_data.x;
+		me.y = m_data.y;
+		me.z = 0;
+		glm::vec3 hit;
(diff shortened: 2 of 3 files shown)


mackal:
Can you add an optional SQL that does the update? I'm pretty sure this shouldn't be a required SQL :P

TheGrandPackard:
The SQL above is optional and without setting the max_z to -99,999 for any ground spawns, no change in the behavior of ground spawns will occur. The query I provided is a convenience to opt-in any and all ground spawns that have a random spawn location (variance in X or Y location). Any static ground spawn locations will not be affected by any of these source changes or sql queries.

mackal:
There is a folder at utils/sql/git/optional that would be a good place for stuff like this so they don't have to go read a PR or commit.

TheGrandPackard:
Ah, I'm tracking now. I created the optional SQL file in that folder.

noudess:
Can something similar be done for objects (tradeskill stations and such)?

N0ctrnl:
This seems to work just a little too well. The few zones I've tested it in, the items seem to be placed right on the Z axis, but they're very hard to see. Maybe add like 0.05 to the Z location or something to move them up just a tad?

TheGrandPackard:
@noudess This could definitely be done for objects. With the changes to how the zone boots up, the FindClosestZ code can be applied when loading those. The calculations I do are limited in scope to random ground spawns for now.

TheGrandPackard:
@N0ctrnl It looked fine for the skinning rocks in Eastern Wastes, almost like they were dropped into the snow. Thanks for checking with other item models. I added the 0.1 for better visibility.

PR #562, opened 2016-10-09
Fix rogue merchant usage under sneak.

noudess:
Rogue sneak was not taken into account while checking each of the merchants items against faction requirements, resulting in rogues only seeing temporary items on merchants when using sneak/buy.


noudess:
Don't pull this yet.  Having some odd issues buying once list is displayed.

noudess:
Ok, found and fixed the issue.

PR #566, opened 2016-10-18
Optional rule to grant HoTT to newly created characters

N0ctrnl:
This adds a rule - Character:GrantHoTTOnCreate - that if turned on will give all newly created characters the group leadership ability Health of Target's Target without the need to spend group ability points.


N0ctrnl:
If there's a better way to do this globally, somebody please speak up :)

mackal:
I think I would rather add scripting support to add leadership (if they don't exist already)

That way everything is more flexible.

N0ctrnl:
I looked for that first, but didn't find anything. This was about the best I could come up with, but I'm sure there's a better way that could be coded.

PR #573, opened 2016-11-19
Allow mobs to use defensive procs that are found in npc_spells table.

noudess:
I noticed the defensive_proc field in npc_types.  But when filled it, they never fired.

I believe this is the change needed.


noudess:
Added a fix to fishing messages rather than starting a new branch for that.  Just shows actual fish name instead of "caught something.." coming out every time.

mackal:
There are issues with this PR. (mainly checking way too often and NPCs vs anything is already checked so this adds extra checks there too)

PR #581, opened 2016-12-15
Fix potential infinite loop in loginserver's config file reader

SCMcLaughlin:
Some platforms define char to be unsigned by default (motivating example: Raspberry Pi / ARMv7) while GCC headers always define EOF as -1. This leads to an infinite loop when EOF values from fgetc() are stored as 255 and then checked against -1 as a stopping condition.


mackal:
This is why we should use auto. Also int is fgetc(FILE*) return type anyways.

PR #583, opened 2016-12-21
Alliance spell line for chanters was not working.

noudess:
I'm looking for help here to determine if I need more code for this fix.

Currently, chanters could not cast spells like Benevolence as a check was invalidating all PC->NPC benevolent spells.

I used an existing function (IsAllianceSpellLine()) to fix this.  I'm a little worried about other spells that increases faction ending up in something we don't want.  Any advice from  you guys on if this needs more bolting down, or if this is good enough?

Obviously this worked at some point, but seems broken in the current baseline, unless I am missing some setting.

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -3631,7 +3631,7 @@ bool Mob::SpellOnTarget(uint16 spell_id, Mob *spelltar, bool reflect, bool use_r
 
 				}
 
-				if(!IsBeneficialAllowed(spelltar) ||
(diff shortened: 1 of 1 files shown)


af4t:
From the **NPC target's** point of view, should the Alliance spell line even be considered "beneficial"?

noudess:
Yeah, I thought the same thing.  It is the comments I see in various threads and previous "fixes" to this in varying eqemu lines that all refer to it as such.  I believe it is because the mobs never aggro based on it?

noudess:
This is working well on my server.  It seems to work like people remember it on posts I have read.

Issue #543, opened 2016-08-19
Slam not working for non-bash classes?


noudess:
The racial slam command does not seem to function for non bash classes.

From what I understand it should have a chance of stun and 1 HM of damage.

Creating a level 1 barbarian rogue - mashing slam seems to do nothing.

mackal:
This probably broken we I made it so Bards with Furious Refrain wouldn't crash the sever ><

Akkadius:
@mackal did you plan on adjusting this or maybe you did it already?

emagi:
This capability was never supported, unfortunately my code is way off from yours at this point, but maybe this snippet can help (OPCombatAbility bash skill block).  This block is assuming the stun is successful, so any 'chance' to stun should be handled before this.


				bool stunTarget = false;
				
				switch(GetClass())
				{
				case SHADOWKNIGHT:
				case PALADIN:
					{
						const ItemInst *wpn = CastToClient()->GetInv().GetItem(SLOT_PRIMARY);
							if ( wpn && ( wpn->GetItem()->ID == 10099 || // Pally Epic
										wpn->GetItem()->ID == 14383 )  ) // SK Epic
								stunTarget = true;
						break;
					}
				}
					switch(GetBaseRace())
					{
					case BARBARIAN:
					case TROLL:
					case OGRE:
						{
							stunTarget = true;
							break;
						}
					default:
						{
							// Image (2014): this will allow non big races when they bash to stun if they have a shield
							if ( IsClient() && skill == BASH )
							{
								ItemInst* inst = CastToClient()->GetInv().GetItem(SLOT_SECONDARY);
								if ( ( inst != NULL && inst->GetItem()->AC > 0 ) || CastToClient()->GetAA(aa2HandBash) > 0 ) // Image: 5/28/2015, two handed bash (aa2HandBash) addition to stun fix for bash
									stunTarget = true;
							}
							break;
						}
					}

				
				if ( stunTarget )
				{
					// these guys can bash and stun
					if ( who->IsClient() )
						who->CastToClient()->Stun(MakeRandomInt(500,1000), true);
					else if ( who->IsNPC() && !who->CastToNPC()->SpecAttacks[UNSTUNABLE] )
					{
						who->CastToNPC()->Stun(MakeRandomInt(500,1000), true);
					}
				}

emagi:
[...]

PR #591, opened 2017-03-12
Added an optional rule value that forces startzones to be the same as…

SecretsOTheP:
… the bind zone and enabled bind xyz once again.


mackal:
This should resolve #543

PR #594, opened 2017-03-14
Fixed Task Item Handin

TheGrandPackard:
This fixes the task item processing to consume the coin and items passed so that they never make it to the quest execution. This cleans up situations where you want an NPC to handle tasks as well as return unneeded items back.


Akkadius:
I like that you're trying to solve this problem, because it is a problem. However it doesn't take care of all of the scenarios that come up with this problem.

It's a touchy problem because tasks aren't the only thing that present this issue. Sometimes, you actually have two scripts that are doing "Return items", you need to be able to account for those as well
And another case is you have clients that have actually received an item from a quest.

The stop return in the code, while experimental, worked very well in EZ however you have to rework the way your scripts behave fundamentally for it to work correctly.

In the script, I use a half a second window, which is more than fine unless the zone actually hangs at all during the event that an item was handed in, so you can set the window to be a second or two for that reason and be completely safe.

Also you need to have plugin::stop_return(); in your hand in scripts in quests to keep from the global return from handing items back.

Below is a snippet of the modified return items plugin. 

https://gist.github.com/Akkadius/ffb7d4f66c7f9dd5eb812ec89671dc06

I took it out of EZ in the extremely rare instance that it would actually end up duping items and didn't have time to mess with it more. It ran really well for a good year and a half. And before resubmitting this to the live repo, obviously all of the scripts need to be adjusted to work with this, but PEQ doesn't really have the issue of designing their quests to run into this issue to begin with - because they don't safety net their players who do accidentally hand-ins via script.

HTH

TheGrandPackard:
Yeah it definitely sounds like there are some scenarios that I'm not accounting for. I can't think of any scenarios that fit the bill though. Are there perhaps other ways to handle quest hand-ins than the Perl/Lua scripts? If there are some examples you can provide i'd like to expand this to cover them or at least know what the weaknesses are in the quest hand-in logic.

Akkadius:
I provided the scenarios, however if you want a script example of the usage of the modified plugin, this is what a typical one would look like.

This is a quest script someone wrote from EZ:
https://gist.github.com/Akkadius/d056c1d035e3bb648ec1e187a6863000

Essentially what you're doing is setting a timestamp in the source on the NPC to check against when any of these actions happen, so you can see if a task, hand-in, quest summon just happened so you can prevent from returning their items back which is what you're trying to do.

This is the way I did it, and while I don't have time to revisit this issue at the moment, it needs to get fixed gracefully to cover all scenarios for both perl/lua and all other scenarios.

It's not an easy problem, so if you don't want to tackle it EQEmu wide, I don't blame you.

TheGrandPackard:
So in that example, the author is calling the stopreturn after checking the item handing. Won't the check_handin call consume the items that are required for the quest  and already prevent them from being returned? I know that the check_handin function is located another file so it may be different than what i'm using. I'm using the same one in the quest repo and the quests all work (afaik) as expected without needing to call stopreturn or do anything else to prevent the items from getting returned and duped.

https://github.com/EQEmu/Quests-Plugins/blob/master/plugins/check_handin.pl

Akkadius:
Yes, it will, but if you have a global_npc.pl script with EVENT_ITEM that is also doing a return_items as a catch all, you run into the issue of still getting items returned.

You can see the modified return_items function here that I posted before, it accounts for 3 different scenarios.

https://gist.github.com/Akkadius/ffb7d4f66c7f9dd5eb812ec89671dc06

TheGrandPackard:
Ah, yes. I see where that can also cause issues. So if i were to add ```this->EVENT_ITEM_ScriptStopReturn();``` back in, would that isolate the rest of the changes in this PR to only handle the task turn ins better and not impact how the scripts are handled? The servers I run don't have a global_npc.pl script so I removed it because i didn't think it would be relevant any longer, but it would work fine with that call put back.

I'm not looking to invest a ton of time into the many ways that quest files can further do things, I just wanted to make a clean solution for consuming tasks hand-ins to prevent extra work in the quest files and share it.

Akkadius:
Use the modified return_items in tandem with the unmodified emulator source and your problem is resolved the same instead of adding a bunch of proprietary handling in the source.

You can also handle tasks in script so you don't have the script returning items and do your logic completely in script without the need of a modified return_items plugin.

Or lastly, don't use return items on the same script of the NPC that is handling a task update when cash is handed in, or the items that the task check for are handed in.

PR #609, opened 2017-04-15
Adds XTarget for Bot Pets to Bot Owner

spdkils:
Making it so bots pets pass hate up the chain to the bot owner. Also AutoXTarget when bot pets gain hate, and pass 0 hate up to chain to the Client() that controls Bot.

XTarget didn't populate when a bot pet hit things, even if the bot was added to agro, and as soon as the bot does anything it will add to AutoXTarget, but the second trip to the well seemed unessisary.

This way, if any bot, or any pet of a bot does an aggressive act, it passes to the client owner immediately, and seamlessly.

This is not a re-do of the last commit, it's in addition to, to cover Bot pets.

This checks immune agro flags, as well as feigning, as per usual, however other bot cases don't seem to always do that. I used the same logic that was already in place for simple pet checks. Maybe they should all be consistent across all hate gain.

It would be nice to get on XTarget even if you're feigned since it's added feign memory... However, since the AutoXTarget function explicitly says it's not safe, I didn't do that. Tested running with bots in raid, and single group for several hours of playtime, no issues.


mackal:
When you are FD'd on live and send your pet into attack, it will not add it to XTargets. That behavior was kept because it was wanted.

Ideally bots would manage their own xtarget managers like clients rather then jerry rigging stuff like this. (This was something I was going to look into once I worked out the remaining kinks)

spdkils:
I noted when you clear FD memory it removed you from XTarget, so I thought for simplicity if you had any type of memory on any mob it would make sense if  you saw it by your AutoHate Xtarget list.

However if it never functioned like that on live, I'm not worried about it.

Because I play with bots a LOT, I've just been hunting for quality of life issues on Bot servers. Never having XTargets but having hate was a annoying because real players don't function that way.

So this may not be the best solution. I'm just trying to contribute some quality of life things that I can code in, that don't fundamentally change anything too much.

spdkils:
Also if I'm understand you correctly, you would prefer to see some way of just having the bot add the XTarget to the client that owns it, in some bot function, rather than dealing with it in the attack script. So it functions more like the client() where it just merges it's autohaters into the xtarget list, as opposed to hacking it into the attack process..

mackal:
It actually shouldn't remove you, but that was done for simplicity. It should clear xtargets when you stand up if you no longer have aggro. I forget the exact behavior on live.

PR #613, opened 2017-04-23
Fix effects over int16 on items and fix aeduration spells

Natedog2012:


Natedog2012:
AE duration spells never checked range on item clicks.

Akkadius:
Looks good to me quick glance - maybe @mackal can take a quick glance at it as well to see if he sees an issues

mackal:
As long as they're right in the packet translators it should be fine. I know ages ago I was looking at this and was running into issues with clients displaying unknown spells due to a bunch of casting happening all over the place :P

PR #626, opened 2017-06-12
Add CrossZoneSetEntityVariableByClientName

Natedog2012:
Allow entity variables to be sent across zones to other clients.


mackal:
Lua?

Natedog2012:
use the lua format and added it to that as well

PR #628, opened 2017-07-05
Allow pets to use the better of db/oocregen if oocregen rule on.

noudess:
This allows pets to use the better value of oocregen or their db setting for hp_regen_rate (when out of combat).

This also allows you turn turn hp_regen way down in the db, and still get a speedy pet regen when out
of combat.

### DIFF
--- a/zone/npc.cpp
+++ b/zone/npc.cpp
@@ -628,11 +628,11 @@ bool NPC::Process()
 			OOCRegen += GetMaxHP() * oocregen / 100;
 		}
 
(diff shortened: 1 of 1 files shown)


mackal:
Can you remove your name from the comment? And the other persons :P

noudess:
Done.  Was just mimicking the other person.

noudess:
After about 8 hours of playtime with my mage I love this fix.  Allows me to tweak both in combat and ooc regen for pets.

noudess:
Still looks great here.

PR #631, opened 2017-07-10
Dockerfile

joshuacox:
Here is my best guess on where you should start.  Next step would be to figure out your start command (must not daemonize, it must stay open and print logs to stdout by convention, there are ways around that though), and the mysql stuff needs to be eliminated from the install as we'll link or otherwise use an external DB there.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM debian:stretch
+
+ENV EQEMU_DOCKER=20170710 \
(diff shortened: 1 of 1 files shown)


xackery:
If docker is considered, there's a repo i leverage for shits and giggles based on my private repo that may have some stuff tackled if looking for insight @joshuacox https://github.com/xackery/eqdocker

joshuacox:
thanks!

Issue #630, opened 2017-07-10
JSON eqemu_config conversion


xackery:
* Add support for eqemu_config using common/json_config code
* Find/Make a binary that converts eqemu_config to json

xackery:
* Locked element will be parsed and convert to Locked = "", which isn't the ideal solution. Needs more logic to convert cleaner.

xackery:
The perl script should do the following:
* [ ] Verify eqemu_config.json does not exist.
* [ ] Download and execute the xmltojson binary to convert eqemu_config.xml to eqemu_config.json.
* [ ] Capture any signals (1) from the executed xmltojson binary, and assume if non-zero, failure (error is outputted to stdout).
* [ ] Optionally delete eqemu_config.xml
* [ ] Parse eqemu_config.json once it exists, and use it for future configuration runs.

Akkadius:
Will be bringing this over into master in the near future

PR #633, opened 2017-07-16
Eqemu config json

xackery:
This is based on #630 
I do not suggest merging this PR until the perl changes are in place, or else owners will need to run the xmltojson-<os> binaries to convert their config, it needs to be ran in the same directory as eqemu_config.xml.

The perl script should do the following:
* [ ] Verify eqemu_config.json does not exist.
* [ ] Download and execute the xmltojson binary to convert eqemu_config.xml to eqemu_config.json.
* [ ] Capture any signals (1) from the executed xmltojson binary, and assume if non-zero, failure (error is outputted to stdout).
* [ ] Optionally delete eqemu_config.xml
* [ ] Parse eqemu_config.json once it exists, and use it for future configuration runs.


xackery:
Closing, and PR to a new branch.

PR #639, opened 2017-07-21
Mob::ResistSpell on fear spells now passes UseChaMod=true

noudess:
The function ResistSpell itself mentions fear and how these get applied.


Akkadius:
Need to cleanup these commits

PR #640, opened 2017-07-21
Repair fear pathing issues with Z and FEar resists now apply cha adjustments as they should.

noudess:
Two quick bug fixes.


noudess:
This causes issues.  Can revert for now or wait a couple hours while I test diff approach.

PR #643, opened 2017-07-22
My "fix" for the fear spell getting dropped caused many fleeing errors

noudess:
Undid all changes except the one about dropping fear when no pathing could be found.


mackal:
I think the reason it was fading it was because it assumed it was stuck which could be exploited like a stun.

PR #650, opened 2017-07-31
Add fix for GetGroundZ and fix some constant references.

noudess:
When GetGroundZ used GetBestZ, I have logs where 1028 turned into -489...
All the way down to the map grid.

Using ClosestZ grabs the more sane value.

### DIFF
--- a/zone/command.cpp
+++ b/zone/command.cpp
@@ -7257,7 +7257,7 @@ void command_bestz(Client *c, const Seperator *sep) {
 
 		float best_z = zone->zonemap->FindBestZ(me, &hit);
 
-		if (best_z != -999999)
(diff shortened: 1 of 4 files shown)


noudess:
May want to hold off on merge.  This may cauise a need to tweak FixZ().  I am looking at it.

Akkadius:
I didn't jump on this because its adjusting a core functionality of FixZ and there's a ton of scenarios that need to be accounted for and tested

noudess:
Been running ok for us so far.  I'm dropping the offset_z to 0 (locally) to see if it just works without now.

noudess:
Reduced this to constant fixes as Akka merged in the rest.

PR #659, opened 2017-09-12
if() fixes

Congdar:
This is my first try at git fork/pull request so I thought I would start small.

The first if() fix calls boolean IsAugmentSlotAvailable() and compares it to not be equal to -1
This case will always be true since the int conversion for bool is 1 or 0 so why even make the call?
I'm hoping my change is what was really intended.

The second if() fix does not do what the comment says.
If the call to GetGroup() returns 0xFFFFFFFF (not in a group) this if will not return; like wanted.

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -3065,7 +3065,7 @@ void Client::Handle_OP_AugmentItem(const EQApplicationPacket *app)
 			}
 			else
 			{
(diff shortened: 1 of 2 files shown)


mackal:
Parentheses aren't needed in the change in raids.cpp

Congdar:
Not needed for the compiler due to operator precedence but nice for showing humans full intent, but the ! was the issue.

Congdar:
So... what are the next steps to get this in?

Akkadius:
Can we close this PR and then re-submit it with the PR intentions in the commit description? IE: What exactly is this fixing. Thank you

PR #662, opened 2017-09-24
Rule to dictate if you can teach yourself lang.

noudess:
The current code allows teaching yourself languages that are at least a 1 skill.

Some of us believe that in the past, you needed someone higher than you to teach.

This new rule defaults to current behaviour.

Set to false, it requires someone else to teach.

+RULE_BOOL(Skills, SelfLanguageLearning, true)


joligario:
I seem to remember that as long as you were equal skill to each other, you could train each other by "leapfrogging" by 1.

mackal:
Live works like if this rule is off :P Can skill yourself up as long as you're in a group. No need to leap frogging or anything.

PR #669, opened 2017-10-28
Added a #petname command for renaming only pets

noudess:
For GMs that want to give players the ability to name only their own pets.

### DIFF
--- a/zone/command.cpp
+++ b/zone/command.cpp
@@ -375,6 +375,7 @@ int command_init(void)
 		command_add("task", "(subcommand) - Task system commands",  150, command_task) ||
 		command_add("tattoo", "- Change the tattoo of your target (Drakkin Only)", 80, command_tattoo) ||
 		command_add("tempname", "[newname] - Temporarily renames your target. Leave name blank to restore the original name.", 100, command_tempname) ||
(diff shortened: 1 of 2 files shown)


mackal:
They actually have pet name store stuff now. Might be better to kind of give better support for that (stored etc) than just temp naming.

noudess:
So I copied the #tempname, but honestly it is not temporary unless you issue the command again with blank.  I just stuck with what the old function used for a comment and behavior, except that it only works on the clients pet.

noudess:
Well if you mean a new pet will inherrit the name, then yes, it is temporary.  But I'd want pets to have unique names as I go.

mackal:
Live allows you to buy a potion to give your pet a name. They will always spawn with this name. I think it would better if this emulated that. Then we can get around to adding support for the official stuff (I'm not sure which clients support it, I'm pretty sure RoF2 does)

noudess:
I see your point.  

I do want this on my server regardless, as I just want to give my players this.  So I guess either commit it as a second way to go,  or reject it and I'll keep my command local.

chris-statzer:
I would also like to use this functionality. As long at the level needed to used it is high and it had to be enabled per server I don't see a problem here right?

noudess:
I kinda agree with @chris-statzer seems like adding this and doing what @mackal suggests works.  Some servers may want one vs the other and it is status access controlled.  Is ok?

PR #673, opened 2017-11-22
Upgrade weaponless NPC offense

noudess:
NPCs greater than level 39 are using H2H for melee if they don't have a weapon.  With the melee system
using weapon skill as the base for offense, this means level 50 weaponless warriors were starting with a base offence (from skill) of 100 (or whatever your db skillcap is for H2H.

This code adjusts melee offense up to the npcs best melee skill if they are weaponless, but uses the weapon skill if they have one.  I believe we want all level 50s warriors hitting the same, all other things being equal.


mackal:
Ideally NPCs would have all their skills set rather than just magic.

noudess:
I believe all this change does is use the NPCs best skill, if the NPC has no weapon - to keep H2H skillcaps from dwarfing a mass of Mobs.  At least that's what I intended, nothing further like giving them fake values.

mackal:
Yep. I'm just saying ideally we would set the skill for every skill on every NPC without hitting skill caps, which would remove the need for this :P

But until a time we can set an NPC's skill, this should work.

PR #679, opened 2017-12-21
Fixed Ubuntu 17.10 compile

SecretsOTheP:
This is a compile fix so that fmt can be build on Ubuntu 17.10.

Other notes: uuid-dev is also not included in Ubuntu 17.10 by default. You can apt-get install uuid-dev to get the required libraries.


mackal:
This issue is fixed by updating fmt

SecretsOTheP:
I agree it'd be a better fix to update it, but this is more a hacky fix for now.

PR #683, opened 2017-12-26
Adjust xp loss on death if using the level_based_mods.

noudess:
+		// Death in levels with xp_mod (such as hell levels) was resulting
+		// in losing more that appropriate since the loss was the same but
+		// getting it back would take way longer.  This makes the death the
+		// same amount of time to recover.  Will also lose more if level is
+		// granting a bonus.


noudess:
To be a bit more clear.  With this turned on, all deaths should take the same amount of grinding to recover from.  Before this, with level based mods, deaths during an xp adjusted level were either too hard or too easy to recover based on the unchanging-xp-lost.

Issue #623, opened 2017-06-08
Sun's Corona requires a Fire Beetle Eye to cast


mackal:
My druid can cast this spell line just fine without one in her inventory, but live doesn't let me without having one!

daerath:
Can you re-confirm?  Alla's lists the Fire Beetle Eye as a Focus and not as a Reagent.

mackal:
Alla is silly is my guess. Looking at PEQ editor Sun's Corona has Fire Beetle Eye in "No Expend Reagent 1" this means it requires a Fire Beetle Eye present to cast (which isn't consumed)

You need the item in your inventory, not bank :P

daerath:
Gotcha.  Does it get used up on live?

mackal:
No, you just need it

daerath:
I'm confused then.  When the ID for a fire beetle eye is added to the components1 value for that spell, attempts to cast it fail if you don't have it in your inventory.  If it is also listed as a noexpendreagent1 it doesn't get consumed.

Is the real issue that the emu should also be checking the noexpendreagents values (and it isn't)?

mackal:
Because that's not how it's suppose to work. Without changing the spell data, it should work as I said.

I haven't actually look far enough into fixing, why I opened the issue.

daerath:
Ok.  I'll work under the assumptuon that the spell data is correct.  I'll look at the code tomorrow.

daerath:
I see the issue in the code and also confirmed this by testing Banshee Aura.  It also requires a component (Pearl, not consumed), but the Emu doesn't care if you have one or not.  Data looks the same (-1s for components, but the ID for Pearl is listed as a non-consumable).

mackal:
Make sure you don't break bard songs! (main reason I haven't bothered)

mackal:
Okay, so it looks like from the details I can gleam from the client, bards have special behavior here. The NOEXPENDREAGENT for bards is kind of just convention and doesn't act the same for other clients :P (they're actually checking the spell's skill to get the item type needed)

daerath:
Yup, I kept the bard bits mostly the same.  I made one minor change (removed it from a loop that made no sense because it just did the exact same check each iteration).  Going to do some playtesting over the next couple of days to make sure it works.

daerath:
Withdrew pull request.

PR #685, opened 2018-01-07
* Suns Corona requires a fire beetle eye (not consumed).

daerath:
#623


mackal:
The data is right, the code is wrong.

PR #694, opened 2018-01-13
NPCs were waiting very long after spawn to cast

noudess:
This makes newly spawned NPCs have a chance to cast.  It augments my previous fix to finalize the fix.

### DIFF
--- a/zone/mob_ai.cpp
+++ b/zone/mob_ai.cpp
@@ -493,6 +493,11 @@ void Client::AI_Start(uint32 iMoveDelay) {
 	SetFeigned(false);
 }
 
+void NPC::AI_RestartSpellTimer() {
+	Log(Logs::Moderate, Logs::Spells, "Triggering RestartSpellTimer :: Mob %s - %u", this->GetCleanName(), 300);
+	AIautocastspell_timer->Start(300, false);
+}
+
 void NPC::AI_Start(uint32 iMoveDelay) {
 	Mob::AI_Start(iMoveDelay);
 	if (!pAIControlled)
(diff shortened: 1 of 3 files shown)


zone/mob_ai.cpp
```diff
@@ -1762,6 +1768,10 @@ void Mob::AI_Event_Engaged(Mob* attacker, bool iYellForHelp) {
 
 	SetAppearance(eaStanding);
 
+	// Idle timer is huge, gives NPCs an early disadvantage when 1st entering
+	// combat.
+	this->CastToNPC()->AI_RestartSpellTimer();
```
Akkadius: Just do the timer inline here instead of using a function, it doesn't have any params anyways

noudess: Make it an inline function in the header? @Akkadius

noudess: Sorry, not experienced with inline - seen it in headers ,,,  Would want it in header in case needs to be called elsewhere?

noudess: Oh you mean just use AIautocastspell_timer->Start(300, false); inline.

I think we want the logging as it happens for mobs?  Also since it is already a function for mobs - seems more consistent?

Also the function it is being called in does not have access to the timer if I recall, without changes.

Akkadius: You don't need to CastToNPC() because you're already in NPC. Since this is a NPC class timer - you can call AIautocastspell_timer->Start(300, false); inline where the function call was made

noudess: This function is mob:: not npc.  I think it is not accessable.

Akkadius: image

It is confusing with the diff - an AI function isn't typically 'Mob' you can CastToNPC inside the mob function with an if NPC check before the timer. This code needs to be cleaned up in general

PR #696, opened 2018-01-14
Bug fix/suns corona required component

daerath:
#623 #685 

Component checking logic was not properly handling noexpendreagents.  They were only being used if the spell also had one or more components as well.

Fixed that problem and refactored to make the code more readable.  It also skips component checks if a spell has no components.

Removed a redundant log statement (informational only) if a bard song had a non-instrument component.  The song ignores the component, so the log was more of an FYI.  If that sort of check really matters it should be done at boot-up, not every time someone sings a bard song with a non-instrument component.


mackal:
Move AreReagentsRequired/AreNonExpendableReagentsRequired to clients and make them take a spell ID or a const references to a SPDat_Spell_Struct.

I don't like passing a bunch of pointers around like that :P

daerath:
I'll move those two into client and pass the spellId instead of explicitly passing components and noexpendable arrays

mackal:
Can probably refactor out even the whole item consuming bit into a function with these too. (I think that's a reason you have an out array)

daerath:
I can move the hasReagents || hasNonExpendableReagents code, no problem

daerath:
Ok, moved some code around.  Cleaner now than it was before.

Akkadius:
I'm going to let @mackal check off on this one for merge

Issue #610, opened 2017-04-17
2 slot sewing kit can make 4 item combine (generic issue also true)


noudess:
Looks  like the code does not enforce container size when all the components do not fit.

noudess:
To give better detail.  This happens when you click a recipe and the GUI loads the items in.  It loads more items into the container (virtually I guess) than it can hold.

noudess:
Take a small sewing kit and making a gnoll fur patch.

daerath:
HandleAutoCombine does not perform any checks against container size or # of components required in the recipe.  However, I don't know if that matters on live so I'm going to check that.

mackal:
I'm pretty sure live doesn't define recipes to use generic typed containers either. It's just an item ID of a container (I do think they do a size check anyways)

noudess:
Second check unneeded as they can't put the items in without the recipe since they won't fit?

daerath:
The second check is more to catch people packet hacking

noudess:
ah thx

PR #702, opened 2018-01-21
Bug fix/recipes ignore container size

daerath:
#610


mackal:
database.GetItem() may return a nullptr.

daerath:
Updated.  If a nullptr is returned, the ID provided is logged as an error and the method exits.

mackal:
Hmm, would it be better to default to 10?

daerath:
My only concern about defaulting to 10 when the item ID in the packet cannot be found is that I'm not sure how that could happen short of a hacked packet.  That's based on my (possibly incorrect) assumption that the server sends the item IDs to the client, and then the client sends them back and that's where the data comes from in that method.

So I'm not opposed, but I'm not sure of the data flow in this case.

daerath:
After we spoke, this now defaults to 10 if the container isn't found.

mackal:
@Akkadius can you merge this if the SQL looks good to you?

PR #709, opened 2018-02-12
Scaled AC and ATK if npc is scaling.

noudess:
On mobs where max_level is set,  AC and ATK were not being scaled.  This makes a large difference.  I had mobs that were level 7, max_level 9 and mobs that were just 9.  The level 9 mobs were a challenge for PCs around their level.  PCs could take 3 level 7 mobs scaled to 9 easily.

This makes them all the same based on scaled level.


KinglyKrab:
Looks good to me, going to merge.

PR #712, opened 2018-02-20
Fix for experiance loss past certain levels

dencelle:
Max level through experience is 112

### DIFF
--- a/common/string_util.cpp
+++ b/common/string_util.cpp
@@ -432,6 +432,12 @@ const char *ConvertArray(int input, char *returnchar)
 	return returnchar;
 }
 
+const char *ConvertArrayU(uint32 input, char *returnchar)
+{
+	sprintf(returnchar, "%u", input);
+	return returnchar;
+}
+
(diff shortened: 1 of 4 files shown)


mackal:
Well, if sending this string with the data correctly doesn't work, then maybe we shouldn't support higher levels than we should.

dencelle:
this is kind of future proofing, HoT+ on live were level 90+
currently if you try to level past 86ish you will get reset back to 86 due to a overflow
as for the Message_StringID change, i'm not 100% sure why you'd be mad about changing that...

dencelle:
canceling out, seems more or less i'm getting a huge fu... not sure what person's toe i stepped on but eh... i can easily just keep to my self and keep my contributions private

mackal:
Sorry I have not had the time to sit down and look at the reason for the code change. I just didn't like moving away from using the StringID. The rest of the changes will take much longer to analyze. Sorry no one could accommodate you soon enough.

PR #714, opened 2018-02-23
Added rule to remove pet reagent cost

xackery:
A simple rule, enabled by default, where when disabled reagent costs check if the spell has SE_SummonPet as an effect in it. If it does, it bypasses the normal logic (for malachite, bone chips, tiny daggers, etc)


mackal:
You could edit your spell file ... but whatever.

PR #721, opened 2018-03-03
Make old style rogue poison closer to live.

noudess:
There were a few things wrong with old style rogue APPLY poison.

1) The failure rate was way off.  It is very seldom for APPLY poison to fail.  I ran some tests on live and got these results:

A level 18 rogue (1st level to get apply poison) starting at skill 1, was 28/30 attempts.  The first fail was at apply skill 4 after 7 successes.  This was for any poison he was high enough to use.  A level 68 rogue was 21/21 on a poison that was above both her make and apply skill.  No AAs.

2) Skillups on live can occur, even on poisons too high to apply.   They get eaten, and nothing applies, but skillups are possible.  Verified this via testing.

3) The poison always gets eaten on live, success or failure.  We did not eat them on certain conditions.

4) Our code was allowing the applying of poisons with a required level above our level.  Fixed those to fail as they do on live,

The one area I DID NOT fix.  On live, old school poisons are 1 shot.  On 1st hit, they cast a spell of the poison and they are gone.  1 shot.  Done.  Our code does not do this yet, but I like how we do it, and I believe we match P1999.  I am leaving it as a proc which you lose on zoning.  If someone wants to be 100% old world poison compliant (still acts like this today) then we should add code to the logic to do the one-time fire and maybe make the existing logic optional via a rule.

I have a brief excel table of data to backup what I found and changed if anyone wants it.  I suspect this is just going unused for the most part, which is a shame.

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -2903,10 +2903,13 @@ void Client::Handle_OP_ApplyPoison(const EQApplicationPacket *app)
 			// Poisons that use this skill (old world poisons) almost
 			// never fail to apply.  I did 25 applies of a trivial 120+
 			// poison with an apply skill of 48 and they all worked.
+			// Also did 25 straight poisons at apply skill 248 for very
+			// high end and they never failed.
+			// Apply poison ranging from 1-9, 28/30 worked for a level 18..
 			// Poisons that don't proc until a level higher than the
-			// rogue simply won't apply at all, as done above now.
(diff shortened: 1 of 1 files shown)


mackal:
Can you remove #define MIN(x,y) ((x)<y?(x):(y))? It's not needed!

noudess:
Ah, yes.  Not using it anymore.

PR #731, opened 2018-05-01
Update entity.cpp

thalic:
Fix underwater boats for Maidens_Voyage, Icebreaker, Bloated_Belly, Sea_King, SirensBane and small boats


mackal:
Can you use defines in common/races.h? (Add if needed)

fryguy503:
After playing with this, it seems like we might want to check if the npc is in/on the water map. Currently pathing, spawning and such are broken in Water based areas as it throws the NPC to the floor. I fixed this in PoWater by excluding the zone from FixZ but this will not solve zones like Sirens that are half/half water/land.

thalic:
Unfortunately at the moment no time to improve the code

PR #733, opened 2018-05-11
It is not possible anymore to start sneaking underwater/lava.

fryguy503:


mackal:
20030710.html
98:<li>Sneak - Sneak will now work under water.</li>

mackal:
This could be rule based, but clients expect to be able to sneak underwater so we should have it off by default.

PR #736, opened 2018-06-01
VS 2017 Community wants namespaces of 3 common/compression functions s…

af4t:
…pecified in calls in zone/map.cpp.  The functions are defined in common/compression as being in namespace EQEmu:
InflateData
EstimateDeflateBuffer
DeflateData

There's likely a better way to do this, but this was easy

### DIFF
--- a/zone/map.cpp
+++ b/zone/map.cpp
@@ -996,7 +996,7 @@ bool Map::LoadMMF(const std::string& map_file_name, bool force_mmf_overwrite)
 	fclose(f);
 
 	std::vector<char> rm_buffer(rm_buffer_size);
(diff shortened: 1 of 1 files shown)


joligario:
I'm using VS2017 Community and I'm not having any issues. Is there a specific scenario going on here?

Uleat:
This is with Map MMF files enabled and not the 'standard' build.

Compile without Map MMF support (cmake) for now and I'll bring it up again.

af4t:
"This is with Map MMF files enabled and not the 'standard' build." ... missed seeing that item defaulted to checked.

Closing this out myself till Uleat can get back to it.

PR #744, opened 2018-07-19
Faction War Fixes

fryguy503:
Help prevent Faction Wars from Charmed mobs.


zone/mob_ai.cpp
```diff
@@ -1184,7 +1184,11 @@ void Mob::AI_Process() {
 		{
 			if (AI_target_check_timer->Check())
 			{
-				if (IsFocused()) {
+				if (IsNPC() && (!IsPet() || (HasOwner() && GetOwner()->IsNPC())) && !CastToNPC()->WillAggroNPCs()) {
```
Akkadius: Indentation seems off here?

PR #747, opened 2018-07-21
Fixing #resetaa to actually remove aa's

fryguy503:


mackal:
resetaa shouldn't reset all AAs. Stuff like the MPG group trial AA, DoN progression, etc shouldn't be reset.

fryguy503:
That is why you get paid the big bucks @mackal I didn't even think of those. I will close this and add an exclusion system.

PR #748, opened 2018-07-21
Pet Feign Death functionality

fryguy503:


mackal:
GetAA() shouldn't be used, SPA 281 is the percent chance. Use it. Also they don't use clean name :P

fryguy503:
Will work on revamping it to Mob

PR #753, opened 2018-07-21
Added random graveyard corpse placement within +/- 20 units on X/Y ax…

fryguy503:
…is to help spread corpses out.


mackal:
Please don't use rand()

PR #757, opened 2018-07-22
Purfy Body Resurrection Effects fix

fryguy503:
Purify Body was removing Resurrection Effects and it should not.
Added a !IsResurrectionEffects check to the logic to make sure Purify Body does not remove Resurrection Effects.


mackal:
This should probably be checking no dispel flag instead.

PR #758, opened 2018-07-22
Fixing zonewide /assist bug

fryguy503:
Prevents players from /assist <target> to see if a mob is up due to error or no error messages.


fryguy503:
I will fix the formatting and resubmit

PR #760, opened 2018-07-22
untargetable npcs like horses shouldn't get hit by AoEs

fryguy503:


mackal:
Auras get hit by AEs :P

PR #769, opened 2018-08-25
Prevent Gate if near bind location (Rule) and Heal on Gate (Rule)

fryguy503:
RULE_INT(NPC, NPCGatePercent, 5) // % at which the NPC Will attempt to gate at.
RULE_BOOL(NPC, NPCGateNearBind, false) // Will NPC attempt to gate when near bind location?
RULE_INT(NPC, NPCGateDistanceBind, 75) // Distance from bind before NPC will attempt to gate
RULE_BOOL(NPC, NPCHealOnGate, true) // Will the NPC Heal on Gate.
RULE_REAL(NPC, NPCHealOnGateAmount, 25) // How much the npc will heal on gate if enabled.


mackal:
Can you restore the behavior from #766 ?

PR #770, opened 2018-08-28
Removed unneeded calcs from FixZ if Client()

noudess:
Just pushed the variables and the call to GetFixedZ() inside the conditional so we only do it when needed.


Akkadius:
Addressed in https://github.com/EQEmu/Server/commit/b92e83a465639126383c4217eac3c050cfedbf0e

PR #776, opened 2018-09-17
Fun #nodelay [on/off] [milliseconds] gm command

chazix:
Works for yourself or your target, clients can be persistent with the nodelay smallint field, also works on npcs

#nodelay on ms => turns nodelay on and sets milliseconds between [0-65535], value < 0 indicates off
#nodelay off => turns nodelay off sets milliseconds to -1
For persistent usage between zoning and logging in you'll need to add: a nodelay smallint field with default value -1 in the account table

[09/17/18] => Added in the SQL required and manifest items

### DIFF
--- a/common/memory_buffer.h
+++ b/common/memory_buffer.h
@@ -124,7 +124,7 @@ namespace EQEmu
 
 	class OutBuffer : public std::stringstream {
 	public:
-		inline size_t size() { return tellp(); }
+		inline size_t size() { return static_cast<size_t>(tellp()); }
 		void overwrite(OutBuffer::pos_type position, const char *_Str, std::streamsize _Count);
 		uchar* detach();
 	};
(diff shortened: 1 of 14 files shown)


Akkadius:
Missing `DB_VERSION`, see https://github.com/EQEmu/Server/wiki/Database-Schema-Migrations#adding-new-required-schema-migrations

Also, if we're going to be adding a new field to `character_data` I know in context it is referring to your nodelay functionality in this sense but it makes more sense to keep it contextual to a simple `attack_timer`. Which can still have considerable delay as acceptable values and would rather just name the commands as such as well ie #set_attack_timer and then keep your #nodelay command but also just have it set literally no delay

Once those you roll those changes it should be good for merge

chazix:
sounds good, thanks :) I'll add those in this week after work

PR #780, opened 2018-10-08
Pets always try and update owner, if hp changed

noudess:
SendHPUpdate never got called for non-engaged pets.  This could lead to issues between mobs and when resting.

### DIFF
--- a/zone/npc.cpp
+++ b/zone/npc.cpp
@@ -747,8 +747,10 @@ bool NPC::Process()
 
 	/**
 	 * Send HP updates when engaged
+	 * Pets always - master wants to see the healing
 	 */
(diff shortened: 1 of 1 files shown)


mackal:
This will still have issues with with resting NPCs. NPCs do have HP regen so ... but maybe targetting will get it updated too anyways or being on xtargets. Would need to do more investigating.

noudess:
Yeah I agree about resting NPCs.  I thought about that, but figured we'd save alot of computation by not adding the NPCS and the pet info is much more likely to be relevant.

Akkadius:
Fixed in https://github.com/EQEmu/Server/commit/72bf4b7e3cb17f6fbf17f17f166708e3ba57a504

PR #784, opened 2018-10-20
Fear pathing fixups, random path rewrite, and nektulos / general roam…

SecretsOTheP:
Fear pathing fixups - works a hell of a lot more reliably, still not perfect. Random path rewrite, and nektulos / general roambox fixups where best Z axis is below the ground. Use nearest Z instead.

### DIFF
--- a/zone/fearpath.cpp
+++ b/zone/fearpath.cpp
@@ -157,41 +157,19 @@ void Mob::ProcessFlee()
 
 void Mob::CalculateNewFearpoint() {
 	if (RuleB(Pathing, Fear) && zone->pathing) {
-		auto Node = zone->pathing->GetRandomLocation();
+		auto Node = zone->pathing->GetRandomLocation(glm::vec3(GetX(), GetY(), GetZ()));
 		if (Node.x != 0.0f || Node.y != 0.0f || Node.z != 0.0f) {
-
-			++Node.z;
 			m_FearWalkTarget  = Node;
 			currently_fleeing = true;
 
 			return;
 		}
 
+		Stun(6000);
+
 		Log(Logs::Detail,
 			Logs::Pathing,
 			"No path found to selected node. Falling through to old fear point selection.");
 	}
-
-	int   loop = 0;
-	float ranx, rany, ranz;
-
-	currently_fleeing = true;
-	while (loop < 100) //Max 100 tries
-	{
-		int ran = 250 - (loop * 2);
-		loop++;
-		ranx    = GetX() + zone->random.Int(0, ran - 1) - zone->random.Int(0, ran - 1);
-		rany    = GetY() + zone->random.Int(0, ran - 1) - zone->random.Int(0, ran - 1);
-		ranz    = FindGroundZ(ranx, rany);
-		if (ranz == BEST_Z_INVALID)
-			continue;
-		float fdist = ranz - GetZ();
-		if (fdist >= -12 && fdist <= 12 && CheckCoordLosNoZLeaps(GetX(), GetY(), GetZ(), ranx, rany, ranz)) {
-			break;
-		}
-	}
-
-	if (currently_fleeing)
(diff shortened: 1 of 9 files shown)


SecretsOTheP:
Linux portability issue fixed.

zone/mob_ai.cpp
```diff
@@ -1661,7 +1663,7 @@ void NPC::AI_DoMovement() {
 			destination.x = roambox_destination_x;
 			destination.y = roambox_destination_y;
 			destination.z = m_Position.z;
-			roambox_destination_z = GetFixedZ(destination) + this->GetZOffset();
+			roambox_destination_z = zone->zonemap->FindClosestZ(destination, nullptr) +  this->GetZOffset();
```
Akkadius: Changes look good overall first glance, though I'm not sure why you swapped this particular line out since it effectively does the same thing, however it goes through some checks under GetFixedZ that we don't exactly want to ignore. What were you trying to fix with this?

zone/fearpath.cpp
```diff
@@ -157,41 +157,19 @@ void Mob::ProcessFlee()
 
 void Mob::CalculateNewFearpoint() {
 	if (RuleB(Pathing, Fear) && zone->pathing) {
-		auto Node = zone->pathing->GetRandomLocation();
+		auto Node = zone->pathing->GetRandomLocation(glm::vec3(GetX(), GetY(), GetZ()));
```
Akkadius: Do you know if this selects valid locations? We were going to eventually look into this to see if it is at least grabbing random points inside of geometry or on tiles

zone/mob_ai.cpp
SecretsOTheP: GetFixedZ chooses the best ground Z. Roamboxes are often used in zones with slopes and hills, some of the old world zones such as legacy (1.0) Nektulos actually have their 'ground' way below the world.

To illustrate the issue, here's a poor mspaint drawing:

The mob goes into the hill, and their BestZ location is actually /below/ the normal field. We typically use this in ourdoor zones, so BestZ is inappropriate to use in this case.

If we need a clone of GetClosestZ in Mob that does the other checks, I can do that, but I don't think it's a huge deal personally.

zone/fearpath.cpp
SecretsOTheP: Yes - I rewrote the function to get the closest point near the closest geometry near a location.
I've tested it in hellish indoor scenarios (gukbottom near evil eye) and open fields (northkarana) and it works great.

zone/mob_ai.cpp
Akkadius: I don't really want to clone functionality as it should all go through similar aggregate logic `GetFixedZ`. I'm looking at the code at the moment to review

If the current logic bound to lowest `ground` all the time then we'd have serious issues. The entire objective is to find the best Z relative to the current Z with a slight search offset so that should be the ground closest as you are suggesting and should be that way with what we already do

Both `FindClosestZ` and `FindBestZ` do the same "first step" raycast call with both having different secondary attempts. If both of these methods are being used properly with appropriate Z offsets we shouldn't see either of them hitting their second raycast attempt. The other main difference is that `FindClosestZ` will not immediately return out of the first "hit" results its finds which is weird and expensive

I'm not objecting this by any means its more of the intention of hitting one focal point of logic so we don't have fragmentation and our Z finding logic is consistent, that's all

I'd leave it for now but it makes me want to debug the difference between these two functions. I have other things in head of this before I can take a look. If you have time to sort it out that would be helpful but not expected

zone/fearpath.cpp
Akkadius: Sweet! I'm excited to see this

zone/mob_ai.cpp
SecretsOTheP: The difference is that it does a check to see the top or bottom Z is closer. This is so that the 'ground' z is not prioritized over the bottom z, which is what FindBestZ does.

image

PR #805, opened 2019-01-10
SaveCharacterTribute misses tribute id 0

prestanneth:
Aura of clarity has an id of 0, Never gets saved to db so after zoning tribute is not reapplied.


prestanneth:
Note have had a chat to Uleat, this ideally requires a db update at some point to change default value of id in character_tribute, as its currently defaulting to 0 also.

joligario:
Yeah that will be significant for db change. For simplicity, they should never have set an id to 0...

PR #806, opened 2019-01-10
Reimplemented pre-nerf Ranger stationary damage bonus.

KinglyKrab:
This will allow Developers to disable/enable the stationary requirement.


mackal:
The " && RuleB(Combat, ArcheryBonusRequiresStationary)" part is redundant, isn't it? Should function the same with just adding the or-not bit.

KinglyKrab:
Yeah, I can change it.

PR #817, opened 2019-01-25
Add RemovePermaProc to Perl API and RemovePermaProcFromWeapon to Mob

mthogan:


mthogan:
TravisCI failed due to Internal Server Error. Code works. Used in production on FlexQuest server.

PR #818, opened 2019-01-30
Properly disable IntAggroThreshold if UseLevelAggro is true

isenior:
The documentation for rule Aggro:UseLevelAggro in ruletypes.h states
"RULE_BOOL(Aggro, UseLevelAggro, true) // MinAggroLevel rule value+ and Undead will aggro regardless of level difference. (this will disabled Rule:IntAggroThreshold if set to true)"

This wasn't happening though. Since the conditions were &&'ed together, if any of them failed, it would cause us to fall into the "else" block containing the intelligence checks.

### DIFF
--- a/zone/aggro.cpp
+++ b/zone/aggro.cpp
@@ -320,36 +320,35 @@ bool Mob::CheckWillAggro(Mob *mob) {
 	int heroicCHA_mod = mob->itembonuses.HeroicCHA/25; // 800 Heroic CHA cap
 	if(heroicCHA_mod > THREATENLY_ARRGO_CHANCE)
 		heroicCHA_mod = THREATENLY_ARRGO_CHANCE;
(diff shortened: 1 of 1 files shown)


isenior:
Yeah, it kinda makes sense as it is. Like, it does the level checks, and then for any gray monsters below your always aggro threshold, it will do an intelligence check.

If that's desired behavior, then we could probably just remove  "(this will disabled Rule:IntAggroThreshold if set to true)" from ruletypes.h since it's not true and is. misleading to someone investigating aggro issues.

noudess:
Again, whatever you all agree on.  I agree comment should be removed if we leave it as is.

noudess:
So @isenior  , is removing the comment good enough?  If so, I suggest you resubmit this that way and we can get it merged.  If not, I defer to the devs.

PR #821, opened 2019-02-02
Remove boost

adam4813:
This PR removes the requirement for boost completely.

I do not have the proper ability to test this, so I will leave it here until I can.

The reason for the premature PR is to show the potential of removing boost, but at the cost of swapping luabind. This in itself isn't a bad thing as luabind hasn't had a solid commit in 4 years.

If this PR is a wont-fix that's fine, or if you can point me in the best direction to test I will happily do so.


mackal:
We use boost elsewhere too.

adam4813:
Drats that's what I had feared. Though I am up for removing it elsewhere to. None of the CMake generated server code used boost at all, just luabind.

Akkadius:
Removing boost is certainly a welcome change, unfortunately we don't have any automated tests to validate all of this so some manual testing is going to need to be done

mackal:
Removing boost just to remove boost isn't something I can get behind. Replacing what we're using with something better suited is something I can get behind.

Luabind probably should be replaced, but that's a different topic. And I know @KimLS has looked into replacing it with sol2, which is something I think is good idea.

adam4813:
Removing boost when it isn't actually used (except luabind) seems a prudent course of action for any project to reduce dependency and build times.

I am not active in the project and am very new to the project, so I do 100% understand that my action seems rash and I don't want to start an argument. I will accept what is decided by those who are active in development, but do look at the required code changes outside of the luabind. There was 1 include that needed to be removed. There was a few other lines that were needed to make the new luabind work.

However from my experience installing it on windows was a pain to get boost to actually download (thanks sourceforge).

KimLS:
I agree with removing boost if there's no issues otherwise (I've already removed it in the movement code).  That said this is a huge diff that's gonna be difficult to go over quickly.

adam4813:
99% of it is swapping luabind. The actual changes are like 15 lines to make it work with the new luabind fork.

Akkadius:
@adam4813 have you had further opportunity to do some testing with all of these changes?

PR #822, opened 2019-02-06
Updated profanity manager to only redact in 'public' channels

Uleat:
** Hold for discussion **

### DIFF
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -896,10 +896,6 @@ void Client::ChannelMessageReceived(uint8 chan_num, uint8 language, uint8 lang_s
 		language = 0; // No need for language when drunk
 	}
 
-	// Censor the message
-	if (EQEmu::ProfanityManager::IsCensorshipActive() && (chan_num != 8))
-		EQEmu::ProfanityManager::RedactMessage(message);
-
 	switch(chan_num)
 	{
 	case 0: { /* Guild Chat */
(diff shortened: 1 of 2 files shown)


Uleat:
Bumping for conversation..

zone/client.cpp
```diff
@@ -965,16 +964,22 @@ void Client::ChannelMessageReceived(uint8 chan_num, uint8 language, uint8 lang_s
 				}
 			}
 
+			if (EQEmu::ProfanityManager::IsCensorshipActive())
+				EQEmu::ProfanityManager::RedactMessage(message);
+
 			if (!worldserver.SendChannelMessage(this, 0, 4, 0, language, message))
-			Message(0, "Error: World server disconnected");
+				Message(0, "Error: World server disconnected");
```
dencelle: this should be using Chat::White instead of 0, new standard

Uleat:
:P 

Yeah, this will definitely need to be updated before a merge.

Conversation-wise, I'd like to see input on the actual principal of redacting speech and what other features would be viable in this implementation.

I'm thinking a client notice (message) upon logging in if the system is active. Kinda like a disclaimer/rule to keep the admin (and us) out of trouble.

dencelle:
that seems like a good idea, live does this in a slightly different way though, instead of redacting the message they "blur" it out (random string of letters that make no sense)... instead of redacting, maybe we should use that method? or maybe when it does get redacted, tell the user that sent the message that it was redacted via a message?

Akkadius:
This has conflicts that need to be settled

Akkadius:
In general, this PR looks fine. I'm not sure why we need to prefix `EQEmu::` with every call to this service

`ProfanityManager` is enough on its own to declare its responsibility, and the whole project is EQEmu scoped

Some conflicts would need to be resolved in order to have this be mergeable

cc @Uleat

Akkadius:
Closing this for now until @Uleat wants to revive this PR

PR #824, opened 2019-02-07
Reference the defender melee mitigation, not attacker

briankinney:
This function would cause defender's spell-based MeleeMitigationEffect to increase damage dealt and the attacker's item and aa-based MeleeMitigationEffect to increase damage dealt. New code should cause all three forms of defender MeleeMitigationEffect to reduce damage dealt.


mackal:
Spell bonus is 100% a negative number to reduce damage (so need to do addition)

mackal:
You are right it needs to reference the defender in the other cases, luckily, no AAs use that SPA so this never ran into issues (and the itembonus.MeleeMitigationEffect is never set in ItemBonuses so it's always 0)

PR #830, opened 2019-03-12
Adding title and suffix commands to bots

prestanneth:
Making use of the currently unused title and suffix fields in botdata table.
^bottitle
^botsuffix


KinglyKrab:
Need to add a required SQL to add the “title” and “suffix” columns to the bot table.

prestanneth:
The `bots` table hasn't been used for some years.
ULeat changed the schema back in 2015 to align more with the player profile one.
The new `bot_data` table contains these title/suffix fields but they are unused. 
Selecting bots menu option from akkas script emu updater thing will auto change the table schema.
https://github.com/EQEmu/Server/blob/master/utils/sql/git/bots/required/2015_09_30_bots.sql

info:
http://www.eqemulator.org/forums/showthread.php?t=40091

schema changes from *_old-> new: 
http://wiki.eqemulator.org/i?Module=Pastebin&Paste=TtvuBFHN

KinglyKrab:
Ah okay, wasn’t aware the table had the columns.

prestanneth:
Closing this before it gets merged as have rewrote using char arrays instead with some better management ,along with adding surname changes, list by account with owner, and saylinks for clicky spawning of "char owned" bots via botlist.

PR #832, opened 2019-03-25
Added model field to npc_types to allow using a diff graphic without losing race

noudess:
There are a few places in the items db, and other places where race is a factor.  Over time, the race field has evolved and sometimes new values are used for old races, like gnoll and new gnoll.  If we change an  npc to the new model, we lose the race effect (like gnollbane).

This PR adds a new field to npc_types which can  remain completely unused, but if you want to use an alternate model to an npc, putting that model # in the new field will do that, and impact only the client representation, not anything else like faction or bane,

I also greatly simplified the logic inside send_illusion.

This also fixes a bug where ignite-bones, a necro illusion spell changes an NPCs texture.  On death, the NPC chnages from skeleton back into original race, but the wrong texture.  This is because aside from race and gender, all other npc fields were overwritten and lost.


zone/mob.cpp
```diff
@@ -1692,151 +1694,77 @@ void Mob::SendIllusionPacket(
...
+		new_hairstyle = hairstyle		= CastToClient()->GetBaseHairStyle();
+		new_luclinface = luclinface		= CastToClient()->GetBaseFace();
+		new_beard = beard				= CastToClient()->GetBaseBeard();
+		new_aa_title = aa_title			= 0xFF;
+		new_drakkin_heritage 			= drakkin_heritage	
+										= CastToClient()->GetBaseHeritage();
+		new_drakkin_tattoo = 			drakkin_tattoo	
+										= CastToClient()->GetBaseTattoo();
+		new_drakkin_details 			= drakkin_details	
+										= CastToClient()->GetBaseDetails();
```
Akkadius: Formatting is crazy here

zone/mob.cpp
```diff
@@ -1692,151 +1694,77 @@ void Mob::SendIllusionPacket(
...
+	new_eyecolor2 = (in_eyecolor2 == 0xFF) ? GetEyeColor2() : in_eyecolor2;
+	new_hairstyle = (in_hairstyle == 0xFF) ? GetHairStyle() : in_hairstyle;
+	new_luclinface = (in_luclinface == 0xFF) ? GetLuclinFace() : in_luclinface;
+	new_beard = (in_beard == 0xFF) ? GetBeard() : in_beard;
+	new_drakkin_heritage = 
+		(in_drakkin_heritage == 0xFF) ? GetDrakkinHeritage() : in_drakkin_heritage;
+	new_drakkin_tattoo = 
+		(in_drakkin_tattoo == 0xFF) ? GetDrakkinTattoo() : in_drakkin_tattoo;
+	new_drakkin_details = 
+		(in_drakkin_details == 0xFF) ? GetDrakkinDetails() : in_drakkin_details;
```
Akkadius: Needs to be 0xFFFFFFFF

zone/mob.cpp
```diff
@@ -1692,151 +1694,77 @@ void Mob::SendIllusionPacket(
...
+	new_haircolor = (in_haircolor == 0xFF) ? GetHairColor() : in_haircolor;
+	new_beardcolor = (in_beardcolor == 0xFF) ? GetBeardColor() : in_beardcolor;
+	new_eyecolor1 = (in_eyecolor1 == 0xFF) ? GetEyeColor1() : in_eyecolor1;
+	new_eyecolor2 = (in_eyecolor2 == 0xFF) ? GetEyeColor2() : in_eyecolor2;
+	new_hairstyle = (in_hairstyle == 0xFF) ? GetHairStyle() : in_hairstyle;
+	new_luclinface = (in_luclinface == 0xFF) ? GetLuclinFace() : in_luclinface;
+	new_beard = (in_beard == 0xFF) ? GetBeard() : in_beard;
+	new_drakkin_heritage = 
+		(in_drakkin_heritage == 0xFF) ? GetDrakkinHeritage() : in_drakkin_heritage;
+	new_drakkin_tattoo =
```
Akkadius: Needs to be 0xFFFFFFFF

zone/mob.cpp
```diff
@@ -1692,151 +1694,77 @@ void Mob::SendIllusionPacket(
...
-		size = in_size;
-	}
+	new_haircolor = (in_haircolor == 0xFF) ? GetHairColor() : in_haircolor;
+	new_beardcolor = (in_beardcolor == 0xFF) ? GetBeardColor() : in_beardcolor;
+	new_eyecolor1 = (in_eyecolor1 == 0xFF) ? GetEyeColor1() : in_eyecolor1;
+	new_eyecolor2 = (in_eyecolor2 == 0xFF) ? GetEyeColor2() : in_eyecolor2;
+	new_hairstyle = (in_hairstyle == 0xFF) ? GetHairStyle() : in_hairstyle;
+	new_luclinface = (in_luclinface == 0xFF) ? GetLuclinFace() : in_luclinface;
+	new_beard = (in_beard == 0xFF) ? GetBeard() : in_beard;
+	new_drakkin_heritage =
```
Akkadius: Needs to be 0xFFFFFFFF

zone/mob.cpp
```diff
@@ -1112,7 +1114,7 @@ void Mob::FillSpawnStruct(NewSpawn_Struct* ns, Mob* ForWho)
 	ns->spawn.spawnId	= GetID();
 	ns->spawn.curHp	= static_cast<uint8>(GetHPRatio());
 	ns->spawn.max_hp	= 100;		//this field needs a better name
-	ns->spawn.race		= race;
+	ns->spawn.race		= (use_model) ? use_model : race;
```
Akkadius: First glance I'm not really use what `use_model` means, if the purpose of this is to be able to provide a "Temporary race" why not just have it be `temporary_race` which makes it far more obvious to new eyes

noudess: So if you have a gnoll in highpass - the base db uses the new race for (Gnoll New).  It does this not because the mob isn't a gnoll, but because they want the new graphics.  The use_model field lets you change race back to the old gnoll race (making gnollbase weapons work, as well as any other stuff in the db or code that applies to gnolls) and put the model to use in this new field.

Basically it decouples model from race for any mob where it is filled in.

zone/mob.cpp
noudess: Doing 2 assignments in one statement in both cases.  Updating to be on one line (is that what you seek)?

zone/mob.cpp
noudess: Good catch

zone/mob.cpp
noudess: Good catch.

zone/mob.cpp
noudess: I think I got them all now.  Just drakkin stuff right?  The others are 0xFF everywhere.

PR #839, opened 2019-04-12
Rule based faction thresholds

briankinney:
Changing hard-coded faction threshold values can be disruptive to player experience. This change allows for each server to opt into updated values.

Any comments appreciated.


noudess:
Seems like a very good idea for those that want to customize.

mackal:
Values in common/ruletypes.h need to be correct default values, not old values.

briankinney:
The old values were chosen as defaults for backwards compatibility. I think this is more important to most servers than some idea of "correctness" of faction thresholds. Adjusting everything to account for the new thresholds is something I believe that server operators should be able to opt into if they are particularly interested in accurately emulating Everquest and willing to put in the extra work.

mackal:
@noudess wrote some SQL to do that work for the server ops. The correct values need to be the default values.

briankinney:
I understand that some work has been done to make the transition to the new values seamless but I think that this has not been achieved: on one server I have played on players suddenly became kos to factions that previously tolerated them after the code was updated with a build that has the new values hard-coded. It is likely that some amount of disruption could have been prevented by more careful server operations but this is clear evidence to me that seamless update will require extra work for some server operators. I'm not sure if it is feasible or even possible to anticipate every possible case that might cause disruptive changes to character factions. If you are concerned that the SQL updates will cause problems if the new faction levels are not applied I would suggest that those updates also be made optional.

mackal:
As it stands, I'm not merging this. A bug in the implementation was just fixed as well, maybe that could have been related to what you were seeing, but the correct values need to be default.

briankinney:
I will change it as you suggest but I urge you to reconsider your opinion. I think that taking this path will create more unnecessary work not only for server operators but also for you. From what I can see there is a very good reason to take the old values as default that outweighs the benefit of making a backwards-incompatible change.

mackal:
The new values are correct. The old values were wrong. It's what we got to deal with.

PR #846, opened 2019-05-08
New SpawnConditions OnChange action allowing condition_value  to change spawngroup

noudess:
There are some cases on live, that our current spawn_conditions options do not handle.

In Lake Rathe for eaxmple, if the zone starts at night, all undead are spawned in the gnoll camp.  As time passes, these mobs do NOT spawn again as gnolls in the daytime.  The change only occurs if an undead is killed and the change from night to day happens while that spawn point is awaiting respawn.

I looked at solving this with existing options, but our day/night solution doesn't allow this.  The biggest hurdle is our day/nigh solution forces us to create (2) spawn points, on top of each other, with different conditions.  We turn one off and the other on, when a day/night barrier changes.  These two spawn2 points do not know about each other.

To offer a new option, I created a new on_change value for spawn_conditions (4) which is called DoUpdateSpawnGroup.

To use this mechanism, you need only one (single) spawn2 entry for each spot that changes based on the VALUE of a single condition.  So for these gnoll/undead spawns I use condition 1 for all the spawn2 entries.  Then I have 2 values for condition 1 that are set in spawn_events (1 and 2).  When the condition (in this case day/night transition) occurs, on Oncgange:DoUpdateSpawnGroup changes the spawngroup used for the spawn2 location but does nothing else.  If a mob is killed, and then that spot respawns after one of these transitions occurs, the new spawngroup is used, producing the correct mobs.

To do this, a new mandatory (yet can be empty if you dont use this feature) was created.  spawn2_groups.

This table has 3 fields:  spawn2id, condition_value, spawngroupid

This allows for an infinate # of spawn groups at any location based on condition_values set by spawn_events.

This change does nothing to those who dont want to use it.  If you do want to use it, you simply need to follow the above guidelines.  If you want to change day/night to use this method, dont forget to make the correlating changes in DayNight.pl in global quests.

Ive been running this for over a week now.,  It runs great for the gnoll/undead camp.  I plan to use it for any new day/night stuff I intend to create - that behaves as this camp does on live.


xackery:
I think the two most widely used dev tool guis for spawn stuff other than #commands is https://github.com/ProjectEQ/peqphpeditor and https://github.com/Akkadius/EQEmuEOC

noudess:
I will add examples to wiki if approved.

Akkadius:
@noudess this seems like a problem that we should definitely solve for, I am glad you continue to make efforts to improve our system

**Few thoughts**

We already have 7 spawn related tables of which store a lot of this related information already. I think we should extend and use the existing tables flexibly instead of creating yet another table. Spawns are not a complex subject matter and we've over complicated our interface with it with layered ambiguity

**Code Review**

- Formatting is off with indents in many places
- Please do not use very abbreviate variables for SQL statements or variable names, use the full world even if it is verbose, it makes it incredibly hard to read and understand intent and we have plenty of that in the source as it is
- `2019_05_08_spawn_condition_onchange4.sql` has an unnecessary number at the end
- New database table Schema's should have full words and snake_case with no abbreviations

noudess:
What we need is multiple spawngroups per spawn2 to accomplish this goal.  Any solution using existing tables would be limited to a fixed # of spawngroups.  I can go that way, but it will mean limitations in a 1-many situation.

noudess:
As to  indents is there a standard (4 vs 8 space tabs)?

noudess:
Actually what my solution does is allow n spawngroups per spawn2, one per spawn2-condition-value pair.  The existing system is flawed and without redoing it, I think the new table is the best way.  It is an optional feature as well.  I'll probably just have to keep remerging locally, unless someone comes up with a better way.  If you change your mind, or have a suggestion, I'll do the code review changes.

noudess:
Adding a table wasn't my first choice either, I get that.

xackery:
Do you know if this situation happens a lot? If not.. couldn't we just tap into the EVENT_SPAWN quest system to replace the NPC and have more flexibility for one-offs like this?

Keeps our database simpler, too.

mackal:
I think it would require too emulation of the spawn table to just do it in quests.

noudess, Akkadius, and I discussed some other possible solutions that should work, hopefully it will pan out.

PR #847, opened 2019-05-28
Changes to allow filtering of GetNPCType() based on condition_value.

noudess:
Every spawn2 entry has a condition and a condition value.  The existing code filters whether or not the spawn2 location spawns based on the conditon and its value.  The value must currently be greater or equal to the current condition_value for the condition specified.  This is an on/off filter.

This PR introduced a secondary filter based on the condition_value.  If the spawn2 does fire (is enabled as per above), spawn_entry is searched as normal to pick an NPC (GetNPCType()).  This new change added a column, condition_value_filter.  The default value is 1, which is always on (the top level filter already deals with on/off).  Any other value entered guarantees that the NPC indicated by its spawnentry entry is only spawned when the spawn2 condition is exactly equal to the value supplied.

This has been added to support various concepts.  Specifically for me, lakje rathe gnolls/undead.  These mobs occupy the same location, but do not depop/repop based on day night, rather, if killed, the next spawn that occurs naturally in that location will spawn based on the condition value at the time of the spawn.

In my example, I no longer need 2 conditions for night/day.  I use one condition (1) and have 2 values (1 and 2).  One is the default (night) and 2 is day.  Spawnentry used the "do nothing on change" option.  They simply pop based on the value.


noudess:
Updated manifest and version, as will be needed for merge now.

noudess:
Updated manifest and version again.

mackal:
You forgot the SQL file!

noudess:
Added the sql - how does that trigger and run - manifest? Im getting old dude.

mackal:
Would it be possible to rework this so the default is 0? The default of 1 kind of just looks like it's magic numbers :P

noudess:
I feel like you are I still aren't in sync on what this value is.  Can we talk voice maybe?

Before the change. a condition_id/value pair were tied to a spawn2 point.  This allowed the spawn point itself to be on or off, but did not allow a selection of which mobs based on the value.  As a workaround, most servers have multiple spawn2 entries per loc, so that one can be enabled during the day, and one at night.  All of these spawn2 entries typically had a value of 1 (on) or 0 (off). 

My change allows a single spawn2 entry to be enabled/disabled as normal (0 - disabled, non-zero enabled) - while allowing a choice of what mobs based on the value (1-x).

All the entries in spawnentry are enabled, by one of the non-zero values.

If 0 was a default, that meant on, then the condition_value in spawn2 would conflict.

noudess:
Not sure why I had to do conflict resolution just now.  Something else change in mainline?  Did you reformat some lines?

PR #850, opened 2019-06-05
small quality of life fixes

dencelle:
sorry i wasn't able to split this into 2 different pull requests, couldn't figure out how to

1 adds isdooropen boolian to lua functions, this allows quest scripts to check if the door is in the open state or close state. not exactly a huge deal, but it does help when your dealing with doors like you'd find in sleepers tomb.. making both doors open and close at the same time is kinda nice

2 adds the ability to go past level 86, currently there is a overflow issue if you level past 86 while this isn't a issue for most servers since they don't generally go past that level, it does pose a issue for the future of peq, during the house of thule expansion, the level cap goes from 85 to 90 

most of the work is simply debug output things and sanity checks, the only real part that matters is the change from using atoi to atoll when pulling info from the database

### DIFF
--- a/common/eq_packet_structs.h
+++ b/common/eq_packet_structs.h
@@ -953,8 +953,7 @@ struct PlayerProfile_Struct
 /*2384*/	char				title[32];			// length might be wrong
 /*2416*/	char				suffix[32];			// length might be wrong
 /*2448*/	uint32				guildid2;			//
-/*2452*/	uint32				exp;				// Current Experience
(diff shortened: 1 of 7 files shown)


mackal:
You need to create a branch for each specific PR and make a request with those.

I'm going to have to say no here because we really should just be switching XP to a signed 64 bit int (I believe they just uses MSVC's __int64)

dencelle:
closing for now till i have this more refined

PR #853, opened 2019-06-09
fixed opcode for deleteing a item out of a player's inventory slot

dencelle:
OP_MoveItem causes a dysec if you try to move to -1 slot on client side, need to use OP_DeleteItem


Uleat:
Travis passed at the time of this comment..going ahead with merge.

Another inventory desync accounted for - nice job dencelle!

PR #857, opened 2019-06-14
Large warp detection

dencelle:
this is a fairly basic warp detection, I've tested it personally, not getting any false positives as of the time of this post, but still have it set to just make a record of when it catches a large warp until i can say for sure that it won't feed back any false positives.

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -4369,7 +4369,15 @@ void Client::Handle_OP_ClientUpdate(const EQApplicationPacket *app) {
 	}
 	
 	PlayerPositionUpdateClient_Struct *ppu = (PlayerPositionUpdateClient_Struct *) app->pBuffer;
-	
+	ZonePoint *Closest_ZP = zone->GetClosestZonePointWithoutZone(m_Position.x, m_Position.y, m_Position.z, this, 50);
+	if (!Closest_ZP && !cheat_timer.Enabled() && !IsMQExemptedArea(zone->GetZoneID(), m_Position.x, m_Position.y, m_Position.z)) {
+		float distance = Distance(glm::vec3(m_Position), glm::vec3(ppu->x_pos, ppu->y_pos, ppu->z_pos));
+		if (distance >= 140.0f) {
+			CheatDetected(CheatTypes::MQWarp, ppu->x_pos, ppu->y_pos, ppu->z_pos);
+		}
+	} else {
+		cheat_timer.Disable();
+	}
 	/* Boat handling */
 	if (ppu->spawn_id != GetID()) {
 		/* If player is controlling boat */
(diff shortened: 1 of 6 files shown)


zone/client_packet.cpp
```diff
@@ -14765,3 +14773,177 @@ void Client::Handle_OP_ResetAA(const EQApplicationPacket *app)
...
+	default:
+		char* hString = nullptr;
+		MakeAnyLenString(&hString, "Unhandled HackerDetection flag with location %.2f, %.2f, %.2f.", GetX(), GetY(), GetZ());
+		database.SetMQDetectionFlag(this->account_name, this->name, hString, zone->GetShortName());
+		safe_delete_array(hString);
+		break;
+	}
+}
+
+bool Client::IsMQExemptedArea(uint32 zoneID, float x, float y, float z)
```
Akkadius: Instead of "exempt areas" I would prefer to have coverage over "exempt behaviors"

For example if we are performing a door click and warping, or performing a normal action and warping, we need to be able to trigger the timer to let the server consider that you are performing normal behavior. Manually specifying zones and areas is far too much manual work and maintainability that is prone to producing false positives

zone/mob.h
```diff
@@ -68,6 +68,7 @@ enum class eSpecialAttacks : int {
 
 class Mob : public Entity {
 public:
+	Timer cheat_timer;
```
Akkadius: This needs to be more descriptive, something more like `normal_warp_grace_period_timer`

zone/zoning.cpp
```diff
@@ -446,27 +446,34 @@ void Client::ProcessMovePC(uint32 zoneID, uint32 instance_id, float x, float y,
 
 	switch(zm) {
 		case GateToBindPoint:
+			cheat_timer.Start(3500, false);
```
Akkadius: I would like to see the 3500 a constant that we reference somewhere, preferably if we don't have a dedicated class managing this logic I would prefer to see it in `emu_constants.h`, something like `EQEmu::Warp::GracePeriodNormal`

zone/client_packet.cpp
```diff
@@ -14765,3 +14773,177 @@ void Client::Handle_OP_ResetAA(const EQApplicationPacket *app)
...
+{
+	switch (CheatType)
+	{
+	case MQWarp:
+		if (RuleB(Zone, EnableMQWarpDetector) && ((this->Admin() < RuleI(Zone, MQWarpExemptStatus) || (RuleI(Zone, MQWarpExemptStatus)) == -1)))
+		{
+			Message(13, "Large warp detected.");
+			char hString[250];
+			sprintf(hString, "/MQWarp with location %.2f, %.2f, %.2f", GetX(), GetY(), GetZ());
+			database.SetMQDetectionFlag(this->account_name, this->name, hString, zone->GetShortName());
```
Akkadius: This needs to log to its own log category, I would create a MQDetection logging category and log there as well as the database table

zone/client_packet.cpp
```diff
@@ -14765,3 +14773,177 @@ void Client::Handle_OP_ResetAA(const EQApplicationPacket *app)
 	}
 	return;
 }
+
+void Client::CheatDetected(CheatTypes CheatType, float x, float y, float z)
```
Akkadius: WarpDetected vs CheatDetected, remove the switch and keep it explicit

dencelle:
Being forced to take this down, i'll place it back up after i'm done with this legal stuff

PR #858, opened 2019-06-15
Updating linux installer script

jalavoy:
This brings the installer script up to date and changes a few logical problems. For example running apt 50 times is a ton slower than running it once, same with dnf.

the yum install for RHEL was also installing 1777 undeeded perl and lua modules for no reason. Now it installs 53.

### DIFF
--- a/utils/scripts/linux_installer/install.sh
+++ b/utils/scripts/linux_installer/install.sh
@@ -146,61 +146,75 @@ if [[ "$OS" == "Debian" ]]; then
 
 elif [[ "$OS" == "red_hat" ]]; then
 	# Do RedHat / CentOS stuff
-	# Add the MariaDB repository to yum
-cat <<EOF > /etc/yum.repos.d/mariadb.repo
-# MariaDB 10.1 CentOS repository list - created 2016-08-20 05:42 UTC
-# http://downloads.mariadb.org/mariadb/repositories/
-[mariadb]
-name = MariaDB
-baseurl = http://yum.mariadb.org/10.1/centos7-amd64
-gpgkey=https://yum.mariadb.org/RPM-GPG-KEY-MariaDB
-enabled=1
-gpgcheck=1
-EOF
 	# Install prereqs
-	yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
-	yum -y install deltarpm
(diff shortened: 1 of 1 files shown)


Akkadius:
Hey @jalavoy thanks for taking the initiative

The reason why we didn't include the installs on one line is in the case specifically for Debian - the same lines are used for Ubuntu and Debian. because of this and distributions changing a package would go missing and then bomb the entire install, OR we have a package listed in the list that is available on Debian but not in Ubuntu and would break one and not the other. At least with the individual lines it may error but continue in the script. Yes, there is the `apt-get` `--ignore-missing` flag but we also had inconsistent behavior with that flag working as well so it wasn't a huge deal to list the commands out line by line, it makes it readable and maybe costs a few more seconds at most. For this reason I would want the Debian / Ubuntu block to the remain the same

Because Fedora and Redhat are definitely more targeted, I'm fine with it being in one command but I would prefer each package to be on its own line like:

```
dnf -y install \
	cmake \
	open-vm-tools \
	vim
```

I also would like to see proof that the final state of your PR runs through a 100% successful, operational server install. Each time we've made tweaks to this with new OS's we go through the whole routine

I would also like to make note that there are plans to revise some of the installer in the near future, most of it will remain the same but we have some changes to be incorporating soon

Kind regards,
Akk

jalavoy:
This all sounds reasonable to me. I’ll update my end later tonight. 

How would you like me to demonstrate the changes are working? Or is that done in your CI pipeline?

Akkadius:
Our testing has always been manual simply because testing the same containers over and over only does so much. We only usually run into issues when distribution upgrades happen, even then they may be in the middle of changing packages around and may or may not be final when that happens so its not something that can be simply automated anyways

We usually bring up VM's on our nodes and just run the scripts and check them in a while, make sure we can launch world / zone and everything is functional in game with scripts etc. 

That being said I will probably at least look at hooking up a series of distributions to go through it all through a CI at some point so we at least have that as a measurable pass and run it through some basic automated tests. It's just never been a huge priority and has worked for thousands of installs without issues

jalavoy:
I went ahead and spun up brand new Centos 7 and Fedora 30 VM's and applied the current version of the PR against them. Both of them built without errors and resulted in a running server. 

Also reverted the apt changes for you.

This iteration should probably be good to go. Let me know if anything else needs tweaking.

Akkadius:
Hey @jalavoy - not ignoring this, I had surgery last week and then not long after recovering I drove 6 hours north out of cell range and now I'm back into the business week

It's on my list to run through the Ubuntu/Debian side of this and merge this in

Thank you!

jalavoy:
No worries my dude. Lmk if you need me to update anything. Hope you’re feeling better.

PR #859, opened 2019-06-16
Convert constant check in level_command() to a Rule

jwinky:
Add GM:MinStatusToLevelTarget rule, which determines the GM status needed to use the #level command on your target.

### DIFF
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -205,6 +205,7 @@ RULE_CATEGORY_END()
 RULE_CATEGORY(GM)
 RULE_INT(GM, MinStatusToSummonItem, 250)
 RULE_INT(GM, MinStatusToZoneAnywhere, 250)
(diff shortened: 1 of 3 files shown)


KinglyKrab:
When adding rules we usually like to include a required/optional SQL file depending upon the use of the rule. This allows server operators to see this rule in their database rather than manually adding it or simply not knowing of its existence. Other than that, looks good to me.

joligario:
Or in a case like this, it should be ok because it is already hard-coded and he didn't change the default.

jwinky:
Happy to update the SQL also!  I figured I'd likely miss something so starting with a simple PR.

jwinky:
I checked db_update_manifest.txt but did not add an entry as other recent optional SQLs do not have one either.  If that's preferred I'm happy to add that as well.

PR #862, opened 2019-06-19
Container lock inconsistent state fix

fryguy503:


mackal:
Why by char ID and not entity ID (which should be faster look up)

PR #863, opened 2019-06-24
Change BaneDmgAmt from int8 to int32 in item_data.h to match the client-specific structs.

KinglyKrab:


mackal:
Hmm tit is a byte. (not saying no, just stating it :P )

KinglyKrab:
Yeah, in the Titanium struct it's all commented out for some reason.
Line 3443 (common/titanium_structs.h): `//	int BaneDmgAmt;`

But in the others it seems to be fine.
Line 4045 (common/uf_structs.h): 
`int32 BaneDmgAmt;`


Line 3889 (common/sof_structs.h): 
`int32 BaneDmgAmt;`


Line 3967 (common/sod_structs.h): 
`int32 BaneDmgAmt;`


Line 4566 (common/rof_structs.h): 
`int32 BaneDmgAmt;`


Line 4626 (common/rof2_structs.h): 
`int32 BaneDmgAmt;`

mackal:
The tit file doesn't matter though, for this :P because of how it's serialized. I was just checking old MQ2, which could be wrong.

mackal:
It's 100% a single byte in tit :P

KinglyKrab:
So is it fine the way it was? Will the other clients just use their struct data and Titanium will default to the int8 here?

Uleat:
I'm not familiar with the internals of titanium since it was built on an older core library...

So long as ti doesn't adjust the packet pointer, we should be able to 'overload' the field and only have an incorrect value there.

Have you guys tested the effect of the change yet? Increasing damage to >255?

KinglyKrab:
On RoF2 I found some issues actually with changing it to int32.

At 128, 250, or 255 bane damage on body type 3 (undead) it doesn't appear on the item at all.

At 10,000 damage it appears, but it appears as +16 Bane Damage.

Not really sure why that is.

mackal:
You missed updating common/shareddb.cpp, you would also have to re-run shared_memory.

PR #865, opened 2019-06-26
Add mod hook for AddLevelBasedExp

dencelle:


dencelle:
Being forced to take this down, i'll place it back up after i'm done with this legal stuff

PR #871, opened 2019-07-18
Fix issues with OP_SpecialMesg handling

mackal:
This should prevent any optimizations being done on the "1 char string"

This also fully documents the packet and expands the uses of
quest::say/QuestSay

Looking for comments on quest API changes before pushing to master.


mackal:
Hmm I realize the lua constants being under Journal_SpeakMode and Journal_Mode are a bit annoying to type, I'll change these I guess :P

PR #877, opened 2019-08-05
Light wield for Npcs/mobs

noudess:
NPC/Mobs do seem to be willing to equip light sources that are primary/secondary when traded or as loot.  Made the change.


mackal:
Does this end affect combat at all?

noudess:
I guess only in that if you give a pet a light item and he has a primary/secondary slot open, it will use it for light.

mackal:
It's not gonna like enable dual wield on NPCs if you give them two of them?

noudess:
I'll test that.  Great catch.

noudess:
Yes, you were correct.  They don't check for weapon in attack code for offhand, they just check for a texture, then specifically cancel the attack if is a shield.  I fixed it so it rejects any non 1handed weapon.

PR #883, opened 2019-08-13
Update questmgr.cpp

noudess:
Put back Say() for non initiated quest::say() calls like those in an EVENT_WAYPOINT_ARRIVE sub.


mackal:
Mob::Say isn't feature compatible, so I'll get this working w/o it.

noudess:
Okies!

PR #885, opened 2019-08-14
Warp detection

dencelle:
this should be a fairly light weight way of detecting warps.
since its not distance dependent like we've always done in the past, we shouldn't get any false positives.

### DIFF
--- a/common/eq_packet_structs.h
+++ b/common/eq_packet_structs.h
@@ -5422,7 +5422,7 @@ struct UpdateMovementEntry {
 	/* 00 */	float Y;
 	/* 04 */	float X;
 	/* 08 */	float Z;
-	/* 12 */	byte type;
+	/* 12 */	uint8 type;
 	/* 13 */	unsigned int timestamp;
 	/* 17 */
 };
--- a/common/eq_packet_structs.h
+++ b/common/eq_packet_structs.h
@@ -5418,6 +5418,15 @@ struct SayLinkBodyFrame_Struct {
 /*056*/
 };
 
+struct UpdateMovementEntry {
+	/* 00 */	float Y;
+	/* 04 */	float X;
+	/* 08 */	float Z;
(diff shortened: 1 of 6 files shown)


mackal:
I don't think we need the worldserver.SendEmoteMessage

zone/client_packet.cpp
```diff
@@ -5865,6 +5865,30 @@ void Client::Handle_OP_Fishing(const EQApplicationPacket *app)
 	// forage for.
 }
 
+void Client::Handle_OP_FloatListThing(const EQApplicationPacket *app) {
+	UpdateMovementEntry* m_MovementHistory = new UpdateMovementEntry[(app->size) / sizeof(UpdateMovementEntry)];
```
mackal: This leaks on the line below

zone/client_packet.cpp
```diff
@@ -5865,6 +5865,30 @@ void Client::Handle_OP_Fishing(const EQApplicationPacket *app)
...
+	m_MovementHistory = (UpdateMovementEntry*)app->pBuffer;
+	for (int index = 0; index < (app->size) / sizeof(UpdateMovementEntry); index++) {
+		switch (m_MovementHistory[index].type) {
+		case 3:
```
mackal: Can we make an enum instead of magic numbers?

common/eq_packet_structs.h
```diff
@@ -5418,6 +5418,15 @@ struct SayLinkBodyFrame_Struct {
 /*056*/
 };
 
+struct UpdateMovementEntry {
+	/* 00 */	float Y;
+	/* 04 */	float X;
+	/* 08 */	float Z;
+	/* 12 */	byte type;
```
mackal: We usually just use uint8 for this, can that be changed for consistency?

Akkadius:
Closing because I know this was not complete and there were other efforts in progress

PR #895, opened 2019-09-07
git submodule init/update must be executed from the top level working…

mibastian:
… tree

Build is failing as the submodules are not installed.


Akkadius:
Good catch

PR #896, opened 2019-09-07
Build on all cores

mibastian:
nproc is part of the coreutils so no further yum/apt package needed.
One could argue to build on nproc-1 or even nproc/2 for SMT systems, but that would need handling for nproc=1)
Building the latest source on an 8 core CentOS7 qemu VM (E5-2690 v2) takes 2m12s on all cores.


Akkadius:
The reason why this isn't on by default is because so many people ran into under powered server issues with the amount of memory some of our template based luabind code takes thus bombing out the entire compilation process and ultimately killing the installer. Yes 1 core takes longer but it is far more guaranteed from a user-end perspective

PR #906, opened 2019-09-26
Modified eqemu_server.pl to use new login server schema

alimalkhalifa:
Has a matching pull request in Akkadius/EQEmuInstall

* Error was found by Maze

### DIFF
--- a/utils/scripts/eqemu_server.pl
+++ b/utils/scripts/eqemu_server.pl
@@ -1489,9 +1489,9 @@ sub do_windows_login_server_setup {
     print "[Install] Done... \n";
 
     print "[Install] Pulling down Loginserver database tables...\n";
-    get_remote_file($install_repository_request_url . "login_server_tables.sql", "db_update/login_server_tables.sql");
+    get_remote_file($eqemu_repository_request_url . "loginserver/login_util/login_schema.sql", "db_update/login_schema.sql");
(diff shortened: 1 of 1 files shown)


Akkadius:
This needs to reference the source of truth here (Download the raw):

https://github.com/EQEmu/Server/blob/master/loginserver/login_util/login_schema.sql

alimalkhalifa:
Updated to pull from source of truth

alimalkhalifa:
Updated urls to use Variables

utils/scripts/eqemu_server.pl
```diff
@@ -1518,9 +1518,9 @@ sub do_linux_login_server_setup {
     print "\n Done... \n";
 
     print "[Install] Pulling down Loginserver database tables...\n";
-    get_remote_file($install_repository_request_url . "login_server_tables.sql", "db_update/login_server_tables.sql");
+    get_remote_file("https://github.com/EQEmu/Server/blob/master/loginserver/login_util/login_schema.sql", "db_update/login_schema.sql");
```
Akkadius: This needs to reference the raw file, not the Github HTML representation of this file

alimalkhalifa:
Fixed, thanks for pointing out @Akkadius .  Also squashed the commits

PR #911, opened 2019-10-07
Old school apply poison was allowing multiple poisons at once.

noudess:
Old school poisons should be confined to one active poison.  The code was allowing up to MAX_PROCS after applying multiple poisons.

This code adds the proc normally if no other poison procs are already in place, or replaces the old one, which matches live behaviour.

No changes were made to the new poison code.


mackal:
better than it was, but I figured out some more implementation details for noudness to implement later.

noudess:
I'll open an issue and try to capture what you suggested.  This shouldn't take me too long, but I need to get familiar with the code a bit more.

PR #923, opened 2019-10-22
Eye of Zomm needed support in OP_ClientUpdate.

noudess:
Eye of Zomm eyeball was not showing up on other clients screens when it moved.

Added:

- Support for the OP_ClientUpdate when it is coming in for the eye (or any other client controlled mob that uses OP_ClientUpdate (boats are different and are left alone for now)
- Added support to movement manager to include a parameter that allows ignoring a particular client.  Talked to @KimLS  about this and he's ok.
- Changed eye of zomm to not have the flag "follow" set, as once we added support for position, the eye started resisting and heading back to the caster :package: 

Note:

In working on this, I noticed that client controlled boats are not working.  @Uleat found a packet entry that is not being propogated up, that helps with this issue.  I am working in a new branch to understand and repair that.  I was going to do that in this PR as well, but on discovering that boats use a diff mechanism now (send two spawnIDs in the OP_Client update) that the code is different enough as to not share much code.


noudess:
Changes made to use a variable instead of eye-name for efficiency as requested. @mackal 

I used a separate variable, as eye of zomm and being on a boat can be done simultaneously on live. 

I have fixes, which based on the spawn2Id field that @Uleat  gave me a heads up about, repair client position updates (to other clients).  I am not checking these in, as I believe @Uleat  is working on the decoding process for these fields in a more global fashion that my work (which only addressed RoF2).

Issue #932, opened 2019-11-13
How trading with NPC should function under different conditions


huffin:
Things that happen currently that should not:
1.) When you complete a trade with an NPC, they turn towards you, they should not. They should only turn to you when you engage in conversation or combat.
2.) You can currently start a trade with an NPC while in combat with that NPC, you should not be able to even open a trade with an NPC that is engaged in combat with you.
3.)You can currently finish/complete  a trade with an NPC while you are in combat with them. This should not be possible.

**Parsed the following conditions**

-Sneaking behind an NPC I am kos to; does not turn towards me
[Wed Nov 13 15:10:24 2019] You gain experience!
[Wed Nov 13 15:10:24 2019] You receive 7 silver from Trooper Mozo.

-sneaking behind npc, initiate trade, drop sneak, npcs aggro, i attempt to complete the trade, it fails
[Wed Nov 13 15:10:56 2019] Trooper Mozo regards you indifferently -- You could probably win this fight. (Lvl: 50)
[Wed Nov 13 15:10:58 2019] You stop sneaking.
[Wed Nov 13 15:11:01 2019] Trooper Mozo says, 'Fear the blades of the Iksar!'
[Wed Nov 13 15:11:01 2019] Trooper Taer says, 'Fear the blades of the Iksar!'
[Wed Nov 13 15:11:01 2019] Trooper Grouko says, 'Fear the blades of the Iksar!'
[Wed Nov 13 15:11:03 2019] Trooper Mozo says, 'I have no need for this, Probard. You can have it back.'

-when in combat, you just cannot open a trade with the NPC that is in combat with you,no emotes

-[Wed Nov 13 15:13:19 2019] You say, 'open trade window with no sneak no invis before they aggro- got attacked(because initiating trade should drop invis) - faded memories (bard)(invis) trade window still open - he then accepted trade'

-[Wed Nov 13 15:14:32 2019] You say, 'open trade - invis drops - get invis before npc aggro's - you are now invis and can finish trade because the NPC never even aggro'd'

noudess:
My comments on this issue (4 years after it's creation):

#1 I don't think I have a concern with this one way or the other.  At least not enough to change the code.  The code change was provided in a PR that was closed due to the contriibutor not replying.

#2 I do not think this is possible anymore (at least not on RoF2).  Testing and could not do it.

#3 Is this really true on live?  We'll do some testing.

PR #933, opened 2019-11-13
Mobs will not trade anymore if they are currently in combat.

fryguy503:
This commit should address the second bullet-point in issue #932


mackal:
I don't think there should be a message.

PR #937, opened 2019-12-07
Advance client position while on boat

noudess:
This is a collaboration between myself and @Uleat.   I added code to reflect client updates (incremental) correctly once a client enters a boat.  @Uleat made it fit in much more elegantly.


Uleat:
Noudess and I have tested this code and it does prevent the player from 'disappearing' when boarding a boat.

More changes coming as the boat system is figured out.

PR #941, opened 2020-01-04
Fix #reloadworld message.

KinglyKrab:


zone/command.cpp
```diff
@@ -4252,10 +4252,15 @@ void command_corpsefix(Client *c, const Seperator *sep)
 
 void command_reloadworld(Client *c, const Seperator *sep)
 {
-	c->Message(Chat::White, "Reloading quest cache and repopping zones worldwide.");
+	int world_repop = atoi(sep->arg[1]);
+	if (world_repop == 0)
```
Akkadius: I know this is legal, but I prefer we put brackets on our calls

PR #943, opened 2020-01-08
Reworked EVENT_COMMAND handling and addressed script commands not pre…

Uleat:
…sent during help use


Uleat:
Added/Updated 'Events' docs

PR #951, opened 2020-01-20
Implemented SE_SummonCorpseZone.

KinglyKrab:
- Summons all of a targeted player's corpses if they are your group member.
- Will summon a players corpses from all the zones they have died in similar to quest::summonallplayercorpses(uint32 char_id, const glm::vec4& position);

### DIFF
--- a/zone/spell_effects.cpp
+++ b/zone/spell_effects.cpp
@@ -1820,6 +1820,63 @@ bool Mob::SpellEffect(Mob* caster, uint16 spell_id, float partial, int level_ove
 
 				break;
 			}
+			case SE_SummonCorpseZone:
+			{
+#ifdef SPELL_EFFECT_SPAM
+				snprintf(effect_desc, _EDLEN, "Summon Corpse Zone");
+#endif
+				if (!IsNPC()) {
+					Client* client_target = this->CastToClient();
+					Group* group = entity_list.GetGroupByClient(client_target);
+					if(group) {
+						if(!group->IsGroupMember(client_target)) {
+							if (caster != this) {
+								caster->Message(Chat::Red, "Your target must be a group member for this spell.");
+								break;
+							}
+						}
+					} else if (caster) {
+						Raid *raid = entity_list.GetRaidByClient(caster->CastToClient());
+						if(raid) {
+							uint32 group_id = 0xFFFFFFFF;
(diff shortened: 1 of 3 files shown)


zone/spell_effects.cpp
```diff
@@ -1820,6 +1820,63 @@ bool Mob::SpellEffect(Mob* caster, uint16 spell_id, float partial, int level_ove
 
 				break;
 			}
+			case SE_SummonCorpseZone:
+			{
+#ifdef SPELL_EFFECT_SPAM
```
Akkadius: Can we remove this? If we're going to utilize logging, please just use `LogSpells("Message")`

zone/spell_effects.cpp
```diff
@@ -1820,6 +1820,63 @@ bool Mob::SpellEffect(Mob* caster, uint16 spell_id, float partial, int level_ove
...
+				snprintf(effect_desc, _EDLEN, "Summon Corpse Zone");
+#endif
+				if (!IsNPC()) {
```
mackal: Wouldn't IsClient() be better here?

zone/spell_effects.cpp
```diff
@@ -1820,6 +1820,63 @@ bool Mob::SpellEffect(Mob* caster, uint16 spell_id, float partial, int level_ove
...
+							if (caster != this) {
+								caster->Message(Chat::Red, "Your target must be a group member for this spell.");
+								break;
+							}
+						}
+					} else if (caster) {
+						Raid *raid = entity_list.GetRaidByClient(caster->CastToClient());
+						if(raid) {
+							uint32 group_id = 0xFFFFFFFF;
+							group_id = raid->GetGroup(caster->GetName());
```
Akkadius: Can we inline `group_id` here

```uint32 group_id = raid->GetGroup(caster->GetName()); ```

`GetGroup` returns `0xFFFFFFFF` if there is no group anyways, I'd prefer if we did a check for `group_id` > 0 and `group_id` < MAX_RAID_GROUPS because we can get overflowed values back (4294967295), if we ever had this return 0 or -1 the current code would not work correctly

zone/spell_effects.cpp
mackal: That's just how we do spell effect shit, I think it's fine for now. Maybe we can switch them all over to LogSpells at another time :P

PR #953, opened 2020-01-20
_GetRunSpeed did not correctly report aa mods for Clients.

noudess:
I actually believe this was some old cut-n-paste error.  aa_mod was being set to a total of all (3) caps - like the previous line.

#showstats now correctly displays movement rate for people with runspeed AAs.


Akkadius:
@mackal if you confirm that this lines up with what you see in the client - I'm good with this one

PR #958, opened 2020-01-27
Implement consent for group/raid/guild and add Auto Consent support

hgtw:
Refactors consent to be more live accurate

Message sent to owner and receiver for each zone a corpse is in
Corpses now store consent list instead of clients holding corpse list
Consent throttling added
Message strings and colors updated
Removed reporting invalid consent targets

### DIFF
--- a/world/zoneserver.cpp
+++ b/world/zoneserver.cpp
@@ -1064,29 +1064,34 @@ void ZoneServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p) {
 		ServerOP_Consent_Struct* s = (ServerOP_Consent_Struct*)pack->pBuffer;
 
 		ZoneServer* owner_zs = nullptr;
-		if ((s->instance_id != 0 && (owner_zs = zoneserver_list.FindByInstanceID(s->instance_id))) ||
-		    (s->instance_id == 0 && (owner_zs = zoneserver_list.FindByZoneID(s->zone_id))))
-		{
+		if (s->instance_id == 0) {
+			owner_zs = zoneserver_list.FindByZoneID(s->zone_id);
+		}
+		else {
+			owner_zs = zoneserver_list.FindByInstanceID(s->instance_id);
+		}
+
+		if (owner_zs) {
 			owner_zs->SendPacket(pack);
 		}
-		else
-		{
+		else {
 			LogInfo("Unable to locate zone record for zone id [{}] or instance id [{}] in zoneserver list for ServerOP_Consent_Response operation", s->zone_id, s->instance_id);
 		}
 
-		if (s->consent_type == EQEmu::consent::Normal)
-		{
+		if (s->consent_type == EQEmu::consent::Normal) {
 			// send the message to the client being granted or denied permission
-			if (ClientListEntry* cle = client_list.FindCharacter(s->grantname))
-			{
+			ClientListEntry* cle = client_list.FindCharacter(s->grantname);
+			if (cle) {
 				ZoneServer* granted_zs = nullptr;
-				if ((cle->instance() != 0 && (granted_zs = zoneserver_list.FindByInstanceID(cle->instance()))) ||
-				    (cle->instance() == 0 && (granted_zs = zoneserver_list.FindByZoneID(cle->zone()))))
-				{
-					// avoid sending twice if owner and granted are in same zone
-					if (granted_zs != owner_zs) {
-						granted_zs->SendPacket(pack);
-					}
+				if (cle->instance() == 0) {
+					granted_zs = zoneserver_list.FindByZoneID(cle->zone());
+				}
+				else {
+					granted_zs = zoneserver_list.FindByInstanceID(cle->instance());
+				}
+				// avoid sending twice if owner and granted are in same zone
+				if (granted_zs && granted_zs != owner_zs) {
+					granted_zs->SendPacket(pack);
 				}
 			}
 		}
--- a/world/zoneserver.cpp
+++ b/world/zoneserver.cpp
@@ -1057,110 +1057,37 @@ void ZoneServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p) {
 		break;
 	}
 	case ServerOP_Consent: {
-		// Message string id's likely to be used here are:
-		// CONSENT_YOURSELF = 399
-		// CONSENT_INVALID_NAME = 397
-		// TARGET_NOT_FOUND = 101
-		ZoneServer* zs;
-		ServerOP_Consent_Struct* s = (ServerOP_Consent_Struct*)pack->pBuffer;
-		ClientListEntry* cle = client_list.FindCharacter(s->grantname);
-		if (cle) {
-			if (cle->instance() != 0)
-			{
-				zs = zoneserver_list.FindByInstanceID(cle->instance());
-				if (zs) {
-					zs->SendPacket(pack);
-				}
-				else
-				{
-					auto pack = new ServerPacket(ServerOP_Consent_Response, sizeof(ServerOP_Consent_Struct));
-					ServerOP_Consent_Struct* scs = (ServerOP_Consent_Struct*)pack->pBuffer;
-					strcpy(scs->grantname, s->grantname);
-					strcpy(scs->ownername, s->ownername);
-					scs->permission = s->permission;
-					scs->zone_id = s->zone_id;
-					scs->instance_id = s->instance_id;
-					scs->message_string_id = 101;
-					zs = zoneserver_list.FindByInstanceID(s->instance_id);
-					if (zs) {
-						zs->SendPacket(pack);
-					}
-					else {
-						LogInfo("Unable to locate zone record for instance id [{}] in zoneserver list for ServerOP_Consent_Response operation", s->instance_id);
-					}
-					safe_delete(pack);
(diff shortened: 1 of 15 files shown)


hgtw:
The biggest change is now /consent and /deny result in searching every zone for a character's corpses to be updated instead of just sending one message for the granted client. I added the throttling that live has so hopefully that's not a big impact.

The only issue I've seen with testing is that the long zone name in messages is going to be truncated for some zones (ferubi for example).  I matched the size of the ConsentResponse_Struct buffer in the server op struct just to be consistent.

mackal:
I don't think that implements raid/group correctly. It should just be, as long as they're in the same raid/group you are currently in, it will consent them. I'm also 99% sure this needs to be saved to the DB and restored on load.

hgtw:
On live it seems to snapshot the group/raid id on to the corpse at the time of consent. Here's some notes from brief testing.

auto consent guild:
   die, log guildmember that was offline from guild in, drag works
auto consent raid:
  die while in a raid, /corpse drag works, disband dragger from raid, drag doesn't work
if owner of corpse goes offline:
  dragging corpses still works as long as player is still in same guild/raid (group untested)
disbanding and forming a new raid:
  1) if corpse owner disbands, corpse still draggable by players in original raid
  2) if a dragger raid disbands, corpse no longer draggable by that player
  3) if dragger re-invited to same raid, corpse draggable (even if original owner not reinvited to raid)
  4) if raid fully disbanded and everyone creates a new one, corpse not draggable
  (original raid id seems to be stored on corpse)
/consent guild, then /consent specific player:
   /deny player, player can still drag until /deny guild
   /deny guild instead, player can still drag until /deny player

And yea, as far as DB goes it's not persistent for dynamic zones/server restarts. I didn't feel like the added complexity was worth it, but I can try to add that if the rest of it checks out

mackal:
Okay, I didn't actually do tests like that, just made assumptions based on it staying on. It is in the Options UI :P which I'm pretty sure I've never had to turn back on on live.


Seems they went the easy route and just snap shotted based on time of corpse creation. Adding to DB shouldn't be hard. I'll check tomorrow to make sure everything is documented where the values should go in packets (unless you verify before I can)


I also haven't fully reviewed the code, so this isn't me saying code is good yet XD

mackal:
Actually, this is already saved to the DB.

hgtw:
Oh, yes the Auto Consent options backend was already implemented for the DB so those options save.

I misread it as saving the consent ids to the database, which after sleeping on it didn't make as much sense (those ids would be invalid on server restarts anyway, the worst that happens for dynamic zones is needing to issue a new consent)

mackal:
Have you tested graveyard (including graveyards that zone) interaction on live and how it compares with this code?

hgtw:
It looks like only guild consent persists when a corpse is summoned cross zone on live. Raid and Group consent both stopped working after summoning the corpse to another zone
All consents persisted when a corpse was summoned within the same zone
(tested with guildlobby and sk Summon Remains AA)

Currently with this code all consents are lost on corpse summons (unless the corpse is summoned within zone through GMMove, like /corpse does)

mackal:
Did you test actual graveyards, like most pop zones or ldons for graveyards that zone?

hgtw:
I did but only for the guild consent case (guild consent persisted). I tested that before I discovered group/raid weren't persistent across zones so I'll need to test that case again. Are there any graveyards faster than the 20min pop zones that you're aware of?

mackal:
I thought they were 15 mins? But no, I don't :P

hgtw:
Graveyards have the same behavior as summoning corpses. Abandoning an ldon  and waiting for the corpse to pop outside instance left guild consent but dropped the others. Graveyard within ponightmare left all consents enabled

mackal:
so you just need to make guild stay?

hgtw:
Yea I think just the guild id needs to be stored on the corpse in DB to match live here. An interesting note is that live doesn't even update existing corpses for leaving a guild. I was still able to drag a guild consented corpse even though I guild disbanded the owner before summoning it.

Summoning within a zone is still not going to persist group/raid like live does though unless the corpse is moved with GMMove. I think that has more to do with how corpse summoning spells/skills handle in-zone corpses (maybe a future optimization there, I haven't really looked at it). I don't like the idea of trying to store changeable group or raid ids for a corpse in DB to work around that

world/zoneserver.cpp
```diff
@@ -1057,110 +1057,37 @@ void ZoneServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p) {
...
 	case ServerOP_Consent_Response: {
-		// Message string id's likely to be used here are:
-		// CONSENT_YOURSELF = 399
-		// CONSENT_INVALID_NAME = 397
-		// TARGET_NOT_FOUND = 101
 		ServerOP_Consent_Struct* s = (ServerOP_Consent_Struct*)pack->pBuffer;
-		if (s->instance_id != 0)
+
+		ZoneServer* owner_zs = nullptr;
+		if ((s->instance_id != 0 && (owner_zs = zoneserver_list.FindByInstanceID(s->instance_id))) ||
```
mackal: I know this takes up less space, but I still don't like doing assignments in if statements, they're just easy to miss on quick reads :P

hgtw: Do you feel the same about declaration assignments in conditions (for limiting scope) or do you want those changed too

e.g.` if (Client* client = ...)`

mackal: Correct.

zone/client.cpp
```diff
@@ -6254,6 +6255,53 @@ void Client::DragCorpses()
 	}
 }
 
+void Client::ConsentCorpses(const char* consent_name, bool deny)
+{
```
mackal: You wanna add a null check? I know none of the current code should cause issues, but it's not like this is in a hot path or anything, so whatever, better safe than sorry :P

zone/client.cpp
```diff
@@ -6254,6 +6255,53 @@ void Client::DragCorpses()
...
+	if (strcasecmp(consent_name, GetName()) == 0) {
+		MessageString(Chat::Red, CONSENT_YOURSELF);
+	}
+	else if (!consent_throttle_timer.Check()) {
+		MessageString(Chat::Red, CONSENT_WAIT);
+	}
+	else {
+		auto pack = new ServerPacket(ServerOP_Consent, sizeof(ServerOP_Consent_Struct));
+		ServerOP_Consent_Struct* scs = (ServerOP_Consent_Struct*)pack->pBuffer;
+		strcpy(scs->grantname, consent_name);
```
mackal: Wanna switch these to our gross strn0cpy functions?

zone/corpse.h
```diff
@@ -142,6 +147,9 @@ class Corpse : public Mob {
 	int32		player_kill_item; /* Determines if Player Kill Item */
 	uint32		corpse_db_id; /* Corpse Database ID (Player Corpse) */
 	uint32		char_id; /* Character ID */
+	uint32		consent_group_id = 0; /* consented group id */
```
mackal: I think we need to re-evaluate if we should allow this feature (non-static data member initializers), I think there was some issues with earlier GCC versions?

Akkadius: This is technically C++11 - which is what we're at; so I'm good with it if everyone else is, its reduces constructor redundancy

zone/corpse.cpp
```diff
@@ -647,6 +660,29 @@ void Corpse::DepopPlayerCorpse() {
 	player_corpse_depop = true;
 }
 
+void Corpse::AddConsentName(const char* add_name)
```
Akkadius: Is there a reason why we didn't just use `std::string consent_name` ?

zone/corpse.cpp
```diff
@@ -647,6 +660,29 @@ void Corpse::DepopPlayerCorpse() {
...
+		for (const auto& n : consent_names) {
+			if (strcasecmp(n.c_str(), add_name) == 0) {
+				return;
+			}
+		}
+		consent_names.emplace_back(add_name);
+	}
+}
+
+void Corpse::RemoveConsentName(const char* rem_name)
```
Akkadius: Same thing here, is there a reason we didn't just use `std::string consent_player_name`

zone/corpse.h
```diff
@@ -142,6 +147,9 @@ class Corpse : public Mob {
...
+	uint32		consent_raid_id  = 0; /* consented raid id  */
```
Akkadius: I know we have comments in other properties of the class, preferably we'd like to just have the variable or function describe itself instead of putting comments if we can

zone/corpse.h
```diff
@@ -160,6 +168,7 @@ class Corpse : public Mob {
 	Timer		corpse_graveyard_timer;
 	Timer		loot_cooldown_timer; /* Delay between loot actions on the corpse entity */
 	EQEmu::TintProfile item_tint;
+	std::vector<std::string> consent_names;
```
Akkadius: Would it make more sense for this to be `consented_player_names` ? I could mistake this as a few different things

Akkadius:
Looks overall solid to me, try not to use abbreviations for variables and just use the full name/description even if it is verbose

Left a few comments, once those are set - we're good

Thanks!

hgtw:
Yep, local testing all checked out

PR #962, opened 2020-02-03
Implement DefaultGuild Rule

noudess:
This allows server owners to assign all new characters created to a default guild of their choosing,

Default behavior is that the rule if off - as normal.

I see this good for:

- local servers where everyone is in same guild
- all servers that want new characters to be born into a community.

### DIFF
--- a/common/database.cpp
+++ b/common/database.cpp
@@ -343,15 +343,9 @@ bool Database::ReserveName(uint32 account_id, char* name) {
 	int guild_id = RuleI(Character, DefaultGuild);
 
 	if (guild_id != 0) {
-		int char_id=-1;
-		query = StringFormat("select `id` FROM `character_data` WHERE `name` = '%s'", name);
-		results = QueryDatabase(query);
-		for (auto row = results.begin(); row != results.end(); ++row) {
-			char_id = atoi(row[0]);	
-			}
-
-		if (char_id > -1) {
-			query = StringFormat("INSERT INTO `guild_members` SET `char_id` = %i, `guild_id` = '%i'", char_id, guild_id);
+		int character_id=results.LastInsertedID();
+		if (character_id > -1) {
+			query = StringFormat("INSERT INTO `guild_members` SET `char_id` = %i, `guild_id` = '%i'", character_id, guild_id);
 			results = QueryDatabase(query);
 			if (!results.Success() || results.ErrorMessage() != ""){
 				LogInfo("Could not put character [{}] into default Guild", name);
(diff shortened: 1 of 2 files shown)


Akkadius:
This can easily be done via script

noudess:
So this should not be done in code then?  Where would I add the scripting?  I need it at character creation not login - unless there is some way to detect that they're new?

Akkadius:
Here is a script example, you use a one time flag when a player initially connects using a flag

```perl
sub EVENT_CONNECT {
	if ($ulevel == 1) {
		my $flag_key = $client->CharacterID() . "-character-creation";
		my $character_creation_flag_set = quest::get_data($flag_key);
		if ($uguild_id == 0 && !$character_creation_flag_set) {
			quest::setguild(int guild_id, int guild_rank_id);
			quest::set_data($flag_key, 1);
			# Do another character creation things
		}
	}
}
```

It's not that other servers wouldn't use this, its about the most flexible way of handling a common use-case. We can merge this in; but I see a few things needing adjustments

This can be solved for using scripts and then still give you flexibility to do other things

We can also easily export a `EVENT_CHARACTER_CREATE` event and have that triggered one time at the player script level so you can utilize logic via `global_player` - utilize existing quest script calls and then call additional things. To me I see the use in having an event like this added so that it gives server operators the full range of quest API methods to not only add a default guild but add whatever else to the character creation process that they deem fit

common/database.cpp
```diff
@@ -338,6 +338,27 @@ bool Database::ReserveName(uint32 account_id, char* name) {
 	query = StringFormat("INSERT INTO `character_data` SET `account_id` = %i, `name` = '%s'", account_id, name);
 	results = QueryDatabase(query);
 	if (!results.Success() || results.ErrorMessage() != ""){ return false; }
+	
+	// Put character into the default guild if rule is being used.
+	int guild_id = RuleI(Character, DefaultGuild);
+
+	if (guild_id != 0) {
+		int char_id=-1;
```
Akkadius: Please use full `character_id` and use the ID from the insert above via `results.LastInsertedID()` instead of making an additional unnecessary call

common/database.cpp
```diff
@@ -338,6 +338,27 @@ bool Database::ReserveName(uint32 account_id, char* name) {
...
+		query = StringFormat("select `id` FROM `character_data` WHERE `name` = '%s'", name);
+		results = QueryDatabase(query);
+		for (auto row = results.begin(); row != results.end(); ++row) {
+			char_id = atoi(row[0]);	
+			}
+
+		if (char_id > -1) {
+			query = StringFormat("INSERT INTO `guild_members` SET `char_id` = %i, `guild_id` = '%i'", char_id, guild_id);
```
Akkadius: Please use `fmt::format` and inline it with `QueryDatabase` if it is single line

See: https://github.com/EQEmu/Server/blob/f73f72b2b2d242d4eedb904d9ce2e95bc2a5541e/zone/spawngroup.cpp#L189

KinglyKrab:
I'm fine with this, I understand it could be done inside a script, but I believe this is a lot easier for some newer server operators to use.

common/database.cpp
noudess: I followed the example of the code I was changing.  If I change more code, I'd asked to keep the changes as small as possible.  If I change less code, but stay with the calls being used, I'm asked to use a different call.

common/database.cpp
noudess: I had no idea this existed.   Changing it now.

PR #964, opened 2020-02-05
Feature - Hot Reload

Akkadius:
This only is implemented for quests to start, can easily be extended to many other data types

HRM client CLI utility is in another codebase that will be available for download and documented soon

### DIFF
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -766,7 +766,7 @@ RULE_CATEGORY_END()
 RULE_CATEGORY(HotReload)
 RULE_BOOL(HotReload, QuestsRepopWithReload, true, "When a hot reload is triggered, the zone will repop")
 RULE_BOOL(HotReload, QuestsRepopWhenPlayersNotInCombat, true, "When a hot reload is triggered, the zone will repop when no clients are in combat")
-RULE_BOOL(HotReload, QuestsResetTimersWithReload, true, "When a hot reload is triggered, the zone will repop")
+RULE_BOOL(HotReload, QuestsResetTimersWithReload, true, "When a hot reload is triggered, quest timers will be reset")
 RULE_CATEGORY_END()
 
 #undef RULE_CATEGORY
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -762,6 +762,12 @@ RULE_CATEGORY(Logging)
 RULE_BOOL(Logging, PrintFileFunctionAndLine, false, "Ex: [World Server] [net.cpp::main:309] Loading variables...")
 RULE_CATEGORY_END()
 
+RULE_CATEGORY(HotReload)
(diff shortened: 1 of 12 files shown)


zone/zone_reload.cpp
```diff
@@ -0,0 +1,46 @@
...
+ *
+ */
+
+#include "zone_reload.h"
+#include "quest_parser_collection.h"
+
+void ZoneReload::HotReloadQuests()
+{
+	BenchTimer timer;
+	timer.reset();
```
mackal: It's not required to call BenchTimer::reset() here.

Akkadius: Looks like some copy pasta; will remove, good eyes

xackery:
I love this feature.

Just to toss into consideration, instead of naming this feature set hot reload, we look at extending #reloadquest, adding in optional arguments to the command to emulate these behaviors? Then console emulates this behavior as well.

`#reloadquest <isForced> <isRepop> <timer> <attempts>`

* isForced = force a quest reload. If false, player combat will abort this command
* isRepop = on reload success, also repop the zone
* timer = time (in seconds) until this command is ran.
* attempts = number of times to attempt to reload, if 0, try infinite

if you pass a bare #reloadquest, it executes `#reloadquest true false 0 0`, to emulate behavior expected as of now

Akkadius:
The HRM utility does not go through the #command interface; it will communicate over the back channel API automatically through listening to file changes

The idea is to take away from having to type `#reloadquest` at all and it all happens automatically to give developers a tight feedback loop.

The idea would also extend to other static data types where if there is a change detected; the HRM module will automatically issue a reload for that zone (also tightening the feedback loop)

image

image

This will be extended to web API functionality as well from the admin panel

common/ruletypes.h
```diff
@@ -763,6 +763,12 @@ RULE_CATEGORY(Logging)
 RULE_BOOL(Logging, PrintFileFunctionAndLine, false, "Ex: [World Server] [net.cpp::main:309] Loading variables...")
 RULE_CATEGORY_END()
 
+RULE_CATEGORY(HotReload)
+RULE_BOOL(HotReload, QuestsRepopWithReload, true, "When a hot reload is triggered, the zone will repop")
+RULE_BOOL(HotReload, QuestsRepopWhenPlayersNotInCombat, true, "When a hot reload is triggered, the zone will repop when no clients are in combat")
+RULE_BOOL(HotReload, QuestsResetTimersWithReload, true, "When a hot reload is triggered, the zone will repop")
```
prestanneth: RULE_BOOL(HotReload, QuestsResetTimersWithReload, true, "When a hot reload is triggered, quest timers will be  reset")

Akkadius: Good eye! When I sit down to finish this all of I'll sort out the copy pastas and get all the docs typed up around this feature

PR #971, opened 2020-02-10
Eqemu server db updates

Uleat:
Top four commits... (all but most recent have notes associated with the commit.)

Starting with earliest:

- Change to address issue of bot updates not allowing mainstream updates and not setting `bots_version` correctly.

- Removed unneeded code relating to bot rules and commands. (server auto-injects/-removes them now.)

- Rework up update process to allow 'current state' evaluation of database when determining whether an update needs to be applied.

- Added missing `return` statement.


alimalkhalifa:
I don't know where I get off on commenting on PRs but I like this 😄 makes a lot more sense to check first then run logic

Uleat:
I'm gonna tweak a couple of things..then, probably merge it in this weekend.

KinglyKrab:
The Perl looks fine to me, logic is also sound. I was having the issue described here myself.

I haven't tested the Perl, but as long as it fixes the issue and doesn't introduce new ones, I approve. 👍

PR #972, opened 2020-02-14
Added skills rebuild to NPC::SetLevel

xackery:
Reasoning: 
image
left is a mob scaled from lvl 14 to 70, you can see the calculated stats are off vs the Elite Guardian of Ro on right.

After this fix:
Screen Shot 2020-02-14 at 3 29 34 AM

So, now it's possible to "scale" an NPC using my scaling plugin code.

### DIFF
--- a/zone/npc.cpp
+++ b/zone/npc.cpp
@@ -2200,7 +2200,13 @@ void NPC::SetLevel(uint8 in_level, bool command)
 	if(in_level > level)
 		SendLevelAppearance();
 	level = in_level;
-	SendAppearancePacket(AT_WhoLevel, in_level);
+	SendAppearancePacket(AT_WhoLevel, in_level);	
+	
+	int r;
+	for (r = 0; r <= EQEmu::skills::HIGHEST_SKILL; r++) {
+		skills[r] = database.GetSkillCap(GetClass(), (EQEmu::skills::SkillType)r, level);
+	}
+
 }
 
 void NPC::ModifyNPCStat(const char *identifier, const char *new_value)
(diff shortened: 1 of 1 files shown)


mackal:
This isn't scaling skills exactly like the ctor does.

xackery:
I'm not sure what ctor is.. from what I could deduce, when a mob is created (NPC::NPC) <https://github.com/EQEmu/Server/blob/2db47adf7b92762ad21df108d0d303e87a5eb0c7/zone/npc.cpp#L303> this is the logic it follows. There's additional modifiers to dual wield and double attack applied after, which I can put in. I mainly wanted this as a baseline to begin the process. Let me know what other areas of improvement or approached would be preferred, if any.

mackal:
Basically, live NPCs follow such a common rate of those that it's very likely those skills seem to be set like the overrides for the vast majority of NPCs, so it should be here too. I would also like a way to skill this re-skilling, just in case someone doesn't want to.

xackery:
Yeah, we chatted about this a bit in coders, and this may have adverse effects on people using SetLevel and expecting skills not to scale.

Perhaps a NPC::RecalculateSkills() ?
So SetLevel retains it's old behavior, and a follow up call to RecalculateSkills() will align skills to the level of an NPC, in cases like mine where an NPC is dynamically being created..

prestanneth:
PR - After skill update goes from 9 to 36, but other guy has 64?

xackery:
That was another concern that I resolved, @prestanneth . Can be safely ignored

xackery:
This PR is likely to be closed, and we'll work towards a new one with a less aggressive impact.

xackery:
Closing this, will do a new PR.

PR #973, opened 2020-02-16
Add Necromancer pet spell effect id to pet spell reagent check

TheGrandPackard:
The rule that was created does not include create undead pet spell effects, which have a different id than standard pets (magician, enchanter, shaman...) and therefore still require bone chips for Necromancers and Shadowknights. This will fix that functionality.


KinglyKrab:
Looks good to me. 👍

PR #974, opened 2020-02-16
Added a rule to disallow click training of tomes and fall back to handins.

noudess:
The default for the rule will be to keep behavior the same for all servers.

Only setting it to true (default is false) will disable click-to-train.

Also, if servers add a MeleeSkillLearn to book to their books Table, whatever text they put there will end up as the Item description:   This addresses an issue I found way back in a 2010 post:

http://www.eqemulator.org/forums/showthread.php?p=264511#post264511

image


mackal:
This seems good to me.

KinglyKrab:
Looks good. 👍

mackal:
Mercs and Bots still use that ItemEnduranceRegenCap rule ... (they need to be rewritten I guess)

noudess:
> Mercs and Bots (edit: oh, clients too lol, missed that) still use that ItemEnduranceRegenCap rule ... (they need to be rewritten I guess)

I hope you know what's going on there?  I added that rule back - so this would compile?  But I dont know why it was gone in the first place?

Just tell me what to do.

PR #977, opened 2020-02-20
CheckIncreaseSkill() for Throwing within combat abilities

prestanneth:
Moved CheckIncreaseSkill() for Throwing (51) so it gets called during combat abilities as currently they skip the function that contains the skill up check.

Combat abilities call DoThrowingAttackDmg() function directly,  whereas regular throws get to the above function via a call to ThrowingAttack() where skill up check resides.

Tested on a zerker volley ability, and monk ranged throws.
The only thing noted was that the skill up check does fire for each proc of a volley.
General consensus is that that is how it should be, i cant find any evidence nor remember either way.

/thoughts?


alimalkhalifa:
Just for the sake of argument, why not put in CheckIncreseSkill in line 1010 instead? Would save us all those is client checks and I find it more logical to check at skill use time instead of at dmg time.

alimalkhalifa:
Just for the sake of argument, why not put in CheckIncreseSkill in line 1010 instead? Would save us all those is client checks and I find it more logical to check at skill use time instead of at dmg time.

alimalkhalifa:
Just for the sake of argument, why not put in CheckIncreseSkill in line 1010 instead? Would save us all those is client checks and I find it more logical to check at skill use time instead of at dmg time.

mackal:
I'm not saying this is bad or anything, just that I'm not sure if this is the best place to put it. I just haven't had time to sit down and see if there is a better spot yet. I'm not sure @alimalkhalifa suggestion is best either.


If anyone else needs some help deciding on best spot, this does need to happen before hit/miss checks etc. It shouldn't matter if dmg happens or not. If attempt of skill to be used, it should check for skill ups.

prestanneth:
Putting it in 1010 would do nothing. That whole functions not called in combat abilities as it doesn't follow the projectile attack route. They are fire off of spells as a spell effect SE_SkillAttack.
So its not about damage in itself, it just happens to be the function it calls directly:

`	case EQEmu::skills::SkillThrowing:`
`	caster->DoThrowingAttackDmg(this, nullptr, nullptr, spells[spell_id].base[i],spells[spell_id].base2[i], focus,  ReuseTime);`


[spell_effects.cpp](https://github.com/EQEmu/Server/blob/master/zone/spell_effects.cpp#L2333)

alimalkhalifa:
Thanks for the insight! I was trying to trace it last night but got too sleepy. Would it make sense to put a second checkskillincrease there instead? Thoughts?

prestanneth:
Ideally it wouldn't be necessary. Makes more sense to put it in one place.
The main reasoning for its current place in pr is due to :
1. The function is called directly.
2. Put it after the LaunchProjectile return() as:
 "If LaunchProjectile is false this function will do archery damage on target,
	otherwise it will shoot the projectile at the target, once the projectile hits target
	this function is then run again to do the damage portion"

I was worried some edge case might call it twice if was done prior to above.

Been testing out Rangers etc as they have same problem, like spells that use skill 7 archery don't call the skillincrease check either.
OPCombatAbility has some checks specific to classes/skills, but plops rangers/zerkers/warriors/etc together and just does a combined check on kicks only.
Could split it out there to a more class/skill check, but it doesn't cover all scenarios either. 
Gets hit when using the "ranged attack" button, but not when using /autoattack .. 
/ffs nightmare really..
Mostly I wish to god the client didnt kill the zone if you breakpoint for more than 2 seconds..

alimalkhalifa:
Any thoughts on moving all skill increase checks to the opcombatability and autoattack methods?

mackal:
> Any thoughts on moving all skill increase checks to the opcombatability and autoattack methods?

that wouldn't fix the issue this is addressing.

PR #979, opened 2020-02-22
Send packet uncompressed if zlib deflates to a larger size than input

hgtw:
commit message has more information
script used for testing: https://gist.github.com/hgtw/8f0eac12dcaa9e86bb1f004f707d81f9


alimalkhalifa:
I think this check is worth the time comparing sizes.

However, do we have a way to permanently disabled compression on certain packets by type? Maybe pp should never be compressed?

Akkadius:
image

PR #981, opened 2020-03-01
Improved vscode devcontainer setup

xackery:
This adds:
* docker-in-docker support
* moved devcontainer Dockerfile to https://github.com/xackery/eqemudevcontainer (accessible via eqemu/devcontainer) to speed up build process (existing Dockerfile ran all commands locally, and takes significant time to build the container for development)
* Added extensions to streamline spinning up environment
* using eqemu/server for the runtime, so valgrind/gdb can be used
* manage docker network (eqemu) for spin down/up so each task is managed within vscode 
Screen Shot 2020-02-29 at 9 41 32 PM
* database priming and eqemu_server script execution

Things to add later:
* eqemu_config.json and login.json for this dev environment (the address/localaddress fields must be manually set due to each instance running in it's own container, which was needed for port allocation)


alimalkhalifa:
Can we have a VS2019 version as well? :D

alimalkhalifa:
I don't see how this can hurt anyone. I vote let us merge this

PR #984, opened 2020-03-05
Fix /split while in a raid

mackal:
This still doesn't add support for /autosplit in a raid, how should that
work?

This changes the Raid::SplitMoney to take a group ID and fails when
provided with RAID_GROUPLESS. This does change behavior, but I'm not
sure if it was ever used so ...

Reported here: http://www.projecteq.net/forums/index.php?threads/cant-split-when-raiding.16514/#post-78810

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -13327,13 +13327,17 @@ void Client::Handle_OP_Split(const EQApplicationPacket *app)
 	Split_Struct *split = (Split_Struct *)app->pBuffer;
 	//Per the note above, Im not exactly sure what to do on error
 	//to notify the client of the error...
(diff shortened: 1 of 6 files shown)


Akkadius:
Seems odd to force a call to a group through the raid class when we could call the group directly

Do we not have access to a simple group object even though the group is in the raid?

mackal:
No, that's the main issue with raids :P

Akkadius:
Until we have a cleaner way of interfacing with raids / groups I guess I'm fine with it. My only beef is `gid` just use full variable names `group_id`

mackal:
Other raid code uses `gid` for the variable name. So it's at least consistent with that :P

PR #985, opened 2020-03-07
Deleted characters will no longer show in the guild window.

KinglyKrab:


mackal:
I think this is fine, but I think Akkadius would be a better person to do a full review

Akkadius:
Looks good!

PR #986, opened 2020-03-08
Faction conning indifferent regardless of faction level in no-combat zones.

regneq:


mackal:
You can grind up POK faction and they'll still /con indiff, also affects pricing, so kind of important.

PR #987, opened 2020-03-08
Add GetGroupIDByCharID and GetGuildIDByCharID to Perl/Lua.

KinglyKrab:


mackal:
This seems reasonable. I don't think there really is a better universal way to do this besides hitting the DB.

KinglyKrab:
Yeah, I had to use DBI to even do this for some instancing stuff. I likely need to add GetRaidIDByCharID() as well in the future.

common/database.cpp
```diff
@@ -2158,6 +2158,28 @@ uint32 Database::GetGuildIDByCharID(uint32 character_id)
...
+uint32 Database::GetGroupIDByCharID(uint32 character_id)
+{
+	std::string query = fmt::format(
+        SQL(
+            SELECT groupid
+            FROM group_id
+            WHERE charid = '{}'
+        ),
+        character_id
+    );
```
Akkadius: Formatting?

zone/questmgr.h
```diff
@@ -251,6 +251,8 @@ class QuestManager {
 	const char* varlink(char* perltext, int item_id);
 	std::string saylink(char *saylink_text, bool silent, const char *link_name);
 	const char* getguildnamebyid(int guild_id);
+    int getguildidbycharid(uint32 char_id);
```
Akkadius: Formatting: indent

zone/questmgr.cpp
```diff
@@ -2876,6 +2876,20 @@ const char* QuestManager::getguildnamebyid(int guild_id) {
...
 
+int QuestManager::getguildidbycharid(uint32 char_id) {
+    if (char_id > 0)
+        return database.GetGuildIDByCharID(char_id);
+    else
+        return 0;
+}
+
+int QuestManager::getgroupidbycharid(uint32 char_id) {
+    if (char_id > 0)
```
Akkadius: No need to do an if else here, just return 0 default always at the bottom since its implied and just return out like you are if we return a database value

Formatting: Please use brackets

common/database.cpp
KinglyKrab: Are you referring character_id and charid or the newlines?

Akkadius: image

Closing parenthesis

KinglyKrab: Ah, okay, will resolve.
My editor's formatting is off.
Using VSCode with TabSize 4 and not using Spaces as Tabs.

KinglyKrab:
Should be formatted properly and everything now, going for a merge.

PR #988, opened 2020-03-09
Add GetRaidIDByCharID to Perl/Lua.

KinglyKrab:


common/database.cpp
```diff
@@ -2180,6 +2180,22 @@ uint32 Database::GetGroupIDByCharID(uint32 character_id)
 	return atoi(row[0]);
 }
 
+uint32 Database::GetRaidIDByCharID(uint32 character_id) {
+	std::string query = fmt::format(
```
Akkadius: If the statement could fit on one line easily like this; it's fine to inline it and not use the `SQL()` macro

The `SQL` macro is designed to allow multi-line statements without requiring cludgy formatting like quotes

Not saying it needs to change this time but just FYI

KinglyKrab: Thanks for the heads up, will keep that in mind.

Thought it was just general practice now.

PR #990, opened 2020-03-11
Modified eqemu_server.pl to pull latest PEQ

alimalkhalifa:
* Also added an escape to skip pulling Maps (tighter feedback loop if
developing)

### DIFF
--- a/utils/scripts/eqemu_server.pl
+++ b/utils/scripts/eqemu_server.pl
@@ -52,6 +52,10 @@ if (-e "eqemu_server_skip_update.txt") {
     $skip_self_update_check = 1;
 }
 
+if (-e "eqemu_server_skip_maps_update.txt") {
+    $skip_self_maps_update_check = 1;
+}
+
 #::: Check for script self update
 check_xml_to_json_conversion() if $ARGV[0] eq "convert_xml";
 do_self_update_check_routine() if !$skip_self_update_check;
@@ -460,7 +464,7 @@ sub do_installer_routines {
         get_remote_file($install_repository_request_url . "libmysql.dll", "libmysql.dll", 1);
     }
 
-    map_files_fetch_bulk();
+    map_files_fetch_bulk() if !$skip_self_maps_update_check;
     opcodes_fetch();
     plugins_fetch();
     quest_files_fetch();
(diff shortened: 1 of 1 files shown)


utils/scripts/eqemu_server.pl
```diff
@@ -1705,26 +1709,22 @@ sub fetch_server_dlls {
...
     find(
         sub { push @files, $File::Find::name unless -d; },
         $start_dir
     );
     for my $file (@files) {
         $destination_file = $file;
-        $destination_file =~ s/updates_staged\/peq_db\///g;
-        if ($file =~ /peqbeta|player_tables/i) {
+        $destination_file =~ s/updates_staged\/peq_db\/peq-dump\///g;
+        if ($file =~ /create_tables_content|create_tables_login|create_tables_player|create_tables_queryserv|create_tables_state|create_tables_system/i) {
```
Akkadius: Why not simply run the `create_tables_all` here ?

utils/scripts/eqemu_server.pl
```diff
@@ -1705,26 +1709,22 @@ sub fetch_server_dlls {
...
             print "[Install] DB :: Installing :: " . $destination_file . "\n";
             get_mysql_result_from_file($file);
         }
     }
-
-    #::: PEQ DB baseline version
-    print get_mysql_result("DELETE FROM db_version");
-    print get_mysql_result("INSERT INTO `db_version` (`version`) VALUES (9130);");
```
Akkadius: I'm glad you caught this

utils/scripts/eqemu_server.pl
alimalkhalifa: I had it doing all but ran into pwd issues. Am I missing something? 😑

Akkadius: You need to be in the context of the directory to source the `all` file as it is referring to files in the same directory context

This is fine

Did you test a full server install with this and did a cursory functionality test?

alimalkhalifa: Yup, my dev box is running this install and all works fine so far!

alimalkhalifa: 1. Tested windows only though

PR #991, opened 2020-03-12
7z tweak

Uleat:
Complete - left for final tweaking

### DIFF
--- a/common/database/database_dump_service.cpp
+++ b/common/database/database_dump_service.cpp
@@ -95,7 +95,7 @@ bool DatabaseDumpService::IsTarAvailable()
  */
 bool DatabaseDumpService::Is7ZipAvailable()
 {
-	std::string version_output = execute("7z -help");
+	std::string version_output = execute("7z --help");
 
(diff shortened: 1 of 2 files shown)


common/database/database_dump_service.cpp
```diff
@@ -93,19 +97,30 @@ bool DatabaseDumpService::IsTarAvailable()
...
+	return false;
 }
 
 /**
  * @return
  */
 bool DatabaseDumpService::HasCompressionBinary()
 {
-	return IsTarAvailable() || Is7ZipAvailable();
+	return IsTarAvailable() || Is7ZipAvailable(std::string());
```
Akkadius: Is there a reason we're passing an empty string here? Can we just as well remove it ?

Uleat:
We need a means of knowing which command path was successfully found - if any.

Otherwise, we have to iterate the list again when we call for the actual the compression.

You could end up calling for compression twice without a means of knowing which one to use.

Uleat:
Dirty...

PR #992, opened 2020-03-12
#petsetname newname command

prestanneth:
#petsetname [newname] to set a persistent name , or command by itself to revert to generated name.

zoning/camping/death/new pet etc retain custom name until its cleared.
Suspended pets get custom name upon unsuspending. 
When pet is initially created the original generated name stays in place, with new name being stored in a separate field on pet[0]

left temp version #petname in place for now.

### DIFF
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -5663,7 +5663,7 @@ void Client::SuspendMinion()
 	else
 	{
 		uint16 SpellID = CurrentPet->GetPetSpellID();
-		std::string OrigName = database.GetOriginalPetName(0, this->character_id);
+		std::string original_name = database.GetOriginalPetName(0, this->character_id);
 		if (SpellID)
 		{
 			if (m_suspendedminion.SpellID > 0)
@@ -5694,7 +5694,7 @@ void Client::SuspendMinion()
 					CurrentPet->GetPetState(m_suspendedminion.Buffs, m_suspendedminion.Items, m_suspendedminion.Name);
 				}
 
-				strn0cpy(m_suspendedminion.Name, OrigName.c_str(), 64);
+				strn0cpy(m_suspendedminion.Name, original_name.c_str(), 64);
 				MessageString(Chat::Magenta, SUSPEND_MINION_SUSPEND, CurrentPet->GetCleanName());
 
 				CurrentPet->Depop(false);
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -5599,15 +5599,15 @@ void Client::SuspendMinion()
 
 	int AALevel = GetAA(aaSuspendedMinion);
 
-	if(AALevel == 0)
+	if (AALevel == 0)
 		return;
 
-	if(GetLevel() < 62)
+	if (GetLevel() < 62)
 		return;
 
-	if(!CurrentPet)
+	if (!CurrentPet)
 	{
-		if(m_suspendedminion.SpellID > 0)
+		if (m_suspendedminion.SpellID > 0)
 		{
 			if (m_suspendedminion.SpellID >= SPDAT_RECORDS) {
 				Message(Chat::Red, "Invalid suspended minion spell id (%u).", m_suspendedminion.SpellID);
@@ -5620,13 +5620,13 @@ void Client::SuspendMinion()
 
 			CurrentPet = GetPet()->CastToNPC();
 
-			if(!CurrentPet)
+			if (!CurrentPet)
 			{
 				Message(Chat::Red, "Failed to recall suspended minion.");
 				return;
 			}
 
-			if(AALevel >= 2)
+			if (AALevel >= 2)
 			{
 				CurrentPet->SetPetState(m_suspendedminion.Buffs, m_suspendedminion.Items);
 
@@ -5664,23 +5664,23 @@ void Client::SuspendMinion()
 	{
 		uint16 SpellID = CurrentPet->GetPetSpellID();
 		std::string OrigName = database.GetOriginalPetName(0, this->character_id);
-		if(SpellID)
+		if (SpellID)
 		{
-			if(m_suspendedminion.SpellID > 0)
+			if (m_suspendedminion.SpellID > 0)
 			{
-				MessageString(Chat::Red,ONLY_ONE_PET);
+				MessageString(Chat::Red, ONLY_ONE_PET);
 
 				return;
 			}
-			else if(CurrentPet->IsEngaged())
+			else if (CurrentPet->IsEngaged())
 			{
-				MessageString(Chat::Red,SUSPEND_MINION_FIGHTING);
+				MessageString(Chat::Red, SUSPEND_MINION_FIGHTING);
 
 				return;
 			}
-			else if(entity_list.Fighting(CurrentPet))
+			else if (entity_list.Fighting(CurrentPet))
 			{
-				MessageString(Chat::Blue,SUSPEND_MINION_HAS_AGGRO);
+				MessageString(Chat::Blue, SUSPEND_MINION_HAS_AGGRO);
 			}
 			else
 			{
@@ -5693,7 +5693,7 @@ void Client::SuspendMinion()
 				if (AALevel >= 2) {
 					CurrentPet->GetPetState(m_suspendedminion.Buffs, m_suspendedminion.Items, m_suspendedminion.Name);
 				}
-				
+
 				strn0cpy(m_suspendedminion.Name, OrigName.c_str(), 64);
 				MessageString(Chat::Magenta, SUSPEND_MINION_SUSPEND, CurrentPet->GetCleanName());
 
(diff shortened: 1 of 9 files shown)


KinglyKrab:
Seems like something that could be handled by a quest and doesn't necessarily need to be a GM command. Code looks fine though, will leave it up to Akkadius or someone else to merge it in.

alimalkhalifa:
I agree that this seems like it is a very specific command. Maybe we can ask the server ops to chime in? Is this a core feature that they would like or something better served being a lua plugin

alimalkhalifa:
discordsnip

alimalkhalifa:
However,

image

zone/pets.cpp
```diff
@@ -405,6 +405,19 @@ void Mob::MakePoweredPet(uint16 spell_id, const char* pettype, int16 petpower,
 
 	entity_list.AddNPC(npc, true, true);
 	SetPetID(npc->GetID());
+	// save updated clients petinfo
+	this->CastToClient()->Save(1);
```
mackal: This isn't safe.

zone/zonedb.cpp
```diff
@@ -3770,15 +3770,25 @@ void ZoneDatabase::SavePetInfo(Client *client)
 		petinfo = client->GetPetInfo(pet);
 		if (!petinfo)
 			continue;
+		
+		std::string Orig = GetOriginalPetName(pet,client->CharacterID());
+		if (Orig == "")
```
mackal: Can check against std::string::empty instead. (Technically this is creating another string object currently, no idea if the compiler is smart enough to avoid it though)

zone/client.cpp
```diff
@@ -5711,6 +5709,20 @@ void Client::SuspendMinion()
 			return;
 		}
 	}
+		//get db and struct in sync.
+		if (GetPet() && GetPet()->CastToNPC()->GetPetSpellID() && !dead) {
```
mackal: Formatting is off here

zone/zonedb.h
```diff
@@ -451,6 +451,8 @@ class ZoneDatabase : public SharedDatabase {
 	bool		SetSpecialAttkFlag(uint8 id, const char* flag);
 	bool		GetPetEntry(const char *pet_type, PetRecord *into);
 	bool		GetPoweredPetEntry(const char *pet_type, int16 petpower, PetRecord *into);
+	std::string GetCustomPetName(uint32 char_id);
```
mackal: Formatting is off here

alimalkhalifa:
I suggest char[64] for the name as that seems to be the standard in eqemu

zone/client.cpp
```diff
@@ -5663,7 +5663,7 @@ void Client::SuspendMinion()
 	else
 	{
 		uint16 SpellID = CurrentPet->GetPetSpellID();
-
+		std::string OrigName = database.GetOriginalPetName(0, this->character_id);
```
alimalkhalifa: Variable name: should be something like `original_name`

zone/zonedb.cpp
```diff
@@ -3770,15 +3770,25 @@ void ZoneDatabase::SavePetInfo(Client *client)
 		petinfo = client->GetPetInfo(pet);
 		if (!petinfo)
 			continue;
+		
+		std::string Orig = GetOriginalPetName(pet,client->CharacterID());
```
alimalkhalifa: Varname: `original_name` verbose but descriptive 😍

zone/zonedb.cpp
```diff
@@ -3946,6 +3956,40 @@ void ZoneDatabase::LoadPetInfo(Client *client)
...
+std::string ZoneDatabase::GetOriginalPetName(int petid, uint32 char_id) {
+	std::string query = fmt::format("SELECT `petname` FROM character_pet_info WHERE `pet` = {} and char_id = {} LIMIT 1", petid, char_id);
+	auto results = QueryDatabase(query);
+	if (!results.Success()) {
+		return "";
+	}
+	if (results.RowCount() != 1)
+		return "";
+	auto row = results.begin();
+	std::string namefound;
```
alimalkhalifa: Varname again

zone/zonedb.h
prestanneth: Hrmm, whole files formatting is messed up/inconsistent.
[https://github.com/EQEmu/Server/blob/master/zone/zonedb.h#L426](https://github.com/EQEmu/Server/blob/master/zone/zonedb.h#L426)

Ironically i see my edit like:
image

zone/client.cpp
prestanneth: yup corrected..

zone/zonedb.cpp
prestanneth: 4 or so methods could be used, but yeah looking at emu code, .empty is used much more than anything else. It uses a size comparative, so possibly slightly less overhead.
Switched to s.empty() for consistency.

zone/pets.cpp
prestanneth: I should wrap inside 
if (this->IsClient()) { 

per say:
https://github.com/EQEmu/Server/blob/master/zone/groups.cpp#L324

prestanneth:
Interesting, I see folks have done it via quests. Saving to buckets? I figured it would be preferable to make source version work better. either or i guess.

_### I agree that this seems like it is a very specific command. Maybe we can ask the server ops to chime in? Is this a core feature that they would like or something better served being a lua plugin_

Clarifying, the #petname function is already in the emu, but it's not very good, as in it only lasts as long as the pet. Not sure how that its useful in current form.
pet dies name gone, you zone name gone, suspended pet doesn't get it etc.. 
You have to macro button it and click it every time you make a new pet, which is irritating.

Maybe i'm biased, being a mage/druid combo mostly forever.. but i reckon you'd be hard pressed to find a pet class player who doesn't want to name it without clicking a button every time.

mackal:
It should be a server function given that https://items.eqresource.com/items.php?id=64705 exists. This just needs to be able to do a quest call (the spell is scripted)


The actual details of how this item works is actually really dumb. It permanently attaches an NPC to you that constantly manages the pets name. Sometimes this NPC goes missing on live and stops working :P.  We obviously don't need to follow the exact mechanism they use, but we should have something implemented to make it work.

alimalkhalifa:
This PR has been open for a long time. I have seen the light and agree.

Would we be ok to merge this if OP takes logic out of command and into somewhere more appropriate (maybe Pet class) then uses the method from both command and a lua/perl binding?

noudess:
#petname already exists

zone/zonedb.cpp
```diff
@@ -3946,6 +3956,40 @@ void ZoneDatabase::LoadPetInfo(Client *client)
 	}
 }
 
+
+std::string ZoneDatabase::GetCustomPetName(uint32 char_id) {
```
Akkadius: Can be simplified

Example:
```
std::string ZoneDatabase::GetCustomPetName(uint32 character_id) {
	std::string query = fmt::format("SELECT `custompetname` FROM character_pet_info WHERE `pet` = 0 and character_id = {} LIMIT 1", character_id);
	auto results = QueryDatabase(query);
	if (results.Success() && results.RowCount() != 1) {
		auto row = results.begin();
		return row[0];
	}

	return std::string();
}
```

Akkadius: `character_id` versus `char_id`

prestanneth:
image


Updating as hadnt for a bit, and i know this pings the discord bot every day.
My works gone mental busy with this virus shit at the moment so haven't spent much time here.
.
.
Really appreciated the detailed feedback , my work needs to take some lessons from that.. we have like no guidelines at all, which i guess makes for sloppiness.
Have tidied up code significantly based on the comments, and combined old function and new one, which was working well, but **only** using a separate field.
.

I went down the rabbit hole yesterday of trying to to use the current _petname_ field rather than new one..
Its sounded simple enough but not in my world... per image above, db gets like that if store the name in same field, so i lose track-ability of whether if its custom or not. 
(trying to avoid non custom getting stuck with same name). Then i got in a mess trying to get head around it.
.
.
.
**I feel the simplest thing to do right now is close out pr,  so its not spamming discord, and as mentioned above, I don't know when i will find time to make it better for submission, but would intend to when can.**
.
.
.
.
.
.

**_saved notes:**
calling /pet get lost or "leave" button,  screws everything up.
A player may have set a custom name, which we want to keep during get lost call, but as its same petname field, we lose the knowlege that it is custom.

Field may contain: 
**kebanab000** [client initial]
**BobThePet** [custom]
**Kenanab**  [if calling `GetRandPetName()` ]
This means the next time a pet is created, the pet would always have the same name as last time, fine for custom, but would lose all uniqueness about normal random client names.
i.e if previous pet was called Kenanab, next pet will be called Kenanab as it maybe a custom, may not be, no way of knowing.
.
.
.

**noted issues in code to work with or around and find a whole new way of tracking:**

 `target->TempName(new_name);` // strips the 000 from any names, ( It would be sole difference between a custom or client initial generated one ).
.
` GetRandPetName()` doesnt add the 000 to name, we could tag it on and manage it, though feel this is somewhat weak way to track it.
 .
 Zoning, unsuspending, creating new, all "creates" a new pet, have to add a lot of checks here to figure out what to do with the name.
 all ways needs more db calls to manage, with extra checks on spell_id column for example, as its non 0 when there is a pet, 0 when there is no pet, null when there is no record yet.
.
 Where no db records at all, client generates own name, but in current iteration the pet creation name check hits `GetRandPetName()` , due to db name being blank, needs a null check added for empty records 
.
 client packet `Handle_OP_PetCommands `gets out of sync with petinfo struct in client and says messages like "hi bob is following you master", but pets name on screen and in db is peter 
(haven't figured out yet why but is only on first ever pet)
.
need to add `setname `calls to db to cover off `getpetstate()` calls during client saves, as it tries to mess you up, resetting name to `getname()` which is out of sync in brand new pet.
.
check live if generated pet names are maintained across zoning, not on main mage has name pot applied, check alt._

PR #997, opened 2020-03-21
Fix for invisibility not killing pets

fryguy503:
Generalizing SetInvisible function to handle all kinds of invis and handle pets similarly for all of them


mackal:
This isn't live like. This is classic like though.

alimalkhalifa:
Do we not want this to be flagged with a rule?

PR #1006, opened 2020-03-29
[BUG] Barter was allowing the purchase of bags. Bags could contain an…

fryguy503:
…ything and it would be traded.

Longterm to match Live's functionality, the search function in the buyer window should not return bags or no drop items at all.

### DIFF
--- a/zone/trading.cpp
+++ b/zone/trading.cpp
@@ -2602,6 +2602,11 @@ void Client::SellToBuyer(const EQApplicationPacket *app) {
 		return;
 	}
 
+    if(item->IsClassBag()) {
+        Message(13, "That item is a Bag.");
+        return;
+    }
+
 	if(!item->Stackable) {
 
 		for(uint32 i = 0; i < Quantity; i++) {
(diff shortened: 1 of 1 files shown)


zone/trading.cpp
```diff
@@ -2602,6 +2602,11 @@ void Client::SellToBuyer(const EQApplicationPacket *app) {
 		return;
 	}
 
+    if(item->IsClassBag()) {
```
Akkadius: `if (` vs `if(`

zone/trading.cpp
```diff
@@ -2602,6 +2602,11 @@ void Client::SellToBuyer(const EQApplicationPacket *app) {
...
+        Message(13, "That item is a Bag.");
```
Akkadius: Use `Chat::Red` vs `13`

zone/trading.cpp
```diff
@@ -2973,6 +2978,9 @@ void Client::UpdateBuyLine(const EQApplicationPacket *app) {
 		else if(item->NoDrop == 0)
 			Message(Chat::Red, "Buy line %s disabled as the item is NODROP.", ItemName);
 
+        else if(item->IsClassBag())
+            Message(13, "Buy line %s disabled as the item is a Bag.", ItemName);
```
Akkadius: Use `Chat::Red` vs `13`

fryguy503: Got me on those, copied from my code pre Chat::Color

Akkadius:
Still some weird formatting on this one; tagging @Uleat and @mackal for reviews

mackal:
This is fine. It fixes the exploit, but as the PR says, it's not a complete fix. (This code should probably still exist just in case of packet forging non-sense I guess)

Akkadius:
Seems like formatting got way weirder on this PR

fryguy503:
> Seems like formatting got way weirder on this PR

Per Mackal I just reformatted the two functions and let CLion autoformat it all.

PR #1007, opened 2020-03-31
Improved Random Name Generator

Akkadius:
While testing character creation; I got tired of unreliable name generation. This has shown to be more consistent, clean, re-usable and gives us options to use it elsewhere easily

Use name generator here: https://github.com/skeeto/fantasyname

http://rinkworks.com/namegen/


mackal:
http://www.eqemulator.org/forums/showthread.php?t=42794

Issue #998, opened 2020-03-21
Invis vs Undead/Animals semi-permanent if cast and then char zones


noudess:
If you cast Invisibility to Undead or Invisibility to Animals on yourself, and then zone, you can cast at will in the new zone and neither spell will drop.   You can nuke and buff so long as you do not attack and you are not hit, and the buffs will not drop.

joligario:
This is on RoF clients only.

alimalkhalifa:
This issue is ready to close I believe?

noudess:
My "fix" is really just a hack.  I think someone with client knowledge can figure out why, after zoning, the client isn't taking these down when casting is initiated, like it does with invis.

PR #1009, opened 2020-03-31
Hack to fix RoF2 perma IVU/IVA bug after zoning.

noudess:
This fixes the problem, but I believe someone with better knowledge of the client than I can find a more "correct" solution.

Posted as a PR per @joligario  to solve it for now.  Was also asked to put SE_Invisibility in there so that this "hack" is a catch all.

https://github.com/EQEmu/Server/issues/998

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -4017,6 +4017,14 @@ void Client::Handle_OP_CastSpell(const EQApplicationPacket *app)
 		//Message(Chat::Red, "You cant cast right now, you arent in control of yourself!");
 		return;
 	}
+	
+	// Hack for broken RoF2 which allows casting after a zoned IVU/IVA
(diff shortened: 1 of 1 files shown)


noudess:
> I guess this is fine if it's causing issues.

It was the only way I could find to fix it.  I thought though with your intimate knowledge of the client you could figure out why after zoning the client doesn't take IVI and IVA down like it does Invis?  

I know this is a hack, I just could not figure out how to make the client do the right thing like is already does for invis.   Apparently this is only RoF2 (or our handling on RoF2).

alimalkhalifa:
Likely, there will be a missing packet or a missing flag on one of the
packets, or it might actually send a packet saying that you have IVU/IVA.

Worth capturing packets and seeing what it does.

PR #1010, opened 2020-03-31
Added proximity check for feared clients

noudess:
@Uleat made these changes so that a feared entity can still trigger proximity events.  I've tested this since 3/18 and its working great for me.

### DIFF
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -161,6 +161,7 @@ RULE_BOOL(Character, DismountWater, true, "Dismount horses when entering water")
 RULE_BOOL(Character, UseNoJunkFishing, false, "Disregards junk items when fishing")
 RULE_BOOL(Character, SoftDeletes, true, "When characters are deleted in character select, they are only soft deleted")
 RULE_INT(Character, DefaultGuild, 0, "If not 0, new characters placed into the guild # indicated")
+RULE_BOOL(Character, ProcessFearedProximity, true, "Processes proximity checks when feared")
 RULE_CATEGORY_END()
 
 RULE_CATEGORY(Mercs)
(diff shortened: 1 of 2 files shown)


mackal:
hmmm .... is this actually correct for live? I feel like we should double check live tasks ...

noudess:
I think that's why he added the rule.  As far as I know this is for custom servers.   I used it as a solution to chars being feared through locked doors.  Now, I can control that via invisible npcs and a quest file.

Uleat:
I had a couple of big issues come up and haven't gotten back to this in a while.

Akkadius and I were having a conversation about how/where this should be implemented..so, it's still a wip.

Default would be to false..I only had it true here so noudess wouldn't have to build and then play with database settings.

As far as behavior, this probably is not live-like..but, it does give an option so that unique situations can be handles - like in noudess's case.

Akkadius:
If it’s not live like then gate the rule as false by default and I’m good with it

noudess:
I put a PR into @uleat's branch to change the default value.

alimalkhalifa:
For argument's sake, would we want to instead punch holes in the navmesh dynamically when we have locked doors?

mackal:
I think that might be "easier said than done" atm.


We do need door support in mesh ... live uses doors to modify geometry sometimes and causes a bunch of other issues :(

alimalkhalifa:
I am aware 😑 looked through the code. We need to integrate recast and dynamically alter the meshes, and for that to work, the server needs to learn the concept of space, mesh collision, raytracing etc. Easier said than done indeed but might have to be done at some point

noudess:
I'd certainly be happy to see me not need a quest  mob at each door, but I am very grateful for this workaround.

Issue #849, opened 2019-06-03
In many areas, water mobs can attack those on land above them.


noudess:
In qcat and in rivervale, to name two spots, pirahnas attack up through the ground (qcat) and the bridges (rivervale).

#bestz returns in water in these areas every time.

Seems to have cropped up since we moved to the base/nav/water scheme but I can't say directly which.

Rivervale walk near the bridge or on it around X: -240, Y: 248, Z: 4 when pirahnas are under. 

Qcat the areas around water near the zonein from qeynos by bank.

4eyebiped:
Happens in cabeast as well.  Will get aggro near the water ways and will still be hit on the bridges there as well.

noudess:
This was  pretty bad in runnyeye as well, by the underground rivers.

noudess:
Also on bridge in feerrott.

noudess:
I have a fix in testing.

It removes client from aggro lists of mobs that are water only when they emerge from a water region.

It Prevents npcs from aggroing clients outside of a water region, if that npc is a water only npc.

It heals the npc to full if the client runs from the water and the npc is no longer engaged (as on live)

PR #1017, opened 2020-04-02
Fix Issue #849.

noudess:
https://github.com/EQEmu/Server/issues/849

npcs flagged as underwater_only:

- no longer aggro PCs outside of the water
- PC that flees from the underwater_only npc onto another region type are removed from the hate list of the underwater_only npc
- npcs that have no remaining hate list after PCs flee to land heal as per live


zone/entity.cpp
```diff
@@ -3294,6 +3294,30 @@ void EntityList::ClearAggro(Mob* targ)
...
 }
 
+//removes "targ" from all hate lists of mobs that are water only.
+void EntityList::ClearWaterAggro(Mob* targ)
+{
+	Client *c = nullptr;
+	if (targ->IsClient())
+		c = targ->CastToClient();
+	auto it = npc_list.begin();
+	while (it != npc_list.end()) {
```
Akkadius, line 3304: Crazy formatting happening here

noudess: Duplicated the function above:

//removes "targ" from all hate lists, including feigned, in the zone
void EntityList::ClearAggro(Mob* targ)
{
	Client *c = nullptr;
	if (targ->IsClient())
		c = targ->CastToClient();
	auto it = npc_list.begin();
	while (it != npc_list.end()) {

Do you want me to fix both to have braces?

noudess: I agree.  I hate implied braces.  Too much time working on a military project where "do as they did" was enforced.  Fixing now.

noudess: I fixed the function I added and the one I cloned.  This entire file has a ton of single line ifs without braces.  I feared breaking something if I changed everything.

PR #1020, opened 2020-04-06
Added getcharnamebyid(char_id) to Perl/Lua.

KinglyKrab:
Perl: quest::getcharnamebyid(char_id);
Lua: eq.get_char_name_by_id(char_id);

### DIFF
--- a/common/database.cpp
+++ b/common/database.cpp
@@ -925,6 +925,22 @@ void Database::GetCharName(uint32 char_id, char* name) {
 	}
 }
 
+const char* Database::GetCharNameByID(uint32 char_id) {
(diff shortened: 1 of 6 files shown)


common/database.cpp
```diff
@@ -925,6 +925,22 @@ void Database::GetCharName(uint32 char_id, char* name) {
 	}
 }
 
+const char* Database::GetCharNameByID(uint32 char_id) {
+	std::string query = fmt::format("SELECT `name` FROM `character_data` WHERE id = '{}'", char_id);
```
Akkadius, line 929: `id` does not need quotes around it because its an integer, query plan in the optimizer does change when things are quoted

KinglyKrab: Word, saw examples using it and some without, will fix. 👍

PR #1031, opened 2020-04-11
Code Duplicated? Second instance is in a more correct place 20 or so lines down

prestanneth:
Should first "sanity check" be where it is or remove per PR. 
Makes the next if{} else{} statement pointless doesn't it as value will never be 0.7

for easier understanding:
image

### DIFF
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -163,10 +163,6 @@ Mob::Mob(
 	size          = in_size;
 	base_size     = size;
 	runspeed      = in_runspeed;
-	// neotokyo: sanity check
-	if (runspeed < 0 || runspeed > 20) {
-		runspeed = 1.25f;
-	}
 
 	// clients -- todo movement this doesn't take into account gm speed we need to fix that.
 	base_runspeed = (int)((float)runspeed * 40.0f);


mackal:
This is going to throw off the calculations of base_runspeed, isn't it?

prestanneth:
`base_runspeed = (int)((float)runspeed * 40.0f);`
In testing i found no difference, though at the above point in code there's nothing in **std emu** ever uses the less than 0, > than 20 check. So when it gets here everything's as per database runspeed which is inside these "parameters".

If the check is left in place as a "sanity check," and something was set to a lower speed, it would be overwrite, so the following is never used then.
`if (runspeed == 0.7f) {
		base_runspeed  = 28;
		walkspeed      = 0.3f;
		base_walkspeed = 12;
		fearspeed      = 0.625f;
		base_fearspeed = 25;
		// npcs
	}`


bolded the **std emu** as the only thing i can find that ever hits the "if portion of statement" with a runspeed of 0.7 is bots (after i commented out the check.)
`bot_npc_type->runspeed = 0.7f;`

If first sanity is left in place then bots also would not hit the above as would be set to 1.25?
likewise if first one is left, second one becomes unnecessary..

Dunno if uleat has any insights/thoughts..

Uleat:
That sanity check is a remnant of the old movement speed calculations.

I 'think' it's Haynar's code we're using now.

The first 'sanity check' bounds anything passed into the ctor .. most likely, database values.

Without the first one, it could be possible to skew numbers so that walk speed is faster than runspeed.

To me, it looks like a bad merge and the second is the extra .. but, without thorough testing, I couldn't say for sure...

Bots are a special case since we treat them as other clients to the client.

Otherwise, their runspeed would be the npc default of 1.25f.

mackal:
given what @Uleat said, I think it maybe best to leave it, it shouldn't be too much of a performance hit either way.

prestanneth:
No worries, will leave it be, thanks for inputs.

PR #1032, opened 2020-04-11
Win64 eqemu_server.pl

alimalkhalifa:
Changed binaries to x64 and got rid of wget.exe for windows, switched to
native WebClient

Uses zipped binaries at Akkadius/EQEmuInstall until appveyor is sorted


alimalkhalifa:
Changed source branch, before doing something silly at master

PR #1035, opened 2020-04-13
Fix errant merchant full messages and add logging for later use

noudess:
This change fixes cases where a merchant has at one point used slot 255 and slot 255 is still full but there are plenty of other slots open for use due to player buyback.  Without this fix, the code will throw an error saying the merchant is full.

This fix:

- Utilizes empty merchantlist_temp slots when available, regardless of where they are in the list
- Added LogInventory() statements throughout to aid in any later debugging that might be needed (though this has been running on my server for months)
- Added a DumpIt function which will dump the entire merchant inventory (currently unused) for later debugging if neeeded.
- Fixed a handful of if statements that were missing braces.

Obviously this fix is mainly for those that have rule:  World:ClearTempMerchantlist set to false.

### DIFF
--- a/zone/zone.cpp
+++ b/zone/zone.cpp
@@ -330,81 +330,154 @@ bool Zone::LoadGroundSpawns() {
 	return(true);
 }
 
-int Zone::SaveTempItem(uint32 merchantid, uint32 npcid, uint32 item, int32 charges, bool sold) {
-	int freeslot = 0;
-	std::list<MerchantList> merlist = merchanttable[merchantid];
-	std::list<MerchantList>::const_iterator itr;
-	uint32 i = 1;
-	for (itr = merlist.begin(); itr != merlist.end(); ++itr) {
-		MerchantList ml = *itr;
-		if (ml.item == item)
-			return 0;
-
-		// Account for merchant lists with gaps in them.
-		if (ml.slot >= i)
-			i = ml.slot + 1;
-	}
+void Zone::DumpIt(uint32 npcid) {
 	std::list<TempMerchantList> tmp_merlist = tmpmerchanttable[npcid];
 	std::list<TempMerchantList>::const_iterator tmp_itr;
-	bool update_charges = false;
 	TempMerchantList ml;
-	while (freeslot == 0 && !update_charges) {
-		freeslot = i;
-		for (tmp_itr = tmp_merlist.begin(); tmp_itr != tmp_merlist.end(); ++tmp_itr) {
-			ml = *tmp_itr;
-			if (ml.item == item) {
-				update_charges = true;
-				freeslot = 0;
-				break;
-			}
-			if ((ml.slot == i) || (ml.origslot == i)) {
-				freeslot = 0;
-			}
-		}
-		i++;
+
+	for (tmp_itr = tmp_merlist.begin(); tmp_itr != tmp_merlist.end(); ++tmp_itr) {
+		ml = *tmp_itr;
+		
+		LogInventory("slot[{}] Orig[{}] Item[{}] Charges[{}]", ml.slot, ml.origslot, ml.item, ml.charges);
 	}
-	if (update_charges) {
+}
+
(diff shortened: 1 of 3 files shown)


zone/zone.cpp
```diff
@@ -330,81 +330,154 @@ bool Zone::LoadGroundSpawns() {
 	return(true);
 }
 
+void Zone::DumpIt(uint32 npcid) {
```
Akkadius, line 333: This need to be more specific to merchants, eg: `Zone::DumpMerchantList`

noudess: Good catch.  That was going to be a temporary function, but I figured leave it in, in case details are needed in the future.  Fixing.

noudess: Could someone also request a review from @Uleat ?  I don't have the ability to do that formally.

zone/zone.cpp
```diff
@@ -330,81 +330,154 @@ bool Zone::LoadGroundSpawns() {
...
-			ml = *tmp_itr;
-			if (ml.item == item) {
-				update_charges = true;
-				freeslot = 0;
-				break;
-			}
-			if ((ml.slot == i) || (ml.origslot == i)) {
-				freeslot = 0;
-			}
+	uint32 first_empty_slot=0; // Save 1st vacant slot while searching..
```
KinglyKrab, line 372: Formatting.

zone/zone.cpp
```diff
@@ -330,81 +330,154 @@ bool Zone::LoadGroundSpawns() {
...
+				if (sold) {
+					LogInventory("Total charges is [{}] + [{}] charges", ml.charges, charges);
+					ml.charges = ml.charges + charges;
+				}
+				else {
+					ml.charges = charges;
+					LogInventory("new charges is [{}] charges", ml.charges);
+				}
+				if (!ml.origslot) {
+					ml.origslot = ml.slot;
```
KinglyKrab, line 402: Spacing around all these conditons maybe? I know it's like that by default, but doesn't mean we shouldn't clean it up given the chance.

zone/zone.cpp
```diff
@@ -330,81 +330,154 @@ bool Zone::LoadGroundSpawns() {
...
+		first_empty_slot = idx;
+
+		// Find an ununsed mslot
+		slots.clear();
+		for (tmp_itr = tmp_merlist.begin(); tmp_itr != tmp_merlist.end(); ++tmp_itr) {
+			ml3 = *tmp_itr;
+			slots.push_back(ml3.slot);
+		}			
+		slots.sort();
+		uint32 first_empty_mslot=0;
```
KinglyKrab, line 453: Formatting.

zone/zone.cpp
noudess: I don't see the formatting issue here?  Looking at the full context, this line is lined up correctly. Am I missing something?

zone/zone.cpp
noudess: I'm unclear what you're asking for.

zone/zone.cpp
noudess: Also unclear.  Maybe we see something different based on tab settings?

zone/zone.cpp
noudess: Spaces around the =?  Ok fixed that,.

zone/zone.cpp
noudess: Put spaces around the '='

Uleat:
Doesn't look like it'll affect anything but merchant.

If any issues pop up, they should be easy enough to trace.

zone/zone.cpp
KinglyKrab: Spacing around the conditionals. A newline between some of them. If they're not part of the same logic flow they don't need to be grouped together without newline gaps.

noudess: Okies,

PR #1041, opened 2020-04-17
Add missing break for OP_AppCombined switch statement

peterigz:
This randomly popped into my head that I meant to mention a while back when I was trying to fix the network issue on Windows server. When I was looking at the network code I noticed this switch statement missing a break for OP_AppCombined so some of those packets were spilling over into session request. I remember getting excited because I thought maybe that had something to do with the issue at the time, but it wasn't the case in the end. The missing break doesn't seem to have any major adverse affects but thought I'd quickly point it out with this pull request while I remember.


Akkadius:
Good eye Pete!

I'll nudge KLS to take a look at this one to seal it up

PR #1042, opened 2020-04-21
Fix for static zone issue with mobs congregating after last player zones

noudess:
In static zones, if IDLE_WHILE_EMPTY is used (which is the default), when the last player leaves the zone, mobs cease to path immediately.

This PR adds code to allow these mobs to return to their spawn points before suspending mob::process.  In my testing 60 seconds is a good value for this rule, but servers might want more/less depending on preference.

This resolves issues like trains remaining at zone lines for the next unfortunate PC that zones in, days later.

### DIFF
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -280,6 +280,7 @@ RULE_BOOL(Zone, UseZoneController, true, "Enables the ability to use persistent 
 RULE_BOOL(Zone, EnableZoneControllerGlobals, false, "Enables the ability to use quest globals with the zone controller NPC")
 RULE_INT(Zone, GlobalLootMultiplier, 1, "Sets Global Loot drop multiplier for database based drops, useful for double, triple loot etc")
 RULE_BOOL(Zone, KillProcessOnDynamicShutdown, true, "When process has booted a zone and has hit its zone shut down timer, it will hard kill the process to free memory back to the OS")
+RULE_INT(Zone, SecondsBeforeIdle, 60, "Seconds before IDLE_WHEN_EMPTY define kicks in")
 RULE_CATEGORY_END()
 
 RULE_CATEGORY(Map)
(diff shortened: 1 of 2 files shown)


alimalkhalifa:
I like the thought but would advise against only resuming the pathing system.
I'd suggest to A- use this system but B- have another longer timeout that warps all mobs back to spawn point and pause pathing.

noudess:
> I like the thought but would advise against only resuming the pathing system.
> I'd suggest to A- use this system but B- have another longer timeout that warps all mobs back to spawn point and pause pathing.

This doesn't only resume pathing.  It processes mobs normally after the zone empties for specified seconds.

Also, this code already resumes pathing of grids that despawn so that these mobs dont always spawn and path when someone zones in.  This just allows trained mobs to return through normal processing.

alimalkhalifa:
I'm aware, but the issue you will face is that the world might timeout with the mobs having only pathed part ways, or maybe had some factional issue where they got stuck fighting a griffon or something.

If say in 5 minutes hard timeout hits, it just warps all mobs back to spawn and the zone is fresh available for the next player to wander in.

alimalkhalifa:
I'm aware, but the issue you will face is that the world might timeout with the mobs having only pathed part ways, or maybe had some factional issue where they got stuck fighting a griffon or something.

If say in 5 minutes hard timeout hits, it just warps all mobs back to spawn and the zone is fresh available for the next player to wander in.

alimalkhalifa:
I'm aware, but the issue you will face is that the world might timeout with the mobs having only pathed part ways, or maybe had some factional issue where they got stuck fighting a griffon or something.

If say in 5 minutes hard timeout hits, it just warps all mobs back to spawn and the zone is fresh available for the next player to wander in.

noudess:
I'd argue that this fix lets the world owner decide how long to allow normal processing to take place.  You could set the rule to 300 if you wanted a 5 minute settle time.

I assert my change resolves a bug.  If someone wants further tuning (this problem has been here for years), such as a zone reset, that can be added later.  To me, this is a simple, natural, way for the zone to resume, and the use of a higher value can handle the problem you note.  Although, I really doubt anyone will notice anything after 60 seconds.  A complete warp back seems heavy handed when letting mobs path back, for a configurable time, feels more natural and configurable.  It's also an extension of a working mod.

PR #1045, opened 2020-04-30
Fishing update to match live 2018 patch

noudess:
From July 28th, 2018 patch:

**- Changed a number of fishing pole items to be unbreakable. Fishing poles that have a chance to break while fishing now display the Expendable flag when inspected.**

From EQTrader's research:

http://mboards.eqtraders.com/eq/forum/tradeskills/fishing/368408-unbreakable-fishing-poles

My own research.  I was able to learn that only poles tagged as magic were unbreakable.

I could have coded this, using the magic flag, but that would not have resulted in the "Expendable" tag on the item.  This is a capture from live this week on a breakable pole:

image

As such, I found the only way I could get this tag, in lower case, like live. was to set the expendablearrow field in items.  Perhaps the field should get a name change, but I didn't want to break anything else.

This is working and tested on my server. 

Re: the sql, I played with updates using item type and item name instead of using specific ids.  However too many exceptions existed, like a pail that is tagged as 36 and poles not called poles, etc.

This sql should keep all standard poles breakable.

### DIFF
--- a/common/item_instance.h
+++ b/common/item_instance.h
@@ -103,7 +103,7 @@ namespace EQEmu
 		bool IsAugmentSlotAvailable(int32 augtype, uint8 slot) const;
 		inline int32 GetAugmentType() const { return ((m_item) ? m_item->AugType : 0); }
 
-		inline bool IsExpendable() const { return ((m_item) ? ((m_item->Click.Type == item::ItemEffectExpendable) || (m_item->ItemType == item::ItemTypePotion) || (m_item->ExpendableArrow == 1)) : false); }
+		inline bool IsExpendable() const { return ((m_item) ? ((m_item->Click.Type == item::ItemEffectExpendable) || (m_item->ItemType == item::ItemTypePotion)) : false); }
 
 		//
 		// Contents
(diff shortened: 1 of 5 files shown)


mackal:
I don't think this is the expendable arrow flag controlling this

noudess:
> I don't think this is the expendable arrow flag controlling this

Well, setting it makes our server display the correct graphic.  If you can find another field, tell me and I'll fix it.  It works perfectly.

mackal:
Need to sit down and see if it is in the clients ... Might not be, if it isn't, we can hack the client to show anyways, but server side we should figure out whatever this unknown30 should be called and use that instead of hacking server side to use the arrow flag

noudess:
> Need to sit down and see if it is in the clients ... Might not be, if it isn't, we can hack the client to show anyways, but server side we should figure out whatever this unknown30 should be called and use that instead of hacking server side to use the arrow flag

I agree, but we don't have an unknown30.  So is it a new field we need to add?

noudess:
Modified but not complete.  On hold while @mackal adds support to patchers and items.

Akkadius:
What is still pending with this

mackal:
We need to fully implement SubType for items.

noudess:
I can try this coming week, but was daunted when I first looked at loading a new column into items code.

Akkadius:
New column merged; nice collab guys; awaiting changes post new field

noudess:
Closing this.  Cleaner to open a new one based on current mainline.

PR #1046, opened 2020-04-30
Flying

noudess:
Before this change, flying mobs would attempt to traverse any means they could to get to a target.  If they could not reach the target, they would use whatever stuck_behavior was in place.

This change only impacts mobs that:

- have flymode set to flying
- have a specific target (aka they were pulled or assisting)
- that target can be seen with LoS

In the above case, they will fly directly at the target.  The FlyTo mode also utilizes faster updates to avoid the effect of flying super high and then dropping down at the last minute, as the MoveTo only updates every 5 seconds for ground.  This new FlyTo method also does not use FixZ at all, as they are midair during pathing.

These mobs, if they survive and path back home, will resume ground travel.   This was done as LoS for flying needs a Z to see if they can simply fly there, and without a target we have no size to assume for the target.  I feel it is reasonable for returning mobs to walk back.

This change makes pulling mobs on balconies, that can fly, very elegant and these spots dont have to rely on portals or stuck mode anymore.

### DIFF
--- a/zone/mob_movement_manager.cpp
+++ b/zone/mob_movement_manager.cpp
@@ -263,6 +263,146 @@ protected:
 	double m_total_v_dist;
 };
 
+class FlyToCommand : public IMovementCommand {
+public:
+	FlyToCommand(float x, float y, float z, MobMovementMode mob_movement_mode)
+	{
+		m_distance_moved_since_correction = 0.0;
+		m_move_to_x                       = x;
+		m_move_to_y                       = y;
+		m_move_to_z                       = z;
+		m_move_to_mode                    = mob_movement_mode;
+		m_last_sent_time                  = 0.0;
+		m_last_sent_speed                 = 0;
+		m_started                         = false;
+		m_total_h_dist                    = 0.0;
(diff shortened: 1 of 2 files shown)


zone/mob_movement_manager.cpp
```diff
@@ -936,6 +1078,16 @@ void MobMovementManager::UpdatePath(Mob *who, float x, float y, float z, MobMove
 	else if (who->IsUnderwaterOnly()) {
 		UpdatePathUnderwater(who, x, y, z, mob_movement_mode);
 	}
+	// If we can fly, and we have a target and we have LoS, simply fly to them.
+	// if we ever lose LoS we go back to mesh run mode.
+	else if (target && who->GetFlyMode() == GravityBehavior::Flying &&
```
Akkadius, line 1083: Formatting should be cleaned up but I'm still going to push this through

noudess: @Akkadius Will fix formatting and push.  Is it  just the closing brace is indented too far?

PR #1047, opened 2020-05-06
Don't scale 0 values

isenior:
In this one code path, values could be added to an NPCs hp and max_dmg if they had been set to 0 with the intention of being dynamically scaled. This would cause those fields to be ignored when later scaling the enemy.


mackal:
This level function is live-like, where the stuff @Akkadius added isn't. It's a dev tool, unsure on proper path to take, but I assume if we have the 0 values and max level set, we still want the live like scale function to happen?

PR #1050, opened 2020-05-08
Identify SubType in Item structs in some clients

mackal:
This maybe in older clients as well, but I couldn't verify it and those
clients never did anything with this field so it doesn't matter.

MQ2 calls this SubClass, but we call the field it's a "sub" of Type so I
figured we'd call it SubType.

We still need to rename the DB field and handle it server side etc

### DIFF
--- a/common/patches/rof2_structs.h
+++ b/common/patches/rof2_structs.h
@@ -4804,8 +4804,8 @@ struct ItemQuaternaryBodyStruct
 	int32 HealAmt;
 	int32 SpellDmg;
 	int32 Clairvoyance;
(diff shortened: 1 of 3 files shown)


mackal:
This should be unk219 in the item DB

mackal:
> Looks good. What all is this used for other than fishing pole stuff?

Yes, the fishing pole thing is using sub types, we will have to use a hack as well, I just wanted to get the struct changes merged instead of sitting on them lol

mackal:
There is a spell effect focus limiter (at least in the clients I could identify it in) that checks sub type :P

PR #1054, opened 2020-05-10
Add Item SubType to data structures and DB

mackal:
This can be used by other devs to implement other features now (fishing pole changes, Ff_ItemClass, etc)


mackal:
@noudess

noudess:
Thank you!  Once merged, I'll remove the sql stuff from my PR and make the remaining changes.

common/item_fieldlist.h
```diff
@@ -203,3 +203,4 @@ F(procname)
 F(wornname)
 F(focusname)
 F(scrollname)
+F(subtype)
```
mackal, line 206: should we order this better? Does it matter?

noudess: I defer to you.  I'll add the accessor and change my PR tomorrow.

PR #1061, opened 2020-05-11
New fishing

noudess:
Utilized field added by @mackal (SubType - previously unk219) to make sure fishing pole has a chance of breaking before rolling.

@mackal code also fixes client to show Expendable on poles that break:

image


noudess:
Added code for patchers after @mackal reminded me that I had edited my local db, which was making me believe I was done, rather than the truth that I still needed to overload ExpendableArrow in the patchers.

Added optional SQL for servers that want to make the arrows that were made unbreakable in 2018 unbreakable on their servers.

PR #1067, opened 2020-05-20
Prevent mob from moving from point of aggro until after calls for help are issued.

noudess:
Mob_ai was using the engaged flag to decide when to start navigating toward a target.  This flag is a reflection of the hatelist being non-empty.  As is stood, the hatelist is added to early in the process, and calls for help have not yet been processed.  This resulted in the mob moving, sometimes a significant distance, before calls for help were sent out.

I changed mob_ai to only start moving toward the target after the CombatEvent has been set (which is done by all calls to AI_Event_Engaged).

This wasn't enough, as all special attacks were using callforhelp=false, which resulted in AI_Event_Engaged not calling for help, but setting the flag.

I changes all special attacks to use callforhelp=true and this seems to work.

Pulling via throwing or arrows now properly calls for help BEFORE mob_ai code starts moving the pulled mob toward the target.  Seems good, but seek advice.  

I will be testing this starting now on my server with a player load.

### DIFF
--- a/zone/mob_ai.cpp
+++ b/zone/mob_ai.cpp
@@ -1319,7 +1319,7 @@ void Mob::AI_Process() {
 						FaceTarget();
 					}
 				}
-				else if (AI_movement_timer->Check() && target) {
+				else if (AI_movement_timer->Check() && target && CastToNPC()->GetCombatEvent()) {
 					if (!IsRooted()) {
 						LogAI("Pursuing [{}] while engaged", target->GetName());
 						RunTo(target->GetX(), target->GetY(), target->GetZ());
--- a/zone/mob_ai.cpp
+++ b/zone/mob_ai.cpp
@@ -1321,7 +1321,7 @@ void Mob::AI_Process() {
 				}
 				// mob/npc waits until call for help complete, others can move
 				else if (AI_movement_timer->Check() && target &&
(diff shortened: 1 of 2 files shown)


noudess:
This does - in fact - break sneak/pull - but I'm not sure the code was originally written to make sneak pull work?

noudess:
This is working well on my server.  Mobs now complete their call for help before running off to engage.

noudess:
This needs to be held up.  Found an issue with pets.

Akkadius:
These changes seem reasonable; however the biggest thing that sticks out that we can't keep the way it is right now is the `IsPet` call. This is not necessarily because of your changes but its not a good idea at all with the way `IsPet` is currently written that we put that in a hot path like this because it constantly hits the entity list and it **really** shows its ugly head on a highly populated server

If we're going to use `IsPet` in a hot path like this; I'd like us to also make a change that we set the `IsPet` to a constant value that doesn't hit the entity list every single time its called; so once a Mob becomes a pet we set it once then recall that value continually

Everything else seems fine here

noudess:
GetOwnerID() identifies pets and mercs without entity list hits.  That work?

Akkadius:
That should be good; make sure to update the changelog

PR #1071, opened 2020-05-25
SQL to add faction mods for guktan and drakkin races, and Agnostic mods

noudess:
These mods were missed by the massive faction conversion some time ago.  
These mods are post-conversion - so don't use them unless you've converted.


Akkadius:
@noudess should PEQ be running these? Let me know in Discord

PR #1072, opened 2020-05-28
Updated CMakeLists to default to MariaDB instead of MySQL

alimalkhalifa:
This makes windows x64 compile hands off (no need to edit cmake parameters to get a successful compile)


Uleat:
This will fix a known issue with 32-bit builds as well.

There does seem to be a problem with MariaDB using the MySQL c/api lib .. but, not the other way around.

Akkadius:
Yeah this is something I had spoke to Ali about needing to be addressed after we redid the installer to be x64. We need our setups to be smooth

PR #1073, opened 2020-05-28
Repair #grid show command

noudess:
I had been trying to use #grid show in game to help debug some pathing issues.  I quickly noticed that in many cases, the node names did not match, or even come close to the grid nodes.

After investigation, the code was trying to spawn nodes named aaa_bbb where aaa is the grid # and bbb is the node number.  This was failing in a number of ways.  The creating of the NPC was stripping all numbers, leaving the node named '_'.  Then, the entity list code was making it unique, adding 000, 001, 002 on the end.  The resulting nodes were all _000, _001, _002 where the values were not guaranteed to match the node numbers, and in faction, when used in repetition, sometimes node 0 ended up being some arbitrary available number.  This is due to timing and the cleaning up of used values.

Also, if a grid reuses the same location, the values are impossible to read as the nodes stack  onto each other.

This PR replaces the names with text.  It also staggers the nodes that are literally on top of each other so they are readable.  For those nodes, Z is altered for readability.

This video shows how the nodes look now.   Clicking the link fails (I don't know why),  Cut-n-paste it into a browser and it works fine.

[http://norrath.no-ip.org/gridshow.mp4](url)

### DIFF
--- a/zone/command.cpp
+++ b/zone/command.cpp
@@ -2477,7 +2477,9 @@ void command_grid(Client *c, const Seperator *sep)
 			glm::vec4 node_position = glm::vec4(atof(row[0]), atof(row[1]), atof(row[2]), atof(row[3]));
 
 			std::vector<float> node_loc {
-					node_position.x, node_position.y, node_position.z
+					node_position.x, 
+					node_position.y, 
+					node_position.z
 			};
 
 			// If we already have a node at this location, set the z offset
--- a/zone/command.cpp
+++ b/zone/command.cpp
@@ -2438,10 +2438,10 @@ void command_grid(Client *c, const Seperator *sep)
 		}
 
 		std::string query = StringFormat(
-			"SELECT `x`, `y`, `z`, `heading`, `number`, `pause` "
+			"SELECT `x`, `y`, `z`, `heading`, `number` "
 			"FROM `grid_entries` "
 			"WHERE `zoneid` = %u and `gridid` = %i "
-			"ORDER BY `number` ",
+			"ORDER BY `number`",
 			zone->GetZoneID(),
 			target->CastToNPC()->GetGrid()
 		);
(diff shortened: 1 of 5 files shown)


Akkadius:
@noudess this video link doesn't work

noudess:
> @noudess this video link doesn't work

Yeah, as I said right above the link, I couldnt figure out how to make github use the link, but if you cut-n-paste the text URL into a browser, it does resolve.

Akkadius:
@noudess this looks like a great addition! 

Left a few comments; other than that lets get merge conflicts and comments addressed and get this in!

zone/command.cpp
```diff
@@ -2471,18 +2471,29 @@ void command_grid(Client *c, const Seperator *sep)
...
-				static_cast<uint32>(target->CastToNPC()->GetGrid()),
-				static_cast<uint32>(atoi(row[4])),
-				static_cast<uint32>(atoi(row[5]))
-			);
-			npc->SetFlyMode(GravityBehavior::Flying);
-			npc->GMMove(node_position.x, node_position.y, node_position.z, node_position.w);
+			glm::vec4 node_position = glm::vec4(atof(row[0]), atof(row[1]), atof(row[2]), atof(row[3]));
+
+			std::vector<float> node_loc {
+					node_position.x, node_position.y, node_position.z
```
Akkadius, line 2480: Indent seems to be off here; can we newline each of the values?

Example

```
std::vector<float> node_loc {
	node_position.x, 
	node_position.y, 
	node_position.z
};
```

zone/npc.cpp
```diff
@@ -1068,14 +1068,20 @@ bool NPC::SpawnZoneController()
...
-NPC * NPC::SpawnGridNodeNPC(std::string name, const glm::vec4 &position, uint32 grid_id, uint32 grid_number, uint32 pause) {
+void NPC::SpawnGridNodeNPC(const glm::vec4 &position, int32 grid_number, int32 zoffset) {
+
 	auto npc_type = new NPCType;
 	memset(npc_type, 0, sizeof(NPCType));
 
-	sprintf(npc_type->name, "%u_%u", grid_id, grid_number);
-	sprintf(npc_type->lastname, "Number: %u Grid: %u Pause: %u", grid_number, grid_id, pause);
+	std::string str_zoffset = numberToWords(zoffset);
+	std::string str_number= numberToWords(grid_number);
```
Akkadius, line 1077: Space `=`

zone/npc.cpp
```diff
@@ -1068,14 +1068,20 @@ bool NPC::SpawnZoneController()
...
+
+	strcpy(npc_type->name, str_number.c_str());
+	if (zoffset != 0) {
+		strcat(npc_type->name, "(Zoffset)");
```
Akkadius, line 1081: `Zoffset` seems really confusing to fresh eyes when not coming from the perspective of looking at the code

Can we use something like `(stacked)` so people know these are on the same grid point, its just displayed as stacked?

Thank you

zone/command.cpp
noudess: Done.

zone/npc.cpp
noudess: Done.

zone/npc.cpp
noudess: Done

PR #1076, opened 2020-06-21
Add new zone methods to Perl/Lua.

KinglyKrab:
Add eq.zone(short_name) to Lua.
Add eq.zone_group(short_name) to Lua.
Add eq.zone_raid(short_name) to Lua.
Add quest::zonegroup(short_name) to Perl.
Add quest::zoneraid(short_name) to Perl.

### DIFF
--- a/zone/embparser_api.cpp
+++ b/zone/embparser_api.cpp
@@ -530,6 +530,32 @@ XS(XS__zone) {
 	XSRETURN_EMPTY;
 }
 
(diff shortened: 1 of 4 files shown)


Akkadius:
This is great! Only thing I would say is that this would be nice to be able to invoke on a client object specifically instead of using implied initiator via the quest objects

Akkadius:
Make sure to update the changelog with your PR changes!

PR #1085, opened 2020-06-29
Add GetDisplayAC() to Perl/Lua.

KinglyKrab:
Pulled from a [Gist](https://gist.github.com/fe1c0e77a5f9c40d6ce037c8efe2cf9a) by Mackal.


Akkadius:
Is this something that should be in the dev tools window instead?

KinglyKrab:
It could be used there for sure.

PR #1090, opened 2020-07-02
Issue 743 - WorldDatabase::GetStartZone now checks isTitanium when creating the SQL

ECXevian:
This is a simple change to handle Titanium sending in player choice instead of a zone id.


joligario:
Looks good.

PR #1093, opened 2020-07-06
Add hot reload saylinks as well as a rule to determine if you bypass the saylinks.

KinglyKrab:


Akkadius:
This is missing the hot reload bits and duplicates functionality that is already covered by `SendEmoteMessageRaw` 

`SendEmoteMessageRaw` can send with a minimum account status which can cover the messages needed in the hot reload functionality

KinglyKrab:
SendStatusMessage allows a min/max status, not just minimum status, while not necessarily 100% useful here, it is useful elsewhere in the future more than likely.

Akkadius:
Yeah I see the difference; but I've never seen a use case where we need to set a "max status" for a message or anything; we'd be better off just removing the duplicated message logic

Akkadius:
Nice work!

PR #1096, opened 2020-07-13
Added Rule to enable AA Ding()

noudess:
Added rule to allow for a DING() sound when an AA point is earned.  This will allow low xp/classic like servers where AA dings are still a major event to play the sound when the AA is earned.  Default is off to remain as code is currently written.


joligario:
Since you are adding this, you may want to add ability to customize 3 options: Silent (Default), Ding, Fanfare.

noudess:
> Since you are adding this, you may want to add ability to customize 3 options: Silent (Default), Ding, Fanfare.

How do I code Fanfare?

joligario:
Disregard. Must not be a method for that.

mackal:
Doesn't tit play a ping sound for AAs?

noudess:
> Doesn't tit play a ping sound for AAs?

Does it?  I'm on ROF2.  Do I need to make it client based?

PR #1097, opened 2020-07-19
Null ptr fixes

SecretsOTheP:
Few fixes I've never put back upstream from classless involving deleted pointers that are not set to null. The codebase really needs a bigger review of "delete" usage vs "safe_delete" macro usage, but that could have complications that extend beyond just a mass find/replace, so don't do that (also not everything crashes that does that!)

First, fix null pointers in group/raid removal code. This is to avoid an instance where a group or raid is removed in that instance of the zone and then becomes a dangling pointer that could be later used in the same server tick.

Fix issue where npc_types could become null in the cache but remain referenced, causing a crash due to invalid memory accessed later in the code (IE; pets that are spawned by NPC Type ID, and mercenaries/bots are especially bad with this.)

Don't do Group::Update if there is no member in zone to update for. This fixes an issue where the zone would crash upon receiving an update about a member that doesn't exist.

Set the leader to null in various places:
When the group is disbanded, set the leader to null.

When setting a new raid leader, make sure we have a new raid leader. If we don't, avoid a crash and disband the raid. It's better than zones falling apart, and will resolve itself on the next VerifyRaid call.

If a member zones, set the leader pointer to nullptr. This fixes an issue where the leader pointer is freed later (MemberZoned normally cleans up the Client object), but referenced by other entities, allowing the leader to be used in the same server process tick, post-cleanup - as the leader won't exist.

### DIFF
--- a/zone/entity.cpp
+++ b/zone/entity.cpp
@@ -2476,7 +2476,7 @@ void EntityList::RemoveAllGroups()
 	while (group_list.size()) {
 		auto group = group_list.front();
 		group_list.pop_front();
-		delete group;
+		safe_delete(group);
 	}
 #if EQDEBUG >= 5
 	CheckGroupList (__FILE__, __LINE__);
@@ -2488,7 +2488,7 @@ void EntityList::RemoveAllRaids()
 	while (raid_list.size()) {
 		auto raid = raid_list.front();
 		raid_list.pop_front();
-		delete raid;
+		safe_delete(raid);
 	}
 }
 
@@ -2838,7 +2838,7 @@ bool EntityList::RemoveGroup(uint32 delete_id)
 	}
 	auto group = *it;
 	group_list.erase(it);
-	delete group;
+	safe_delete(group);
 	return true;
 }
 
@@ -2850,7 +2850,7 @@ bool EntityList::RemoveRaid(uint32 delete_id)
 		return false;
 	auto raid = *it;
 	raid_list.erase(it);
-	delete raid;
+	safe_delete(raid);
 	return true;
 }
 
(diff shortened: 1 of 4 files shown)


mackal:
The changes to entity.cpp and zone.cpp should have no affect, they stuff they're setting to null are gone the next line.

SecretsOTheP:
> 
> 
> The changes to entity.cpp and zone.cpp should have no affect, they stuff they're setting to null are gone the next line.

This is false. If a player dies via a spell hit that affects a group, the pointer to the Group object that is retrieved via GetGroup through the Client class of the initial target will potentially point to corrupted memory depending on the order entities are iterated through. See: Group::CastGroupSpell in spells.cpp and the raid version of said function. Normally, this wouldn't be an issue as most of EQ's spells do not cause damage. However for custom servers (like mine, where the fix is coming from) we also use this target type for damage. Not to mention lua points indirectly to that data, so a wild script could cause the same issue. (as it *also* did on a custom server, and potentially could affect PEQ in the distant future)

In the case of the zone.cpp changes (where NPCType's pointer becomes garbage, but not freed) it is because clear does not remove the pointer, only the reference in the map. The pointer itself (itr->second) is independent of the map and calling erase on a map does not remove the contents, only the reference in the map. Beyond that, the same pointer is used in npc_type.h (twice, actually!) and will return invalid npc type data on the chat command #npctype_cache reload being issued.

See:
https://github.com/EQEmu/Server/blob/master/zone/npc.h#L483

Either way, these two locations, even if there was nothing wrong with the code in question, should still be using the safe_delete macro for nothing else but code consistency.

PR #1100, opened 2020-07-27
Allow quests to turn on mob processing in empty zones

noudess:
This PR allows quest writers to write code that allows activities to take place in zones with no clients.

I'm adding this, so that periodic wars, or movement based on other changes in the world can take place in zones that have no population when the changes or activities are desired.

A quest that wants to make sure that mob movement and processing take place - rergardless of the zone being empty, enable the provessing via quest::processmobswhilezoneempty(1) or the lua counterpart eq.process_mobs_while_zone_empty(true).

Once the quest has completed the quest needs to return the processing to normal state with the same call passing 0 or false.

Or these quest routines could be used by local zone_controller or other spawns to enable processing for a zone permanently.

The default is of course off - so nothing should be impacted if these calls are not used.

I'll add to the docs and the changelog if approved.

These changes allow ownership of territory in game to change - while zones are empty.

### DIFF
--- a/zone/embparser_api.cpp
+++ b/zone/embparser_api.cpp
@@ -2070,6 +2070,19 @@ XS(XS__repopzone) {
 	XSRETURN_EMPTY;
 }
 
+XS(XS__processmobswhilezoneempty);
+XS(XS__processmobswhilezoneempty) {
+	dXSARGS;
+	if (items != 1)
+		Perl_croak(aTHX_ "Usage: quest::processmobswhilezoneempty(bool on)");
+
+	bool ProcessingOn = ((int) SvIV(ST(0))) == 0 ? false : true;
+
+	quest_manager.processmobswhilezoneempty(ProcessingOn);
+
+	XSRETURN_EMPTY;
+}
+
 XS(XS__npcrace);
 XS(XS__npcrace) {
 	dXSARGS;
@@ -5614,6 +5627,7 @@ EXTERN_C XS(boot_quest) {
 	newXS(strcpy(buf, "playersize"), XS__playersize, file);
 	newXS(strcpy(buf, "playertexture"), XS__playertexture, file);
 	newXS(strcpy(buf, "popup"), XS__popup, file);
+	newXS(strcpy(buf, "processmobswhilezoneempty"), XS__processmobswhilezoneempty, file);
 	newXS(strcpy(buf, "pvp"), XS__pvp, file);
 	newXS(strcpy(buf, "qs_player_event"), XS__qs_player_event, file);
 	newXS(strcpy(buf, "qs_send_query"), XS__qs_send_query, file);
--- a/zone/entity.cpp
+++ b/zone/entity.cpp
@@ -515,7 +515,7 @@ void EntityList::MobProcess()
 			mob_settle_timer->Disable();
 		}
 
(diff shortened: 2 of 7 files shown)


Akkadius:
I like the idea of this, however I'm wondering if it would make sense to literally just have this set at the zone level (zone table) to dictate the behavior of a zone. Seems like we have zones (boats) that would also benefit from just having this set regardless as well

What would be the scenarios that if we wanted to use this in quests that we wouldn't just set it at the zone level

noudess:
For me, I have an event that happens once an hour, I want it to occur.  So the timer fires, the event runs, and then I turn off processing again.  So the zone only eats CPU for a minute or so.

The zone wide thing also sounds good - for other things like boats.  But that would be making the zone always use CPU.   Sounds like a good, but different feature?

thalic:
This would allow us to reproduce quests like the Ring War much better. Or quests where an NPC runs from zone to zone. And of course finally work boats without endless timer fumbling. I could have a lot of fun with this.

noudess:
> This would allow us to reproduce quests like the Ring War much better. Or quests where an NPC runs from zone to zone. And of course finally work boats without endless timer fumbling. I could have a lot of fun with this.

Thalic, as far as npcs running from zone to zone, one of my past patches helped with that.  As long as your NPC has a pathgrid with the flag set to despawn at the end, the NPC will already traverse the zone and despawn while the zone is empty.  Assuming the zone is either (dynamic and not timed out) or static.

```if (zone->process_mobs_while_empty || numclients > 0 ||
			mob->GetWanderType() == 4 || mob->GetWanderType() == 6 ||
			mob_settle_timer->Enabled()) {```

**Comment by thalic** [comment]
Noudess, how cool is that? I completely missed that one.
With your patch one could already realize boats, despawn them according to the pathgrid. Cool :-)

With that in mind, how about we have both? The functionality that this change already has and the suggestion from Akkadius to control it zone-wide. It's probably a lot more resource-efficient to do one or a few quests than it is to do all the quests in a zone. Depending on the use case you need one or the other. 

And while we are at it: there is already an option in the source with which you can set whether empty zones are idle or not ("#define IDLE_WHEN_EMPTY"). Then we would have four options in the zone configuration table: 

0. do nothing if empty (current mode - I know "nothing" is not entirely accurate)
1. process only the flaged quests  in zone
3. process all quests in zone
4. process all quests and all mobs  in zone

**Comment by noudess** [comment]
To be super clear here.

Quests still run in static or empty dynamic zones that have not yet shutdown.  It's mob movement and processing that does not.  Timers are firing, signals stack up.  That's why in zones line rivervale Mangler spams when someone enters the zone after a long time being empty.  I acutally fixed that in my quests on my server with this:

```
    if ($timer eq "mangler")
        {
        quest::say("<Bark, bark, bark! Hoooooowwwwwwl!>");
        if (plugin::zoneClientCount() > 0)
            {
            quest::signal(19049,1);
            }
        }
```
That stops the signals from going out and stacking up, as they are part of the processing that does not happen.

I am down with adding a zone flag that says a zone is always on, if thats what you and @Akkadius  want, but as you said, boats will already work using the grid strategy I added months ago.  These grids are also great for  unique mobs like Fippy, so he doesn't always spawn when someone enters the zone, but timing is preserved.

PR #1102, opened 2020-08-09
[Combat] AE Ramp now allows Max Melee Range

fryguy503:
RuleR AERampageSafeZone - Reduce AE Ramp range by this amount to allow for a small safe zone on AE Ramp.


mackal:
My only concern is how is this suppose to actually work. There is some weird oddities I found on live trying to experiment with melee range to fix this issue and I'm unsure if any of those issues actually cause this ...

fryguy503:
I can tell you it works for the last few years on my server, but as for why melee range does what it does on live, I could not tell you. All i hear is about how not being able to max melee range AE Ramp on PEQ makes some events hard to impossible to do under normal circumstances and many resort to cheesing events due to it. Figured I would share my fix and it can be improved on at a later time.

mackal:
I know live has some cache mechanic of melee range (if you shrink, your melee range changes but server side it doesn't) no idea if that is part of the issue.


We also need to actually fix melee range to be correct to start with ...

fryguy503:
I agree if that is the case, then a Issue/Milestone should be set for the community to work on it, but until then, would it not benefit the player base to have a temporary stop-gap implemented until then?

mackal:
I also had a hard time reliably reproducing this on live (at least as reliable this code is) when I was testing it ...

Issue #1104, opened 2020-08-15
lootdrop_entries min_level cleanup/added support


xackery:
12:26 PM] Akkadius: We should probably properly label those fields though
[12:27 PM] Akkadius: trivial_min_level trivial_max_level
12:27 PM] Akkadius: And we can add a normal min_drop_level max_drop_level

change lootdrop_entries min_level/max_level to  trivial_min_level trivial_max_level
add support for min_drop_level max_drop_level to check npc's level before adding drops

PR #1105, opened 2020-08-15
Lootdrop Filtering Adjustments

Akkadius:
* Add new logging category "Loot"
* Adds fields `npc_min_level` and `npc_max_level` to `lootdrop_entries` which runs checks against the current NPC as long as the field values are non-zero
* Fields `minlevel` and `maxlevel` are renamed to `trivial_min_level` and `trivial_max_level` respectively for `lootdrop_entries` for item drops to be filtered based on the player level
* Before, `trivial_max_level` defaulted to `127` in the database, this became an issue on some servers and some scenarios, the default has been adjusted to simply be 0 and only check if the value is non-zero


xackery:
closes #1104

PR #1107, opened 2020-08-16
[Aggro] PercentalHeal should be included in aggro generation

fryguy503:
Per January 15th 2003 Patch Notes:
"NPCs will now notice Druid and Shaman Percentage Heal spells in a manner similar to traditional healing spells."

Added SE_PercetalHeal to the same CheckHealAggroAmount as SE_CurrentHP to be similar to traditional healing spells.


mackal:
Is this another case of them using "similar" to mean exactly :P

PR #1113, opened 2020-08-19
Fix memory leak for virtual zone line branch

mackal:


freyonsecurity:
I tested the code with about 100 zone switches. There were zero problems with it.

mackal:
That's not gonna show this memory leak.

PR #1118, opened 2020-08-28
Spell Fizzle for < LDON expansions via lua_mod

alimalkhalifa:
** PLEASE SQUASH COMMITS ON MERGE **

This PR adds Mob:GetSkill() to the Lua API, necessary for fizzle calculation.

The new combat code is in utils/mods/expansions_combat.lua.

This needs to be copied to eqemu/mods/expansions_combat.lua and in the same folder you will need a LoadOrder.txt file with the filename to load. eg:

LoadOrder.txt:
``
expansions_combat.lua
```

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -82,6 +82,7 @@ Copyright (C) 2001-2002 EQEMu Development Team (http://eqemu.org)
 #include "string_ids.h"
 #include "worldserver.h"
 #include "fastmath.h"
+#include "lua_parser.h"
 
 #include <assert.h>
 #include <math.h>
@@ -734,6 +735,16 @@ bool Mob::CheckFizzle(uint16 spell_id)
 
 bool Client::CheckFizzle(uint16 spell_id)
 {
+#ifdef LUA_EQEMU
+	bool ignoreDefault = false;
+	bool fizzle = LuaParser::Instance()->CheckFizzle(this, spell_id, spells[spell_id], ignoreDefault);
+
+	if (!fizzle) {
+		return false;
+	} else if (ignoreDefault) {
+		return true;
+	}
+#endif
 	// GMs don't fizzle
 	if (GetGM()) return(true);
 
--- a/utils/mods/expansions_combat.lua
+++ b/utils/mods/expansions_combat.lua
@@ -4,72 +4,170 @@ function CheckFizzle(e)
     e.IgnoreDefault = false;
     return e;
 
-  else -- Classic Fizzle -- Based on p2002eq/wfh's Fizzle code
+  else -- Classic Fizzle -- Based on TAKP's fizzle formula
     e.IgnoreDefault = true;
     local client = e.self;
-    local spell_id = e.spell_id;
     local spell = e.spell;
-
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by mackal** [comment]
We should be fixing the "native" code as well. Lua support is good though. Just need to tell custom servers they can play with it if they want :P

**Comment by alimalkhalifa** [comment]
I am a bit reluctant to touch the native code because I worry about breaking custom servers again. I am OK if we have a plan with the server owners whereby they will have a CheckFizzle override implementing current logic.

**Comment by mackal** [comment]
Fix native code. Provide Lua example that is current broken implementation.

**Comment by freyonsecurity** [comment]
How about a rule that decides whether to use the old native code or the new one? I think we have done that in other places. The rule could point to the old code by default.

**Comment by alimalkhalifa** [comment]
Well the last time there was a combat overhaul, the old code was dumped in a Lua mod. I'm leaning towards doing the same once I have the whole Combat revamp done so as to not disrupt operations. Instead I'll keep going in the optional Lua first then we can decide if we want to port it all to C++

**Comment by KimLS** [comment]
So ignoring the "you should fix it the right way" stuff from above, as purely an extension to lua I don't see why this can't be merged?

**Comment by xackery** [comment]
once merged I believe @alimalkhalifa will be testing it on my bountyeq server, expect additional mods after.

And yeah, we can look into the native fizzling too, maybe wrap old way as a rule? 

to start would love to abstract it to lua mods so that custom pvp spell logic can be lace in too.

File: zone/spells.cpp [hunk 1]
```diff
@@ -734,6 +735,16 @@ bool Mob::CheckFizzle(uint16 spell_id)
...
 {
+#ifdef LUA_EQEMU
+	bool ignoreDefault = false;
+	bool fizzle = LuaParser::Instance()->CheckFizzle(this, spell_id, spells[spell_id], ignoreDefault);
+
+	if (!fizzle) {
+		return false;
+	} else if (ignoreDefault) {
+		return true;
+	}
```
xackery, line 746: this may be incorrect.
ignoreDefault is a bool to override default behavior with lua or not.
```suggestion
	if (ignoreDefault) {
		return fizzle;
	}
```

xackery (changes requested):

zone/spells.cpp
alimalkhalifa: You are probably right. I'll check it in the morning.

Its a bit of an edge case that I didn't test: when you check fizzle via an override but then still want to use the native function? It's a bit redundant, right?

xackery: it's how all other mods work. May as well retain the same pattern. Your lua code seems to honor it already, but the C++ integration is pipelined to one use case. Behavior may be abnormal when folks do want a double fizzle check or wahtever odd combination.

Akkadius: Did this get followed up on?

Akkadius: @alimalkhalifa bump

thalic (approved):
I have tested the LUA extension. Technically it works.  Whether the behavior of it's era is accurate, I honestly can't really check. But the fizzle rate is now more like what I remember (I was very active between Classic and Velious).

PR #1130, opened 2020-10-25
SendIllusion Update Internal Values

Akkadius:
Fix scenarios where quest calls to SendIllusion also update internal values so that new clients that zone in see the correct appearance

Lots of appearance calls ultimately end up using SendIllusion; this will allow appearance to stick dynamically in scripts

Example script

```perl
sub EVENT_SPAWN {
	quest::settimer("set_appearance", 1);
}

sub EVENT_TIMER {
	if ($timer eq "set_appearance") {
		quest::stoptimer($timer);
		my $texture = int(rand(4));
		my $gender  = int(rand(2));
		my $race    = int(rand(12));
		$npc->RandomizeFeatures(1);
		$npc->SendIllusion($race, $gender, $texture, 0);
		$npc->WearChange(7, 97);
		$npc->WearChange(8, 98);
	}
}

### DIFF
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -1857,7 +1857,7 @@ void Mob::SendIllusionPacket(
 		}
 	}
 
-	// update internal values fo mob
+	// update internal values for mob
 	size             = (in_size <= 0.0f) ? GetSize() : in_size;
 	texture          = new_texture;
 	helmtexture      = new_helmtexture;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: zone/mob.cpp [hunk 1]
```diff
@@ -1859,6 +1857,21 @@ void Mob::SendIllusionPacket(
 		}
 	}
 
+	// update internal values fo mob
```
mackal, line 1860: spelling

PR #1136, opened 2020-11-06
Update the SummonItem call to default charges for QuestRewards

mackal:
This should summon the item at max charges, if it has charges


thalic (approved):
Tested, works without problems. Thanks!

PR #1137, opened 2020-11-17
Save the rest_timer sooner to prevent issues

mackal:
We need to save the remaining time sooner to prevent circumstances
described here:

http://www.projecteq.net/forums/index.php?threads/raid-out-of-combat-regen-timers-resetting-on-non-raid-mobs.16114/


thalic (approved):
I was not aware of the problem, but I was able to reproduce it on the old code. With the PR this is fixed.

PR #1138, opened 2020-11-19
Remove the extra erroneous message (DoT landing message) at mob death if mob dies of DoT.

noudess:
If you land a dot, like Heat Blood, you get this:

Soandso 's blood simmers.

Currently, if that dot does the killing damage, you get another repeat message at the time of death:

Soandso 's blood simmers.

This PR removes the spell from the Death_Struct if it was a DoT, and that message is eliminated.


mackal:
So this at least confirms it was the death struct? Hmm, I think we should be able to make this work correctly without overwriting the spell. Now that I know you confirmed the place that causes it, I can look more in depth into this to see if there is a better way :P

noudess:
Ok.  Whatever you feel is correct.  I don't see any loss of anything here, unless there is supposed to be a different message.  It looks clean so far.

noudess:
> I'm not saying no to this, just that it needs more investigating.

Yep, glad for that.  You know the entire picture, I just shotgunned it.

thalic:
From the player's perspective I have tested it and can confirm that the change works.

noudess:
I seem to still get an occasional extra message - looks like when multiple effects wear off.  But the message is still related to the DoT that made the kill.

noudess:
I think this errant message must be created by more than 1 thing.  My changes lessened it greatly, but I am still seeing the message sometimes at death, when it _looks_ to me like the DoT damage caused the death. 

So this patch helps but is not the whole story.

noudess:
I'd say 80%+ of the extra messages are gone.  I am still left with the occasional problem:

`[Sun Nov 29 16:12:10 2020] Frau hit a shadowed man for 25 points of non-melee damage.
[Sun Nov 29 16:12:10 2020] Your Affliction spell has worn off of a shadowed man.
[Sun Nov 29 16:12:10 2020] Your Engulfing Darkness spell has worn off of a shadowed man.
[Sun Nov 29 16:12:10 2020] Your Siphon Strength spell has worn off of a shadowed man.
[Sun Nov 29 16:12:10 2020] Your Choking spell has worn off of a shadowed man.
[Sun Nov 29 16:12:10 2020] A shadowed man sweats and shivers, looking feverish.
[Sun Nov 29 16:12:10 2020] You have slain a shadowed man!
`

Akkadius:
Let us know what you find

noudess:
I don't see any adverse effects and all the extra messages are gone. but I defer to what @mackal  finds with a client dump.

mackal:
I'd rather check what live is sending I guess. It's a bit harmless (extra message) so kind of whatever. It might be that DOTs don't send a spell ID, or we just have the damage type wrong for DOT damage in the packet.

noudess:
Yeah, I am triggering on the messages for an audio trigger, which is why I want it, but I'm happy to wait for you to find the right thing to do, if you can tell.  I appreciate it.

thalic:
Tested again and still satisfied with the PR.

Akkadius (approved):

PR #1140, opened 2020-11-26
Add safeguard for issue with mobs pathing to 0,0,0 on teleport nodes with values 0,0,0

SecretsOTheP:
Under any teleport node with 0.0f for x/y, mobs will end up at 0,0,0 due to a bug that does not check teleport nodes in the else statement that prevents the mob from having that route added.

Ignore the Jan 13, 2019 commit, it's reverted in the merge commit. Only the two changes are in question here.


fryguy503:
I smoke tested this on my Rebase and problematic mobs such as Statue and Rathe Council worked correctly! No more "Yeeting" to 0 0 0 on path back.

PR #1146, opened 2020-12-09
Feature/underworld

mackal:
This is used to handle falling off of airplane and zoning so we don't have to hack around it


mackal:
This field is responsible for controlling some collision, specifically when running over an edge that would result you going "underground." For example in paineel to fall into the hole.

joligario (approved):
No obvious issues. Good find!

thalic:
I am not quite sure how to test this?

mackal:
> I am not quite sure how to test this?

You can try airplane, do #depopzone to depop our current work around and run off the edge and see what happens.

Akkadius (approved):
LGTM

PR #1160, opened 2020-12-30
Add optional rule to disable Developer Tools Window.

KinglyKrab:
This is just for people who don't want to have to continue disabling the menu on each zone, don't want the menu, or have their own scripted menu.


Akkadius:
I don't mind adding more options to disable things, however just a few items of feedback

* Rules that are missing get automatically injected into the database when world boots up, no need to make a required migration for it
* Developer window can be shut off per character but the messages still show up - I'm not opposed to combining the setting for this to keep things simple

I'll put up a separate PR to fulfill your same needs shortly

Akkadius:
Covered in: https://github.com/EQEmu/Server/pull/1161

KinglyKrab:
Yeah, I know they automatically get injected, the SQL is for those who use old source and may want to merge some of this code manually. Player by player, zone by zone basis was just not very user friendly. But whatever works is fine.

Akkadius:
The mentioned PR addresses your wants, you can shut it off at an account level and server wide

PR #1161, opened 2020-12-30
[DevTools] Improve DevTools Toggling Options

Akkadius:
This PR improves DevTools toggling options

DevTool Window toggling has been unified into simply "DevTools"

* When DevTools are toggled off at an account level, windows that display on target, messages that display on target and the zone-in menu will no longer show
* DevTools can optionally be shut off globally in rules for all GMs


KinglyKrab:
Only issue I have with this is I don't think this will stop the menu from printing even when all the tools are "disabled".

See https://github.com/EQEmu/Server/pull/1161/files#diff-8541f36e60ccf19a8ff067c78dd04929f95503850299a869b54fd28e49fb8ed1R9433

Issue #368, opened 2015-02-08
/makeleader invalidates member[0] as the group leader


joligario:
I don't have more time to investigate, so I'm adding here.

PR #1164, opened 2020-12-31
[Bug] Fix for Group Leader Disband Issue

fryguy503:
Added public/private class for oldleadername so we can save the previous leader name when the entity is destroyed then allow us to transfer leadership.

Adjusted DelmemberOOZ and in zone functions to include removal of the old leader when disbanding.

### DIFF
--- a/common/eqemu_logsys.h
+++ b/common/eqemu_logsys.h
@@ -120,6 +120,7 @@ namespace Logs {
 		Loot,
 		Expeditions,
 		DynamicZones,
+		Group,
 		MaxCategoryID /* Don't Remove this */
 	};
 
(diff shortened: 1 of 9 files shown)


joligario:
Nice! Hopefully this fixes #368

fryguy503:
We have been using it for ages, I believe the original base design was from our Mac source which was forked from TAKP 4-5 years ago. We adjusted a few things, but there is still the odd group bug when lots of rapid leave/joins happen and people have to just reform the group.

Akkadius (commented):
Please update eqemu_log_aliases to have the Group logging aliases

mackal (approved):

Akkadius:
@fryguy503 please update the changelog

Akkadius (approved):

PR #1165, opened 2020-12-31
[Bug] Prevent Bards from auto equip loot when using instrument

fryguy503:
Bug Fix:

This prevents where a bard has a 2hand only instrument in secondary which prevents anything from being equipped in primary, it is possible to autoloot a main hand weapon and equip it with a 2-hand instrument, this will prevent that scenario from happening.


fryguy503:
Bug Fix:

This prevents where a bard has a 2hand only instrument in secondary which prevents anything from being equipped in primary, it is possible to autoloot a main hand weapon and equip it with a 2-hand instrument, this will prevent that scenario from happening.

Akkadius (approved):
LGTM; the formatting is a bit gnarly so I would prefer to have this cleaned up but otherwise thanks for the PR @fryguy503

Akkadius:
@fryguy503 please update changelog

PR #1166, opened 2020-12-31
[Bug] Barter was allowing the purchase of bags. Bags could contain…

fryguy503:
… anything and it would be traded.


Akkadius (approved):

Akkadius:
@fryguy503 be sure to update the changelog

PR #1170, opened 2021-01-03
#ww

splose:
Casts a spell on everyone currently online

### DIFF
--- a/zone/command.cpp
+++ b/zone/command.cpp
@@ -438,6 +438,7 @@ int command_init(void)
 		command_add("wp", "[add/delete] [grid_num] [pause] [wp_num] [-h] - Add/delete a waypoint to/from a wandering grid", 170, command_wp) ||
 		command_add("wpadd", "[pause] [-h] - Add your current location as a waypoint to your NPC target's AI path", 170, command_wpadd) ||
 		command_add("wpinfo", "- Show waypoint info about your NPC target", 170, command_wpinfo) ||
+		command_add("ww", "Casts the provided spell ID to all players currently online. Use caution with this!", 250, command_ww) ||
 		command_add("xtargets",  "Show your targets Extended Targets and optionally set how many xtargets they can have.",  250, command_xtargets) ||
 		command_add("zclip", "[min] [max] - modifies and resends zhdr packet", 80, command_zclip) ||
 		command_add("zcolor", "[red] [green] [blue] - Change sky color", 80, command_zcolor) ||
@@ -737,6 +738,16 @@ void command_logcommand(Client *c, const char *message)
 /*
  * commands go below here
  */
+void command_ww(Client *c, const Seperator *sep)
+{
+	if (sep->arg[1][0] && Seperator::IsNumber(sep->arg[1])) {
+		int spell_id = atoi(sep->arg[1]);
+		quest_manager.WorldWideCastSpell(spell_id, 0, 0);
+		worldserver.SendEmoteMessage(0, 0, 15, fmt::format("<SYSTEMWIDE MESSAGE> A GM has cast {} world-wide!", GetSpellName(spell_id)).c_str());
+	}
+	else
+		c->Message(Chat::Yellow, "Usage: #ww <spellid>");
+}
 void command_setstat(Client* c, const Seperator* sep){
 	if(sep->arg[1][0] && sep->arg[2][0] && c->GetTarget()!=0 && c->GetTarget()->IsClient()){
 		c->GetTarget()->CastToClient()->SetStats(atoi(sep->arg[1]),atoi(sep->arg[2]));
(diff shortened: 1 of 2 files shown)


mackal:
ww is a bad name.

fryguy503:
Suggestion: Maybe add it as an argument for #cast - e.g. #cast ##### world

splose:
Changed it up a little bit.. still learning how to use GitHub.

I'm open to suggestions but please try to be more constructive than "bad name". Thanks!

Akkadius:
I'd be fine if there was a command that was #worldwide that ended up having a bunch of sub-commands for GM utilities. It would need to print a help menu to inform the user what is available underneath

Example #worldwide 

* cast <spell>
* moveallplayers <zone>
* etc.

As this PR currently stands, there's merge conflicts that need to be resolved, otherwise thanks for taking the time to contribute

splose:
If you leave it I'll do that will be good practice for me..

Should I keep it all in this PR or open another?

Akkadius:
It would be good for you to figure out how to handle merge conflicts; @KinglyKrab should be able to help you through resolving them

If the merge is too gnarly some people just cut another branch and re-add their changes and re-PR

This should be an easy merge conflict to resolve

KinglyKrab (approved):
Looks good to me.

mackal:
I would prefer this wasn't merged without addressing the comment I left ...

Akkadius:
@KinglyKrab I would have to echo what @mackal said; I left an open suggestion - we shouldn't push PR's through as it voids the point of the process

KinglyKrab:
We can revert if necessary, he changed the name from #ww to #wwcast and removed the merge conflicts, figured it was good to go.

splose:
I'm a bit confused.

Seems you can't make multiple forks so my understanding was that this needed to be pushed through so I could pull it down again and start working on expanding the command in place of what is currently there. Then when done do another PR. I was under the impression that's the way this process is supposed to work.

@mackal's comment was pretty useless and unhelpful. Even still I did address by changing the name. I addressed @Akkadius' response by fixing the merge conflicts. What else needs to be done?

mackal:
I thought both of my comments were very useful, and one of them you 100% understood. The other one was that the provided spell ID needs verification. I even gave an example of another command that takes a user provided spell ID and we do some checks to make sure it's not garbage.


You need to create multiple branches to address the issue you had with git and PRs. Basically, you fork, create a branch with your changes etc and do the same for other fixes you want to open PRs for. Just leave your master up-to-date with upstream.

splose:
Yeah no. You said "ww is a bad name" then cried it got merged without your consent. Those were your only 2 comments prior to this one. So far you're 0 for 3 on any of your comments being useful.

009bbf93f7c75524ec60200bcbeae504

Since you're nitpicking so hard I decided to do some testing in-game to see if checking if the argument is a number its enough. None of these crashed or had any adverse effects so I think it's all good. You can close this now :)

Akkadius:
@splose you should have thicker skin and take the higher road instead of going on the offense with comments that don’t help anyone

If Mackal’s suggestion wasn’t very helpful, then sure communicate that to him. It doesn’t need to devolve into pathetic garbage; the pull request process is one that requires peer review feedback so that we can collectively increase the quality of what goes into the source that the entire community uses, it doesn’t matter how amazing of a coder you are. Feedback doesn’t and shouldn’t be received in a offensively harsh manner, even if someone isn’t giving you the most constructive or descriptive feedback, you can elicit it from other reviewers

“#ww is a bad name” is a valid concern because I agreed above that its not very descriptive, but I agree with you that it doesn’t help you much in terms of feedback, so let me give more context

https://github.com/EQEmu/Server/pull/1170#issuecomment-753694305

Commands should not be cryptic with acronyms, it should be #worldwide because we have a system that is made to alias commands that feel long to type for this reason

Remember that we have a `command_aliases` table for short command inputs, so that #worldwide could be aliased to #ww or other shorthand’s while keeping the descriptive, readable command

In general, titles, labels, function names should make themselves obvious to the end user. Quite a few people gave the feedback of a different command name so it doesn’t pass the smell test. This is simple feedback and a simple change

You don’t need to revert the PR; thanks for making your PR and making an effort to contribute back to the community, it is appreciated and I’m sure someone will definitely make use of your command - however, please get rid of the bullshit, you’re better than that

splose:
Yeah I'm not seeing it that way. My skin was plenty thick this entire PR until he got up in arms about a comment that he didn't even make not being addressed.

Seems like you're telling me I should've made the entire #worldwide command with all that extra functionality in this PR. Which seems strange because I've watched people complain about multiple changes / functionality not being in seperate PRs. Seems like I wouldn't be able to satisfy everyone either way.

My line of thinking was that this PR had been fulfilled. Which seems to be the case since it doesn't need to be reverted. I was using branches for my changes and it would error out when trying to update master. That's the only reason I wanted this pushed through. So I could update my master and make the command you suggested in place of this one.

Seems like I was right and everyone is huffing and puffing for no reason.

Akkadius:
I think you need to work on receiving feedback, this isn’t a personal thing. This is simple, you’re making it bigger than it really is

My very clear suggestion was a name change so that sub commands can be added later; not now

PR #1173, opened 2021-01-03
[Bugfix] Very edge case Guard awareness latency

Akkadius:
When mobs are being pulled to Guards; they lag in their awareness to their surrounding because they are idle and not scanning aggressively (Every tic) whereas idle NPCs only do full scans once a minute

Flipped the flag for NPC's to add themselves to others close list when they are moving. This will ensure that any moving entity always updates those around them

Also added in-source comments around the "close mob" system

### DIFF
--- a/zone/entity.cpp
+++ b/zone/entity.cpp
@@ -2693,6 +2693,36 @@ void EntityList::RemoveAuraFromMobs(Mob *aura)
 }
 
 /**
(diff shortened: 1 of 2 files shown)


KinglyKrab (approved):
I've encountered this, always wondered what the fix was. 👍

KimLS (approved):

PR #1174, opened 2021-01-04
[Boats] Remove code override of boat flymode and update dbs to reflect it

noudess:
Old code remained to set boats to flymode 4 after flymode as added to npc_types as a field.

I removed that code override, and updated the db  so that all boats that the code used to override are set to the correct flymode.

This allows servers to change some boats to not use flymode 4 if they choose, due to unsizable boats like 553 not performing well while floating i regards to some docks.

Note that this PR does not alter 553 in the db, because up until last weeks PR (mine), 553 was not in IsBoat().    I added it assuming that was correct (and it is, since IsBoat() controls pathing as well) but by adding it my 553 boats cease to line up.

Rergardless, this PR puts flymode in the db in control, and changes dbs so that no one should see any impact (other than having the flexibility to change any boat if they like).

@thalix1337 You'll want to set race 553 to flymode 4 in your db if you prefer it higher up, but I am guessing you'll see no change in your boats.  Thank you for helping me get mine working.

### DIFF
--- a/common/version.h
+++ b/common/version.h
@@ -34,7 +34,7 @@
  * Manifest: https://github.com/EQEmu/Server/blob/master/utils/sql/db_update_manifest.txt
  */
 
-#define CURRENT_BINARY_DATABASE_VERSION 9159
+#define CURRENT_BINARY_DATABASE_VERSION 9160
 
 #ifdef BOTS
 	#define CURRENT_BINARY_BOTS_DATABASE_VERSION 9027
--- a/common/version.h
+++ b/common/version.h
@@ -34,7 +34,7 @@
  * Manifest: https://github.com/EQEmu/Server/blob/master/utils/sql/db_update_manifest.txt
  */
 
-#define CURRENT_BINARY_DATABASE_VERSION 9160
+#define CURRENT_BINARY_DATABASE_VERSION 9159
 
 #ifdef BOTS
 	#define CURRENT_BINARY_BOTS_DATABASE_VERSION 9027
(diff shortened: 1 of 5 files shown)


Akkadius:
I don't mind this change, however we should consider making it so those who have no idea how any of this stuff works and they make a boat that the proper flymode is set. 

Can we have flymode for boats is still set in the source and still have database overrides

Have options is great, but having sane defaults that don't have to be touched is far more ideal

noudess:
> I don't mind this change, however we should consider making it so those who have no idea how any of this stuff works and they make a boat that the proper flymode is set.
> 
> Can we have flymode for boats is still set in the source and still have database overrides
> 
> Have options is great, but having sane defaults that don't have to be touched is far more ideal

Ok, so this PR changes their db to use the correct modes.  After that, they shouldn't need to think about it.  I need the code not to override on some of my boats.  So in your scenario, the code still overrides, but only if flymode is -1 in the db?

Akkadius:
```
So in your scenario, the code still overrides, but only if flymode is -1 in the db?
```

Correct

noudess:
> ```
> So in your scenario, the code still overrides, but only if flymode is -1 in the db?
> ```
> 
> Correct

Ok.  I can do that.  Good idea.  Should I remove changing the db_version and their databases, since it will now work without?

Akkadius:
Yep, we should be able to retract the DB migration

noudess:
> Yep, we should be able to retract the DB migration

Maybe just reject this PR straight up, and I'll create a new clean branch with 1 commit?

Akkadius:
That’s up to you, I’m most likely going to squash the commits on this branch regardless when we merge

noudess:
Okay done.  I had to move the override to npc.c, as the mob constructor doesn't know about the db flymode setting.

It's interesting... The NPC constructor takes iflymode as an argument, but if its in the db as something else, that overrides.  Our change is kinda the opposite.

Let me know what you think.

PR #1179, opened 2021-01-09
Fix zone startup crash if maps not installed

hgtw:
Just a minor inconvenience when developing without maps installed, shouldn't affect normal servers

### DIFF
--- a/zone/mob_ai.cpp
+++ b/zone/mob_ai.cpp
@@ -1637,7 +1637,7 @@ void NPC::AI_DoMovement() {
 					roambox_destination_y,
 					m_Position.z + 15
 				);
-				if (!zone->watermap->InLiquid(position)) {
+				if (zone->HasWaterMap() && !zone->watermap->InLiquid(position)) {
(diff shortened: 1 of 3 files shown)


Akkadius:
Curious why you wouldn’t just have maps?

KinglyKrab (approved):
Looks good to me, fixing crashes that only affect a few is still fixing a crash. Not sure how many would encounter not having maps, but a worthwhile fix nonetheless. Good work. 👍

mackal:
We shouldn't crash. But wouldn't it be better to just fail to load if we have no map file? Shit doesn't work very well without them :P

hgtw:
While working on expeditions I spread development with another computer and laptop rather than the vm I usually used and cloning the map repo on all of them wasn't necessary for what I was doing. It doesn't crash every zone so I only ran into it later when zoning around to start documenting a usage example and update quests.

Someone in discord a few months ago had these crashes when they had maps installed in old/wrong paths so it fixes that edge case too.

PR #1182, opened 2021-01-16
Add Race/Class Name and Bitmask Methods.

KinglyKrab:
Adds GetClassBitmask(), GetClassName(), GetRaceBitmask(), and GetRaceName() to Perl/Lua.

### DIFF
--- a/zone/lua_client.cpp
+++ b/zone/lua_client.cpp
@@ -134,6 +134,16 @@ void Lua_Client::SetBaseGender(int v) {
 	self->SetBaseGender(v);
 }
 
+int Lua_Client::GetClassBitmask() {
+	Lua_Safe_Call_Int();
+	return GetPlayerClassBit(self->GetClass());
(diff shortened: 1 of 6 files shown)


mackal:
I think you forgot the implementations of the functions added to the Client class.

KinglyKrab:
Forgot to push Lua. Going to modify. Not merge ready.

KinglyKrab:
This is ready for another lookover. Pushed Lua up.

mackal:
Looking over it again, I wonder if we should have GetBaseRaceBitmask and a GetRaceBitmask (so rename current one?)

KinglyKrab:
Well my thought process there was you'd only ever need to get a bitmask for their REAL race, as a bitmask for a non-player race is useless. So using BaseRace there is fine in my opinion.

mackal (approved):

PR #1185, opened 2021-01-20
Add HasDisciplineLearned() to Perl/Lua.

KinglyKrab:
- Add $client->HasDisciplineLearned(spell_id) to Perl.
- Add client:HasDisciplineLearned(spell_id) to Lua.

### DIFF
--- a/zone/effects.cpp
+++ b/zone/effects.cpp
@@ -678,6 +678,16 @@ void Client::ResetDisciplineTimer(uint32 timer_id) {
 	SendDisciplineTimer(timer_id, 0);
 }
 
+bool Client::HasDisciplineLearned(uint16 spell_id) {
+	bool has_learned = false;
+	for (auto index = 0; index < MAX_PP_DISCIPLINES; ++index) {
+		if (GetPP().disciplines.values[index] == spell_id) {
+			has_learned = true;
+		}
+	}
+	return has_learned;
+}
+
 void Client::SendDisciplineTimer(uint32 timer_id, uint32 duration)
 {
 	if (timer_id < MAX_DISCIPLINE_TIMERS)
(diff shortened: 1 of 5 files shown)


zone/effects.cpp
```diff
@@ -678,6 +678,16 @@ void Client::ResetDisciplineTimer(uint32 timer_id) {
 	SendDisciplineTimer(timer_id, 0);
 }
 
+bool Client::HasDisciplineLearned(uint16 spell_id) {
+	bool has_learned = false;
+	for (auto index = 0; index < MAX_PP_DISCIPLINES; ++index) {
+		if (GetPP().disciplines.values[index] == spell_id) {
+			has_learned = true;
```
mackal, line 685: We can just return true here and return false later.

mackal (changes requested):

zone/effects.cpp
KinglyKrab: Figured one point of return was better than 2.

Akkadius (approved):

zone/effects.cpp
mackal: well, if the first discipline checked is the one you're searching for, it's gonna still search the rest of them. Ideally you return early, you stop searching. Currently this always has the same complexity, where on average we should be hitting O(n/2).

Akkadius:
Current pr has conflicts to resolve, also has suggestions. Lgtm otherwise

zone/effects.cpp
mackal: We could also just use std::find_if.

KinglyKrab: I'll just add a break in the loop.

mackal (approved):

mackal:
Conflicts still nee to be resolved it appears.

KinglyKrab:
Merging based on approval and passing all checks. Going to squash to one commit.

PR #1193, opened 2021-01-26
Use heading in GMMove and change FaceTarget not to turn boat

noudess:
Mob::GMMove was ignoring heading.  Fixed to set it.

If a server allows boats to be targetted (so you can see boat name), any hail of the boat was turning the boat.  Disabled this for boats.

### DIFF
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -1734,6 +1734,7 @@ void Mob::GMMove(float x, float y, float z, float heading, bool SendUpdate) {
 	m_Position.x = x;
 	m_Position.y = y;
 	m_Position.z = z;
+	SetHeading(heading);
 	mMovementManager->SendCommandToClients(this, 0.0, 0.0, 0.0, 0.0, 0, ClientRangeAny);
 
 	if (IsNPC()) {
@@ -2820,6 +2821,11 @@ bool Mob::HateSummon() {
 }
 
 void Mob::FaceTarget(Mob* mob_to_face /*= 0*/) {
+
+	if (IsBoat()) {
+		return;
+	}
+
 	Mob* faced_mob = mob_to_face;
 	if(!faced_mob) {
 		if(!GetTarget()) {


KinglyKrab (approved):
Looks good to me.

mackal:
Hmm, I'm a bit cautious of adding that boat bit. How many places would we need to add it before every problematic FaceTarget calls?

noudess:
> Hmm, I'm a bit cautious of adding that boat bit. How many places would we need to add it before every problematic FaceTarget calls?

This is added inside FaceTarget, so anyone that calls it is covered.

mackal:
> > Hmm, I'm a bit cautious of adding that boat bit. How many places would we need to add it before every problematic FaceTarget calls?
> 
> This is added inside FaceTarget, so anyone that calls it is covered.

I just feel like that might be excessive. Is there 0 times when a boat ends up calling FaceTarget that we don't want them? That's what you gotta convince me :P

noudess:
> > > Hmm, I'm a bit cautious of adding that boat bit. How many places would we need to add it before every problematic FaceTarget calls?
> > 
> > 
> > This is added inside FaceTarget, so anyone that calls it is covered.
> 
> I just feel like that might be excessive. Is there 0 times when a boat ends up calling FaceTarget that we don't want them? That's what you gotta convince me :P

I cannot think of any case where we want a boat to face a target.  I'd wager that nearly all servers have them untargetable.

noudess:
I looked at this further.   To avoid what I am running into, I could just alter NPC::DoQuestPause and  put the check for boat there, so that it does not call FaceTarget.  

If this is the preferred change, let me know.

I still like it where it is, as things like trading, attack, etc could all cause FaceTarget if the db settings on boats allowed it, and I still can't think of a reason we would want the boat to face a target.

But let me know and I'll defer.

KimLS (approved):
I see nothing wrong with this.

PR #1202, opened 2021-01-31
RandomPath improvements

regneq:
RandomPath grid type improvement

RandomPath has been changed to the following behavior:

* When stopped at a waypoint, it will now use the waypoint's pause timer if it is non-zero.  If it is zero, it will use the first waypoint's pause timer if that is non-zero.  If both are zero it will use 38 seconds.  Previously it would always use the first waypoint's pause timer if it was non-zero.  This new way is what it should have been from the start, really.

* If any waypoint is flagged as a centerpoint, then the NPCs will always return to that first randomly selected centerpoint waypoint after moving.  I.e. every other move will be to the first waypoint if the bool is set.

If any waypoint has a negative number (-1, -10, etc) in the pause timer. It will not be included in the waypoints to be random.  Instead the npc will just path through it everytime.  This is useful for a path to only select a small area to pause at before returning to it's centerpoints or another random waypoints in between it.


zone/waypoints.cpp
```diff
@@ -385,8 +385,49 @@ void NPC::CalculateNewWaypoint()
...
 			if (cur_wp == patrol) // reutilizing patrol member instead of making new member for this wander type; here we use it to save a random waypoint
 			{
+				if (!Waypoints[cur_wp].centerpoint)
+				{
+					// if we have arrived at a waypoint that is NOT a centerpoint, then check for the existence of any centerpoint waypoint
+					// if any exists then randomly go to it otherwise go to one that exist.
+					std::vector<wplist> random_centerpoints;
+					for (auto& w : Waypoints)
+					{
+						wplist wpl = w;
```
mackal, line 395: This shouldn't be necessary, just directly access it through the w reference.

zone/waypoints.cpp
```diff
@@ -385,8 +385,49 @@ void NPC::CalculateNewWaypoint()
...
+
 				while (patrol == cur_wp)
-					patrol = zone->random.Int(0, Waypoints.size() - 1);
+				{
+					// Setting a negative number in pause of the select waypoints will NOT be included in the group of waypoints to be random.
+					// This will cause the NPC to not stop and pause in any of the waypoints that is not part of random waypoints.
+					std::vector<wplist> random_waypoints;
+					for (auto& w : Waypoints)
+					{
+						wplist wpl = w;
```
mackal, line 422: Same here.

mackal (changes requested):

mackal (approved):

KinglyKrab (approved):

Akkadius (approved):

PR #1219, opened 2021-02-02
NPC ability to open doors while moving now configuarble in npc_types

noudess:
This PR allows server operators to change the behavior of some mobs to not open doors.

The PR in itself changes no behavior, unless the operator changes the new field, can_open_doors to 0.  The npc_types is updated so that all npcs  _can_ open doors, which is the behavior before this PR.

My reason for wanting this is to quiet down home cities, where rodents are opening doors like there is no tomorrow, wearing out hinges, and my speakers.

### DIFF
--- a/zone/zonedb.cpp
+++ b/zone/zonedb.cpp
@@ -2525,7 +2525,8 @@ const NPCType *ZoneDatabase::LoadNPCTypesData(uint32 npc_type_id, bool bulk_load
 		"npc_types.stuck_behavior, "
 		"npc_types.model, "
 		"npc_types.flymode, "
-		"npc_types.always_aggro "
+		"npc_types.always_aggro, "
+		"npc_types.can_open_doors "
 		"FROM npc_types %s",
 		where_condition.c_str()
 	);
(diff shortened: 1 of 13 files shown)


zone/zonedb.cpp
```diff
@@ -2728,6 +2729,7 @@ const NPCType *ZoneDatabase::LoadNPCTypesData(uint32 npc_type_id, bool bulk_load
 		temp_npctype_data->use_model        	= atoi(row[110]);
 		temp_npctype_data->flymode          	= atoi(row[111]);
 		temp_npctype_data->always_aggro	        = atoi(row[112]);
+		temp_npctype_data->can_open_doors		= atoi(row[113]);
```
KinglyKrab, line 2732: Should return a bool here just to be on the safe side.
```cpp
temp_npctype_data->can_open_doors		= atoi(row[113]) != 0 ? true : false;

KinglyKrab (approved):
Looks good to me, just had one small comment.

zone/zonedb.cpp
mackal: Just need the `!= 0` bit, it's already a bool with just that. (We do that in a lot of places)

noudess: Done.   Fixed my previous field as well.

KinglyKrab: 👍

Akkadius:
Closing for preference of https://github.com/EQEmu/Server/pull/1231

PR #1223, opened 2021-02-06
[Repositories] Add expedition repositories

Akkadius:


hgtw (approved):

KinglyKrab (approved):

common/repositories/base/base_character_expedition_lockouts_repository.h
```diff
@@ -0,0 +1,362 @@
...
+#include "../../string_util.h"
+
+class BaseCharacterExpeditionLockoutsRepository {
+public:
+	struct CharacterExpeditionLockouts {
+		int         id;
+		int         character_id;
+		std::string expedition_name;
+		std::string event_name;
+		std::string expire_time;
```
hgtw, line 42: It should generate a time_t for DATETIME fields instead of a string

common/repositories/base/base_character_expedition_lockouts_repository.h
```diff
@@ -0,0 +1,362 @@
...
+
+	static CharacterExpeditionLockouts NewEntity()
+	{
+		CharacterExpeditionLockouts entry{};
+
+		entry.id                   = 0;
+		entry.character_id         = 0;
+		entry.expedition_name      = "";
+		entry.event_name           = "";
+		entry.expire_time          = current_timestamp();
```
hgtw, line 116: I don't see this method in the source, maybe an artifact from the autogen for DATETIME columns?

Akkadius: Not sure where this came from tbh

Akkadius: It's coming from the MySQL default value column; that's why. Should still probably change how this outputs a bit. Easy enough to change either way

image

Akkadius:
```perl
        my $default_value = 0;
        if ($column_default ne "NULL" && $column_default ne "") {
            $column_default =~ s/'/"/g;
            $default_value = $column_default;
        }
        elsif ($column_default eq "''") {
            $default_value = '""';
        }
        elsif ((trim($column_default) eq "" || $column_default eq "NULL") && $column_type =~ /text|varchar/i) {
            $default_value = '""';
        }
```

PR #1231, opened 2021-02-08
[Door Opening] Rule to let configure Animal Door Opening

Akkadius:
This PR allows a server operator to turn off animals opening doors if they want

Default live behavior is for animals to open doors when in proximity

```
RULE_BOOL(NPC, AnimalsOpenDoors, true, "Determines or not whether animals open doors or not when they approach them")
```

**Show your Work**

**Default Value** (Animals can open doors)

image
image

Rodent opened door

**Rule Disabled** (Animals can't open doors)

image

Rodent warped through door


noudess (approved):
Thanks for making a compromised solution that fits needs without bloating the field list.

noudess (approved):
Looks good.

PR #1239, opened 2021-02-09
[Game Play Mechanic] Very simple implementation of Sneak Pull

noudess:
Sneak pulling is a technique I used for years with my rogue on live, dating way back to 1999.

There are many different interpretations of how this works.  This thread covers alot of the ideas within the thread and with references to Monkly Business and The Safehouse.

https://www.project1999.com/forums/showthread.php?t=260642

For my server and from my memory, this implementation gets me close to my experience from live.

This is behind a rule, so no base behavior will change unless you enable it.

If it is enabled, then  successful sneaking reduces the assist range.  If behind the mob, it negates it - but you must remain sneaking while pulling.  I spent many years hurling throwing weapons while peddling backwards with sneak/hide on with sow. If I peddled back out of aggro range before the missile hit, remaining sneaking, I often got 1 mob if its back was turned.

I like this implementation, and I'll just keep it local if the community doesn't want it.  

I do think it's useful and think some server operators would turn it on.

I know P1999 did alot of work on sneak pull, but I don't know their details.

### DIFF
--- a/zone/npc.cpp
+++ b/zone/npc.cpp
@@ -3241,7 +3241,7 @@ void NPC::AIYellForHelp(Mob *sender, Mob *attacker)
 		float assist_range = (mob->GetAssistRange() * mob->GetAssistRange());
 
 		if (RuleB(Combat, EnableSneakPull) && attacker->sneaking) {
-			assist_range=RuleI(SneakPullAssistRange);
+			assist_range=RuleI(Combat, SneakPullAssistRange);
 			if (attacker->IsClient()) {
 				float clientx = attacker->GetX();
 				float clienty = attacker->GetY();
--- a/zone/npc.cpp
+++ b/zone/npc.cpp
@@ -3240,6 +3240,7 @@ void NPC::AIYellForHelp(Mob *sender, Mob *attacker)
 
 		float assist_range = (mob->GetAssistRange() * mob->GetAssistRange());
 
+		// Implement optional sneak-pull
 		if (RuleB(Combat, EnableSneakPull) && attacker->sneaking) {
 			assist_range = RuleI(Combat, SneakPullAssistRange);
 			if (attacker->IsClient()) {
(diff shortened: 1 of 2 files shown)


zone/npc.cpp
```diff
@@ -3239,6 +3239,18 @@ void NPC::AIYellForHelp(Mob *sender, Mob *attacker)
 		}
 
 		float assist_range = (mob->GetAssistRange() * mob->GetAssistRange());
+
+		if (RuleB(Combat, EnableSneakPull) && attacker->sneaking) {
+			assist_range=(20*20);
```
KinglyKrab, line 3244: Formatting, could just put 400 being it's hard-coded anyway or make it a rule maybe

Others below need spacing in declaration.

zone/npc.cpp
```diff
@@ -3239,6 +3239,18 @@ void NPC::AIYellForHelp(Mob *sender, Mob *attacker)
...
+			if (attacker->IsClient()) {
+				float clientx=attacker->GetX();
+				float clienty=attacker->GetY();
```
KinglyKrab, line 3247: Formatting here and for X.

KinglyKrab (approved):
Looks good to me pending comments and discussion.

zone/npc.cpp
noudess: Done and thx.

zone/npc.cpp
noudess: done and thx!

zone/npc.cpp
```diff
@@ -3239,6 +3239,18 @@ void NPC::AIYellForHelp(Mob *sender, Mob *attacker)
 		}
 
 		float assist_range = (mob->GetAssistRange() * mob->GetAssistRange());
+
+		if (RuleB(Combat, EnableSneakPull) && attacker->sneaking) {
+			assist_range=RuleI(Combat, SneakPullAssistRange);
```
KinglyKrab, line 3244: Spacing here. Other than that. Looks good! 👍

KinglyKrab: Will squash on merge. So more commits is no biggie.

noudess: got it

mackal:
Code wise it's fine, but I have no knowledge of this mechanic.

mackal (approved):

PR #1240, opened 2021-02-10
[Memory Leak] Fix va_list leaks in MakeAnyLenString/AppendAnyLenString

mackal:
Slowly leaking memory for ages!

clang-tidy complained about it


mackal:
These functions should actually just be removed and replaced with more modern better things. But hey, let's not leak at least for now.

Akkadius (approved):

PR #1249, opened 2021-02-13
[Quest API] Resolves traindiscs and scribespells issues in Perl/Lua.

KinglyKrab:
These functions were returning the size of the array of spells, not the amount of spells or disciplines learned.

Was also missing a couple logic checks.

Condensed messages as well and added a count in a message with scribespells similar to traindiscs.

### DIFF
--- a/zone/questmgr.cpp
+++ b/zone/questmgr.cpp
@@ -1111,7 +1111,7 @@ uint16 QuestManager::scribespells(uint8 max_level, uint8 min_level) {
 	}
 
 	if (spells_learned > 0) {
-		std::string spell_message = (spells_learned == 1 ? " a new spell" : fmt::format("{} new spells", spells_learned));
+		std::string spell_message = (spells_learned == 1 ? "a new spell" : fmt::format("{} new spells", spells_learned));
 		initiator->Message(Chat::White, fmt::format("You have learned {}!", spell_message).c_str());
 	}
 	return spells_learned;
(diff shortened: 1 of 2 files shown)


zone/questmgr.cpp
```diff
@@ -1101,9 +1106,16 @@ uint16 QuestManager::scribespells(uint8 max_level, uint8 min_level) {
 			}
 			initiator->ScribeSpell(spell_id, book_slot);
 			book_slot = initiator->GetNextAvailableSpellBookSlot(book_slot);
+			spells_learned++;
+			spell_learned = true;
 		}
 	}
-	return spell_count;
+
+	if (spell_learned) {
```
hgtw, line 1114: Could just check `spells_learned > 0` here and ditch the extra var

hgtw (approved):

thalix1337 (approved):
Verified, traindisc and scribespells now work as expected.

zone/questmgr.cpp
KinglyKrab: True enough. Will do that for disciplines as well. Variable was just for code legibility.

KinglyKrab:
Added changes based on comments and a change to break loop earlier rather than continuing even if book slot is invalid.

joligario (approved):

KinglyKrab:
Removed extra space. Please squash when merged.

KinglyKrab:
Merging based on approvals and this is a bug that needs fixed for custom servers.

PR #1251, opened 2021-02-13
[Cleanup] Nuke Make/AppendAnyLenString

mackal:
These functions were fine back in the day, but we got better more modern things.

Either use bounded safe c-string stuff or fmtlib.

### DIFF
--- a/common/database.h
+++ b/common/database.h
@@ -128,6 +128,7 @@ public:
 	bool SaveCharacterCreate(uint32 character_id, uint32 account_id, PlayerProfile_Struct *pp);
 	bool SetHackerFlag(const char *accountname, const char *charactername, const char *hacked);
 	bool SetMQDetectionFlag(const char *accountname, const char *charactername, const char *hacked, const char *zone);
+	bool SetMQDetectionFlag(const char *accountname, const char *charactername, const std::string &hacked, const char *zone);
 	bool UpdateName(const char *oldname, const char *newname);
 	bool CopyCharacter(
 		std::string source_character_name,
--- a/world/zonelist.cpp
+++ b/world/zonelist.cpp
@@ -333,7 +333,7 @@ void ZSList::SendZoneStatus(const char* to, int16 admin, WorldTCPConnection* con
 			}
 
 			fmt::format_to(out,
-				"#{:-3i} :: {} :: {}:{:-5} :: {:2} :: {}:{} :: {} :: ({})",
+				"#{:<3} :: {} :: {}:{:<5} :: {:2} :: {}:{} :: {} :: ({})",
 				zone_server_data->GetID(),
 				is_static_string,
 				addr.c_str(),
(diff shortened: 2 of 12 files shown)


mackal:
Please double check I converted the format strings correctly :)

hgtw (changes requested):

common/database.h
```diff
@@ -128,6 +128,7 @@ class Database : public DBcore {
 	bool SaveCharacterCreate(uint32 character_id, uint32 account_id, PlayerProfile_Struct *pp);
 	bool SetHackerFlag(const char *accountname, const char *charactername, const char *hacked);
 	bool SetMQDetectionFlag(const char *accountname, const char *charactername, const char *hacked, const char *zone);
+	bool SetMQDetectionFlag(const char *accountname, const char *charactername, const std::string &hacked, const char *zone);
```
hgtw, line 131: Can probably just remove the original method in favor of this one instead of overloading

zone/embparser.cpp
```diff
@@ -1270,25 +1270,23 @@ void PerlembParser::ExportItemVariables(std::string &package_name, Mob *mob)
...
 
 		for (int slot = EQ::invslot::EQUIPMENT_BEGIN; slot <= EQ::invslot::GENERAL_END; slot++) {
-			char *hi_decl = nullptr;
 			int  itemid   = mob->CastToClient()->GetItemIDAt(slot);
 			if (itemid != -1 && itemid != 0) {
-				MakeAnyLenString(&hi_decl, "push (@{$%s{%d}},%d);", hashname.c_str(), itemid, slot);
-				perl->eval(hi_decl);
-				safe_delete_array(hi_decl);
+				// this is really ugly with fmtlib, I think I did it right
+				auto hi_decl = fmt::format("push (@{{${0}{{{1}}}}},{2});", hashname, itemid, slot);
```
hgtw, line 1276: Bet this was fun, looks right though

world/zonelist.cpp
```diff
@@ -333,8 +332,8 @@ void ZSList::SendZoneStatus(const char* to, int16 admin, WorldTCPConnection* con
 				zone_data_string[0] = 0;
 			}
 
-			AppendAnyLenString(&output, &outsize, &outlen,
-				"#%-3i :: %s :: %15s:%-5i :: %2i :: %s:%i :: %s :: (%u)",
+			fmt::format_to(out,
+				"#{:-3i} :: {} :: {}:{:-5} :: {:2} :: {}:{} :: {} :: ({})",
```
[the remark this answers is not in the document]
mackal: Good catch

common/database.h
mackal: I left that just because, although I think removing it is probably fine. (There was some calls that use a string literal still, left it for them)

zone/embparser.cpp
mackal: Gonna not resolve this for now, would like some others to double check it as well.

hgtw (approved):

Akkadius:
Item script interaction the perl side needs to be tested prior to merge here

Akkadius (approved):

Akkadius (approved):

PR #1252, opened 2021-02-15
[Experience] Add exp mod to npc types to let a server op change the exp modifier

KimLS:
Mostly envisioned to be useful for custom content without having a really hacky mod script involved.

### DIFF
--- a/utils/sql/git/required/2021_02_14_npc_exp_mod.sql
+++ b/utils/sql/git/required/2021_02_14_npc_exp_mod.sql
@@ -1,0 +1,1 @@
+ALTER TABLE `npc_types` ADD COLUMN `exp_mod` INT NOT NULL DEFAULT '100' AFTER `always_aggro`;
--- a/zone/zonedump.h
+++ b/zone/zonedump.h
@@ -148,6 +148,7 @@ struct NPCType
 	uint16	use_model;
 	int8	flymode;
 	bool	always_aggro;
+	int     exp_mod;
 };
 
 namespace player_lootitem {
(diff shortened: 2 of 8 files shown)


mackal (changes requested):
You forgot to update CURRENT_BINARY_DATABASE_VERSION in common/versions.h

KimLS:
Updated the version.h

zone/zonedump.h
```diff
@@ -148,6 +148,7 @@ struct NPCType
 	uint16	use_model;
 	int8	flymode;
 	bool	always_aggro;
+	int     exp_mod;
```
thalix1337, line 151: This should probably be uint8, because negative values have no effect, the default value is 100 and the mob already gives no EXP at value 0.

utils/sql/git/required/2021_02_14_npc_exp_mod.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `npc_types` ADD COLUMN `exp_mod` INT NOT NULL DEFAULT '100' AFTER `always_aggro`;
```
thalix1337, line 1: I think for the database type SMALLINT unsigned would be sufficient (0 to 65535), that would be at most 655times the normal EXP. Depending on the amount of EXP, such an increase could otherwise also lead to overflow in "uint32 Client::GetExperienceForKill(Mob *against)".

mackal (approved):

KinglyKrab (approved):

Akkadius (approved):

PR #1254, opened 2021-02-16
[Leak] Fix NPC Scale Manager leaking

mackal:
It should be fine to init further down after the early returns.

This wasn't really a leak that affected performance, but mainly moving
to shutup valgrind


KimLS (approved):
Looks good.

KinglyKrab (approved):

Akkadius (approved):

PR #1256, opened 2021-02-16
Fix zlib-ng compilation errors on newer GCC

mackal:


mackal:
Looks like KLS was able o resolve the issues with newer zlibng, so we don't need this.

PR #1257, opened 2021-02-17
[Scheduler] Event scheduler implementation

Akkadius:
**This PR is in draft and not final; will finalize database schema code changes after other merges**

This initial implementation while fairly solid, is a first release and is easy to change for further use cases and needs down the line

This is something I've wanted for a long time as a server operator, but more specifically out of a now use case of PEQ having scheduled and rotated hot zones

**Server Event Scheduler**

The server has currently two process driven schedulers

* World
  * Broadcast Messages (broadcast) (Stateless)
  * Reload World (reload_world) (Stateless)
* Zone
  * Rule Value Change (rule_change) (Stateful)
  * Content Flag Change (content_flag_change) (Stateful)
  * Hotzone Activate (hot_zone_activate) (Stateful)

Zone maintains state, world does not in the current implementation. For current every current zone event activation, there is an event deactivation event which undoes the state change.

World does not maintain state, it is used for single event triggers and supports cron expressions since the events can be kicked off on a routine basis

### Event Loading

When world or zone boots, you will see the initial load of events. Zone will also reload if World has detected any changes to any of the scheduled events in the database once a minute

```
[Zone] [Scheduler] Loaded Event (1) [Hotzone Rotation Day 1-10] type [hot_zone_activate] start [2/3/2021 3:1:00] end [2/16/2021 4:40:00] cron [] created []
[Zone] [Scheduler] Loaded Event (2) [Hotzone Rotation Day 11-20] type [hot_zone_activate] start [2/3/2021 15:1:00] end [2/15/2021 19:7:00] cron [] created []
[Zone] [Scheduler] Loaded Event (3) [Hotzone Rotation Day 21-33] type [hot_zone_activate] start [2/3/2021 3:1:00] end [2/16/2021 4:40:00] cron [] created []
[Zone] [Scheduler] Loaded Event (4) [Broadcast Test] type [broadcast] start [2/16/2021 22:56:00] end [2/16/2021 22:56:00] cron [] created []
[Zone] [Scheduler] Loaded Event (15) [Broadcast Test Cron] type [broadcast] start [2/16/2021 22:56:00] end [2/16/2021 22:56:00] cron [* * * * * *] created []
[Zone] [Scheduler] Loaded Event (16) [Experience Multiplier] type [rule_change] start [2/16/2021 22:56:00] end [2/16/2021 22:56:00] cron [] created []
[...]

PR #1260, opened 2021-02-20
Add string util search_deliminated_string

mackal:
This function takes a string of deliminated an see if another string is
one of those

This function also verifies it's not finding a substring

The motivational example for this function is the hotzone checking in PR #1257 

Currently our easiest way of doing this task is to split the string, this is rather needlessly expensive, so I wrote this.

Benchmarks: (on 8700k, stock, governor set to performance, optimization level 2 an march not set)

```----------------------------------------------------
Benchmark             Time           CPU Iterations
----------------------------------------------------
bench_split         871 ns        871 ns     796901
bench_search         38 ns         38 ns   18419186
```

### DIFF
--- a/common/string_util.cpp
+++ b/common/string_util.cpp
@@ -125,6 +125,19 @@ std::vector<std::string> SplitString(const std::string &str, char delim) {
 	return ret;
 }
 
+std::string::size_type search_deliminated_string(const std::string &haystack, const std::string &needle, const char deliminator)
+{
+	// this shouldn't go out of bounds, even without obvious bounds checks
+	auto pos = haystack.find(needle);
+	while (pos != std::string::npos) {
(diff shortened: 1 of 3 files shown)


hgtw (approved):
An option to ignore spaces around delimiters would probably be useful for this as well so it could be used for an input string like `"befallen, charasis, dalnir"`

mackal:
> An option to ignore spaces around delimiters would probably be useful for this as well so it could be used for an input string like `"befallen, charasis, dalnir"`

That's a good point, wonder if we should bother with it though, will let others chime in.

Akkadius:
Someone can / should sanitize whitespaces before providing input to this function so its fine as is

Akkadius (approved):

PR #1272, opened 2021-02-25
[Bug Fix] Do not summon pet or set new pet state when PC summoned by a mob

noudess:
This is in response to the discussion in eqemu-coders:

https://discord.com/channels/212663220849213441/557677602706423982/814234259845283870

The "fix" is simple.  From what I read, the pet does not switch from guard to follow on live.  I did not check live myself, I'm relying on the information in the thread.

If the pet should stay on guard, this should fix the issue.

### DIFF
--- a/zone/zoning.cpp
+++ b/zone/zoning.cpp
@@ -590,10 +590,11 @@ void Client::ProcessMovePC(uint32 zoneID, uint32 instance_id, float x, float y, 
 			return;
 		}
 
-		if(GetPetID() != 0) {
+		if(zm != SummonPC && GetPetID() != 0) {
 			//if they have a pet and they are staying in zone, move with them
 			Mob *p = GetPet();
 			if(p != nullptr){
+				p->SetPetOrder(SPO_Follow);
 				p->GMMove(x+15, y, z);	//so it dosent have to run across the map.
 			}
 		}
--- a/zone/zoning.cpp
+++ b/zone/zoning.cpp
@@ -594,7 +594,6 @@ void Client::ProcessMovePC(uint32 zoneID, uint32 instance_id, float x, float y, 
 			//if they have a pet and they are staying in zone, move with them
 			Mob *p = GetPet();
 			if(p != nullptr){
-				p->SetPetOrder(SPO_Follow);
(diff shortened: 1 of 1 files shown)


KinglyKrab (approved):

noudess:
One thought I had, was if summon was GM solicited to add this back in.

Let me know and I can do that below where it checks zone mode.

mackal:
Pet shouldn't be moved either, it should stay in place

noudess:
> Pet shouldn't be moved either, it should stay in place

Ok.  What about a GM summons?  Shall I just move the code to only occur in that instance?

mackal:
> > Pet shouldn't be moved either, it should stay in place
> 
> Ok.  What about a GM summons?  Shall I just move the code to only occur in that instance?
> 
> You realize that once PC is attacked after summons, pet will travel to assist PC?  I agree this is correct, just stating it 'cause perhaps that's why someone added this to begin with.
> 

I'm not sure on the 100% correctly behavior. My pet 100% did not get automoved when I took a portal when my pet was on guard. Does the pet move on live when they have hold and guard on? Or is it because they're moving to attack?

noudess:
> > > Pet shouldn't be moved either, it should stay in place
> > 
> > 
> > Ok.  What about a GM summons?  Shall I just move the code to only occur in that instance?
> > You realize that once PC is attacked after summons, pet will travel to assist PC?  I agree this is correct, just stating it 'cause perhaps that's why someone added this to begin with.
> 
> I'm not sure on the 100% correctly behavior. My pet 100% did not get automoved when I took a portal when my pet was on guard. Does the pet move on live when they have hold and guard on? Or is it because they're moving to attack?

They will start to come as soon as PC is attacked actually, so all we're doing is removing the instamove.  I think you're right that the move be removed.  My only question is, shall we do the move and follow if its a GM summons.

noudess:
After further investigation, things like evac need to move the pet.

I will look at each type of zm and test and post something that covers them all.

Please hold on approval.

noudess:
Ok.  I like this solution better.  It fixes the problem that is being seen by players without impacting all the other cases like evac, succor, etc.

PR #1288, opened 2021-03-07
[Cleanup] Only link Lua with zone

mackal:
Only zone requires to be linked to lua, so let's not link EVERYTHING.

There is probably more clean up we could do, but I know for a fact Lua
is only used by zone.

This probably has some measurable benefits that probably don't really
matter that much :)


CMakeLists.txt
```diff
@@ -348,6 +348,8 @@ IF(UNIX)
 	ENDIF()
 ENDIF()
 
+SET(ZONE_LIBS ${ZONE_LIBS} ${SERVER_LIBS})
```
mackal, line 351: I had no issues with this line when lua was disabled. So I think it's fine.

Akkadius (approved):

Akkadius:
LGTM; can get @KimLS to take a quick look before merging if you want

KimLS (approved):

KimLS:
Looks good to me.

PR #1290, opened 2021-03-09
Remove unneeded calls to DoAnim() for spells/bardsong

noudess:
I found that disciplines like "Focused Will" were causing our characters to stand up on the client, while on the server their sit/stand status was unchanged.  This got us out of sync with the GUI the next time sit/stand was pressed.

Testing Focused Will on live, it does not require the character to stand up, nor does the character do any animation aside from some glowing.  Testing non spell disciplines, both our server and live stand up first.

The change in this PR makes our server behave live like for Disciplines that cast spells for the disciplines I tested.  We still get the glowing effect, but the character sit/stand does not change.

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -2389,10 +2389,6 @@ bool Mob::SpellFinished(uint16 spell_id, Mob *spell_target, CastingSlot slot, ui
 		}
 	}
 
-	if (!spells[spell_id].cast_not_standing) {
-		DoAnim(spells[spell_id].CastingAnim, 0, true, IsClient() ? FilterPCSpells : FilterNPCSpells);
-	}
-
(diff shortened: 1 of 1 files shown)


KinglyKrab (approved):

mackal:
Does this have anything to do with the CAST_NOT_STANDING flag?

noudess:
Since DoAnim() ends up standing the char up, I'd be inclined to change my PR to use that flag instead.

noudess:
Thanks to @mackal suggestions, I looked at these in more detail.  Removing these calls entirely doesn't seem to have any ill effects.  At his suggestion I used two clients and tested to see if animations worked on both without the calls.  They do.  And remove fixes the problem where we were standing chars up when we should not, which was my original goal.

mackal (approved):

PR #1291, opened 2021-03-11
[Crash] Bugfix for zone crash caused by #flymode -1

thalix1337:
Bugfix for zone crash caused by #flymode when invalid values are passed. Also, when #flymode is typed without parameters, the correct syntax is displayed.


KinglyKrab (commented):
Should just atoi the separator argument and check if >= 0 and <= 5.

thalix1337:
> Should just atoi the separator argument and check if >= 0 and <= 5.

I have tried to find the smallest possible change to fix the issue.

mackal:
> Should just atoi the separator argument and check if >= 0 and <= 5.

I agree with this

> > Should just atoi the separator argument and check if >= 0 and <= 5.
> 
> I have tried to find the smallest possible change to fix the issue.

I think just checking IsNumber, atoi, etc is a cleaner fix. Which is better.

KinglyKrab:
> > Should just atoi the separator argument and check if >= 0 and <= 5.
> 
> I have tried to find the smallest possible change to fix the issue.

```cpp
if (strlen(sep->arg[1]) == 1 && sep.IsNumber(1) && atoi(sep->arg[1]) >= 0 && atoi(sep->arg[1]) <= 5)

splose:
^ is the best way to make that code smaller the original is pretty awful.

thalix1337:
> ^ is the best way to make that code smaller the original is pretty awful.

Understood and changed as requested. The PR is tested on my server and works for me. I still have some questions but I will ask them in Discord.

KinglyKrab (approved):
Looks good to me. 👍

splose:
> > ^ is the best way to make that code smaller the original is pretty awful.
> 
> Understood and changed as requested. The PR is tested on my server and works for me. I still have some questions but I will ask them in Discord.

Just to be clear I was referring to the code that was there before you touched it was awful.

Looks good tho.

Akkadius (approved):

PR #1296, opened 2021-03-15
[Boats] Fix x/y offsets from client to reflect EQ x/y instead of boat heading…

noudess:
At the end of  2019 I pushed a [[PR](https://github.com/EQEmu/Server/pull/937)] ( with @Uleat ) which corrected an issue where clients did not properly see other clients that were on a boat.  This was because the x/y sent by the client is an offset from the center of the boat they are on,  when the client is onboard.  We had no code for this, so the moment a client went onto a boat they warped off to close to 0,0 on other client's screens.

I've come to learn, while trying to get cross zone ports from boat to boat working, that the client x,y is also an x,y offset which is based on the boat's heading.  So if the boat has a heading of 0, the x,y offsets are correct in EQ map coordinates, but if the boat is at heading of 128 for example, the x,y sent by the client need to be rotated about the boat's origin to get correct x,y offsets in the EQ map world.

To complicate matters, simple rotation around the boat origin math needed to be altered due to the fact that the EQ X axis is negative to the right, while most axis systems are x positive to the right.

I believe this PR will:

- render clients on the boat in the correct location for other viewing clients
- will not change the view from the rider's perspective as they don't get these updates
- will not fix the drift that we all see.. that is a client issue as best I can tell - where on some angles (mostly non integer) the x,y offsets drift due to rounding (or some other math issue)
- clients actual x and y values will be correct if queried from quests when on a boat (they used to be off by rotation)

The PR also modifies the boat/zomm code slightly to share a variable to clean up that code a little.

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -4411,27 +4411,27 @@ void Client::Handle_OP_ClientUpdate(const EQApplicationPacket *app) {
 
 	PlayerPositionUpdateClient_Struct *ppu = (PlayerPositionUpdateClient_Struct *) app->pBuffer;
 
-	/* Boat handling */
-	if (ppu->spawn_id != GetID()) {
-		/* If player is controlling boat */
-		if (ppu->spawn_id && ppu->spawn_id == controlling_boat_id) {
-			Mob *boat = entity_list.GetMob(controlling_boat_id);
-			if (boat == 0) {
-				controlling_boat_id = 0;
-				return;
-			}
+	/* Non PC handling like boats and eye of zomm */
+	if (ppu->spawn_id && ppu->spawn_id != GetID()) {
+		Mob *cmob = entity_list.GetMob(ppu->spawn_id);
 
+		if (!cmob) {
+			return;
+		}
+
+		if (cmob->IsControllableBoat()) {
+			// Controllable boats
 			auto boat_delta = glm::vec4(ppu->delta_x, ppu->delta_y, ppu->delta_z, EQ10toFloat(ppu->delta_heading));
-			boat->SetDelta(boat_delta);
+			cmob->SetDelta(boat_delta);
 
 			auto outapp = new EQApplicationPacket(OP_ClientUpdate, sizeof(PlayerPositionUpdateServer_Struct));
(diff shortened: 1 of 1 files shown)


zone/client_packet.cpp
```diff
@@ -4476,9 +4473,19 @@ void Client::Handle_OP_ClientUpdate(const EQApplicationPacket *app) {
 			LogError("Can't find boat for client position offset.");
 		}
 		else {
-			cx += boat->GetX();
-			cy += boat->GetY();
+			// Calculate angle from boat heading to EQ heading
+			float theta = fmod(((boat->GetHeading() * 360.0) / 512.0),360.0);
```
mackal, line 4477: We should use the versions of these math functions in std namespace at least.

noudess: Happy to change them. Which ones and what is the benefit?    Just fmod?

mackal: cos and sin as well. the std namespace versions are type overloaded where the C API ones aren't, which means you will incur float to double conversions (the conversion from EQ headings should be floats as well)

Idealy if we could use the FastCos/FastSin that works on EQ Headings it would be better :P

noudess: Thanks.  I'll fix it up tomorrow AM and retest.

Akkadius:
Looks like this one has some comments to be followed up on

thalix1337 (approved):
To be able to comment on the code I don't know enough about C++ yet. But I have tested the function extensively and found no problems. Only that I had to change a script on my server a bit, because I had rotated the rotated coordinates myself - if that makes any sense.

noudess:
I have found something related.  There are times, where the movement manager code sets heading without actually rotating the mob.  This can and will cause errors if the mob is a boat and a client is onboard.

I'm looking to see if  fixing these spots impacts this PR.

If not, I'd like to get this PR pushed through (once I update to use the std library as suggested, and attack the movement manager portion in a new PR.

zone/client_packet.cpp
noudess: How does this look?  I also reduced the # of calls.

freyonsecurity:
Can we please get this merged?

mackal (approved):
I would prefer we didn't have to convert away from EQ headings, but probably not worth figuring out :P

PR #1299, opened 2021-03-16
[Bugfix] Fix for spawn2.version

splose:
setting spawn2.version to -1 will now properly spawn mobs in all zone versions

### DIFF
--- a/zone/spawngroup.cpp
+++ b/zone/spawngroup.cpp
@@ -191,7 +191,8 @@ bool ZoneDatabase::LoadSpawnGroups(const char *zone_name, uint16 version, SpawnG
 				WHERE
 				spawn2.spawngroupID = spawngroup.ID
 				AND
-				spawn2.version = {} and zone = '{}'
+				(spawn2.version = {} OR version = -1)
+				AND zone = '{}'
 				{}
 		),
 		version,


Akkadius (approved):

zone/spawngroup.cpp
```diff
@@ -191,7 +191,8 @@ bool ZoneDatabase::LoadSpawnGroups(const char *zone_name, uint16 version, SpawnG
 				WHERE
 				spawn2.spawngroupID = spawngroup.ID
 				AND
-				spawn2.version = {} and zone = '{}'
+				(spawn2.version = {} OR version = -1)
```
Akkadius, line 194: For future note; not prefixing the table in a join can throw an error in the query parser depending on how the query is built

I am making an assumption this was tested

Akkadius (approved):

zone/spawngroup.cpp
splose: Yeah I tested it.. Looks like in this case its okay, but will keep in mind in the future

91fa7a92e0387e50cdb47dbf10100198

splose: my guess is it is not throwing an error because there is no version field in the spawngroup table?

Akkadius:
Correct, if both tables had the same it would throw

If you tested and it was fine that’s all I care about, just precaution

PR #1302, opened 2021-03-18
[Bug Fix] Add Bounds Checking to OP_LFGCommand Comment Processing

mackal:
Theoretically this could be used to corrupt memory, but they would have
to get extremely lucky to actually execute a successful attack


carb-edward:
> 
> 
> Theoretically this could be used to corrupt memory, but they would have
> to get extremely lucky to actually execute a successful attack

Never tell me the odds.

joligario (approved):

Akkadius (approved):

PR #1304, opened 2021-03-22
[Fix] UpdateFilteredMessageString fix

neckkola:
I was testing 9 messages using FilteredMessageString and GENERIC_9_STRINGS and noticed that when adding the 9th it would cause an exception error and crash zone.exe.  Updated to resolve in both Client::FilteredMessageString and Client::MessageString.  The for (; message_arg[argcount]; ++argcount) length += strlen(message_arg[argcount]) + 1; would try to read message_arg[9] which is invalid.


Akkadius (approved):

mackal:
Oh, I see why it's crashing now and why going 1 over fixes it lol.


I think I'm gonna submit a PR that changes these functions to be a bit cleaner instead of this fix though.

PR #1315, opened 2021-03-29
[Expeditions] Refactor expedition caching

hgtw:
Add common expedition base class

Use repository for zone and world expedition caching

World now stores members and leader as Member objects instead of ids

This improves readability of the caching methods and lets world cache
expedition dzs and members like zone. World also now caches expeditions
as unique_ptr which will be necessary for future dz callback lambdas
that capture 'this' so addresses don't change on cache vector resizes.


Akkadius (approved):
Code looks good!

Issue #737, opened 2018-06-21
bards chain spawn AA vendors


huffin:
when you click the special AA's that summon a vendor (ie summon resupply agent) they just keep spawning them until the bard sings a different song - can definitely crash a zone ><

PR #1317, opened 2021-03-31
[Bug Fix] Misc Bard Song Fixes

mackal:
Bard songs with a mana cost and Vet AAs were repulsing when they shouldn't have.


joligario (approved):

hgtw (approved):
This will partially fix #737 for summon Permutation Peddler.

Spells like Summon Resupply Agent and Summon Banker use SE_Familiar which will still pulse though. Normally those get blocked by buff checks (10 min buff), but clicking those off will still result in bard pulses resummoning them

mackal:
Okay, we can add those too, I didn't think to check that corner case

PR #1319, opened 2021-04-01
Added a check to stop Bard song for Mezz/Stun

RoT-PvP:


joligario (commented):
At quick glance I think you have a logic short circuit issue

RoT-PvP:
> At quick glance I think you have a logic short circuit issue

I'm fairly new to coding and PRs specifically. Any tips or could you further explain exactly what you mean? This is works for my server currently.

mackal:
```cpp
if ((IsStunned() && bardsong != 0) || (IsMezzed() && bardsong != 0)) {
```

is what @joligario was saying it should be, but we can shorten this

```cpp
if (bardsong != 0 && (IsStunned() || IsMezzed())) {
```

joligario (approved):

mackal (approved):

mackal (changes requested):
Okay, after testing on live, a bard holding a song will not always be interrupted when they are stunned. I think fixing whatever is going wrong with stun at https://github.com/EQEmu/Server/blob/master/zone/spells.cpp#L4923 is a better fix. Probably just need to add some bardsong checks I guess since just the plain `InterruptSpell()` call isn't cutting it. the Mob::Mesmerize function probably needs tweaking as well.

mackal (approved):

PR #1324, opened 2021-04-08
[Quest API] Add HTML color tag method to Perl and Lua.

KinglyKrab:
- Add quest::gethexcolorcode(color_name) to Perl.
- Add eq.get_hex_color_code(color_name) to Lua.

Color name is case-insensitive.

Full color list here: https://pastebin.com/rUYKr1ye


mackal (approved):

zone/questmgr.cpp
```diff
@@ -4285,4 +4285,326 @@ std::string QuestManager::secondstotime(int duration) {
 		time_string = fmt::format("{} {}", seconds, second_string);
 	}
 	return time_string;
-}
\ No newline at end of file
+}
+
+std::string QuestManager::gethexcolorcode(std::string color_name) {
+	const std::map<std::string, std::string> colors = {
```
hgtw, line 4291: May want to make the map static also so it's not re-created every call

zone/questmgr.cpp
```diff
@@ -4285,4 +4285,326 @@ std::string QuestManager::secondstotime(int duration) {
...
+		{ "Violet Red2", "#E4317F" },
+		{ "Violet Red3", "#C12869" },
+		{ "Violet Red4", "#7D0541" },
+		{ "White", "#FFFFFF" },
+		{ "Yellow", "#FFFF00" },
+		{ "Yellow1", "#FFFC17" },
+		{ "Yellow Green", "#52D017" }
+	};
+	for (auto color : colors) {
+		if (!strcmp(str_tolower(color.first).c_str(), str_tolower(color_name).c_str())) {
```
hgtw, line 4604: Using `strcasecmp` here instead would avoid having to create the two temporary strings with `str_tolower`

hgtw (approved):
Also an alternative to looping the map is to give the map a case-insensitive comparator and doing normal lookups. Performance difference probably doesn't matter all that much in a quest api though

Akkadius (approved):

PR #1327, opened 2021-04-17
[Dynamic Zones] Rename dynamic zone structs

hgtw:
This renames some dz enum/structs that will no longer be expedition specific for clarity

### DIFF
--- a/common/dynamic_zone_base.h
+++ b/common/dynamic_zone_base.h
@@ -13,6 +13,21 @@
 class Database;
 class ServerPacket;
 
+struct DynamicZoneMember
+{
+	uint32_t id = 0;
+	std::string name;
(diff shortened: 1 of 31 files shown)


KinglyKrab (approved):

mackal (approved):

Akkadius (approved):
Overall LGTM.

It is a bit odd we did a rename of `char_id` to `id` which is a bit more ambiguous. We could have a different internal `id` separate from `char_id` so it feels like we lost a little bit of resolution there. Otherwise I'm good

hgtw:
It felt a bit redundant since the name of the struct is synonymous with characters and implies what it's for. This probably made more sense and looked better when I was replacing `member_id` with `member.id` type of stuff, but I separated and moved around a bunch of commits in the branch so that may no longer be as apparent

PR #1330, opened 2021-04-22
[Rules] Add rule 'GM:MinStatusToBypassLockedServer'

splose:
Rule Name: GM:MinStatusToBypassLockedServer
Default Status: 100
Description: Players >= this status can log in to the server even when it is locked

Believe you used to be able to change this via the old config? Either way should be very useful as the current requirement is 101 to get through a Locked server. Since our default is 100 now you will be able to get in at status 100 but nothing below that obviously.

The plan for me specifically is to create a new field in the accounts table and disconnect this whole thing from GM status entirely (field would be named ignoreLockedServer or bypassLockedServer or something). Don't know if that's too custom for Master source but will provide that as well when I am done with it if anyone wants it.


KinglyKrab (approved):

mackal (approved):

Akkadius (approved):
Good to get rid of magic numbers

PR #1332, opened 2021-04-23
Add a way to bypass locked server without giving client GM status

splose:
Allow certain accounts access to the server when its locked without giving them GM status as mentioned in https://github.com/EQEmu/Server/pull/1330

Not sure how to include the database field.. would I just commit that in the utils folder like anything else? I've included a screenshot of it below. 

I've tested this and it works. Also added it to the debug logging. (screenshots without sensitive information)

11cc49b6610dcc84c470d26c4c0c5932
47052e1d033d49d484e490e6ed9d470f

### DIFF
--- a/common/database.cpp
+++ b/common/database.cpp
@@ -2479,3 +2479,14 @@ bool Database::CopyCharacter(
 	return true;
 }
 
+bool Database::GetBypassFlagByLSAccountID(uint32 lsaccount_id)
+{
+	std::string query = fmt::format("select bypassServerLock from account WHERE lsaccount_id = {}", lsaccount_id);
+	auto results = QueryDatabase(query);
+	auto row = results.begin();
+
+	if (atoi(row[0]))
+		return true;
+
+	return false;
+}
--- a/common/database.h
+++ b/common/database.h
@@ -267,6 +267,9 @@ public:
 	int		CountInvSnapshots();
 	void	ClearInvSnapshots(bool from_now = false);
 
+	/* Bypass Server Lock */
+	bool GetBypassFlagByLSAccountID(uint32 lsaccount_id);
+
 	/* EQEmuLogSys */
 	void	LoadLogSettings(EQEmuLogSys::LogSettings* log_settings);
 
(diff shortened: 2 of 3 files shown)


Akkadius:
Bypass status level can be set to as low as `1` and you don't have to flag anyone as GM. I'd rather not have two knobs to handle the same thing. This may seem pedantic but less columns, fields, settings to affect the same thing the better as we've historically confused people by doing exactly that in many places of the server domain. 

If we're going to consider adding this field, the code variable and the database column name do need to be `snake_case` instead of `camelCase` https://eqemu.gitbook.io/database-schema/#column-naming

For adding new database migrations: https://eqemu.gitbook.io/server/categories/operation/database-schema-migrations

Happy to go deeper into any of this with you if you have questions. Laying down as I'm not feeling too great atm but at least wanted to get PR's while I had a bit of a chance

splose:
Sorry for late response on this.. Been taking a bit of a needed sanity break from EQEmu.

I don't think you're being pedantic I actually agree with you about the two knobs for same thing comment. My original rule (the minstatus one) was meant to be just a temporary solution and this was meant to be the permanent one.

There is a crash associated with this.. My guess is because I'm using SQL to get the status of the DB field and its not fast enough? Not really sure, but if your field is set to 1 you can log in fine.. the next one who tries to log in with a field of 0 will crash world.exe

Haven't had time to sit down and look at it, but I will either today or tomorrow. 

Ideally if we are going to choose one to keep in the source I would prefer this, as I'd like to remove GM status from the equation altogether (setting minstatus rule to 1 is still kinda ehh for me, but is definitely a valid alternative). 

Let me know which one you would prefer and we will move forward that way.

joligario:
Kind of mixing terms which probably should have been mentioned with #1330. Account Status and GM status are two different things. The rule in #1330 probably should be named Account instead of GM.

Personally, I prefer using an account status rather than adding a column to the database.

Database calls are pretty quick - I can't imagine that you are having a race issue causing a crash.

PR #1333, opened 2021-04-23
Fix Perl NPC GetAvoidanceRating() Perl Croak. [skip ci]

KinglyKrab:


joligario (approved):
Could rename to add [skip ci] in the title when compile not necessary.

KinglyKrab:
Gotcha, guess it's too late now.

mackal (approved):

PR #1339, opened 2021-04-28
[Feature] Allow any spawn2 spawned mob to path while zone is idle if new flag is set.

noudess:
This PR simplifies and expands the capability to allow mobs to continue to process even when a zone is idle.

In a PR 4 years ago, I submitted the following PR so that mobs like Fippy, that spawn and then run to the end of a grid and despawn would continue to do this on the set timing regardless of zone status.  This made these spawns more realistic and less hinky (ie,. they used to start there run as soon as the 1st player ran in).

That PR, while helpful, attached the mobs mode of moving while the zone was idle to grid types 4 and 6.  These happened to be the most glaring cases of mobs being too predictable based on the IDLE zone optimizations.

This PR removes that code, and now ANY spawn2 can be set to always process regardless of zone status.  I find this especially useful for named mobs that path but use a roam box, or a grid type other than 4 and 6.

The PR also performs a DB update for any mobs on grid types 4 and 6, so they will continue to behave as they did under the old system without any changes by the server ops.

I talked about this idea with @Akkadius before moving ahead. 

I am not fluent with how the lua_add_spawn_point function works.  Please let know if my changes to that are sufficient.

### DIFF
--- a/zone/entity.cpp
+++ b/zone/entity.cpp
@@ -517,13 +517,15 @@ void EntityList::MobProcess()
 			mob_settle_timer->Disable();
 		}
 
+		Spawn2* s2 = mob->CastToNPC()->respawn2;
+
+		// Perform normal mob processing if any of these are true:
+		//	-- zone is not empty
+		//	-- a quest has turned it on for this zone while zone is idle
+		//	-- the entity's spawn2 point is marked as path_while_zone_idle
+		//	-- the zone is newly empty and we're allowing mobs to settle
 		if (zone->process_mobs_while_empty || numclients > 0 ||
-			mob->GetWanderType() == 4 || mob->GetWanderType() == 6 ||
-			mob_settle_timer->Enabled()) {
-			// Normal processing, or assuring that spawns that should
-			// path and depop do that.  Otherwise all of these type mobs
-			// will be up and at starting positions, or waiting at the zoneline
-			// if they chased the PCs  when idle zone wakes up.
+			(s2 && s2->PathWhenZoneIdle()) || mob_settle_timer->Enabled()) {
 			mob_dead = !mob->Process();
 		}
 		else {
--- a/zone/lua_general.cpp
+++ b/zone/lua_general.cpp
@@ -1732,6 +1732,7 @@ void lua_add_spawn_point(luabind::adl::object table) {
 		uint32 variance;
 		uint32 timeleft = 0;
 		uint32 grid = 0;
+		bool path_when_zone_idle;
 		int condition_id = 0;
 		int condition_min_value = 0;
 		bool enabled = true;
@@ -1841,6 +1842,14 @@ void lua_add_spawn_point(luabind::adl::object table) {
 			}
 		}
 
+		cur = table["path_when_zone_idle"];
+		if(luabind::type(cur) != LUA_TNIL) {
+			try {
+				grid = luabind::object_cast<bool>(cur);
+			} catch(luabind::cast_failed &) {
+			}
+		}
+
 		cur = table["condition_id"];
 		if(luabind::type(cur) != LUA_TNIL) {
 			try {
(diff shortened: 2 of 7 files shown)


zone/spawn2.h
```diff
@@ -32,7 +32,7 @@ class Spawn2
 	Spawn2(uint32 spawn2_id, uint32 spawngroup_id,
 		float x, float y, float z, float heading,
 		uint32 respawn, uint32 variance,
-		uint32 timeleft = 0, uint32 grid = 0,
+		uint32 timeleft = 0, uint32 grid = 0, bool in_path_when_zone_idle=0,
```
KinglyKrab, line 35: Maybe use false here.

zone/spawn2.cpp
```diff
@@ -632,7 +638,8 @@ bool ZoneDatabase::PopulateZoneSpawnList(uint32 zoneid, LinkedList<Spawn2*> &spa
 Spawn2* ZoneDatabase::LoadSpawn2(LinkedList<Spawn2*> &spawn2_list, uint32 spawn2id, uint32 timeleft) {
 
 	std::string query = StringFormat("SELECT id, spawngroupID, x, y, z, heading, "
-                                    "respawntime, variance, pathgrid, _condition, "
+                                    "respawntime, variance, pathgrid, "
+									"path_when_zone_idle, _condition, "
```
KinglyKrab, line 642: Weird formatting.

zone/lua_general.cpp
```diff
@@ -1841,6 +1842,14 @@ void lua_add_spawn_point(luabind::adl::object table) {
 			}
 		}
 
+		cur = table["path_when_zone_idle"];
+		if(luabind::type(cur) != LUA_TNIL) {
+			try {
+				grid = luabind::object_cast<bool>(cur);
```
KinglyKrab, line 1848: Should be path_when_zone_idle not grid.

zone/lua_general.cpp
```diff
@@ -1732,6 +1732,7 @@ void lua_add_spawn_point(luabind::adl::object table) {
 		uint32 variance;
 		uint32 timeleft = 0;
 		uint32 grid = 0;
+		bool path_when_zone_idle;
```
KinglyKrab, line 1735: Can probably initialize as false, not sure if necessary.

KinglyKrab (approved):
Looks good to me.

zone/spawn2.cpp
noudess: Think this is tabs and spaces mixed.  Also best views with 4 char tabs.  Fixing to be all tabs.

zone/spawn2.h
noudess: Fixing.

zone/lua_general.cpp
noudess: Fixing.

zone/lua_general.cpp
noudess: Fixing.

thalix1337:
I can at least already confirm that mobs with grid type 4 and 6 behave as before the patch. i'm going to test intensively on the weekend.

noudess:
I'm happy with my testing.

thalix1337:
So, I tested as best I could and would say everything works as it should.

zone/entity.cpp
```diff
@@ -517,13 +517,15 @@ void EntityList::MobProcess()
 			mob_settle_timer->Disable();
 		}
 
+		Spawn2* s2 = mob->CastToNPC()->respawn2;
+
+		// Perform normal mob processing if any of these are true:
```
Akkadius, line 522: Thanks for leaving comments for the next guy

Akkadius:
Thanks guys for helping Noudess with testing!

LGTM Noud!

noudess:
> Thanks guys for helping Noudess with testing!
> 
> LGTM Noud!

LGTM?  :)

joligario:
(L)ooks (G)ood (T)o (M)e

PR #1343, opened 2021-05-04
[Boats] Prevent client update while on boat if boat turning

noudess:
Updates for client positions while a client is on a boat are received as x/y offsets based on the boats current heading.

When the server rotates the boat, it updates Heading in the boat server instance.  This means all future updates from the client will be processed according to that new heading.

Since the server only approximates rotation progress, there will be errors if we use client updates for PCs on boats while the boat is turning.

This patch tosses out the updates for the client while the boat he/she is on is turning, eliminating that error.

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -4474,6 +4474,8 @@ void Client::Handle_OP_ClientUpdate(const EQApplicationPacket *app) {
 			LogError("Can't find boat for client position offset.");
 		}
 		else {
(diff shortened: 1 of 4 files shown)


KinglyKrab (approved):

thalix1337:
In the game the patch works very well. After 2 hours of boating with three accounts I could not notice any problems.

Akkadius (approved):

PR #1346, opened 2021-05-09
[Bug Fix] #grid delete was deleting the grids everywhere

noudess:
Fix to only delete correct zone grid.


KinglyKrab (approved):
Good catch. 👍

joligario (approved):

PR #1354, opened 2021-05-13
 [Bug Fix] EntityList::AESpell fix for Pacify / Mez

dencelle:
this fixes AE pacify / Mez spells only landing on 4 when it shouldn't have a cap

### DIFF
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -369,7 +369,7 @@ RULE_BOOL(Spells, NPCInnateProcOverride, true, "NPC innate procs override the ta
 RULE_BOOL(Spells, OldRainTargets, false, "Use old incorrectly implemented maximum targets for rains")
 RULE_BOOL(Spells, NPCSpellPush, false, "Enable spell push on NPCs")
 RULE_BOOL(Spells, July242002PetResists, true, "Enable Pets using PCs resist change from July 24 2002")
-RULE_INT(Spells, AOEMaxTargets, 0, "Max number of targets a Targeted AOE spell can cast on")
+RULE_INT(Spells, AOEMaxTargets, 0, "Max number of targets a Targeted AOE spell can cast on. Set to 0 for no limit.")
 RULE_CATEGORY_END()
 
 RULE_CATEGORY(Combat)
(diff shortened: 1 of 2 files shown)


zone/effects.cpp
```diff
@@ -793,6 +793,8 @@ void EntityList::AESpell(
 	}
 	else if (IsTargetableAESpell(spell_id) && is_detrimental_spell && !is_npc) {
 		max_targets_allowed = 4;
+		if (spells[spell_id].effectid[2] == 18 || spells[spell_id].effectid[0] == 31) // pacify and mez
```
Akkadius, line 796: Please use constants here for 18 and 31

`SE_Lull`
`SE_Mez`

From `spdat.h`

Also others who know more than me will need to comment on the Live like relevance of this fix

dencelle: constants added on commit after this

as for it being valid live, Drogerin checked on live using a enchanter to make sure there wasn't a cap for mez and pacify
image

Akkadius: Is there a reason we're restricting the checking here to specifically effect slots 2 and 0 selectively for each type ?

dencelle: pacify is only on slot 2 and mez is only on slot 1, we could check each slot for both types i guess but this seemed the most efficient and costing the least amount of time since we know what slots those effects will always be at

Akkadius (approved):

mackal (approved):

common/ruletypes.h
```diff
@@ -369,6 +369,7 @@ RULE_BOOL(Spells, NPCInnateProcOverride, true, "NPC innate procs override the ta
 RULE_BOOL(Spells, OldRainTargets, false, "Use old incorrectly implemented maximum targets for rains")
 RULE_BOOL(Spells, NPCSpellPush, false, "Enable spell push on NPCs")
 RULE_BOOL(Spells, July242002PetResists, true, "Enable Pets using PCs resist change from July 24 2002")
+RULE_INT(Spells, AOEMaxTargets, 0, "Max number of targets a Targeted AOE spell can cast on")
```
mackal, line 372: Description should probably include something about 0 being unlimited.

dencelle: thanks, i hate writing out descriptions.

KinglyKrab (approved):

joligario (approved):
LGTM

noudess:
LGTM (old dogs can learn new acromyms)

noudess (approved):

PR #1358, opened 2021-05-13
[Expeditions] Store members on dynamic zone

hgtw:
This moves members from expeditions so other systems can use them

Replace expedition_members table with dynamic_zone_members

Move 'EnableInDynamicZoneStatus' rule to DynamicZone namespace

Modify #dz list to show dz members (not instance players) and type name

Move various queries to repository methods


KinglyKrab (approved):

common/dynamic_zone_base.cpp
```diff
@@ -287,3 +308,123 @@ std::unique_ptr<ServerPacket> DynamicZoneBase::CreateServerDzLocationPacket(
...
+	if (it != m_members.end() && it->status != status)
+	{
+		it->status = status;
+		return true;
+	}
+
+	return false;
+}
+
+std::string DynamicZoneBase::GetDynamicZoneTypeName(DynamicZoneType dz_type)
```
Akkadius, line 414: I dig it

Akkadius (approved):

Issue #947, opened 2020-01-19
NPC charming a PC: bugs


huffin:
Currently it is very easy to make an NPC faction war happen. These are some parsed mechanics that need to be implemented to make NPC charm function more accurately and reduce how easy it is to get NPCs to attack each other.

1.)  If after you are charmed, and another spell was already being cast by a different NPC, then that spell lands on you while you are now charmed, that NPC will continue to attack you, it will call for assist (minus the mob that charmed you) and they kill you, you do not fight back, aggro is not transferred to owner NPC.

The main issue with how #1 currently functions on PEQ is that aggro transfers from the charmed PC  to the owner NPC way too easily which starts an instant faction war.

Issue #905, opened 2019-09-26
Charm pets continuing fighting their target after charm break


Isaaru:
This does NOT happen when you cast invis and break the invis, only when the charm breaks naturally.

mackal:
Can you describe how it should work for each situation? Like when you break by x pet should clear aggro and be removed from aggro lists or whatever.

Isaaru:
If any spell happens to hit one npc from the other (doesn't matter which direction), after the charm is broke, they will continue fighting each other. This is easily noticeable with DoTs, but can happen with spellcasters too if spells land post break.

The way it should work is they both hate the player like normal after the break. 

However, there's a wrench in the behavior. If I forcefully break my charm with invis both NPCs immediately hate me, ignoring the above described spell/dot aggro issue.

PR #1363, opened 2021-05-18
[Bug Fix] NPC not breaking charm correctly

dencelle:
#947 and #905
fixes the issue with charm breaking and spells being cast after to cause a faction war. this removes dots to stop faction wars also.

dot removal part needs better testing to ensure it works as intended


mackal:
Stripping dots sounds wrong to me.

dencelle:
seems wrong to me too but if you don't and a dot is still on the other npc when charm is broken a war still happens.
maybe we should leave this as something you can turn on or off according to a rule variable and it being off by default?

dencelle:
I'm not sure, according to everything I've read this problem is still
present on live as a whole and has been attempted to be fixed in a lot of
ways. This was the easiest and lightest weight way I could think of fixing
the issue, more than likely is not the best, ailia gave and idea of doing
something along the lines of a bunch of checks to see if the aggro/damage
should apply or not

GangstaEQ:
The way it functions currently prevents most faction wars. If someone wrote the same check for whether the dot tick should provide  aggro or not that would probably be the preferred solution but how it works currently stops faction wars. +1 upvote for going to master.

zone/spell_effects.cpp
```diff
@@ -4011,6 +4011,33 @@ void Mob::BuffFadeBySlot(int slot, bool iRecalcBonuses)
 				if (IsAIControlled())
 				{
 					// clear the hate list of the mobs
+					for (auto pMob : this->hate_list.GetHateList()) {
```
Akkadius, line 4014: Rename to just `mob` please

Remove the `this->` explicits please

Akkadius (approved):

PR #1365, opened 2021-05-19
[Spells] Added a pet check to Cazic Touch

RoT-PvP:
Causes the DT to target the pet owner if a pet is tanking. On current PEQ pets can eat DTs for their masters allowing mages to chain pets into DT mobs easily which is incorrect.


splose:
This looks good as of right now but have you thought of making an "IsDeathTouchSpell" function that just returns true if its one of the DT spells (theres at least 3 I know of off the top of my head).

mackal:
Pretty sure the DT spell the client has hardcoded functionality for is only 982.

We do need a define for this, random numbers are bad.

fryguy503:
I know in my LUA Scripts for DT, I have at least two entries listed as "Death Touch"

(982: Cazic touch, 2859: Touch of Vinitras)

zone/spells.cpp
```diff
@@ -2044,6 +2044,14 @@ bool Mob::SpellFinished(uint16 spell_id, Mob *spell_target, CastingSlot slot, ui
 	if(!IsValidSpell(spell_id))
 		return false;
 
+	//Cazic Touch targets the pet owner instead of the pet when said pet is tanking.
+	if (RuleB(Spells,CazicTouchTargetsPetOwner) && spell_id == 982 && spell_target->HasOwner()) {
```
Akkadius, line 2048: Another magic number here, at least convert 982 to a constant that lives in at the very least `eq_constants.h` if there isn't another place that is appropriate

eg

```cpp
static const uint32 DB_SPELL_CAZIC_TOUCH = 982;
```

Akkadius: Also minor nit: spacing between `Spells,Cazic...`

zone/spells.cpp
```diff
@@ -2044,6 +2044,14 @@ bool Mob::SpellFinished(uint16 spell_id, Mob *spell_target, CastingSlot slot, ui
 	if(!IsValidSpell(spell_id))
 		return false;
 
+	//Cazic Touch targets the pet owner instead of the pet when said pet is tanking.
+	if ((RuleB(Spells, CazicTouchTargetsPetOwner) && spell_target->HasOwner()) && spell_id == DB_SPELL_CAZIC_TOUCH || spell_id == DB_SPELL_TOUCH_OF_VINITRAS) {
```
Akkadius, line 2048: Your `or` will not work as expected here. It will run this logic IF the rule is enabled and `DB_SPELL_CAZIC_TOUCH` OR the spell is `DB_SPELL_TOUCH_OF_VINITRAS`. Basically, it won't matter whether the rule is enabled or not as long as the spell passing through is TOV

Your spell checks with the or's need to be wrapped in parenthesis to encapsulate the OR inside of the AND

**Original**

```cpp
if ((RuleB(Spells, CazicTouchTargetsPetOwner) && spell_target->HasOwner()) && spell_id == DB_SPELL_CAZIC_TOUCH || spell_id == DB_SPELL_TOUCH_OF_VINITRAS) {
```

**Suggested**

```cpp
if ((RuleB(Spells, CazicTouchTargetsPetOwner) && spell_target->HasOwner()) && (spell_id == DB_SPELL_CAZIC_TOUCH || spell_id == DB_SPELL_TOUCH_OF_VINITRAS)) {
```

For context as well OR's and AND's get processed left to right

According to the C++ Standard

```
1 The && operator groups left-to-right. The operands are both contextually converted to bool (Clause 4). The result is true if both operands are true and false otherwise. Unlike &, && guarantees left-to-right evaluation: the second operand is not evaluated if the first operand is false.
```

and

```
1 The || operator groups left-to-right. The operands are both contextually converted to bool (Clause 4). It returns true if either of its operands is true, and false otherwise. Unlike |, || guarantees left-to-right evaluation; moreover, the second operand is not evaluated if the first operand evaluates to true.
```

Akkadius:
@RoT-PvP You will also need to merge this branch with master and resolve the conflicts in `./common/ruletypes.h`

I appreciate you rolling with the review cycles, it can feel painful as you're learning all of the little nuances of creating PR's, checking in, out, resolving issues, merging but once you have the workflows down its much easier and we have less review cycles and things ultimately take far less time to make their way through

Thank you <3

common/ruletypes.h
```diff
@@ -370,6 +370,7 @@ RULE_BOOL(Spells, OldRainTargets, false, "Use old incorrectly implemented maximu
 RULE_BOOL(Spells, NPCSpellPush, false, "Enable spell push on NPCs")
 RULE_BOOL(Spells, July242002PetResists, true, "Enable Pets using PCs resist change from July 24 2002")
 RULE_INT(Spells, AOEMaxTargets, 0, "Max number of targets a Targeted AOE spell can cast on. Set to 0 for no limit.")
+RULE_BOOL(Spells, CazicTouchTargetsPetOwner, true, "If True, causes Cazic Touch to swap targets from pet to pet owner if a pet is tanking.")
```
KinglyKrab, line 373: Not sure if this is Live-like or not, commenting for discussion.

If not Live-like, should be `false` by default instead of `true`.

zone/spells.cpp
```diff
@@ -2044,6 +2044,14 @@ bool Mob::SpellFinished(uint16 spell_id, Mob *spell_target, CastingSlot slot, ui
...
+		Mob* owner =  spell_target->GetOwner();
+		
+		if (owner)
```
KinglyKrab, line 2051: Formatting, usually we use brackets even with implicit conditionals like this. 

I know other places in the same code don't, but it allows us to better understand the logic flow.

```cpp
if (owner) {
    spell_target = owner;
}
```

common/ruletypes.h
RoT-PvP: I believe it to be live like, anyone with higher faction would be assisted.

zone/spells.cpp
Akkadius: Still need parenthesis wrapped around

```
&& (spell_id == DB_SPELL_CAZIC_TOUCH || spell_id == DB_SPELL_TOUCH_OF_VINITRAS)
```

mackal:
The constants should match what already exists, like here https://github.com/EQEmu/Server/blob/71e9dd5a3ca4c2d577abdddfb41bde22a2c9fab2/common/spdat.h#L28

splose:
Other than the merge conflicts what else is needed to get this ready to merge?

RoT-PvP:
> Other than the merge conflicts what else is needed to get this ready to merge?

Been super busy with finals week, not sure how to fix the merge conflicts.

Akkadius (approved):

PR #1367, opened 2021-05-22
[PVP] Pvp guard assist code. (Guards will assist in PvP based on faction)

RoT-PvP:
feel free to look this over, will only affect client vs. client and client's pet vs client. Rule must be true to work as well.

### DIFF
--- a/common/races.cpp
+++ b/common/races.cpp
@@ -2232,35 +2232,3 @@ bool PlayerAppearance::IsValidWoad(uint16 race_id, uint8 gender_id, uint8 woad_v
 	}
 	return false;
 }
-
-bool IsGuard(uint16 race_id, uint8 texture_value, uint32 primaryfaction)
-{
-	switch (race_id) {
-		case RT_GUARD:
-			if (texture_value == 1 || texture_value == 2)
-				return true;
-				break;
-		case RT_IKSAR_2:
-			if (texture_value == 1)
-				return true;
-				break;
-		case RT_GUARD_2:
-		case RT_GUARD_3:
-		case RT_GUARD_4:
-		case RT_HUMAN_3:
-		case RT_HALFLING_2:
-		case RT_ERUDITE_2:
-		case RT_BARBARIAN_2:
-		case RT_DARK_ELF_2:
-		case RT_TROLL_2:
-		case OGGOK_CITIZEN:
-		case RT_DWARF_2:
-			return true;
-		default:
-		break;
-	}
-	if (primaryfaction == 255 || primaryfaction == 265 || primaryfaction == 333) { //these 3 factions of guards use player races instead of their own races so we must define them by faction.
-		return true;
-	}
-	return false;
-}
--- a/common/races.cpp
+++ b/common/races.cpp
@@ -2233,7 +2233,7 @@ bool PlayerAppearance::IsValidWoad(uint16 race_id, uint8 gender_id, uint8 woad_v
 	return false;
 }
 
-bool IsGuard(uint16 race_id, uint8 texture_value)
+bool IsGuard(uint16 race_id, uint8 texture_value, uint32 primaryfaction)
 {
 	switch (race_id) {
 		case RT_GUARD:
@@ -2258,6 +2258,9 @@ bool IsGuard(uint16 race_id, uint8 texture_value)
 			return true;
 		default:
 		break;
+	}
+	if (primaryfaction == 255 || primaryfaction == 265 || primaryfaction == 333) { //these 3 factions of guards use player races instead of their own races so we must define them by faction.
+		return true;
 	}
 	return false;
 }
--- a/common/races.cpp
+++ b/common/races.cpp
@@ -2233,7 +2233,7 @@ bool PlayerAppearance::IsValidWoad(uint16 race_id, uint8 gender_id, uint8 woad_v
 	return false;
 }
 
-uint32 IsGuard(uint16 race_id, uint8 texture_value)
+bool IsGuard(uint16 race_id, uint8 texture_value)
 {
 	switch (race_id) {
 		case RT_GUARD:
--- a/common/races.cpp
+++ b/common/races.cpp
@@ -2232,3 +2232,32 @@ bool PlayerAppearance::IsValidWoad(uint16 race_id, uint8 gender_id, uint8 woad_v
 	}
 	return false;
 }
+
+uint32 IsGuard(uint16 race_id, uint8 texture_value)
+{
+	switch (race_id) {
+		case RT_GUARD:
+			if (texture_value == 1 || texture_value == 2)
+				return true;
+				break;
+		case RT_IKSAR_2:
+			if (texture_value == 1)
+				return true;
+				break;
+		case RT_GUARD_2:
+		case RT_GUARD_3:
+		case RT_GUARD_4:
+		case RT_HUMAN_3:
+		case RT_HALFLING_2:
+		case RT_ERUDITE_2:
+		case RT_BARBARIAN_2:
+		case RT_DARK_ELF_2:
+		case RT_TROLL_2:
+		case OGGOK_CITIZEN:
+		case RT_DWARF_2:
+			return true;
+		default:
+		break;
+	}
+	return false;
+}
--- a/common/races.h
+++ b/common/races.h
@@ -860,7 +860,7 @@ uint16 GetRaceIDFromPlayerRaceBit(uint32 player_race_bit);
 
 float GetRaceGenderDefaultHeight(int race, int gender);
 
-bool IsGuard(uint16 race_id, uint8 texture_value);
+bool IsGuard(uint16 race_id, uint8 texture_value, uint32 primaryfaction);
 
 // player race-/gender-based model feature validators
 namespace PlayerAppearance
--- a/common/races.h
+++ b/common/races.h
@@ -860,8 +860,6 @@ uint16 GetRaceIDFromPlayerRaceBit(uint32 player_race_bit);
 
 float GetRaceGenderDefaultHeight(int race, int gender);
 
-bool IsGuard(uint16 race_id, uint8 texture_value, uint32 primaryfaction);
-
 // player race-/gender-based model feature validators
 namespace PlayerAppearance
 {
--- a/common/races.h
+++ b/common/races.h
@@ -860,7 +860,7 @@ uint16 GetRaceIDFromPlayerRaceBit(uint32 player_race_bit);
 
 float GetRaceGenderDefaultHeight(int race, int gender);
 
-uint32 IsGuard(uint16 race_id, uint8 texture_value);
+bool IsGuard(uint16 race_id, uint8 texture_value);
 
 // player race-/gender-based model feature validators
 namespace PlayerAppearance
--- a/common/races.h
+++ b/common/races.h
@@ -860,6 +860,7 @@ uint16 GetRaceIDFromPlayerRaceBit(uint32 player_race_bit);
 
 float GetRaceGenderDefaultHeight(int race, int gender);
 
+uint32 IsGuard(uint16 race_id, uint8 texture_value);
 
 // player race-/gender-based model feature validators
 namespace PlayerAppearance
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -163,6 +163,7 @@ RULE_BOOL(Character, SoftDeletes, true, "When characters are deleted in characte
 RULE_INT(Character, DefaultGuild, 0, "If not 0, new characters placed into the guild # indicated")
 RULE_BOOL(Character, ProcessFearedProximity, false, "Processes proximity checks when feared")
 RULE_BOOL(Character, EnableCharacterEXPMods, false, "Enables character zone-based experience modifiers.")
+RULE_BOOL(Character, PVPEnableGuardFactionAssist, true, "Enables faction based assisting against the aggresor in pvp.")
 RULE_CATEGORY_END()
 
 RULE_CATEGORY(Mercs)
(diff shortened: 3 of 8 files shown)


splose:
This really should be a function. 
bool IsGuard(int npcTypeID);


`if (mob->GetRace() == 106 || mob->GetRace() == 112 || mob->GetRace() == 71 || mob->GetRace() == 67 || mob->GetRace() == 44 && mob->GetTexture() == 2 || mob->GetRace() == 1 && mob->GetTexture() == 3 || mob->GetRace() == 1 || mob->GetRace() == 44 && mob->GetTexture() == 1 || mob->GetRace() == 81 || mob->GetRace() == 78 || mob->GetRace() == 1 || mob->GetRace() == 90 || mob->GetRace() == 77 || mob->GetRace() == 92 || mob->CastToNPC()->GetNPCFactionID() == 382 || mob->GetRace() == 93 || mob->GetRace() == 94 || mob->CastToNPC()->GetNPCFactionID() == 778) {`

splose:
This if statement really should be a function. Something like this.

bool IsGuard(int npcTypeID);


`if (mob->GetRace() == 106 || mob->GetRace() == 112 || mob->GetRace() == 71 || mob->GetRace() == 67 || mob->GetRace() == 44 && mob->GetTexture() == 2 || mob->GetRace() == 1 && mob->GetTexture() == 3 || mob->GetRace() == 1 || mob->GetRace() == 44 && mob->GetTexture() == 1 || mob->GetRace() == 81 || mob->GetRace() == 78 || mob->GetRace() == 1 || mob->GetRace() == 90 || mob->GetRace() == 77 || mob->GetRace() == 92 || mob->CastToNPC()->GetNPCFactionID() == 382 || mob->GetRace() == 93 || mob->GetRace() == 94 || mob->CastToNPC()->GetNPCFactionID() == 778) {`

mackal:
Yeah, that needs to be pulled out. We have a bunch of defines for various races in common/races.h, I would use those and add any missing. I would also recommend writing it as a switch, I think it's more readable than a massive or chain.

common/races.h
```diff
@@ -860,6 +860,7 @@ uint16 GetRaceIDFromPlayerRaceBit(uint32 player_race_bit);
 
 float GetRaceGenderDefaultHeight(int race, int gender);
 
+uint32 IsGuard(uint16 race_id, uint8 texture_value);
```
mackal, line 863: This should be a bool

common/races.cpp
```diff
@@ -2232,3 +2232,32 @@ bool PlayerAppearance::IsValidWoad(uint16 race_id, uint8 gender_id, uint8 woad_v
 	}
 	return false;
 }
+
+uint32 IsGuard(uint16 race_id, uint8 texture_value)
```
mackal, line 2236: Bool here too

splose:
Looking much better.. Just gotta switch them to bools. You could also add the faction checks to the IsGuard function that way you literally can just do this (but not a huge deal either way):

```
if(IsGuard(npc_id)) {

}
```

splose:
Ah.. One thing I missed was make sure the rule is false by default.. But I'll approve.

Great job overall, I know you're pretty new at this so you are improving. :)

splose (approved):

RoT-PvP:
> Ah.. One thing I missed was make sure the rule is false by default.. But I'll approve.
> 
> Great job overall, I know you're pretty new at this so you are improving. :)

default should be set, maybe needs another return false; ? I had Mackal look over it yesterday. Thanks for the kind words.

common/ruletypes.h
```diff
@@ -163,6 +163,7 @@ RULE_BOOL(Character, SoftDeletes, true, "When characters are deleted in characte
 RULE_INT(Character, DefaultGuild, 0, "If not 0, new characters placed into the guild # indicated")
 RULE_BOOL(Character, ProcessFearedProximity, false, "Processes proximity checks when feared")
 RULE_BOOL(Character, EnableCharacterEXPMods, false, "Enables character zone-based experience modifiers.")
+RULE_BOOL(Character, PVPEnableGuardFactionAssist, true, "Enables faction based assisting against the aggresor in pvp.")
```
mackal, line 166: This should be default false is what splose is saying. I'm not sure actual behavior here.

RoT-PvP: Project 1999 has guards that assist based on faction, I believe SZ and RZ both had this mechanic on live as well.

Akkadius:
If this is not default live behavior then the rule should be false by default. Overall the code LGTM

zone/attack.cpp
```diff
@@ -1522,6 +1522,27 @@ bool Client::Attack(Mob* other, int Hand, bool bRiposte, bool IsStrikethrough, b
...
 	other->AddToHateList(this, hate);
 
+	//Guard Assist Code
+	if (RuleB(Character, PVPEnableGuardFactionAssist) && this->IsClient() || RuleB(Character, PVPEnableGuardFactionAssist) && this->HasOwner() && this->GetOwner()->IsClient()) {
+		auto& mob_list = entity_list.GetCloseMobList(other);
+		for (auto& e : mob_list) {
+			auto mob = e.second;
+			float distance = Distance(other->CastToClient()->m_Position, mob->GetPosition());
+			if (mob->CheckLosFN(other) && distance <= 70 || mob->CheckLosFN(this) && distance <= 70) {
+				if (mob->CastToNPC()->GetNPCFactionID() == 382 || mob->CastToNPC()->GetNPCFactionID() == 778 || IsGuard(mob->GetRace(),mob->GetTexture())) {
```
Akkadius, line 1532: What are these faction ID's supposed to be ? These should at least tie back to a constant of some kind and pulling these values out of the database to remove magic numbers from the logic

Something like

```cpp
// database factions
const int16 DB_FACTION_WATCHMEN       = 382;
const int16 DB_FACTION_PAINEEL_GUARDS = 778;
```

And at least put it in somewhere like `factions.h` for now

Akkadius: Also if these factions are guards then it could fall underneath your `IsGuard` method and factions can just be an additional check there. 

You don't need to pass parameters because it's already a mob method and you have access to mob-scoped methods and there's not a huge benefit to explicitly passing the parameters

zone/spells.cpp
```diff
@@ -2044,6 +2044,27 @@ bool Mob::SpellFinished(uint16 spell_id, Mob *spell_target, CastingSlot slot, ui
 	if(!IsValidSpell(spell_id))
 		return false;
 
+	//Guard Assist Code
+	if (RuleB(Character, PVPEnableGuardFactionAssist) && IsDetrimentalSpell(spell_id) && this->IsClient() && spell_target != this || RuleB(Character, PVPEnableGuardFactionAssist) && IsDetrimentalSpell(spell_id) && spell_target != this && this->HasOwner() && this->GetOwner()->IsClient()) {
```
Akkadius, line 2048: This also needs some simplification 

* `RuleB(Character, PVPEnableGuardFactionAssist)` is getting checked twice in an OR - Just make this the first check before trying to check anything else. It will reduce unnecessary other checks especially when OR's are included. This isn't a super hot path but we should avoid it
* `IsDetrimentalSpell` gets processed in both OR's (same with `spell_target != this`) so maybe it makes sense to have the Rule check, the detrimental check and the spell_target check be the first if block since regardless of your OR operand, you need to check for both of these things
* Using `this->` is not needed as scope is implied

Example

```cpp
	// Original
	if (RuleB(Character, PVPEnableGuardFactionAssist) && IsDetrimentalSpell(spell_id) && this->IsClient() &&
		spell_target != this ||
		RuleB(Character, PVPEnableGuardFactionAssist) && IsDetrimentalSpell(spell_id) && spell_target != this &&
		this->HasOwner() && this->GetOwner()->IsClient()) {
		// ... loop logic
	}

	// Simplified
	if (RuleB(Character, PVPEnableGuardFactionAssist) && IsDetrimentalSpell(spell_id) && spell_target != this) {
		if (IsClient() || (HasOwner() && GetOwner()->IsClient())) {
			// ... loop logic
		}
	}
```

zone/attack.cpp
```diff
@@ -1522,6 +1522,27 @@ bool Client::Attack(Mob* other, int Hand, bool bRiposte, bool IsStrikethrough, b
 
 	other->AddToHateList(this, hate);
 
+	//Guard Assist Code
+	if (RuleB(Character, PVPEnableGuardFactionAssist) && this->IsClient() || RuleB(Character, PVPEnableGuardFactionAssist) && this->HasOwner() && this->GetOwner()->IsClient()) {
```
Akkadius, line 1526: Similar feedback here as before with the OR block

common/races.h
```diff
@@ -860,6 +860,7 @@ uint16 GetRaceIDFromPlayerRaceBit(uint32 player_race_bit);
 
 float GetRaceGenderDefaultHeight(int race, int gender);
 
+bool IsGuard(uint16 race_id, uint8 texture_value, uint32 primaryfaction);
```
Akkadius, line 863: `IsGuard` is a great function, but it should live in the `NPC` class and should not need any parameters passed to it because everything you need will be scoped within NPC

Your function becomes this (We also need constants in place of the magic numbers for factions)

```cpp
bool NPC::IsGuard()
{
	switch (GetRace()) {
		case RT_GUARD:
			if (GetTexture() == 1 || GetTexture() == 2)
				return true;
			break;
		case RT_IKSAR_2:
			if (GetTexture() == 1)
				return true;
			break;
		case RT_GUARD_2:
		case RT_GUARD_3:
		case RT_GUARD_4:
		case RT_HUMAN_3:
		case RT_HALFLING_2:
		case RT_ERUDITE_2:
		case RT_BARBARIAN_2:
		case RT_DARK_ELF_2:
		case RT_TROLL_2:
		case OGGOK_CITIZEN:
		case RT_DWARF_2:
			return true;
		default:
			break;
	}
	if (GetPrimaryFaction() == 255 || GetPrimaryFaction() == 265 || GetPrimaryFaction() == 333) { //these 3 factions of guards use player races instead of their own races so we must define them by faction.
		return true;
	}
	return false;
}
```

Your utilization of the function would simply end up being `IsGuard()` instead of passing parameters

Because the function becomes NPC scoped, you will either need to cast Mob to NPC (NPC inherits Mob, similarly to Client inheriting Mob) or you will need to explicitly loop through NPC's instead of Mobs and you will already have an NPC type without having to cast

You also want to put the least cost checks first especially in an operation where you're potentially looping many entities. It's much cheaper to check if an NPC is a guard or an NPC at all and not process distance checks at all than it is to run a bunch of distance checks regardless of whether you care about the NPC or not and then run your attribute checking

Example

```cpp
	for (auto &e : entity_list.GetCloseMobList(spell_target)) {
		auto mob = e.second;

		// check for NPC and also cast to NPC after successful validation that it's an NPC
		if (mob->IsNPC() && mob->CastToNPC()->IsGuard()) {

			float distance = Distance(spell_target->CastToClient()->m_Position, mob->GetPosition());
			if (mob->CheckLosFN(spell_target) && distance <= 70 || mob->CheckLosFN(this) && distance <= 70) {
				// ...
			}

		}
	}
```

common/ruletypes.h
KinglyKrab: I would have to agree `false` is better here.

zone/spells.cpp
```diff
@@ -2044,6 +2044,25 @@ bool Mob::SpellFinished(uint16 spell_id, Mob *spell_target, CastingSlot slot, ui
...
+	if (RuleB(Character, PVPEnableGuardFactionAssist) && IsDetrimentalSpell(spell_id) && spell_target != this) {
+		if (IsClient() || HasOwner() && GetOwner()->IsClient()) {
+			auto& mob_list = entity_list.GetCloseMobList(spell_target);
+			for (auto& e : mob_list) {
+				auto mob = e.second;
+				if (mob->IsNPC() && mob->CastToNPC()->IsGuard()) {
+					float distance = Distance(spell_target->GetPosition(), mob->GetPosition());
+					if ((mob->CheckLosFN(spell_target) || mob->CheckLosFN(this)) && distance <= 70) {
+						auto petorowner = this->GetOwnerOrSelf();
+							if (spell_target->GetReverseFactionCon(mob) <= petorowner->GetReverseFactionCon(mob)) {
```
KinglyKrab, line 2057: Formatting, looks like an extra tab or something on these few lines.

zone/npc.cpp
```diff
@@ -3368,3 +3368,35 @@ void NPC::ScaleNPC(uint8 npc_level) {
...
+	case RT_BARBARIAN_2:
+	case RT_DARK_ELF_2:
+	case RT_TROLL_2:
+	case OGGOK_CITIZEN:
+	case RT_DWARF_2:
+		return true;
+	default:
+		break;
+	}
+	if (GetPrimaryFaction() == 255 || GetPrimaryFaction() == 265 || GetPrimaryFaction() == 333) { //these 3 factions of guards use player races instead of their own races so we must define them by faction.
```
KinglyKrab, line 3398: Maybe considering making these less magic similar to the races above?

zone/npc.cpp
```diff
@@ -3368,3 +3368,35 @@ void NPC::ScaleNPC(uint8 npc_level) {
...
+{
+	switch (GetRace()) {
+	case RT_GUARD:
+		if (GetTexture() == 1 || GetTexture() == 2)
+			return true;
+		break;
+	case RT_IKSAR_2:
+		if (GetTexture() == 1)
+			return true;
+		break;
```
KinglyKrab, line 3382: Can remove this break.

zone/npc.cpp
```diff
@@ -3368,3 +3368,35 @@ void NPC::ScaleNPC(uint8 npc_level) {
...
 	npc_scale_manager->ScaleNPC(this);
 }
+
+bool NPC::IsGuard()
+{
+	switch (GetRace()) {
+	case RT_GUARD:
+		if (GetTexture() == 1 || GetTexture() == 2)
+			return true;
+		break;
```
KinglyKrab, line 3378: Can remove this break.

common/ruletypes.h
RoT-PvP: Live like from RZ suggests that this is proper default function

Akkadius (approved):

PR #1368, opened 2021-05-23
[Shared Bank] Shared Bank Disabled Warning Popup

KinglyKrab:
Add additional popup to shared bank warning message, as client-side filters can cause the message to be unseen.
image


RoT-PvP:
Is there anyway to make it so items cannot be placed within the shared bank as well?

splose:
This was definitely needed as you and I had issues on RoN with people putting items in their shared bank even after the chat spam.

splose (approved):

mackal:
We should honestly just have shared banks on by default and not allow extra toons per account by default. Kind of silly to do shit this way. This PR is fine for those who do want it disabled though.

mackal (approved):

Akkadius (approved):

PR #1369, opened 2021-05-24
[Fix] Fix Spell Cast Time reduction issues

mackal:
Remove the overloads that don't make sense (bots probably doesn't make
sense either, but too lazy)

Fix the formulas

Removed the Spells:MaxCastTimeReduction rule since this is HARDCODED in
the client so it doesn't really make sense to have it as a customization
point. If you want to hack the client, change the hardcode as well I
guess.


Akkadius:
Curious, what is this fixing and definitely seems like a touchy change although looks straightforward

Was this a result of something mentioned in Discord lately? Otherwise, :seemsgood:

mackal:
I have no idea where the original code came from. Poked a few people, seems made up /shrug. There on CVS recreation too it seems. I don't think it's real.

Akkadius (approved):

PR #1370, opened 2021-05-28
strict spawn_events now take into account EQ minute.

regneq:
This should fixed the eqtime spawn condition from falling behind.


regneq:
When strict is set in spawn_events, specifies the max EQ minutes into the trigger hour a spawn_event will fire. Going below 3 may cause the spawn_event to not fire.

KinglyKrab (approved):

mackal (approved):

hgtw (approved):

Akkadius (approved):

PR #1371, opened 2021-06-02
Move member compass updates to dz

hgtw:
This method can be moved now that members are on dz


mackal (approved):

Akkadius (approved):
Fairly straightforward

Issue #904, opened 2019-09-26
Invis Undead and Invis Animals do not break charm


Isaaru:

splose:
Fixed pending merge of #1374

PR #1374, opened 2021-06-03
[Bug Fix] Fixed Invis vs Undead / Invis Vs Animals not breaking charm + Added rule for custom capabilities

splose:
This directly addresses https://github.com/EQEmu/Server/issues/904

Apparently the Livelike functionality is ANY invis breaks charm  -- even if you have say a regular mob charmed and you cast IVU. Setting the rule to false will only break charm if your pet cannot see you. (ex. You have a skeleton charmed and cast IVU). Rule description is very clear.

At the current time after the charm breaks the pet starts attacking you which I am told is Livelike behavior.

### DIFF
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -554,6 +554,7 @@ void Mob::SetInvisible(uint8 state)
 		if (RuleB(Pets, LivelikeBreakCharmOnInvis) || (!RuleB(Pets, LivelikeBreakCharmOnInvis) && IsInvisible(formerpet))) {
 			formerpet->BuffFadeByEffect(SE_Charm);
 		}
+
 		LogRules("Pets:LivelikeBreakCharmOnInvis for [{}] | Invis [{}] - Hidden [{}] - Shroud of Stealth [{}] - IVA [{}] - IVU [{}]", GetCleanName(), invisible, hidden, improved_hidden, invisible_animals, invisible_undead);
 	}
 }
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -551,12 +551,8 @@ void Mob::SetInvisible(uint8 state)
 	// Invis and hide breaks charms
 	auto formerpet = GetPet();
 	if (formerpet && formerpet->GetPetType() == petCharmed && (invisible || hidden || improved_hidden || invisible_animals || invisible_undead)) {
-		if (RuleB(Pets, LivelikeBreakCharmOnInvis)) {	
+		if (RuleB(Pets, LivelikeBreakCharmOnInvis) || (!RuleB(Pets, LivelikeBreakCharmOnInvis) && IsInvisible(formerpet))) {
 			formerpet->BuffFadeByEffect(SE_Charm);
-		}
-		else {	
-			if (IsInvisible(formerpet))
-				formerpet->BuffFadeByEffect(SE_Charm);
 		}
 		LogRules("Pets:LivelikeBreakCharmOnInvis for [{}] | Invis [{}] - Hidden [{}] - Shroud of Stealth [{}] - IVA [{}] - IVU [{}]", GetCleanName(), invisible, hidden, improved_hidden, invisible_animals, invisible_undead);
 	}
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -551,7 +551,7 @@ void Mob::SetInvisible(uint8 state)
 	// Invis and hide breaks charms
 	auto formerpet = GetPet();
 	if (formerpet && formerpet->GetPetType() == petCharmed && (invisible || hidden || improved_hidden || invisible_animals || invisible_undead)) {
-		if (RuleB(Pets, LivelikeBreakCharmOnInvis) || (!RuleB(Pets, LivelikeBreakCharmOnInvis) && IsInvisible(formerpet))) {
+		if (RuleB(Pets, LivelikeBreakCharmOnInvis) || IsInvisible(formerpet))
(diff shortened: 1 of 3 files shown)


zone/mob.cpp
```diff
@@ -549,10 +549,17 @@ void Mob::SetInvisible(uint8 state)
 	invisible = state;
 	SendAppearancePacket(AT_Invis, invisible);
 	// Invis and hide breaks charms
-
 	auto formerpet = GetPet();
-	if (formerpet && formerpet->GetPetType() == petCharmed && (invisible || hidden || improved_hidden))
-		formerpet->BuffFadeByEffect(SE_Charm);
+	if (formerpet && formerpet->GetPetType() == petCharmed && (invisible || hidden || improved_hidden || invisible_animals || invisible_undead)) {
+		if (RuleB(Pets, LivelikeBreakCharmOnInvis)) {	//Pets:LivelikeBreakCharmOnInvis = true
```
mackal, line 554: Don't really need this comment here.

zone/mob.cpp
```diff
@@ -549,10 +549,17 @@ void Mob::SetInvisible(uint8 state)
...
+			formerpet->BuffFadeByEffect(SE_Charm);
+		}
+		else {	//Pets:LivelikeBreakCharmOnInvis = false
```
mackal, line 557: Or here.

mackal (changes requested):

zone/mob.cpp
```diff
@@ -549,10 +549,17 @@ void Mob::SetInvisible(uint8 state)
...
+			if(IsInvisible(formerpet))
+				formerpet->BuffFadeByEffect(SE_Charm);
+		}
+		LogRules(fmt::format("Pets:LivelikeBreakCharmOnInvis for [{}] | Invis [{}] - Hidden [{}] - Shroud of Stealth [{}] - IVA [{}] - IVU [{}]", GetCleanName(), invisible, hidden, improved_hidden, invisible_animals, invisible_undead).c_str());
```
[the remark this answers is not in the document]
Akkadius: `fmt::format` can be entirely removed; it's done under the hood in all of the Logging macros to make things simple

zone/mob.cpp
Akkadius: Agree with @hgtw the `IsInvisible` function should simplify and cover all of the OR's in this line

Akkadius (changes requested):
Minor tweak on simplification and we should be good

zone/mob.cpp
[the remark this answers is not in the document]
splose: Oh nice didn't know that. 

I can make this and the other change in a few days.. moving into a new place and don't have my PC with me atm 😞

splose:
The code that @hgtw mentioned simplifying would result in non livelike functionality.

The thing that is 'livelike' is that if you cast any sort of invis the charmed pet breaks even if you are not invisible to it (You have an undead pet and cast regular invis -- livelike it should drop the pet.. The rule I added would keep your pet)

The point of the comments were to make it a bit more intuitive when looking at it but I removed them regardless. I don't mind removing the rule and just going with what was said in hg's original comment as it makes much more sense but I know some people want to emulate Live, including it's bugs/oversights/mistakes.

```
	if (formerpet && formerpet->GetPetType() == petCharmed && (invisible || hidden || improved_hidden || invisible_animals || invisible_undead)) {
		if (RuleB(Pets, LivelikeBreakCharmOnInvis)) {	
			formerpet->BuffFadeByEffect(SE_Charm);
		}
		else {	
			if(IsInvisible(formerpet))
				formerpet->BuffFadeByEffect(SE_Charm);
		}
		LogRules("Pets:LivelikeBreakCharmOnInvis for [{}] | Invis [{}] - Hidden [{}] - Shroud of Stealth [{}] - IVA [{}] - IVU [{}]", GetCleanName(), invisible, hidden, improved_hidden, invisible_animals, invisible_undead);
	}
```

splose:
Was having trouble getting it to show up but I changed the logic a bit.

zone/mob.cpp
```diff
@@ -549,10 +549,14 @@ void Mob::SetInvisible(uint8 state)
 	invisible = state;
 	SendAppearancePacket(AT_Invis, invisible);
 	// Invis and hide breaks charms
-
 	auto formerpet = GetPet();
-	if (formerpet && formerpet->GetPetType() == petCharmed && (invisible || hidden || improved_hidden))
-		formerpet->BuffFadeByEffect(SE_Charm);
+	if (formerpet && formerpet->GetPetType() == petCharmed && (invisible || hidden || improved_hidden || invisible_animals || invisible_undead)) {
+		if (RuleB(Pets, LivelikeBreakCharmOnInvis) || (!RuleB(Pets, LivelikeBreakCharmOnInvis) && IsInvisible(formerpet))) {
```
hgtw, line 554: Just combining the inner if/else blocks like this is what I originally meant (I didn't mean replacing all the invisible checks in the outer if, I realize that would have been a functionality change regarding the rule)

This should now be equivalent to
`if (RuleB(Pets, LivelikeBreakCharmOnInvis) || IsInvisible(formerpet))`
(the second check for the rule being false after the `||` would be implied, anything after the `||` would only be evaluated if the rule was false)

hgtw (approved):

zone/mob.cpp
splose: Done

Akkadius (approved):

PR #1375, opened 2021-06-05
[Bots] Remove hardcoded race-class combinations from bots.

KinglyKrab:
- Allows server operators to directly influence via a database table the classes a specific bot race can be.
- Previously this was hardcoded and required a source modification to do, this will allow those who use the installer or cannot modify their own source to have a higher level of customization with this system.
- Allowed races, classes, and genders have been removed due to redundancy at this point.
- Removed bot-specific race and class name functions and used the ones we use everywhere else.
- Adds a ^viewcombos [race_id] command for bots to view proper race-class combinations.
- Saylink for viewing proper combinations is sent on invalid combination.


mackal (approved):

zone/bot_command.cpp
```diff
@@ -5107,6 +5108,68 @@ void bot_subcommand_bot_clone(Client *c, const Seperator *sep)
...
+	if (helper_command_alias_fail(c, "bot_command_view_combos", sep->arg[0], "viewcombos"))
+		return;
+	if (helper_is_help_or_usage(sep->arg[1])) {
+		std::string window_title = "Bot Races";
+		std::string window_text;
+		std::string message_separator = " ";
+		c->Message(m_usage, "Usage: %s [bot_race]", sep->arg[0]);
+		window_text.append("<c \"#FFFFFF\">Races:<c \"#FFFF\">");
+		for (int race_id = 0; race_id <= 15; ++race_id) {
+			window_text.append(const_cast<const std::string&>(message_separator));
```
hgtw, line 5142: Should be able to just append this string without the const casting here

zone/bot_command.cpp
```diff
@@ -8672,33 +8725,25 @@ uint32 helper_bot_create(Client *bot_owner, std::string bot_name, uint8 bot_clas
...
-		bot_owner->Message(m_fail, "Race '%s' bots are not allowed on this server", GetPlayerRaceName(bot_class));
-		return bot_id;
-	}
-
 	if (!Bot::IsValidRaceClassCombo(bot_race, bot_class)) {
-		bot_owner->Message(m_fail, "'%s'(%u):'%s'(%u) is an invalid race-class combination",
-			Bot::RaceIdToString(bot_race).c_str(), bot_race, Bot::ClassIdToString(bot_class).c_str(), bot_class);
+		const char* bot_race_name = GetRaceIDName(bot_race);
+		const char* bot_class_name = GetClassIDName(bot_class);
+		const char* view_saylink = EQ::SayLinkEngine::GenerateQuestSaylink(fmt::format("^viewcombos {}", bot_race), false, "view").c_str();
```
hgtw, line 8731: This isn't safe because the std::string returned from `GenerateQuestSaylink` will destruct at the end of the statement, so `view_saylink` will become dangling. Would recommend just changing `view_saylink` to `std::string` here

zone/bot_command.cpp
KinglyKrab: Copy and pasted from the code already there.

zone/bot_command.cpp
KinglyKrab: That's fine. I'll make it a string and grab the .c_str() inside the fmt::format instead. Wasn't sure which was better practice. Makes sense.

hgtw (approved):

Akkadius (approved):

PR #1377, opened 2021-06-06
[Spells] Adds a rule to allow right-click memorize from spell scrolls.

KinglyKrab:
Uses similar logic to disciplines with the hardcoded `Tome:` check and stuff, but with `Spell:` items here instead. Locked it behind a rule in case people don't want to use it.

### DIFF
--- a/zone/effects.cpp
+++ b/zone/effects.cpp
@@ -509,6 +509,87 @@ bool Client::TrainDiscipline(uint32 itemid) {
 	return(false);
 }
 
+bool Client::MemorizeSpellFromItem(uint32 item_id) {
+	const EQ::ItemData *item = database.GetItem(item_id);
+	if(item == nullptr) {
+		Message(Chat::Red, "Unable to find the scroll!");
+		LogError("Unable to find scroll id [{}]\n", (unsigned long)item_id);
+		return false;
+	}
+
+	if (!item->IsClassCommon() || item->ItemType != EQ::item::ItemTypeSpell) {
+		Message(Chat::Red, "Invalid item type, you cannot learn from this item.");
+		SummonItem(item_id);
+		return false;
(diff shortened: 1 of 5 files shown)


splose:
Lit. Rule needs to be changed to true by default as this is live functionality.

splose (approved):

mackal (approved):

hgtw (approved):

zone/effects.cpp
```diff
@@ -509,6 +509,87 @@ bool Client::TrainDiscipline(uint32 itemid) {
...
+		return false;
+	}
+
+	if (!item->IsClassCommon() || item->ItemType != EQ::item::ItemTypeSpell) {
+		Message(Chat::Red, "Invalid item type, you cannot learn from this item.");
+		SummonItem(item_id);
+		return false;
+	}
+
+	if(!(
```
Akkadius, line 526: Not necessarily requesting changes here for this PR but converting the char array to a std::string would make it easy to do a string find and simplify the code a bit

Example

```cpp
		std::string item_name = item->Name;
		if (item_name.find("Spell: ") != std::string::npos) {
			std::cout << "found!" << '\n';
		}
```

zone/effects.cpp
```diff
@@ -509,6 +509,87 @@ bool Client::TrainDiscipline(uint32 itemid) {
 	return(false);
 }
 
+bool Client::MemorizeSpellFromItem(uint32 item_id) {
+	const EQ::ItemData *item = database.GetItem(item_id);
```
Akkadius, line 513: Could have used auto here; but the explicitness is good too

Akkadius (approved):
This change is really cool and something I'd be excited for running a custom server

PR #1378, opened 2021-06-06
[Bots] Add Bot scripting capabilities to the source.

KinglyKrab:
- This will allow server operators to interact with bots within a script in Perl or Lua.

### DIFF
--- a/zone/embparser.cpp
+++ b/zone/embparser.cpp
@@ -977,6 +977,14 @@ void PerlembParser::MapFunctions()
 		"package Expedition;"
 		"&boot_Expedition;"
 
+#ifdef BOTS
+		"package Bot;"
+		"our @ISA = qw(NPC);" // Bot inherits NPC
+		"&boot_Mob;" // load our Mob XS
+		"&boot_NPC;" // load our NPC XS
+		"&boot_Bot;" // load our Bot XS
+#endif
+
 		#endif
 		"package main;"
 		"}"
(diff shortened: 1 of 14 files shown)


splose (approved):

mackal:
What benefit does this actually give us? They should be able to be interacted with via just normal NPC scripting, yes?

KinglyKrab:
This will allow us to add bot specific methods to the quest API. This serves as a primary push for the base functionality. Technically you could do it with NPC stuff but that isn't as accurate as being able to specifically grab a bot from the entity list or only the bots on an entity list without a bunch of extra code.

This won't effect many servers at all considering it required bots to be enabled for most of the functionality to be available.

I plan to do a lot of work in regards to bots and this will make that a lot easier. Maybe we could even add a mercenary export for mercenary specific functionality in the future.

mackal:
It's just there wasn't anything added really for the bot quest API yet, so was just wonder what those would be.

KinglyKrab:
Well unlike NPCs, bots have a real inventory structure and stuff like a client. So they needed to take that in to account. The empty Lua and Perl bot stuff will serve as a placeholder so that the functionality doesn't have to be implemented in the future for Bot stuff to make its way in.

splose:
I don't personally use bots and won't after this change.. But I can see the vision and tons of functionality that can come from this hence why I approved initially.

Assuming that people actually use it this could be pretty big for servers with Bots.

hgtw (approved):

Akkadius (approved):
Nice work on the compile guards; tested compiling with a few  different switches on and off

PR #1379, opened 2021-06-06
[Pets] Unhardcode Beastlord pet values.

KinglyKrab:
- Create a Beastlord pets table to allow server operators to easily customize Beastlord pets without a source modification.
- Defaults to similar values as the original switch inside the struct and database table.

### DIFF
--- a/common/database_schema.h
+++ b/common/database_schema.h
@@ -217,6 +217,7 @@ namespace DatabaseSchema {
 			"npc_types_tint",
 			"object",
 			"pets",
+			"pets_beastlord_data",
 			"pets_equipmentset",
 			"pets_equipmentset_entries",
 			"proximities",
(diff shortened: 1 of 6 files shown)


splose (approved):

mackal:
Seems reasonable, but I want @Akkadius to review it

hgtw (commented):
The `pets_beastlord_data` table will need added to `common/database_schema.h` content tables list also

hgtw (approved):

Akkadius (approved):
This is a very nice quality of life change for many server operators who've made tweaks for this in the past :buff: 

KingMort has done this, Dhaymion from Lazarus has done this as well as others

It will be nice to have this driven from the source

I like the DTO pattern around using a struct to hydrate the database data; only thing I would suggest is generating a repository for this new table to do the same thing in the future and you don't have to manually wire any of this up next time

We have a generator script for this if you're interested in diving in, it just generates directly from your database table schema data https://eqemu.gitbook.io/server/in-development/developer-area/repositories

We have a lot of opportunity in our code to simplify bloat by converting over to using repositories

PR #1381, opened 2021-06-08
[Lua/Feature] Add Lua_Mob::GetShuffledHateList

mackal:
This returns the hate list but in a random order. This is useful to
prevent repeated (potentially infinite ...) calls to GetHateRandom()

### DIFF
--- a/zone/lua_mob.cpp
+++ b/zone/lua_mob.cpp
@@ -967,6 +967,23 @@ Lua_HateList Lua_Mob::GetHateList() {
 	return ret;
 }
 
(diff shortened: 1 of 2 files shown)


splose:
Should we do this for Perl as well?

I'm guessing the same issue can occur there?

splose (approved):

mackal:
I don't think this is directly translatable to perl.

hgtw (approved):

Akkadius (approved):

PR #1382, opened 2021-06-11
[Commands] Add #viewzoneloot [item id] command.

KinglyKrab:
- Allows GMs to search for a specific item across all the loot currently available on the spawned NPCs in the zone.
- Specifying item ID 0 will allow GMs to see all the droppable items, I tested in Sanctus Seru (a huge zone) and it sent approximately 1,200 messages, which didn't lag or desync my client.

### DIFF
--- a/zone/command.cpp
+++ b/zone/command.cpp
@@ -14110,21 +14110,20 @@ void command_network(Client *c, const Seperator *sep)
 void command_viewzoneloot(Client *c, const Seperator *sep)
 {
 	std::map<uint32,ItemList> zone_loot_list;
-	std::list<NPC*> npc_list;
+	auto npc_list = entity_list.GetNPCList();
 	uint32 loot_amount = 0, loot_id = 1, search_item_id = 0;
 	if (sep->argnum == 1 && sep->IsNumber(1)) {
 		search_item_id = atoi(sep->arg[1]);
-	} else {
+	} else if (sep->argnum == 1 && !sep->IsNumber(1)) {
 		c->Message(
 			Chat::Yellow,
 			"Usage: #viewzoneloot [item id]"
 		);
 		return;
 	}
-	entity_list.GetNPCList(npc_list);
 	for (auto npc_entity : npc_list) {
-		auto current_npc_item_list = npc_entity->GetItemList();
-		zone_loot_list.insert({ npc_entity->GetID(), current_npc_item_list });
+		auto current_npc_item_list = npc_entity.second->GetItemList();
+		zone_loot_list.insert({ npc_entity.second->GetID(), current_npc_item_list });
 	}
 	for (auto loot_item : zone_loot_list) {
 		uint32 current_entity_id = loot_item.first;
--- a/zone/command.cpp
+++ b/zone/command.cpp
@@ -449,6 +449,7 @@ int command_init(void)
 		command_add("zonebootup", "[ZoneServerID] [shortname] - Make a zone server boot a specific zone", 150, command_zonebootup) ||
 		command_add("zoneinstance", "[instanceid] [x] [y] [z] - Go to specified instance zone (coords optional)", 50, command_zone_instance) ||
 		command_add("zonelock", "[list/lock/unlock] - Set/query lock flag for zoneservers", 100, command_zonelock) ||
+		command_add("viewzoneloot", "[item id] - Allow's you to search a zone's loot for a specific item ID. (0 shows all loot in the zone)", 80, command_viewzoneloot) ||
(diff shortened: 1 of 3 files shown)


zone/command.cpp
```diff
@@ -14106,6 +14107,135 @@ void command_network(Client *c, const Seperator *sep)
...
+	if (sep->argnum == 1 && sep->IsNumber(1)) {
+		search_item_id = atoi(sep->arg[1]);
+	} else {
+		c->Message(
+			Chat::Yellow,
+			"Usage: #viewzoneloot [item id]"
+		);
+		return;
+	}
+	entity_list.GetNPCList(npc_list);
```
mackal, line 14124: https://github.com/EQEmu/Server/blob/f0d0c837108e6925103b268e33d4792ecdc9fe23/zone/entity.h#L514

Use that instead.

zone/command.cpp
```diff
@@ -14106,6 +14107,135 @@ void command_network(Client *c, const Seperator *sep)
...
 }
 
+void command_viewzoneloot(Client *c, const Seperator *sep)
+{
+	std::map<uint32,ItemList> zone_loot_list;
+	std::list<NPC*> npc_list;
+	uint32 loot_amount = 0, loot_id = 1, search_item_id = 0;
+	if (sep->argnum == 1 && sep->IsNumber(1)) {
+		search_item_id = atoi(sep->arg[1]);
+	} else {
```
mackal, line 14117: no arg should probably act the same as passing in 0

zone/npc.h
```diff
@@ -198,6 +198,7 @@ class NPC : public Mob
 	void	RemoveItem(uint32 item_id, uint16 quantity = 0, uint16 slot = 0);
 	void	CheckTrivialMinMaxLevelDrop(Mob *killer);
 	void	ClearItemList();
+	inline ItemList GetItemList() { return itemlist; }
```
mackal, line 201: This results in a lot of copying, probably can return a const reference?

Akkadius (approved):

zone/command.cpp
```diff
@@ -14106,6 +14107,134 @@ void command_network(Client *c, const Seperator *sep)
...
+					"#zone {} {} {} {}",
+					zone_id,
+					npc_x,
+					npc_y,
+					npc_z
+				),
+				false,
+				"Goto"
+			);
+			if (instance_id != 0) {
```
Akkadius, line 14155: The only thing different between this `command_link` block and the one before it is whether or not you're in an instance. Could be combined into just one declaration by ternary operating on the `instance_id` for both the command and the second param

eg

```cpp
std::string command_link = EQ::SayLinkEngine::GenerateQuestSaylink(
	fmt::format(
		"#{} {} {} {} {}",
		(instance_id != 0 ? "zoneinstance" : "zone"),
		(instance_id != 0 ? instance_id : zone_id),
		current_npc->GetX(),
		current_npc->GetY(),
		current_npc->GetZ()
	),
	false,
	"Goto"
);
```

Then, because you're just using one declaration, the `float` `npc_x`, `npc_y` etc no longer becomes necessary because you can just use `current_npc->GetX/Y/Z` etc

zone/command.cpp
```diff
@@ -14106,6 +14107,134 @@ void command_network(Client *c, const Seperator *sep)
...
+		return;
+	}
+	for (auto npc_entity : npc_list) {
+		auto current_npc_item_list = npc_entity.second->GetItemList();
+		zone_loot_list.insert({ npc_entity.second->GetID(), current_npc_item_list });
+	}
+	for (auto loot_item : zone_loot_list) {
+		uint32 current_entity_id = loot_item.first;
+		auto current_item_list = loot_item.second;
+		auto current_npc = (
```
Akkadius, line 14131: This can just be `auto current_npc = entity_list.GetNPCByID(current_entity_id);` since it returns `nullptr` on its own when it fails to find the NPC

You're making two calls to an entity list iteration because you have to check for it once and then fetch it once its valid versus just letting `GetNPCByID` return `nullptr`

Akkadius:
Neat function

Left minor comments

Also I'm not sure if aware but there is a new-ish `Loot` log category that will show what ultimately gets rolled / dropped when NPC's spawn which is very helpful

Obviously this is separate from what your command achieves but worth throwing out there

KinglyKrab:
> 
> 
> Neat function
> 
> Left minor comments
> 
> Also I'm not sure if aware but there is a new-ish `Loot` log category that will show what ultimately gets rolled / dropped when NPC's spawn which is very helpful
> 
> Obviously this is separate from what your command achieves but worth throwing out there

Yeah, I'm aware of the log type, someone had specifically requested a command to view the current loot available in the zone.

All comments should be addressed and ready for merge.

PR #1383, opened 2021-06-11
[Quest API] Add getcleannpcnamebyid(npc_id) to Perl/Lua.

KinglyKrab:
- Allows Server Operators to grab the clean name without having to clean it up in their Perl/Lua.

### DIFF
--- a/common/database.cpp
+++ b/common/database.cpp
@@ -888,7 +888,7 @@ std::string Database::GetCharNameByID(uint32 char_id) {
 	return res;
 }
 
-std::string Database::GetNPCNameByID(uint32 npc_id, bool clean_name) {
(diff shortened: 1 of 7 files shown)


fryguy503 (approved):
LGTM

Akkadius (approved):
In general I'm down with this and fine with merging this

We should try to stray away from setting optional flags on functions to change how something retrieves or mutates data and just have two separate functions. I'm guessing this was just the easier thing to just do in the short term. 

Eg

`GetNPCNameByID(npc_id)`
`GetCleanNPCNameById(npc_id)`

The clean_name sanitization of just underscores won't be enough for names as you could have other non-alpha numeric characters come through this as well. We have `CleanMobName` string_util function for this but it uses annoying character array pointers and isn't std::string friendly

PR #1384, opened 2021-06-11
[Commands] Add #findclass [search criteria] command.

KinglyKrab:
- Allows GMs to find a class by name or ID.
- Modify some verbiage in command messages that were improper.

### DIFF
--- a/zone/command.cpp
+++ b/zone/command.cpp
@@ -2696,26 +2696,36 @@ void command_findclass(Client *c, const Seperator *sep)
 	} else if (Seperator::IsNumber(sep->argplus[1])) {
 		int search_id = atoi(sep->argplus[1]);
 		std::string class_name = GetClassIDName(search_id);
-		if (class_name != std::string("")) {
-			c->Message(Chat::White, "Class %d: %s", search_id, class_name.c_str());
+		if (class_name.length() > 0) {
+			c->Message(
+				Chat::White,
+				fmt::format(
+					"Class {}: {}",
+					search_id,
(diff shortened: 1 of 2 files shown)


fryguy503 (approved):
LGTM

zone/command.cpp
```diff
@@ -2688,10 +2689,51 @@ void command_showskills(Client *c, const Seperator *sep)
...
+		int search_id = atoi(sep->argplus[1]);
+		std::string class_name = GetClassIDName(search_id);
+		if (class_name != std::string("")) {
+			c->Message(Chat::White, "Class %d: %s", search_id, class_name.c_str());
+			return;
+		}
+	} else {
+		const char *search_criteria = sep->argplus[1];
+		int found_count = 0;
+		char class_name[64];
```
Akkadius, line 2706: There's an opportunity to simplify this to use some of the CPP string functionality versus the old C character array API's

See `command_list` for example where string search is using the `std::string` object which has built in functions like `find`

```cpp
/**
 * Filter by name
 */
if (search_string.length() > 0 && entity_name.find(search_string) == std::string::npos) {
	continue;
}
```

Akkadius:
Left minor comments; LGTM otherwise

zone/command.cpp
KinglyKrab: I'll change it when I get home. It was just a copy and paste of #findrace with minor alterations. I can change both for cleanup.

Akkadius (approved):

PR #1387, opened 2021-06-12
[Bug Fix] Fix CMakeLists.txt so compile works.

KinglyKrab:


mackal (approved):

mackal:
Need to include common/types.h in multiple places

PR #1391, opened 2021-06-13
Allow any bag type 51 to be used for Trader

N0ctrnl:
Most commonly this would allow the different color satchels to be used in Trader mode.
PEQ database has 1 item (Yellow Trader's Satchel Token - 35037) marked as type 51, but otherwise only the proper bags are already set.
Bonus of removing the hard-coded ID from source.


N0ctrnl:
Mucked this up somehow. Starting over.

N0ctrnl:
Closing. Will resubmit.

PR #1392, opened 2021-06-13
[Items] Allow any bag type 51 to be used for Trader

N0ctrnl:
Most commonly this would allow the different color satchels to be used in Trader mode.
PEQ database has 1 item (Yellow Trader's Satchel Token - 35037) marked as type 51, but otherwise only the proper bags are already set in that DB.
Bonus of removing the hard-coded ID from source.


Akkadius:
Sweet!

There's an enum I'd prefer to use in place of the magic number `51`

We have an enum in `item_data.h` `BagType` with `BagTypeTradersSatchel` as a value to use instead of `51`

Akkadius (changes requested):
Sweet!

There's an enum I'd prefer to use in place of the magic number 51

We have an enum in item_data.h BagType with BagTypeTradersSatchel as a value to use instead of 51

Also removes the need to have the comments on the line

Akkadius (approved):

N0ctrnl:
Tacked on an update to the test code in bonuses.cpp. Your call if you wanna merge that too or not. Wasn't sure whether to update it or not...

KinglyKrab (approved):

hgtw (approved):

PR #1399, opened 2021-06-13
[Commands] Modify #grid and #wp.

KinglyKrab:
- #grid add will no longer let you put in a duplicate grid.
- Grid nodes now spawn with invul/immune to damage so they cannot be damage accidentally.
- Grid nodes now set an entity variable "grid_id" on spawn.
- This allows grid nodes to be specifically despawned by "grid_id" entity variable, meaning you can view multiple grids at once and not despawn them all accidentally.
- #grid hide will despawn your targeted NPC's Grid nodes.
- #grid add, #grid show, #grid delete, and #grid hide send messages to let GM know what's going on.
- #wp add and #wp delete now send messages to let the GM know what's going on.
- #wpadd now send messages to let the GM know what's going on.

### DIFF
--- a/zone/npc.cpp
+++ b/zone/npc.cpp
@@ -1099,8 +1099,7 @@ bool NPC::SpawnZoneController()
 	return true;
 }
 
-void NPC::SpawnGridNodeNPC(const glm::vec4 &position, int32 grid_number, int32 zoffset) {
-
+void NPC::SpawnGridNodeNPC(const glm::vec4 &position, int32 grid_id, int32 grid_number, int32 zoffset) {
 	auto npc_type = new NPCType;
 	memset(npc_type, 0, sizeof(NPCType));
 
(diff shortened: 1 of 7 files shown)


zone/npc.cpp
```diff
@@ -1112,31 +1111,30 @@ void NPC::SpawnGridNodeNPC(const glm::vec4 &position, int32 grid_number, int32 z
...
+	npc_type->loottable_id = 0;
+	npc_type->texture = 1;
+	npc_type->light = 1;
+	npc_type->size = 1;
+	npc_type->runspeed = 0;
+	npc_type->merchanttype = 1;
+	npc_type->bodytype = 1;
+	npc_type->show_name = true;
+	npc_type->findable = true;
+	strn0cpy(npc_type->special_abilities, "24,1^35,1", 512);
```
hgtw, line 1131: Would setting special abilities work after creating the NPC below so constants could be used? Or does this have a different effect? (I'm not super familiar with this area)
```cpp
npc->SetSpecialAbility(IMMUNE_AGGRO, 1);
npc->SetSpecialAbility(NO_HARM_FROM_CLIENT, 1);

hgtw (approved):

zone/npc.cpp
KinglyKrab: I don't think it has a different effect, I just assumed hard-coding would be fine enough for these grid nodes.

mackal (approved):

PR #1403, opened 2021-06-14
[Feature] Add lua and perl event for test buff

dencelle:


zone/client_packet.cpp
```diff
@@ -14059,6 +14059,7 @@ void Client::Handle_OP_Taunt(const EQApplicationPacket *app)
 
 void Client::Handle_OP_TestBuff(const EQApplicationPacket *app)
 {
+	parse->EventPlayer(EVENT_TEST_BUFF, this, "", 0);
```
mackal, line 14062: We should have a rule guard here (default Handle_OP_TestBuff does nothing and returns, no quest call)

mackal (changes requested):

zone/client_packet.cpp
KinglyKrab: Yeah, example.
```cpp
if (!RuleB(Character, EnableTestBuff)) {
    return;
}

dencelle:
LUA: https://pastebin.com/YjT0ZCFW
PERL: https://pastebin.com/2msZfEZV

mackal (approved):

KinglyKrab (approved):

splose (approved):

Akkadius (approved):

PR #1415, opened 2021-06-14
[Merchants] Add Merchant databucket functionality.

KinglyKrab:
- Allows server operators to limit merchant items based on data bucket values and comparisons.
- Adds 3 columns, `bucket_name`, `bucket_value`, and `bucket_comparison` to `merchantlist` table.
- Bucket is checked based on `character_id-bucket_name`.
- `bucket_comparison` Values are as follows:
  - `bucket_comparison` 0: `bucket_name` == `bucket_value`.
  - `bucket_comparison` 1: `bucket_name` != `bucket_value`.
  - `bucket_comparison` 2: `bucket_name` >= `bucket_value`.
  - `bucket_comparison` 3: `bucket_name` <= `bucket_value`.
  - `bucket_comparison` 4: `bucket_name` > `bucket_value`.
  - `bucket_comparison` 5: `bucket_name` < `bucket_value`.

### DIFF
--- a/zone/client.h
+++ b/zone/client.h
@@ -212,6 +212,15 @@ enum eInnateSkill {
 	InnateDisabled = 255
 };
 
+enum MerchantBucketComparison {
+	BucketEqualTo = 0,
+	BucketNotEqualTo,
+	BucketGreaterThanOrEqualTo,
+	BucketLesserThanOrEqualTo,
+	BucketGreaterThan,
+	BucketLesserThan
+};
+
 const uint32 POPUPID_UPDATE_SHOWSTATSWINDOW = 1000000;
 
 struct ClientReward
--- a/zone/client_process.cpp
+++ b/zone/client_process.cpp
@@ -43,6 +43,7 @@
 #include "../common/skills.h"
 #include "../common/spdat.h"
 #include "../common/string_util.h"
+#include "data_bucket.h"
 #include "event_codes.h"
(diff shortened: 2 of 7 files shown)


splose (approved):
This was created towards the end of RoN and was very useful in creating custom progression merchants or simply hiding stuff from clients.

zone/client.h
```diff
@@ -212,6 +212,15 @@ enum eInnateSkill {
 	InnateDisabled = 255
 };
 
+enum MerchantBucketComparison {
```
mackal, line 215: this should be an `enum class` rather than plain enum. (Eventually I would like to switch all enums over, unless they really don't make sense to)

zone/client_process.cpp
```diff
@@ -824,125 +825,171 @@ void Client::BulkSendInventoryItems()
...
+	uint32 current_merchant_slot = 1;
+	uint8 handy_chance = 0;
+	for (auto merchantlist : current_merchantlist) {
+		if (current_merchant_slot > merchant_slots) {
+			break;
+		}
+		
+		auto merchantlist_slot = merchantlist.slot;
+		auto bucket_name = merchantlist.bucket_name;
+		if (bucket_name.length() > 0) {
```
mackal, line 855: Should be checking if the string is empty instead of length.

mackal (changes requested):

PR #1420, opened 2021-06-16
[Rule] Classic 2HB Animation

splose:
Adds ability to use the ~~old~~ proper 2HB animation and also allows you to do it on a per-zone basis since its rule-based.


KinglyKrab (approved):
<3

Akkadius:
If this is what it should be, should this really be rule based?

Curious what others might know on this animation

mackal:
They changed it on live for some reason. No idea what it's current state is on live, but at some point they used 2H Slashing :|

KinglyKrab:
Live uses 2H Slashing animation now. Used to be 2H piercing animation, which works better for piercing weapons. Just a preference thing and allows old servers to turn off/on the old animation for the plethora of TLP/Classic servers we have. Also it's disabled by default so shouldn't be too big of an issue.

Akkadius:
> Live uses 2H Slashing animation now. Used to be 2H piercing animation, which works better for piercing weapons. Just a preference thing and allows old servers to turn off/on the old animation for the plethora of TLP/Classic servers we have. Also it's disabled by default so shouldn't be too big of an issue.

It's less of it being an issue that its disabled by default and moreso if we can avoid extra knobs and have sane defaults that's the main thing I'm trying to ask at here given Splose's original comment "Adds ability to use the ~~old~~ proper 2HB animation'

Given that what we have currently is live like and this does restore classic functionality it makes sense to have the knob

Akkadius (approved):

PR #1424, opened 2021-06-17
[Commands] Make #maxskills work on target.

KinglyKrab:


splose (approved):
Nice.. this is a lot cleaner than the original.

PR #1427, opened 2021-06-18
[Cheat Detection] Anti-Cheat reimplementation

dencelle:
Currently detects the following:

- MQ2Warp (uses DeltaT and MovementHistory flags)
- MQ2Zone
- MQ2Gate
- MQ2FastMem
- MQ2Ghost

this is a mixture of VZTZ's code back when it got removed and some new methods that have been talked about on the discord.

server operators testing this: please report any false positives you get, while i have tested the code extensively, i can not say with absolute certainty i've thought of every situation the MQ2 client gives.


KinglyKrab:
I'll leave this closed for now due to his clear intention of using this to be malicious to the community. I'll let others pipe in.

Issue #1436, opened 2021-06-22
Casting enchanter illusions shows naked character


thalix1337:
This is from [eqemulator.org](http://www.eqemulator.org/forums/showthread.php?p=266809), found by user **Zandig**.

**Issue**
When casting an illusion spell as an enchanter, the illusion no longer displays the items being worn. 

**Steps to always reproduce it**
High Elf Enchanter with Robe casts "Illusion Gnome" on himself.

**Result**
The character appears without a robe.

**Additional comment from eqemulator.org**
Rolled back to a backup of the source I had (bin folder) which is dated: 05-20-2021 and the Illusion spell is working properly showing the equipment. So definitely something happened in the source code between now and then, at least in the bots version.

ZandigSlaytanic:
Thanks for entering this issue Thalix. Was going to enter it today.

Akkadius:
Posted up https://github.com/EQEmu/Server/pull/1437

Thanks for flagging

Akkadius:
Merged hotfix, closing

PR #1437, opened 2021-06-22
[Hotfix] SendIllusion revert to October

Akkadius:
Revert the rest of the code that was changed back to October until further deeper testing. This function does too many things and is interwoven in spaghetti code

This should address https://github.com/EQEmu/Server/issues/1436 and bring us back to our original zero state but the issues originally reported by @RoT-PvP will still remain which at this point are far smaller than the side effects produced by what's currently there

### DIFF
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -1815,25 +1815,23 @@ void Mob::SendIllusionPacket(
 	new_drakkin_details  = (in_drakkin_details == 0xFFFFFFFF) ? GetDrakkinDetails() : in_drakkin_details;
 	new_aa_title         = in_aa_title;
 
-	bool  reset_features_to_player_profile = IsClient() && in_race == 0;
-	if (reset_features_to_player_profile) {
-		auto c = CastToClient();
-		race                 = c->GetBaseRace();
-		gender               = c->GetBaseGender();
+	// Reset features to Base from the Player Profile
+	if (IsClient() && in_race == 0) {
+		race                 = CastToClient()->GetBaseRace();
+		gender               = CastToClient()->GetBaseGender();
 		new_texture          = texture          = 0xFF;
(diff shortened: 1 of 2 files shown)


mackal (commented):
Only real comment, CastToClient() is a function call (the useless null check is optimized out because this can never be null at least) but it's still a function call, so it makes sense to call it once. And it doesn't get inlined, at least on default optimization levels.

mackal (approved):

noudess (commented):
Reverting the use of a pointer vs calling CastToClient() n times seems unrelated.  Personally I'd leave that alone.

The size code is definitely.. yuk.  But I think the GetRaceDefaultHeight() fixed an issue awhile back when mobs reverted back from illusions.  I think mobs that changed back to original form on death were coming out the wrong size....  Not sure but it rings a bell.

Seems like the changes to spell_effects are more likely to be causing the change..

Just my thoughts  - use or disregard

mackal:
>  But I think the GetRaceDefaultHeight() fixed an issue awhile back when mobs reverted back from illusions.

I'm pretty sure that is correct, so we will probably need something else to fix this issue.

hgtw (approved):

Akkadius:
Appreciate the review comments but this is literally just reverting code back to a previous state verbatim

We need to do additional investigation and more thorough testing with further changes as per original comment

GangstaEQ:
issue still present

Akkadius:
> issue still present

Need to be a bit more specific. This PR reverted back to October. The issue you first brought up would be around because we reverted. The original changes to fix your original issue created more issues than it was worth and needs to be re-investigated

There's a lot of spaghetti logic interwoven with SendIllusion and we need to be careful making changes to it because it can change a lot of state that affects related components. Someone will need investigate further

Issue #1385, opened 2021-06-11
Unify hardcoded spell ID behavior


mackal:
Currently we have these defined in 2 different places. It would be best if we only did this in one place.

https://github.com/EQEmu/Server/blob/6e61f6d0ba264d0c8c9bc3e8bfaf2f1e1912348f/common/spdat.h#L28

https://github.com/EQEmu/Server/blob/6e61f6d0ba264d0c8c9bc3e8bfaf2f1e1912348f/common/eq_constants.h#L441

Personally I'm partial to moving them all over to spdat.h as defines since there are more of those currently.

Akkadius:
I agree; lets keep them in one place

PR #1438, opened 2021-06-25
[Spells Cleanup] Unify and add most hardcoded spell IDs

mackal:
Move the newer stuff added that was recently to the same place all of
our previous spell IDs were defined.

Either of these solutions were good, but I went with defines since it
was less changes

I also added a bunch of stuff the client has hardcoded behavior for, but
not currently implemented by us.

The removed stuff from the command_castspell were reused on live, so I
figured it was best to remove them from the restrictions since they are
no longer test spells


mackal:
This address #1385

Akkadius (approved):
LGTM

Issue #1442, opened 2021-06-27
Evac message falsely claims "Entering South Qeynos".


thalix1337:
**Issue**
Wizards casting "Evacuate" result in the message "Entering South Qeynos" while entering the correct (other) zone.

**Steps to reproduce** (always)
A wizard bound in any zone other than South Qeynos casting Evacuate, resulting in the incorrect client message "Entering South Qeynos".

**Possible fix**
Trust has kindly pointed out that he has a fix in the P2002 repository - which had led to some side effects with aggro.
[no more south qeynos BS on evacs](https://github.com/p2002eq/Game_Server/commit/6fd161278c13faf7914c844441b46b020689c08f).

Valorith:
Pretty sure this happens with druids too...

splose:
Yeah it happens with any succor spell..  pretty sure there's no ideal way to fix it other than hacking the client. 

The proposed fix from before causes really wonky issues with mob aggro.

mackal:
> Yeah it happens with any succor spell.. pretty sure there's no ideal way to fix it other than hacking the client.
> 
> The proposed fix from before causes really wonky issues with mob aggro.

Basically we need to sit down and resolve the corner cases with that, then we should be good, just isn't really worth the effort.

PR #1444, opened 2021-06-29
[Typo Fix] Wizard evac saying the wrong zone

dencelle:
fix for issue brought up by @thalix1337 in issue #1442 this should solve the incorrect succor zone message being shown on the client


dencelle:
closing this since this would cause more issues than the simple typo fix would solve

PR #1446, opened 2021-06-30
Add defines (commented) for further known SPAs

mackal:
Added as comments since that appears what we were doing. This just gets
us caught up to dev post.

### DIFF
--- a/common/spdat.h
+++ b/common/spdat.h
@@ -861,6 +861,12 @@ typedef enum {
 //#define SE_Attack_Accuracy_Max_Percent	518 //
 //#define SE_Luck_Amount				519 //
 //#define SE_Luck_Percent				520 //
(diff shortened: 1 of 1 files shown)


dencelle:
idk if these should be added or this function should be added... but this is in eqgame... its all of the hard coded darkness spells 
image

mackal:
> idk if these should be added or this function should be added... but this is in eqgame... its all of the hard coded darkness spells
> image

unrelated to this PR, but yes, we probably should have that function, but probably after someone figures out why we need it :P (and implements the feature that uses it)

dencelle:
Yeah, its not really needed on our end, I more just wanted to mention it
since it is hard coded client values

fryguy503 (approved):
LGTM

PR #1449, opened 2021-07-06
[Quest API] Sit method

GangstaEQ:
See title.


Akkadius:
This looks fine to me, but we should also add this to Lua

mackal (changes requested):
As Akkadius said, should be exported to lua as well.

splose:
This is actually my code, and unfortunately I don't know how to export it to lua, perhaps one of you can show me?

fryguy503:
> This is actually my code, and unfortunately I don't know how to export it to lua, perhaps one of you can show me?

There are a number of lua implemented snippets here that may be of assistance. https://github.com/EQEmu/Server/pulls?q=is%3Apr+is%3Aclosed+sort%3Aupdated-desc+lua

Akkadius (approved):

PR #1458, opened 2021-07-16
Fixes Bolt spells from hitting the caster in cases where the target dies or zones

GangstaEQ:
See title.

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -1003,8 +1003,8 @@ void Mob::CastedSpellFinished(uint16 spell_id, uint32 target_id, CastingSlot slo
 	}
 
 	if (IsDetrimentalSpell(spell_id) && spells[spell_id].range == 300) { //prevents bolt spells (300 range) from nuking the caster if the target zones or dies before spell finish.
-		target = entity_list.GetMobID(target_id);
-		if (!target) {
+		target = entity_list.GetMob(target_id);
+		if (target == nullptr) {
 			InterruptSpell(spell_id);
 			return;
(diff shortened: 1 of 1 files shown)


GangstaEQ:
So the main thing with this PR for me would be turning this into a spdat method where its IsBoltSpell orrrr doing this check for any detrimental. In what cases should a detrimental spell go through on a target who is no longer in the entity list.

GangstaEQ:
this is broke gotta fix

KayenEQ:
```
		Mob *target = entity_list.GetMob(target_id);
		if (!target) {
			InterruptSpell(spell_id);
			return;
		}
```
Let me review this more closely though since I wrote the bolt code. There is probably a better answer.

mackal:
I feel like this is a pretty poor solution. Should probably be able to prevent the damage much sooner and with less side effects.

GangstaEQ:
> I feel like this is a pretty poor solution. Should probably be able to prevent the damage much sooner and with less side effects.

Got any suggestions? I would rather have a fix or better code instead of "poor implementation" accepting any tips or ideas on how to write this better.

KayenEQ:
This is the appropriate fix. The code was intentionally making it so the caster got nuked if no target. But, there is no actual reason it should be like that given the functionality of that target type. I included the fix in my latest pull request. Thanks GangstaEQ for discovering this bug.

https://github.com/EQEmu/Server/pull/1459/commits/5e049d41e0674674f85ba0d0b8fc8fe403c54ca5

PR #1459, opened 2021-07-16
[Feature] New SPAs pass 2

KayenEQ:
Implemented various new spell effects. (7/16/21)

------------------------------------------------------------------------>
SE_Fc_Cast_Spell_On_Land 481
Focus effect that is checked when a spell is cast on a target, if target has this focus effect and all limiting criteria are met, then the target will cast a spell as specified by the focus. Can be given a roll chance for success. Base1=Chance, Base2=Spellid

Note: This spell has a huge amount of potential applications. See 'Alliance' type spells on live. (ie live spell 50247)
Simple Examples: Every time a heal is cast on a target with this focus effect, the effect will proc causing target to cast an AOE heal. Can also be used for detrimental procs. Every time a direct damage spell is cast on the 'target' with this focus effect, the effect with proc causing a stun to be cast on the 'target's' target..

Implemented associated focus limits seen in live spells.

SE_Ff_CasterClass	485
- Caster of spell on target with a focus effect that is checked by incoming spells must be specified class or classes.
Note do not apply more than of this limit. If multiple classes the base1 value is same as if you had multiple classes on an item.

SE_Ff_Same_Caster	 486 -Caster of the spell on target with a focus effect that is checked by incoming spells 0=Must be different caster 1=Must be same caster. Example: If YOU cast a buff that has Fc_Cast_Spell_on_Land with this limiter set to zero, YOU will not be able to trigger the spell proc.

The following is an associated effect seen with SPA 481

SE_Proc_Timer_Modifier 			512
This provides a way to rate limit the amount of spell triggers generated by SPA 481. For example after 1 successful spell trigger no additional spells can be triggered for 1.5 seconds. Ie. Base=1 and Base2 1500.
Written in a flexible format to allow scaling of multiple different buffs with this effect at same time.
Note, base1 can be any value greater than zero, but needs be at least 1. Base2 is in MS 1500 = 1.5 seconds.
------------------------------------------------------------------------>
SE_Fc_Spell_Damage_Pct_IncomingPC	483 
Focus effect that modifies incoming spell damage by percent. Positive value increases spell damage taken, negative decrease.
Can return a random value from a range of values. Base1=min damage, base2=max damage
Functions like SPA SE_FcSpellVulnerability 296, is stackable with it.

SE_Fc_Spell_Damage_Amt_IncomingPC	484 
Focus effect modifies incoming spell damage by flat amount. Typically adds damage. 

SE_Fearstun 	502
Stun with a max level limit. Normal stun restrictions don't apply. Will not effect targets immune fear or stun immune.
Base1 duration, base2 PC duration (optional if want different length of stun on clients , Max is level limit (max level targets
can be affected)

SE_TwinCastBlocker 39
Previously unused spell effect that is now used on live. Simply, if this effect is present in a spell then the spell can not be twin cast.

SE_PC_Pet_AE_Rampage	 465
Chance for pet to AE rampage with a damage modifier
Base1 % chance to do AE rampage for base2 % of damage each melee round

SE_Pet_Add_Atk 494
Bonus on pet owner which gives their pet increased attack stat

SE_Hatelist_To_Top_Index 477
Chance to be put on top of RAMPAGE list. Base1 % chance 

SE_Hatelist_To_Tail_Index 478
Chance to be put on bottom of RAMPAGE list. Base1 % chance
------------------------------------------------------------------------>
SE_Chance_Best_in_Spell_Grp     469  
Chance to cast highest scribed spell within a spell group. base1=spell chance base2=spellid
All base2 spells share roll chance, only 1 cast. Chance should equal 100 pct, otherwise will just do a roll chance on each spell with chance of failure.
Ie. SpellA has 30% SpellB has 50% and SpellC has 20% pct chance. When applied to target only 1 of these spells will cast, and in this case only the highest ranked spell in the spell group. This is how SPA 340 TrySpellTrigger works except it doesn't check for highest rank in spell group. This effect can be used in conjunction with SPA 340.

SE_Trigger_Best_in_Spell_Grp	470
Chance to cast highest ranked scribed spell within a spell group. Each spell has own chance.

Additional Changes:
Rewrote TrySpellTrigger function used for SPA 340 since it incorporates SPA 469. Improved code so that chance of spell being triggered should be more accurate statistically.
------------------------------------------------------------------------>

### DIFF
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -3797,7 +3797,7 @@ int32 Mob::GetVulnerability(Mob* caster, uint32 spell_id, uint32 ticsremaining)
 	return value;
 }
 
-int16 Mob::GetSkillDmgTaken(const EQ::skills::SkillType skill_used, ExtraAttackOptions *opts)
+int32 Mob::GetSkillDmgTaken(const EQ::skills::SkillType skill_used, ExtraAttackOptions *opts)
 {
 	int skilldmg_mod = 0;
 
@@ -3816,7 +3816,7 @@ int16 Mob::GetSkillDmgTaken(const EQ::skills::SkillType skill_used, ExtraAttackO
 	return skilldmg_mod;
 }
 
-int16 Mob::GetPositionalDmgTaken(Mob *attacker)
+int32 Mob::GetPositionalDmgTaken(Mob *attacker)
 {
 	if (!attacker)
 		return 0;
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -3881,6 +3881,61 @@ int16 Mob::GetHealRate(uint16 spell_id, Mob* caster) {
 	return heal_rate;
 }
 
+void Mob::SetBottomRampageList()
+{
+	auto &mob_list = entity_list.GetCloseMobList(this);
+
+	for (auto &e : mob_list) {
+		auto mob = e.second;
+		if (!mob) {
+			continue;
+		}
+
+		if (!mob->GetSpecialAbility(SPECATK_RAMPAGE)) {
+			continue;
+		}
+
+		if (mob->IsNPC() && mob->CheckAggro(this)) {
+			for (int i = 0; i < mob->RampageArray.size(); i++) {
+				// Find this mob in the rampage list
+				if (this->GetID() == mob->RampageArray[i]) {
+					//Move to bottom of Rampage List
+					auto it = mob->RampageArray.begin() + i;
+					std::rotate(it, it + 1, mob->RampageArray.end());
+				}
+			}
+		}
+	}
+}
+
+void Mob::SetTopRampageList()
+{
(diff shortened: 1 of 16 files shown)


splose (approved):

zone/mob.cpp
```diff
@@ -3856,6 +3937,61 @@ int16 Mob::GetHealRate(uint16 spell_id, Mob* caster) {
 	return heal_rate;
 }
 
+void Mob::SetBottomRampageList()
```
mackal, line 3940: These implementations I think are a bit subpar. We should be able to add hatelist functions and the mob will call them and do it without the whole entity list bit.

mackal (changes requested):

mackal (approved):
Okay, this looks fine, would still like someone else to review it as well since it's A LOT of changes and brains are dumb and stuff.

PR #1466, opened 2021-07-26
Projectiles Updates

KayenEQ:
Fixed bug with spell projectiles  causing angle to not be calculated correctly. Spell projectiles will now fire straight instead of from the side.

Significantly improved all projectile timings. At least up to range of 300 should see more accurate timings for damage/effects occurring upon projectile impact. 

Will be noticed most significantly for all spell projectiles and for longer range archery.

### DIFF
--- a/common/spdat.cpp
+++ b/common/spdat.cpp
@@ -1218,6 +1218,12 @@ bool IsEffectIgnoredInStacking(int spa)
 	case SE_LimitUseType:
 	case SE_GravityEffect:
 	case 425:
+	//Spell effects implemented after ROF2, following same pattern, lets assume these should go here.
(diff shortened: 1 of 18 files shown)


KayenEQ:
Didn't realize had my focus effects branch merged into this as well, which makes seeing the changes confusing.. That branch is already approved for merge.

See that for the isolated changes.
https://github.com/KayenEQ/Server/commit/69c3013c60640e322864c828a06170c4b733933e

fryguy503:
Kayen, can this PR be cleaned up (Or closed/reopened) with just the Projectile changes. Just to keep things consistent for viewing and approvals.

KayenEQ:
Ok I will close and re submit a clean one.

PR #1468, opened 2021-07-26
Projectiles Update

KayenEQ:
Fixed bug with spell projectiles causing angle to not be calculated correctly. Spell projectiles will now fire straight instead of from the side.

Significantly improved all projectile timings. At least up to range of 300 should see more accurate timings for damage/effects occurring upon projectile impact.

Will be noticed most significantly for all spell projectiles and for longer range archery.

### DIFF
--- a/zone/special_attacks.cpp
+++ b/zone/special_attacks.cpp
@@ -929,11 +929,31 @@ bool Mob::TryProjectileAttack(Mob *other, const EQ::ItemData *item, EQ::skills::
 	if (slot < 0)
 		return false;
 
-	float speed_mod = speed;
-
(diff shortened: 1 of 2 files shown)


mackal (approved):
Code wise, this looks fine. I trust @KayenEQ that it looks right in game, but I'll hold off on merging for a bit to give others to actually test in game.

SecretsOTheP:
Looks safe to me. I hate parameter spaghetti (should really be a struct) but we use that elsewhere in the attack code still.

mackal:
Okay, I did test this on my server, looks fine, I'll probably merge this evening unless someone else finds issue.

PR #1472, opened 2021-07-26
[Spells] Bug fix for AOE Harmony/Lull

KayenEQ:
Fixed bug with SPA 30 SE_SE_ChangeFrenzyRad and SPA 86 SE_Harmony allowing those spells to affect NPC's above their level restrictions when cast as a 'Targeted AE' spell (ie. Harmony, Wake of Tranquility) when the targeted NPC was bellow level restricted range, but the NPC's next to them were above it.

Implemented the live like behavior, if this case occurs on live, the buff is not applied to the targets over the level limit and "Your target looks unaffected" message is given. 

Also added additional layer of protection, even if somehow the buff does get on the target, the actual effect will not be applied to an NPC over the level limit.

This bug was originally reported by: Isaaru (7/26/21)

### DIFF
--- a/zone/spell_effects.cpp
+++ b/zone/spell_effects.cpp
@@ -7416,27 +7416,18 @@ void Client::BreakFeignDeathWhenCastOn(bool IsResisted)
 	}
 }
 
-bool Mob::HarmonySpellLevelCheck(int32 spell_id, int target_id, Mob *target)
+bool Mob::HarmonySpellLevelCheck(int32 spell_id, Mob *target)
 {
 	//'this' = caster of spell
-	if (IsClient() && IsHarmonySpell(spell_id))
-	{
-		Mob *spell_target = nullptr;
+	if (!target) {
+		return false;
+	}
 
-		if (target_id)
-			spell_target = entity_list.GetMobID(target_id);
-		else
-			spell_target = target;
-
-		if (!spell_target)
-			return false;
-
-		for (int i = 0; i < EFFECT_COUNT; i++) {
-			// not important to check limit on SE_Lull as it doesnt have one and if the other components won't land, then SE_Lull wont either
-			if (spells[spell_id].effectid[i] == SE_ChangeFrenzyRad || spells[spell_id].effectid[i] == SE_Harmony) {
-				if ((spells[spell_id].max[i] != 0 && spell_target->GetLevel() > spells[spell_id].max[i]) || spell_target->GetSpecialAbility(IMMUNE_PACIFY)) {
-					return false;
-				}
+	for (int i = 0; i < EFFECT_COUNT; i++) {
+		// not important to check limit on SE_Lull as it doesnt have one and if the other components won't land, then SE_Lull wont either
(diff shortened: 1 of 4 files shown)


zone/spell_effects.cpp
```diff
@@ -7415,3 +7415,30 @@ void Client::BreakFeignDeathWhenCastOn(bool IsResisted)
 		MessageString(Chat::SpellFailure,FD_CAST_ON);
 	}
 }
+
+bool Mob::HarmonySpellLevelCheck(int32 spell_id, int target_id, Mob *target)
```
Akkadius, line 7419: I love that you made this a function to help aggregate this check

The thing that I would suggest to simplify this is to just pass the mob target entity in this where you're passing in `target_id`

Eg. If you need to fetch the entity out of band; do it in the place where this function is being called and pass the mob entity into the function 

```cpp
auto target = entity_list.GetMobID(target_id);

// ...
if (!HarmonySpellLevelCheck(spell_id, target)) {

```

zone/spells.cpp
```diff
@@ -224,17 +224,9 @@ bool Mob::CastSpell(uint16 spell_id, uint16 target_id, CastingSlot slot,
...
-		for(int i = 0; i < EFFECT_COUNT; i++) {
-			// not important to check limit on SE_Lull as it doesnt have one and if the other components won't land, then SE_Lull wont either
-			if (spells[spell_id].effectid[i] == SE_ChangeFrenzyRad || spells[spell_id].effectid[i] == SE_Harmony) {
-				if((spells[spell_id].max[i] != 0 && GetTarget()->GetLevel() > spells[spell_id].max[i]) || GetTarget()->GetSpecialAbility(IMMUNE_PACIFY)) {
-					InterruptSpell(CANNOT_AFFECT_NPC, 0x121, spell_id);
-					return(false);
-				}
-			}
-		}
+	if (!HarmonySpellLevelCheck(spell_id, target_id)) {
```
Akkadius, line 227: I would still gate this logic with `IsHarmonySpell`

CastSpell is a bit of a hot path and we are potentially unnecessarily running entity fetch logic and other calculation for things that we may not even be concerned with

I understand you're validating inside the function but validation should be done earlier in the call stack and not left to an edge function

```
if IsHarmonySpell
 -- HarmonySpellLevelCheck(spell_id, entity_list.GetMobID(target_id))
```

zone/spell_effects.cpp
```diff
@@ -7415,3 +7415,30 @@ void Client::BreakFeignDeathWhenCastOn(bool IsResisted)
...
+{
+	//'this' = caster of spell
+	if (IsClient() && IsHarmonySpell(spell_id))
+	{
+		Mob *spell_target = nullptr;
+
+		if (target_id)
```
Akkadius, line 7426: Just a stylistic note for the future to just use brackets instead of implied if checks; it has caused issues in the past

```cpp
if (target_id) {
    // thing
}
```

vs

```cpp
if (target_id)
    // thing
```

KayenEQ:
All comments appreciated. Updated code to reflect suggestions. Much cleaner now. Thanks.

mackal (approved):

PR #1483, opened 2021-07-30
[Spells] Update to SPA 130 and SPA 131 focus calculation, focus code improvements

KayenEQ:
Live like updates.
Update to SPA 130 SE_SpellHateMod: To allow focus calculation to check a random range if indicated by spell file data.
Reminder, focus allows to increase or decrease hatred generated from spell damage.

Update to SPA 131 SE_ReduceReagentCost: To allow focus calculation to check a random range if indicated by spell file data.
Reminder, focus gives a chance not to consume a reagent.

For both, and any focus that uses random effectiveness formula.
If Base and Limit values both are defined. Then Base = min value Limit = max value
If you don't want random range, then Base = value Limit = 0, or set Limit to same as Base value

Added a standard function to do all the random effectiveness calculations in source.

### DIFF
--- a/zone/spell_effects.cpp
+++ b/zone/spell_effects.cpp
@@ -5640,7 +5640,7 @@ int16 Client::GetFocusEffect(focusType type, uint16 spell_id)
 
 	//Improved Healing, Damage & Mana Reduction are handled differently in that some are random percentages
 	//In these cases we need to find the most powerful effect, so that each piece of gear wont get its own chance
-	if(RuleB(Spells, LiveLikeFocusEffects) && (type == focusManaCost || type == focusImprovedHeal || type == focusImprovedDamage || type == focusImprovedDamage2 || type == focusResistRate))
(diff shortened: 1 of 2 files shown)


zone/spell_effects.cpp
```diff
@@ -7459,3 +7365,47 @@ bool Mob::HarmonySpellLevelCheck(int32 spell_id, Mob *target)
...
+	case focusImprovedDamage2:
+	case focusImprovedHeal:
+	case focusManaCost:
+	case focusResistRate:
+	case focusFcDamagePctCrit:
+	case focusReagentCost:
+	case focusSpellHateMod:
+	case focusSpellVulnerability:
+	case focusFcSpellDamagePctIncomingPC:
+	return 1;
```
mackal, line 7384: Use true/false

Akkadius (approved):
Made a few simplifications; LGTM

PR #1485, opened 2021-07-31
[Spells] Added constant labeling to all StatBonuses that use as arrays.

KayenEQ:
Added constant labeling to all StatBonuses that use arrays. It should now be unmistakable what each element of the array represents throughout the source code now. Significantly improves readability of many effects.


Akkadius (approved):
Made adjustments; LGTM

Love the readability changes

PR #1489, opened 2021-08-03
[cmake] Update min cmake version due to fe7cb76

mackal:


Akkadius:
@mackal what was this in regards to? I'm sure lots of windows users will feel pain on this one so wondering why this is needed

mackal:
> @mackal what was this in regards to? I'm sure lots of windows users will feel pain on this one so wondering why this is needed

commit mentioned requires it, unless we change that, we need to force updates anyways since it won't work on older versions.

Akkadius (approved):

PR #1497, opened 2021-08-17
[Spells] Bard Dot Pulse Resist Fix

RoT-PvP:
Previously, if a bard landed the initial "cast" of their bard song the pulses afterward would not have a resist check. There is likely a better way to write this but this is what worked for me so far.

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -2712,15 +2712,25 @@ void Mob::BardPulse(uint16 spell_id, Mob *caster) {
 			return;
 		}
 		//extend the spell if it will expire before the next pulse
-		if(buffs[buffs_i].ticsremaining <= 3) {
+		if(buffs[buffs_i].ticsremaining <= 3 && !IsDetrimentalSpell(spell_id)) {
 			buffs[buffs_i].ticsremaining += 3;
 			LogSpells("Bard Song Pulse [{}]: extending duration in slot [{}] to [{}] tics", spell_id, buffs_i, buffs[buffs_i].ticsremaining);
+		}
+		float resist_check = 0;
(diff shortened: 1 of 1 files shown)


mackal:
This whole pulse "optimization" is just bad and needs to be removed. It causes more problems than it's worth.

splose:
> This whole pulse "optimization" is just bad and needs to be removed. It causes more problems than it's worth.

This is how we give feedback to new people who are obviously trying to learn now? 

Your comment is bad and the opposite of helping. **_Instead of the snarky quotes maybe tell him why its bad, and why "it causes more problems than it's worth."_** Do you feel better sitting there with an IT degree and years of EQEmu experience telling this guy who has been coding for like 2-3 months his code is bad?

When do you realize that putting up PRs is extra work for server operators? Then on top of that having to deal with nonsense like this is a pretty common reason people cite for not contributing and hoarding their code. You do this the most and to anyone who is not in your little clique and whenever anyone tells you where to go after they seem like the bad guy. 

We should be encouraging new people like this and making them want to contribute. The future of the community depends on it. Do better.

Akkadius:
Need folks more versed on intended functionality to give their thoughts on this

mackal:
We can probably call `IsDetrimentalSpell()` only once. It also appears that the resist check is done in `Mob::SpellOnTarget()`. So it's probably better to just come with with a better check to skip the optimization and just call `SpellOnTarget()` without the resist override to get the full behavior.

RoT-PvP:
May have found a bug with this that causes a crash. If a bard is running a song on a mob and then it is charmed by someone he cannot attack the reapplication may cause a crash. Need to verify this.

Akkadius:
Closing this until there is more of a clear path forward on this one

PR #1501, opened 2021-08-21
[Spells] Updated pet suspend code to use spell effect data and bonuses

KayenEQ:
Suspend Minion and related effects were still hardcoded using old AA system.
No actual change in functionality besides that is now more flexible and can be applied like any other spell/bonus, therefore can
give any class at any level these abilities if desired.

SPA 151 SuspendPet updated to use actual spell values from spells_new table.

```
/*
SPA 151 Allows an extra pet to be saved and resummoned later.
Casting with a pet but without a suspended pet will suspend the pet
Casting without a pet and with a suspended pet will unsuspend the pet
effect value 0 = save pet with no buffs or equipment
effect value 1 = save pet with buffs and equipment
effect value 2 = unknown
Note: SPA 308 allows for suspended pets to be resummoned after zoning.
*/
```
Renamed SPA 308 SE_SuspendedMinion to SE_ZoneSuspendMinion
No longer hard coded to the AA.
allow suspended pets to be resummoned upon zoning, base: 1, limit: none, max: none, Calc: Bool

### DIFF
--- a/common/spdat.h
+++ b/common/spdat.h
@@ -514,7 +514,7 @@ typedef enum {
 #define SE_StackingCommand_Block		148 // implemented?
 #define SE_StackingCommand_Overwrite	149 // implemented?
 #define SE_DeathSave					150 // implemented
-#define SE_SuspendPet					151	// *not implemented as bonus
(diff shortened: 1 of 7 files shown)


splose (approved):
Very cool.

Code is pretty simple looks good to me.

Akkadius (approved):

PR #1513, opened 2021-08-31
[Spells] Implement PVP resist and duration overrides

mackal:
Make use of the spell fields. If you don't want these enabled, I recommend just setting the data equal to normal rather than bothering with a rule.

### DIFF
--- a/common/repositories/base/base_spells_new_repository.h
+++ b/common/repositories/base/base_spells_new_repository.h
@@ -199,8 +199,8 @@ public:
 		int         pvpresistcalc;
 		int         pvpresistcap;
 		int         spell_category;
-		int         field181;
-		int         field182;
+		int         pvp_duration;
+		int         pvp_duration_cap;
 		int         pcnpc_only_flag;
 		int         cast_not_standing;
 		int         can_mgb;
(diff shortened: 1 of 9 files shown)


common/repositories/base/base_spells_new_repository.h
```diff
@@ -199,8 +199,8 @@ class BaseSpellsNewRepository {
 		int         pvpresistcalc;
 		int         pvpresistcap;
 		int         spell_category;
-		int         field181;
```
Akkadius, line 202: Did you manually update these repositories or did you use the script to update ?

Kinglykrab (approved):

common/repositories/base/base_spells_new_repository.h
mackal: The script is unusable for me so I manually updated it.

Akkadius: What do you mean, unusable?

Akkadius: How to use is here: https://eqemu.gitbook.io/server/in-development/developer-area/repositories

```
perl ~/code/utils/scripts/generators/repository-generator.pl ~/server spells_new
```
Checked out your code and looks like you got it all manually 100% anyways

Akkadius (approved):

PR #1518, opened 2021-09-04
[Hitpoints] Fix HP update throttling

splose:
Addition to https://github.com/EQEmu/Server/pull/1517

It appears to just be the if statement here.. I upped the throttling numbers 3x and it was still updating much faster than before. 

With the if statement:
https://i.gyazo.com/74f0e6301340412f67fb12c371ac1853.mp4

without:
https://i.gyazo.com/58c791576d268155073a8f1bb0fc70a7.mp4

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -3739,7 +3739,7 @@ void Mob::CommonDamage(Mob* attacker, int &damage, const uint16 spell_id, const 
 
 		//send an HP update if we are hurt
 		if (GetHP() < GetMaxHP())
-			SendHPUpdate(!iBuffTic); // the OP_Damage actually updates the client in these cases, so we skip the HP update for them
+			SendHPUpdate(); // the OP_Damage actually updates the client in these cases, so we skip the HP update for them
 	}	//end `if damage was done`
 
 		//send damage packet...
--- a/zone/client_mods.cpp
+++ b/zone/client_mods.cpp
@@ -337,9 +337,6 @@ int32 Client::CalcMaxHP()
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


Akkadius (approved):
Removed some additional hacks around this

Context from coders

[10:55 PM] Akk: Yeah you're not wrong on that code causing issues. Few things... that code is causing issues because it's a hack on top of a hack. Looks like Uleat introduced it back when the throttle was around to force the throttle update timer to trigger which is going to just cause bouncing and missed updates https://github.com/EQEmu/Server/commit/92d32feb0dd98de4a8460292b9ff84ab11697382. It can be removed per your PR but the member variable last_max_hp and its references should also be removed as well
[10:55 PM] Akk: When I was doing my testing yesterday I was testing with #damage and all works as expected. Although, when you use Manastone it triggers Client::CalcBonuses() which subsequently calls CalcMaxHP(); which zeroes out last_max_hp and causes no HP update packet to be sent at all because last_max_hp is always 0 when SendHPUpdate is called and always returns out without an update. Previous code was relying on the timer to be triggered next process loop in that return block which is not great

image

[10:59 PM] Akk: @chrsschb [Clumsy's World] this change would line up when x64 was introduced but the issue isn't x64 itself
[10:59 PM] Akk: x64 was April 2020 and this change was in Jan 2020
[11:00 PM] Akk: It needs to just be removed and let the hp update logic do what it should normally do with all throttling removed and updates should be accurate to a T unless there are base value disparities between client and server
[11:11 PM] Akk: Looks good, I'll tweak the changes on Splose' PR

image

PR #1519, opened 2021-09-05
[Bots] Fix for Bot Pets Taunting

Isaaru:
Currently bot pets will taunt and there's no way to turn it off. This makes it so pets follow their owner's taunt settings.


mackal:
We might want to add a way to only turn it on for pets or bots I would think. This just makes it so they're both on or off. (Think SHD bot)

Akkadius (approved):

PR #1525, opened 2021-09-08
[Saylinks] Implement Auto Saylink Injection

Akkadius:
This PR adds the ability to automatically inject saylinks into "say" dialogue and client messages that are only invoked through the Quest API

If there is a saylink that already exists in the message, the injector will leave it alone

The reason we don't inject saylinks at the edge (Inside Client::Message) because we would end up with logs and a ton of other message related things with saylinks that have no need to be and adds quite a bit of overhead adding extra saylinks and extra overhead generating them in the process

**Two New Rules**

```cpp
RULE_BOOL(Chat, AutoInjectSaylinksToSay, true, "Automatically injects saylinks into dialogue that has [brackets in them]")
RULE_BOOL(Chat, AutoInjectSaylinksToClientMessage, true, "Automatically injects saylinks into dialogue that has [brackets in them]")
```

**Example Script**

```perl
sub EVENT_SAY {
    quest::say("This is a test that a bracket [without saylinks] can magically now have saylinks [" . quest::saylink("something") . "] [" . quest::saylink("and even more saylinks") . "]");
    $client->Message(15, "This is a client message [bracket1] or [even more brackets] and [" . quest::saylink("an existing saylink") . "]");
}
```

**Result Image**

image

### DIFF
--- a/zone/entity.cpp
+++ b/zone/entity.cpp
@@ -4199,7 +4199,15 @@ void EntityList::QuestJournalledSayClose(Mob *sender, float dist, const char *mo
 	buf.WriteInt32(0); // location, client doesn't seem to do anything with this
 	buf.WriteInt32(0);
 	buf.WriteInt32(0);
-	buf.WriteString(message);
+
+	// auto inject saylinks (say)
+	if (RuleB(Chat, AutoInjectSaylinksToSay)) {
+		std::string new_message = EQ::SayLinkEngine::InjectSaylinksIfNotExist(message);
+		buf.WriteString(new_message.c_str());
+	}
+	else {
+		buf.WriteString(message);
+	}
 
 	auto outapp = new EQApplicationPacket(OP_SpecialMesg, buf);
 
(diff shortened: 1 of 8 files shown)


Kinglykrab (approved):

zone/entity.cpp
```diff
@@ -4199,7 +4199,15 @@ void EntityList::QuestJournalledSayClose(Mob *sender, float dist, const char *mo
 	buf.WriteInt32(0); // location, client doesn't seem to do anything with this
 	buf.WriteInt32(0);
 	buf.WriteInt32(0);
-	buf.WriteString(message);
+
+	// auto inject saylinks (say)
+	if (RuleB(Chat, AutoInjectSaylinksToSay)) {
+		std::string new_message = EQ::SayLinkEngine::InjectSaylinksIfNotExist(message);
+		buf.WriteString(new_message.c_str());
```
mackal, line 4206: `c_str()` is not needed.

splose (approved):

Natedog2012:
Tested and this is great for easy saylinks!

PR #1526, opened 2021-09-08
[Dialogue] Dialogue Window Middleware

Akkadius:
This PR is a completely optional feature that server operators may like to use for a richer dialogue experience

It utilizes the Dialogue Window from https://docs.eqemu.io/quest-api/perl/plugins/diawind

It intercepts the default behavior for `quest::say` and pipes it to a popup dialogue window for richer interaction if server operators choose to do so

**Rules**

```
RULE_BOOL(Chat, QuestDialogueUsesDialogueWindow, false, "Pipes all quest dialogue to dialogue window")
RULE_BOOL(Chat, DialogueWindowAnimatesNPCsIfNoneSet, true, "If there is no animation specified in the dialogue window markdown then it will choose a random greet animation such as wave or salute")
```

* If there are quest saylinks in the quest::say code, the dialogue window render will strip them out so they display normally in the window. 
* If there are multiple saylink responses, they are rendered in the chat window per normal window implementation. Single responses are at the click of a button
* If there is no animation code set in the dialogue window the NPC will animate socially with a handful of different "greet" animations. This can also be shut off behind rule mentioned above

dialogue


common/ruletypes.h
```diff
@@ -593,13 +593,15 @@ RULE_BOOL(Chat, SuppressCommandErrors, false, "Do not suppress command errors by
 RULE_INT(Chat, MinStatusToBypassAntiSpam, 100, "Minimum status to bypass the anti-spam system")
 RULE_INT(Chat, MinimumMessagesPerInterval, 4, "Minimum number of chat messages allowed per interval. The karma value is added to this value")
 RULE_INT(Chat, MaximumMessagesPerInterval, 12, "Maximum value of chat messages allowed per interval")
-RULE_INT(Chat, MaxMessagesBeforeKick, 20, "If an attempt is made to send more than the maximum allowed number of chat messages per interval, the client will be disconnected after this absolute number of messages")
+RULE_INT(Chat, MaxMessagesBeforeKick, 20, "	If an attempt is made to send more than the maximum allowed number of chat messages per interval, the client will be disconnected after this absolute number of messages")
```
Kinglykrab, line 596: Unintentional spacing here?

Kinglykrab (approved):

splose (approved):

PR #1527, opened 2021-09-09
Hack to fix melee chasing fleeing mobs "too far" issues.

noudess:
Just about everyone has seen the "mob is too far away" while chasing fleeing mobs.  Perhaps more so on servers that use a higher percentage for flee ratio.  Upon looking at this, the client updates for client position come in too slowly to accurately compare against a mobs location when fleeing, the earlier a mob flees (and thus the faster the mob is running).

This has been a gripe of melee players for years on my server, and myself.

This patch is a hack, but I don't see a reasonable, or realistic way to fix it any better.  This come from trial and error and looking at the difference in distances that the math is showing, vs. what I see on the screen while chasing.

It dramatically reduces the frustration and looks far more realistic.  Far fewer "you are too far away" despite being right on the target's ass on screen.

The IsBlind() caveat is so that we don't create an exploit where we can hit blinded mobs that don't think they are close enough to turn and engage.

### DIFF
--- a/zone/aggro.cpp
+++ b/zone/aggro.cpp
@@ -851,6 +851,17 @@ bool Mob::CombatRange(Mob* other, float fixed_size_mod, bool aeRampage)
 	size_mod *= RuleR(Combat,HitBoxMod);		// used for testing sizemods on different races.
 	size_mod *= fixed_size_mod;					// used to extend the size_mod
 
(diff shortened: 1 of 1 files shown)


splose (approved):
Looks good to me.. This is something people have wanted fixed for awhile.

Nice to see you back btw!

Kinglykrab (approved):

PR #1529, opened 2021-09-11
[Quest API] Add corpse->GetLootList() and npc->GetLootList() to Perl and Lua.

Kinglykrab:
- Add $corpse->GetLootList() to Perl.
- Add $npc->GetLootList() to Perl.
- Add corpse:GetLootList() to Lua.
- Add npc:GetLootList() to Lua.

Returns an array of item IDs for use with corpse and NPC methods such as HasItem(item_id), CountItem(item_id), and GetFirstSlotByItemID(item_id).

### DIFF
--- a/zone/corpse.cpp
+++ b/zone/corpse.cpp
@@ -1462,12 +1462,12 @@ bool Corpse::HasItem(uint32 item_id) {
 	for (auto current_item  = itemlist.begin(); current_item != itemlist.end(); ++current_item) {
 		ServerLootItem_Struct* loot_item = *current_item;
 		if (!loot_item) {
-			LogError("NPC::CountItem() - ItemList error, null item");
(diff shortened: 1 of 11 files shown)


mackal:
I think the lua stuff should probably match stuff like https://github.com/EQEmu/Server/blob/31ab1d42874fbc39d98c05d5d62096fb0f8610d9/zone/lua_entity_list.cpp#L330

Kinglykrab:
> 
> 
> I think the lua stuff should probably match stuff like
> 
> https://github.com/EQEmu/Server/blob/31ab1d42874fbc39d98c05d5d62096fb0f8610d9/zone/lua_entity_list.cpp#L330

Not a problem, will look at it later today. 👍

mackal (approved):

Akkadius (approved):

PR #1531, opened 2021-09-11
[Quest API] Add EVENT_CONSIDER to Perl and Lua.

Kinglykrab:
- Exports $entity_id in Perl.
- Exports e.entity_id in Lua.

Allows you to perform events on consider for server operators.


mackal (approved):

xackery:
I wonder if it's worthwhile to turn this into a lua hook as well? Namely, being able to hook the native consider message with custom ones and also set that default to false when you don't want original to happen.. Would make custom cons on specific cases possible

PR #1532, opened 2021-09-11
[HP] HP/Stat fixes

KimLS:
So this contains various changes:
- Heroic STA HP calc will now apply after natural durability since this is what the client does.
- The base 5 hp that everyone gets will now apply after natural durability since this is also what the client does.
- Fixed a rounding issue in the mana/endurance calc that would sometimes give the player +1 mana/endurance they shouldn't have.
- Fixed the endurance stats calc to match the client better.
- When a server has a custom stat cap active the new stat caps will be reflected on the client (For non-titanium at least).
    This is done by reserving an AA slot for a "Phantom" AA that will act like Planar Power with the difference between what the client sees and what the server has.
    For example if you have a 1000 cap and a level 80 warrior: normally they have stat cap of 330.  This invisible AA would add the difference of +670 so the client appears to have a cap of 1000.
- AA slots cap now will be 300 (the later clients max) up from 240 (titanium's max).  Any effects that are past the cap will just not be sent for titanium instead.


zone/client_mods.cpp
```diff
@@ -337,7 +340,8 @@ int32 Client::CalcMaxHP()
 		}
 	}
 
-	return max_hp;
+	this->max_hp = std::min(max_hp, 2147483647LL);
```
mackal, line 343: Shouldn't we do something like `std::numeric_limits<decltype(Mob::max_hp)>::max()`

KimLS: We can sure.

KimLS: I changed this in fba333f

Akkadius:
On EZ there's still a pretty large discrepancy between client and server 

image

zone/aa.cpp
```diff
@@ -37,6 +37,12 @@ Copyright (C) 2001-2016 EQEMu Development Team (http://eqemulator.net)
 
 extern QueryServ* QServ;
 
+namespace detail
```
Akkadius, line 40: Wouldn't we want to namespace this `AA` or something similar

zone/aa.cpp
```diff
@@ -933,7 +989,19 @@ void Client::SendAlternateAdvancementPoints() {
...
+	if (RuleI(Character, StatCap) > 0)
+	{
+		aa2->aa_list[i].AA = detail::PhantomStatId;
+		aa2->aa_list[i].value = 1;
+		aa2->aa_list[i].charges = 0;
+		i++;
+	}
+
 	for(auto &aa : zone->aa_abilities) {
+		if (i >= 300) {
```
Akkadius, line 1001: Any reason we aren't just using `MAX_PP_AA_ARRAY` here

zone/aa.cpp
KimLS: I prefer detail for non-public stuff personally.

zone/aa.cpp
KimLS: Agreed

KimLS:
So I figured out the discrepancy with EZ.  With natural durability type effects and HP so high you're causing an int32 overflow in part of the calculation.  I changed the calc serverside to also overflow in the same way and now I'm getting pretty close or exactly close results.

KimLS:
I'm pretty open to suggestions on what the best way forward is.  We could keep the bug in place since it exists in the client.  We could also fix it but the client's HP will still be bugged in that case and not match serverside's HP.

If we keep the bugged hp serverside then that could potentially be a nerf to EZ or any other server that pushes the client beyond it's limits.
If we fix the bug client and server will no longer agree in that situation, the server will always have more HP though.

Akkadius:
> I'm pretty open to suggestions on what the best way forward is. We could keep the bug in place since it exists in the client. We could also fix it but the client's HP will still be bugged in that case and not match serverside's HP.
> 
> If we keep the bugged hp serverside then that could potentially be a nerf to EZ or any other server that pushes the client beyond it's limits.
> If we fix the bug client and server will no longer agree in that situation, the server will always have more HP though.

What are we talking about in terms of discrepancy here? Thousands? Millions? 

If we fix it to match server side how much are we going to be off? 

I can see a world where we gate this behind a rule just to alleviate a major nerf for now and I can see about compensating for the discrepancy in some other means at the content level in EZ

Thanks for taking a look at this

KimLS:
The difference in that screenshot above is nearly 3.5m HP or about 35% of the total.

zone/aa.cpp
```diff
@@ -918,6 +926,54 @@ void Client::SendAlternateAdvancementRank(int aa_id, int level) {
...
+	aai->total_cost = 1;
+	aai->expansion = 0;
+	aai->category = -1;
+	aai->charges = 0;
+	aai->grant_only = 1;
+	aai->total_effects = 7;
+	aai->total_prereqs = 0;
+
+	outapp->SetWritePosition(sizeof(AARankInfo_Struct));
+	for (int i = 0; i < 7; ++i) {
```
Natedog2012, line 966: i = -1; i < 6

zone/aa.cpp
```diff
@@ -918,6 +926,54 @@ void Client::SendAlternateAdvancementRank(int aa_id, int level) {
...
+		outapp->WriteSInt32(SE_RaiseStatCap);
+		outapp->WriteSInt32(diff);
+		outapp->WriteSInt32(0);
+		outapp->WriteSInt32(i + 1);
```
Natedog2012, line 970: outapp->WriteSInt32(SE_RaiseStatCap);
		outapp->WriteSInt32(diff);
		outapp->WriteSInt32(i + 1);
                outapp->WriteSInt32(0);

these are flipped and this corrects it

Natedog2012:
Screenshot_66


With the changes I posted above this is what the #mystats window looks look and the clients HP at level 100

zone/aa.cpp
KimLS: Based on existing AA code what I have is correct the order is:
```
outapp->WriteSInt32(effect.effect_id);
outapp->WriteSInt32(effect.base1);
outapp->WriteSInt32(effect.base2);
outapp->WriteSInt32(effect.slot);
```

zone/aa.cpp
KimLS: It would normally since the client i was using (RoF2) only goes up to 100.  I carefully crafted my base hp to not make it an issue though.

zone/aa.cpp
Natedog2012: Yes and base2 is the "stattype" for that AA..  so the base2 always being 0 will give all the stat raises to "strength" only screenshot below
Screenshot_64



Screenshot_72


Check out the Planar power AA on PeqEditor base2 is the stattype for StatCap

Natedog2012: Now that I think about it, changing it to just 

(i)  and not +1 would be better fix and leaving it ...

0; i < 7

Because stattype starts from 0 anyways

KimLS: Ah I see what you mean now, it isn't so much swapped as it is missing.

Akkadius:
Build failure http://drone.akkadius.com/EQEmu/Server/1698/1/2

KimLS:
I hate that bots are gated by a flag.  Will fix shortly.

Akkadius:
Where are we at on this PR? I know @Natedog2012 was reporting some things and @KimLS had been poking at things but I'm not sure where we left things at

KimLS:
I addressed the issue he brought up.

KimLS:
Weird, will look into that.

Akkadius:
Bumping KLS on this one

Akkadius:
Closing since @KimLS hasn't had time to follow up on this

PR #1534, opened 2021-09-12
[Enhancement] Remove hard coded 200 limit for RoF+ clients merchantlist_temp

noudess:
Temporary merchant lists are limited to 80 pre RoF (per comment in the code).  As of RoF, the client was allowing 200.  I have tested and have found no reasonable boundary to the # of entries.

This code removes the length limitation of what is sent to the client.  The server was already tracking an unlimited list.

The only real limitation is the tinyint(3) type of the field 'slot' in merchantlist_temp.  The same field in merchantlist is int(11).

This PR:

- Removed the 200 limitation of what we can send clients (RoF+)  Pre RoF remains unchanged.
- Raises the datatype for slot in templist to an int(11) to allow for #s greater than 255 (current db storage limit)

### DIFF
--- a/common/version.h
+++ b/common/version.h
@@ -34,7 +34,7 @@
  * Manifest: https://github.com/EQEmu/Server/blob/master/utils/sql/db_update_manifest.txt
  */
 
-#define CURRENT_BINARY_DATABASE_VERSION 9172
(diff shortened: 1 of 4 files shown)


mackal:
This might be writing over memory allocated for other things ...

PR #1538, opened 2021-09-14
[Commands] Overhauled #npcedit.

Kinglykrab:
- Added missing columns like untargetable, show_name, exp_mod, etc.
- Put stats in order of column appearance in table within help message and within code.
- Converted StringFormat to fmt::format.
- Added a GetGenderName() helper method.
- Prettified response messages of nearly every #npcedit option.

All tested and ready to go.

Would like input about possibly changing some of the command arguments to match the table column names more closely, example being "spell" should be "npc_spells_id".

image

### DIFF
--- a/zone/command.cpp
+++ b/zone/command.cpp
@@ -9109,8 +9109,8 @@ void command_npcedit(Client *c, const Seperator *sep)
 		c->Message(Chat::White, "#npcedit special_abilities - Sets an NPC's Special Abilities");
 		c->Message(Chat::White, "#npcedit aggroradius - Sets an NPC's Aggro Radius");
 		c->Message(Chat::White, "#npcedit assistradius - Sets an NPC's Assist Radius");
-		c->Message(Chat::White, "#npcedit featuresave - Saves the NPc's current facial features to the database");
-		c->Message(Chat::White, "#npcedit armortint_id - Set an NPC's Armor Tint ID");
+		c->Message(Chat::White, "#npcedit featuresave - Saves an NPC's current facial features to the database");
+		c->Message(Chat::White, "#npcedit armortint_id - Sets an NPC's Armor Tint ID");
 		c->Message(Chat::White, "#npcedit color [red] [green] [blue] - Sets an NPC's Red, Green, and Blue armor tint");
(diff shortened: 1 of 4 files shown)


joligario (approved):

zone/command.cpp
```diff
@@ -9080,709 +9080,865 @@ void command_npcedit(Client *c, const Seperator *sep)
...
+
+	if (strcasecmp(sep->arg[1], "always_aggro") == 0) {
+        c->Message(Chat::Yellow, fmt::format("NPC ID {} will {} Always Aggro.", npc_id, (atoi(sep->arg[2]) == 1 ? "now" : "no longer")).c_str());
+		std::string query = fmt::format("UPDATE npc_types SET always_aggro = {} WHERE id = {}", atoi(sep->arg[2]), npc_id);
+		content_db.QueryDatabase(query);
+		return;
+	}
+
+	if (strcasecmp(sep->arg[1], "exp_mod") == 0) {
+        c->Message(Chat::Yellow, fmt::format("NPC ID {} now has an Experience Modifier of {}%%.", npc_id, atoi(sep->arg[2])).c_str());
```
Akkadius, line 9884: Some weird indentation going on here and other places above

zone/command.cpp
```diff
@@ -9080,709 +9080,865 @@ void command_npcedit(Client *c, const Seperator *sep)
...
+			target->GetLuclinFace(), target->GetDrakkinHeritage(),
+			target->GetDrakkinTattoo(), target->GetDrakkinDetails(),
+			npc_id
+		);
+		content_db.QueryDatabase(query);
+		return;
+	}
+
+	if (strcasecmp(sep->arg[1], "armortint_id") == 0) {
+        c->Message(Chat::Yellow, fmt::format("NPC ID {} is now using Armor Tint ID {}.", npc_id, atoi(sep->arg[2])).c_str());
```
Akkadius, line 9410: This shouldn't need a `c_str()` but I'm not asking you to change it

Akkadius (approved):
Left a few minor comments, beautiful

zone/command.cpp
Kinglykrab: Not sure why I wouldn't need .c_str(), fmt::format() returns a string and Message needs a const char*.

PR #1546, opened 2021-09-17
[Dialogue] Add support for Dialogue Window buttons.

Kinglykrab:
- Syntax is button_one:button_one_name and button_two:button_two_name, similar to wintype and popupid.
- Add support for negativeid, syntax is negativeid:id. 
- Also changes "mysterious" identifier to "{mysterious}".
- Both button names are required for anything to show up, otherwise it defaults to Yes/No similar to Client::SendFullPopup.

Example:
```pl
sub EVENT_SAY {
    if ($text=~/Hail/i) {
        $client->DialogueWindow("Testing! {mysterious} wintype:1 button_one:Update button_two:Close");
    }
}
```
image

### DIFF
--- a/zone/dialogue_window.cpp
+++ b/zone/dialogue_window.cpp
@@ -182,7 +182,7 @@ void DialogueWindow::Render(Client *c, std::string markdown)
 			auto second_split = split_string(first_split[1], " ");
 			if (!second_split.empty()) {
 				negativeid = second_split[0];
-				LogDiaWindDetail("Client [{}] Rendering negativeid option popupid [{}]", c->GetCleanName(), negativeid);
+				LogDiaWindDetail("Client [{}] Rendering negativeid option negativeid [{}]", c->GetCleanName(), negativeid);
 			}
 
 			if (first_split[1].length() == 1) {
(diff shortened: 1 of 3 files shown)


zone/dialogue_window.cpp
```diff
@@ -98,10 +99,10 @@ void DialogueWindow::Render(Client *c, std::string markdown)
 		}
 	}
 
-	uint32 popup_id       = POPUPID_DIAWIND;
-	uint32 negative_id    = 0;
-	char   *button_name_0 = nullptr;
-	char   *button_name_1 = nullptr;
+	uint32 popup_id = POPUPID_DIAWIND_ONE;
+	uint32 negative_id = POPUPID_DIAWIND_TWO;
+	const char *button_one_name = nullptr;
```
Akkadius, line 104: Any reason we didn't just make these `std::string`'s and then translate them to `c_str` primitives later?

Akkadius:
This is sweet! Nice extension of what's there

Made a minor comment around type usage

Hoping you tested all of the previous cases that used to work so we don't have any regressions around responses and response types

Akkadius (approved):

zone/dialogue_window.cpp
```diff
@@ -171,6 +172,96 @@ void DialogueWindow::Render(Client *c, std::string markdown)
 		}
 	}
 
+	// negativeid
+	std::string negativeid;
+	if (markdown.find("negativeid") != std::string::npos) {
```
Akkadius, line 177: What is this for ?

PR #1553, opened 2021-09-22
[Tradeskill] Fix logic in taught tradeskill recipes

Natedog2012:


zone/tradeskills.cpp
```diff
@@ -810,6 +810,14 @@ void Client::SendTradeskillSearchResults(
 				continue;
 			}
 		}
+		
+		//Check if we need to learn it before sending them the recipe.. 
+		DBTradeskillRecipe_Struct spec;
+		if (content_db.GetTradeRecipe(recipe_id, objtype, someid, this->CharacterID(), &spec)) {
```
Akkadius, line 816: Doing a select query inside of a loop like this is not ideal. We should fetch recipes once and then loop through it in memory to perform this check

Natedog2012: Righto I see a better way to fix it.. but looking at it.. I see other issues

Kinglykrab (approved):
Looks good to me. Can’t believe we never noticed that issue with row data.

Akkadius (approved):

Natedog2012:
> e

Makes me wonder if those checks... row[5] == nullptr even work :D

PR #1563, opened 2021-09-30
[Dialogue] Add support for Dialogue Window titles.

Kinglykrab:
- Syntax is `title:popup_title`
- Custom title allows defaults to be overridden where necessary, like a leaderboard or something.
- Default target to client in case people want to send Dialogue Windows from current client.
- Fix possible issue with markdown.
- Example: Using the word "title" or using any identifier and forgetting the colon.


hgtw (approved):

Akkadius (approved):
Let's make sure to update our docs with all the new markdown

PR #1564, opened 2021-10-01
[Shared Tasks] Always update group members for task kill updates

hgtw:
This reverts the commit that added HandleUpdateTasksOnKill

Solo task updates for group members were being ignored if the client
that killed the npc did not have the same task or had already finished
the kill objective. This was a regression caused by changes to prevent
shared tasks from updating multiple times for every party member.

This changes it so solo task updates for party members are no longer
dependent on the task state of the client that killed the npc. Shared
tasks will still only be explicitly updated once for the client that
kills the npc.


Kinglykrab (approved):

Natedog2012 (approved):
Fixes crash from login kills and regular tasks are working as intended.

PR #1566, opened 2021-10-01
[Quest API] Add client->ReadBookByName(book_name, book_type) to Perl/Lua.

Kinglykrab:
- Add $client->ReadBookByName(book_name, book_type) to Perl.
- Add client:ReadBookByName(book_name, book_type) to Lua.
- Allows server operators to put books in to their database and read from their database instead of storing the values in a script, also allows them to read pre-existing books using a script.

### DIFF
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -10537,9 +10537,7 @@ void Client::ReadBookByName(std::string book_name, uint8 book_type)
 	int length = book_text.length();
 
 	if (book_text[0] != '\0') {
-#if EQDEBUG >= 6
-		LogInfo("Client::ReadBookByName() Book Name: [{}] Text: [{}]", book_name, book_text.c_str());
-#endif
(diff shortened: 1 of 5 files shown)


hgtw (approved):

zone/client.cpp
```diff
@@ -10529,3 +10529,32 @@ void Client::SetDoorToolEntityId(uint16 door_tool_entity_id)
...
 }
+
+void Client::ReadBookByName(std::string book_name, uint8 book_type)
+{
+	int16 book_language = 0;
+	std::string book_text = content_db.GetBook(book_name.c_str(), &book_language);
+	int length = book_text.length();
+
+	if (book_text[0] != '\0') {
+#if EQDEBUG >= 6
```
Akkadius, line 10540: We don't need these preprocessors anymore; just use the logging macros

Akkadius (commented):
LGTM; minor comment

Akkadius (approved):

PR #1570, opened 2021-10-01
[Quest API] Add IsNPCSpawned(npc_ids) and CountSpawnedNPCs(npc_ids) to Perl/Lua.

Kinglykrab:
- Add quest::isnpcspawned(npc_ids) to Perl.
- Add quest::countspawnednpcs(npc_ids) to Perl.
- Add eq.is_npc_spawned(npc_ids) to Lua.
- Add eq.count_spawned_npcs(npc_ids) to Lua.

### DIFF
--- a/zone/entity.cpp
+++ b/zone/entity.cpp
@@ -1202,25 +1202,7 @@ bool EntityList::IsMobSpawnedByNpcTypeID(uint32 get_id)
 
 bool EntityList::IsNPCSpawned(std::vector<uint32> npc_ids)
 {
-	bool is_spawned = false;
-	if (npc_list.empty() || npc_ids.empty()) {
-		return is_spawned;
-	}
-
-	for (auto current_npc : npc_list) {
-		if (
-			std::find(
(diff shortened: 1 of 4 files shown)


zone/entity.cpp
```diff
@@ -1200,6 +1200,52 @@ bool EntityList::IsMobSpawnedByNpcTypeID(uint32 get_id)
...
+	for (auto current_npc : npc_list) {
+		if (
+			std::find(
+				npc_ids.begin(),
+				npc_ids.end(),
+				current_npc.second->GetNPCTypeID()
+			) != npc_ids.end() && 
+			current_npc.second->GetID() != 0
+			) {
+			is_spawned = true;
```
hgtw, line 1219: Can just break early here on the first match

hgtw: Alternatively to avoid code reuse it could just call the function below `return CountSpawnedNPCs() != 0` I think

hgtw (approved):

PR #1573, opened 2021-10-02
[Shared Tasks] Task Kill Update Fix

Akkadius:
This is a slight adjustment to the task kill update logic to not have the receiver of an update absolutely need the context of a task and activity to allow the update to propagate to other clients

Related to https://github.com/EQEmu/Server/pull/1564 (Can be closed)

This keeps logic aggregate in preparation for other work with smart goal lists to be implemented

Tested and confirmed working


zone/client.h
```diff
@@ -1039,6 +1039,7 @@ class Client : public Mob
 	void SendTaskRequestCooldownTimerMessage();
 	void StartTaskRequestCooldownTimer();
 	inline ClientTaskState *GetTaskState() const { return task_state; }
+	inline bool HasTaskState() { if (task_state) { return true; } return false; }
```
hgtw, line 1042: can just `return task_state != nullptr` here

zone/task_manager.cpp
```diff
@@ -1814,3 +1814,96 @@ void TaskManager::SyncClientSharedTaskStateToLocal(
...
+			}
+
+			// Check if there are any active kill activities for this p_task_data
+			auto p_task_data = m_task_data[current_task->task_id];
+			if (p_task_data == nullptr) {
+				return;
+			}
+
+			for (int activity_id = 0; activity_id < p_task_data->activity_count; activity_id++) {
+				ClientActivityInformation *client_activity = &current_task->activity[activity_id];
```
hgtw, line 1841: We should really combine all the update methods that re-use these activity checks at some point, I think they only differ in specifying the TaskActivityType and maybe some goal stuff, but it's duplicated across like 6 methods

hgtw (approved):

zone/task_manager.cpp
Akkadius: It was actually substantially worse at one point. There's a lot I'd like to change about the legacy task system structure

PR #1577, opened 2021-10-02
[Quest API] Add GetIPExemption(), GetIPString(), and SetIPExemption(exemption_amount) to Perl/Lua.

Kinglykrab:
- Add $client->GetIPExemption() to Perl.
- Add $client->GetIPString() to Perl.
- Add $client->SetIPExemption(exemption_amount) to Perl.
- Add client:GetIPExemption() to Lua.
- Add client:GetIPString() to Lua.
- Add client:SetIPExemption(exemption_amount) to Lua.

Will make plugin::IP unnecessary and allow people to get readable IP string easier, as well as set/get IP exemptions from Perl and Lua.

### DIFF
--- a/common/database.cpp
+++ b/common/database.cpp
@@ -2271,6 +2271,35 @@ int Database::GetIPExemption(std::string account_ip) {
 	return RuleI(World, MaxClientsPerIP);
 }
 
+void Database::SetIPExemption(std::string account_ip, int exemption_amount) {
+	std::string query = fmt::format(
+		"SELECT `exemption_id` FROM `ip_exemptions` WHERE `exemption_ip` = '{}'",
+		account_ip
+	);
+
+	auto results = QueryDatabase(query);
+	uint32 exemption_id = 0;
+	if (results.Success() && results.RowCount() > 0) {
+		auto row = results.begin();
+		exemption_id = atoi(row[0]);
+	}
+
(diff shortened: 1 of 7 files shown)


common/database.cpp
```diff
@@ -2271,6 +2271,35 @@ int Database::GetIPExemption(std::string account_ip) {
...
+		exemption_id = atoi(row[0]);
+	}
+
+	if (exemption_id != 0) {
+		query = fmt::format(
+			"UPDATE `ip_exemptions` SET `exemption_amount` = {} WHERE `exemption_ip` = '{}'",
+			exemption_amount,
+			account_ip
+		);
+	} else {
```
Akkadius, line 2293: Minor stylistic feedback is that the else can be defined before the if block and the else block can be tossed. Eg.

```cpp
query = fmt::format(
	"INSERT INTO `ip_exemptions` (`exemption_ip`, `exemption_amount`) VALUES ('{}', {})",
	account_ip,
	exemption_amount
);

if (exemption_id != 0) {
	query = fmt::format(
		"UPDATE `ip_exemptions` SET `exemption_amount` = {} WHERE `exemption_ip` = '{}'",
		exemption_amount,
		account_ip
	);
}
```

Akkadius (approved):

PR #1590, opened 2021-10-05
[Spells] Update SPA158 Reflect

KayenEQ:
Updated 'Reflect' spell effect to use current spell files additional modifiers.

**Limit** value is a resist modifier, where positive values works to decrease resist chance (further decreases any resist modifiers, ie value of 100 on a spell with a -50 resist modifier, will make it a -150 resist modifer)

**Max** value is the percent of base damage that is reflected back at caster (50 = 50pct of damage), if set to 0 will do full damage.

Note: Spell table as a field called 'Reflectable' this must be true for a spell to reflect.

**Mechanics** updated, when reflecting back damage spells the actual damage reflected is BASE damage of the spell, this is
the lowest possible value before any formula or focus effects are applied. No focus at all are applied to damage spells. Also note we have am old RULE that regulate what type of targets/spells reflect can be applied to do.

Spells:ReflectType Reflect type. 0=disabled, 1=single target player spells only, 2=all player spells, 3=all single target spells, 4=all spells. [DEFAULT IS 3]

Based on live testing, AOE spells that have the spell file field 'Reflectable' as true, are able to be reflected. Thus if you want a live like reflect mechanic change the the rule to 4.

Support now for live like Reflect AA's

When a spell is a cast on a player with multiple reflect abilities. It will check spells buffs first, and choose one with highest chance. If that fails will check AA's and then for items effects.

```
	/*
		Reflect
		base= % Chance to Reflect 
		Limit= Resist Modifier (+Value for decrease chance to resist) 
		Max= % of base spell damage (this is the base before any formula or focus is applied)
		On live any type of detrimental spell can be reflected as long as the Reflectable spell field is set, this includes AOE.
		The 'caster' of the reflected spell is owner of the reflect effect. Caster's focus effects are NOT applied to reflected spell.
		
		reflect_effectiveness is applied to damage spells, a value of 100 is no change to base damage. Other values change by percent. (50=50% of damage)
		we this variable to both check if a spell being applied is from a reflection and for the damage modifier.
		
		There are a few spells in database that are not detrimental that have Reflectable field set, however from testing, they do not actually reflect.
	*/
```

SE_Reflect 158 reflect casted detrimental spell back at caster, base: chance pct, limit: resist modifier (positive value reduces resists), max: pct of base dmg mod (50=50pct of base)

### DIFF
--- a/common/spdat.h
+++ b/common/spdat.h
@@ -511,6 +511,15 @@ enum NegateSpellEffectType
 	NEGATE_SPA_SPELLBONUS_AND_AABONUS     = 5,
 	NEGATE_SPA_ITEMBONUS_AND_AABONUS      = 6,
 };
+//Used for rule RuleI(Spells, ReflectType))
+enum ReflectSpellType
+{
+	REFLECT_DISABLED                  = 0,
+	REFLECT_SINGLE_TARGET_SPELLS_ONLY = 1,
+	REFLECT_ALL_PLAYER_SPELLS         = 2,
+	RELFECT_ALL_SINGLE_TARGET_SPELLS  = 3,
+	REFLECT_ALL_SPELLS                = 4,
+};
 
 enum SpellTypes : uint32
 {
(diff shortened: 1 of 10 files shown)


mackal (approved):

Kinglykrab (approved):

KayenEQ:
Figured it out. Turns out the item stat Extra Spell Damage on the person reflecting is applied to the reflected damage on the target.
Also updated a portion of the Extra Spell Damage formula to reflect new information gained while testing this.

PR #1592, opened 2021-10-05
[Spells] Update to SPA 157 Spell Damage Shield

KayenEQ:
Updates to Spell Damage Shield effect.

Reminder of functionality, if you cast a spell at a target with this effect, the caster will receive a flat amount of non-melee damage specified by the base value of SPA 157.

Spell Damage Shield can now be applied from items and AAs

Implemented spells_new 'field160' using variable 'feedbackable' this field determines if a spell can trigger a spell DS. The source code now checks against this field to determine if a spell DS can be triggered from the given spell.

Fixed a random bug that was causing NPC's shielding stat from items to be checking for SpellDamageShield instead of SpellShield. I assume this was causing NPC's not to properly get SpellShield bonuses.

SE_SpellDamageShield	157	 causes non-melee damage on caster of a spell, base: Amt DS (negative), limit: none, max: unknown (same as base but +)

### DIFF
--- a/common/spdat.cpp
+++ b/common/spdat.cpp
@@ -1427,6 +1427,7 @@ int GetSpellStatValue(uint32 spell_id, const char* stat_identifier, uint8 slot)
 	else if (id == "descnum") { return spells[spell_id].descnum; }
 	else if (id == "effectdescnum") { return spells[spell_id].effectdescnum; }
 	else if (id == "npc_no_los") { return spells[spell_id].npc_no_los; }
+	else if (id == "feedbackable") { return spells[spell_id].reflectable; }
 	else if (id == "reflectable") { return spells[spell_id].reflectable; }
 	else if (id == "bonushate") { return spells[spell_id].bonushate; }
 	else if (id == "endurcost") { return spells[spell_id].EndurCost; }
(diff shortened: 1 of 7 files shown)


common/spdat.cpp
```diff
@@ -1427,6 +1427,7 @@ int GetSpellStatValue(uint32 spell_id, const char* stat_identifier, uint8 slot)
 	else if (id == "descnum") { return spells[spell_id].descnum; }
 	else if (id == "effectdescnum") { return spells[spell_id].effectdescnum; }
 	else if (id == "npc_no_los") { return spells[spell_id].npc_no_los; }
+	else if (id == "feedbackable") { return spells[spell_id].reflectable; }
```
Kinglykrab, line 1430: Is this intended to be `spells[spell_id].reflectable` here?

Kinglykrab (approved):
Looks good other than one minor comment.

Edit: Seem to be missing SQL update to change column name in spells_new table?

KayenEQ:
I have avoided updating the field names in the table for a bit now cause unsure if they will break any ones editing tools. There are a lot of fields that need to names updated.

common/spdat.cpp
KayenEQ: resolved

PR #1598, opened 2021-10-09
[Bug Fix] Fix Character Recast Type -1 saving to database.

Kinglykrab:
Not sure if this is used for anything else, but this will at least stop the errors with items with a recast type of -1.

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -2533,7 +2533,7 @@ bool Mob::SpellFinished(uint16 spell_id, Mob *spell_target, CastingSlot slot, ui
 				database.UpdateItemRecastTimestamps(
 				    CastToClient()->CharacterID(),
 					recast_type,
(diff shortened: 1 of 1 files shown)


KayenEQ:
I recall I always used -1 when I wanted unlimited recasts

Kinglykrab:
> I recall I always used -1 when I wanted unlimited recasts without a timer

So would the best way be to just not send the value to the database at all if it’s -1 and keep the column as unsigned but the function as signed?

mackal:
> > I recall I always used -1 when I wanted unlimited recasts without a timer
> 
> So would the best way be to just not send the value to the database at all if it’s -1 and keep the column as unsigned but the function as signed?

no, it's best to just not use -1, this is an actual thing on live (I'm not sure how it works sadly) that does work differently than a positive number.

Akkadius:
Yeah we should just not save it to the database if it is -1

PR #1602, opened 2021-10-14
[Bug Fix] Fix Dialogue Window Custom Titles.

Kinglykrab:
Titles were limited to single words due to the way the parsing works.

Titles will now be handled by the following syntax: `*Title*`


Akkadius:
Lets sync up on this one this weekend

Issue #1461, opened 2021-07-19
Summoned and charmed pets lose their innate magic attack ability when equipping non magical weapons


MidnightMusicClub:
Rusty, fine steel, etc weapons remove a summoned or charmed pet's ability to use its innate magical attack. If you charm an NPC that is holding a non magical weapon that NPC will not be able to hit other NPCs with the hand the weapon is equipped in.

mackal:
Just so I don't have to dig through the code, can you clarify what should be live like behavior and what our current behavior is?

MidnightMusicClub:
Live like would retain the pet's ability to damage NPCs that require a magic attack no matter what is in their hands. The way it is now handing unenchanted weapons to pets makes them unable to damage other NPCs that require magic attacks. It seems NPCs are using the same code for this functionality as PCs. There should at least be a rule that allows the innate magic attack to pass thru unenchanted held items.

MidnightMusicClub:
Anyone else done any testing on this one?

MidnightMusicClub:
This is on live, Erollisi/Nameless server. Charmed undead frog with FS hammer, crushing another undead frog. The way it is now on emu, the primary hand would be getting invulnerable messages because of the FS hammer but as long as the offhand did not have another item in it, it would still be able to hit the mob. If I was to give him another weapon he would no longer be able to do damage to the other undead frogs. https://gyazo.com/1c2b82dc82ed68254f176f0ddca2f2ea

PR #1606, opened 2021-10-15
[Bug Fix] Fixes IMMUNE MELEE NONMAGICAL logic

KayenEQ:
Addresses: https://github.com/EQEmu/Server/issues/1461

Fixes issues with pets losing innate ability to hit IMMUNE MELEE NONMAGICAL flagged NPC's when given non magical weapons. (This doesn't happen on live)

Fixed bug where special attacks like BASH or KICK would ignore IMMUNE MELEE NONMAGICAL and other potential immune states.

Updated logic, any NPC flagged with a special attack 'SPECATK_MAGICAL' that allows them to hit IMMUNE MELEE NONMAGICAL will be checked first thus never will fail the check regardless of weapons or any other rules.

Note Live testing showed PETS and Charmed pets gain ability to hit IMMUNE NPC's when the pet level is >= Level 10
RuleI(Combat, PetAttackMagicLevel) this is an OLD rule and its default has been set to 30 (unknown if this used to be lives level).

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -990,32 +990,31 @@ int Mob::GetWeaponDamage(Mob *against, const EQ::ItemData *weapon_item) {
 
 	//check to see if our weapons or fists are magical.
 	if (against->GetSpecialAbility(IMMUNE_MELEE_NONMAGICAL)) {
-		if (weapon_item) {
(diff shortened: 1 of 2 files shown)


noudess:
Love this one.  Have seen this type of issue alot.

PR #1608, opened 2021-10-15
[Quest API] Add EVENT_LOOT_ZONE to zone_controller

Natedog2012:
One stop shop for loot checking across zone


Natedog2012:
> This looks useful.
> 
> Can we export client entity ID as well?

Is already exports $client to this for perl and e.other for lua

Kinglykrab:
Been testing this event, seems like it'd be more useful to have $corpse_id instead of $corpse which is the name of the corpse minus numbers. This would allow operators to get the corpse looted from by this ID. Could also change EVENT_LOOT to do this as well.

PR #1616, opened 2021-10-18
[Combat] Updates to IMMUNE_MELEE_NONMAGICAL mechanics

KayenEQ:
Upon further review, it appears all NPC's on live level 10 or above, are given the ability to hit other NPC's that are flagged IMMUNE_MELEE_NONMAGICAL.

Live examples below uses Wisps in North Karana which are flagged IMMUNE_MELEE_NONMAGICAL
Example: Lv 9 Wisp  vs Charmed Lv 9 Wisp = Neither can hit each other both immune
Example: Lv 10 Wisp  vs Charmed Lv 9 Wisp = NPC can hit charmed pet, Pet can not hit NPC
Example: Lv 10 Wisp  vs Charmed Lv 10 Wisp = NPC can hit charmed pet, Pet can  hit NPC
Example: Lv 9 Wolf vs Charmed Lv 9 Wisp = Wolf can not hit Wisp, Wisp pet can hit wolf
Example: Lv 12 Lion vs Charmed Lv 9 Wisp = Lion can hit the Wisp, Wisp can hit lion

Following rule is important to implement because otherwise you can Charm a pet that has IMMUNE_MELEE_NONMAGICAL
and it would be immune to melee from other NPC's. Now with below rule as long as they are above required level, the NPC's can fight back as they do on live.

Added: RuleI(Combat, NPCAttackMagicLevel)  This gives all NPC's (including PETS) over the level ability to attack targets that are flagged with IMMUNE_MELEE_NONMAGICAL.  This is set to level 10 to be live like

For live like effect both these rules should be same level.

Also improved logic for applying ability to hit IMMUNE_MELEE_NONMAGICAL from weapons and armor properties to remove some loop holes that I found occurring from non weapons allowing NPC's to bypass restriction unintendedly. Confirmed on live for NPC's only actual weapons in primary or second will give them ability to hit.


KayenEQ:
Changed it to only use one rule which will cover both NPC's and Pets.

PR #1644, opened 2021-10-25
[Saylinks] In-Memory Saylink Lookups

Akkadius:
We've been using saylinks forever. Every single time a saylink appears in a message its a database hit. This adds up substantially especially with heavy saylink usage and even more with automatic saylink injection.

This change keeps a static global bucket to store cached saylinks and loads them on zone bootup. This works fine for now; I'd like to see this get centralized to an instance and store it there rather than using a global but the saylink class is almost purely static right now

* When a saylink is not in the database, its stored in the database and cached
* When a saylink is in the database but not in the cache, its fetched from the database and then cached in the zone (Think a saylink was created in another zone)
* When zone boots, saylinks are cached into memory from the database (# commands are ignored because there is a lot of noise and not often used links)

Lookups subsequently are insanely fast and no longer cause the slight "chug" effect when displaying a bunch of saylink heavy messages as long as the content is static

### DIFF
--- a/common/say_link.cpp
+++ b/common/say_link.cpp
@@ -467,7 +467,7 @@ std::string EQ::SayLinkEngine::InjectSaylinksIfNotExist(const char *message)
 
 void EQ::SayLinkEngine::LoadCachedSaylinks()
 {
-	auto saylinks = SaylinkRepository::All(database);
+	auto saylinks = SaylinkRepository::GetWhere(database, "phrase not like '%#%'");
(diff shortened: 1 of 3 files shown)


Akkadius:
> This can also be updated to do id lookups via the cache since every client saylink click is a query
> 
> https://github.com/EQEmu/Server/blob/ef5124d7564511a6585df4635712cc29e8251341/zone/client_packet.cpp#L8521

Certainly could; going to leave for another PR though

PR #1653, opened 2021-10-30
[Loginserver] Code Cleanup and Tweaks

Akkadius:
**Changes**

* [Code Cleanup] Change member variables to be `m_` prefixed
* [Code Cleanup] Remove licensing information from loginserver related files
* Add human readable addresses to client login log statements
* Sanitize world server names so only a specific allowed set of characters can show up in the server list
* Force legacy mode for `login.eqemulator.net` to avoid user mistakes
* Add more logging information for World when Loginserver::SendInfo is triggered
* Set a remote address for World if it is not sent to the Loginserver based on a non-empty local address field
* Key worldserver lookups by both long name and short name instead of just short name. Probably move this to closer to memory lookups later
* Move validation constants
* Set servers to be authorized in all unregistered allowed validation cases
* Fix `server_short_name` to response with short name instead of long name in `/v1/servers/list` GET endpoint
* Add `world_id` to `/v1/servers/list` endpoint

### DIFF
--- a/loginserver/database.cpp
+++ b/loginserver/database.cpp
@@ -25,11 +25,6 @@ Database::Database(
 	std::string name
 )
 {
-	user = user;
-	pass = pass;
-	host = host;
-	name = name;
-
 	uint32 errnum = 0;
 	char   errbuf[MYSQL_ERRMSG_SIZE];
 	if (!Open(
@@ -55,8 +50,8 @@ Database::Database(
  */
 Database::~Database()
 {
-	if (database) {
-		mysql_close(database);
+	if (m_database) {
+		mysql_close(m_database);
 	}
 }
 
--- a/loginserver/database.cpp
+++ b/loginserver/database.cpp
@@ -363,40 +363,40 @@ Database::DbWorldRegistration Database::GetWorldRegistration(
 		login_world_server_admin_id
 	);
 
-	Database::DbWorldRegistration world_registration{};
+	Database::DbWorldRegistration r{};
 
 	auto results = QueryDatabase(query);
 	if (!results.Success() || results.RowCount() != 1) {
-		return world_registration;
+		return r;
 	}
 
 	auto row = results.begin();
 
-	world_registration.loaded                  = true;
-	world_registration.server_id               = std::stoi(row[0]);
-	world_registration.server_description      = row[1];
-	world_registration.server_list_type        = std::stoi(row[3]);
-	world_registration.is_server_trusted       = std::stoi(row[2]) > 0;
-	world_registration.server_list_description = row[4];
-	world_registration.server_admin_id         = std::stoi(row[5]);
+	r.loaded                  = true;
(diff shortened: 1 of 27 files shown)


loginserver/database.cpp
```diff
@@ -45,10 +25,10 @@ Database::Database(
...
 )
 {
-	this->user = user;
-	this->pass = pass;
-	this->host = host;
-	this->name = name;
+	user = user;
+	pass = pass;
+	host = host;
+	name = name;
```
hgtw, line 31: These members need renamed, otherwise it's just local self assignment

hgtw: Alternatively they could be put in the constructor initializer list list and left not-renamed but I think renaming is more clear

Akkadius: Good eye; these members don't do anything anyways so I will remove them.

There's a few other places in the loginserver code where member variables are used to pass data around and its bizarre

loginserver/world_server.cpp
```diff
@@ -670,41 +672,32 @@ bool WorldServer::HandleNewLoginserverInfoValidation(
...
 	}
-	else if (strlen(new_world_server_info_packet->server_short_name) >= max_server_short_name_length) {
-		LogError("Handle_NewLSInfo error [server_short_name] was too long | max [{0}]", max_server_short_name_length);
+	else if (strlen(new_world_server_info_packet->server_short_name) >= MAX_SERVER_SHORT_NAME_LENGTH) {
+		LogError("Handle_NewLSInfo error [server_short_name] was too long | max [{0}]", MAX_SERVER_SHORT_NAME_LENGTH);
 		return false;
 	}
-	else if (strlen(new_world_server_info_packet->server_version) >= max_server_short_name_length) {
-		LogError("Handle_NewLSInfo error [server_version] was too long | max [{0}]", max_server_version_length);
+	else if (strlen(new_world_server_info_packet->server_version) >= MAX_SERVER_SHORT_NAME_LENGTH) {
```
hgtw, line 691: Shouldn't this be `>= MAX_SERVER_VERSION_LENGTH`

Akkadius: You have eyes of a hawk good sir; fixed

world/login_server.cpp
```diff
@@ -366,15 +366,15 @@ bool LoginServer::Connect()
 				}
 				else {
 					LogInfo(
-						"Could not connect to Legacy Loginserver: [{0}:{1}]",
-						LoginServerAddress,
-						LoginServerPort
+					"Could not connect to Legacy Loginserver: [{0}:{1}]",
+					m_loginserver_address,
+					m_loginserver_port
```
hgtw, line 371: Unintentional loss of indent here I think

Akkadius: Fixed

hgtw (approved):

PR #1666, opened 2021-11-02
[Spells] Update to Charm target restriction code

KayenEQ:
When casting charm the check for if charm could be applied for specific target restriction was not being done at the correct location in the code. This was causing issues due to having to remove the buff after it was applied if target checks failed.

As verified on live,  it should be checking restrictions as soon as you start casting and then it quickly cancels the cast.

Also, updated the fail messages to reflect live messages.

*This does not solve the other charm bug involving targeting.

### DIFF
--- a/zone/spell_effects.cpp
+++ b/zone/spell_effects.cpp
@@ -8428,17 +8428,21 @@ bool Mob::PassCharmTargetRestriction(Mob *target) {
 	
 	if (target->IsClient() && IsClient()) {
 		MessageString(Chat::Red, CANNOT_AFFECT_PC);
+		LogSpells("Spell casting canceled: Can not cast charm on a client.");
 		return false;
 	}
 	else if (target->IsCorpse()) {
+		LogSpells("Spell casting canceled: Can not cast charm on a corpse.");
 		return false;
 	}
 	else if (GetPet() && IsClient()) {
 		MessageString(Chat::Red, ONLY_ONE_PET);
+		LogSpells("Spell casting canceled: Can not cast charm if you have a pet.");
 		return false;
 	}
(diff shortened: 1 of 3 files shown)


zone/spell_effects.cpp
```diff
@@ -8440,6 +8419,31 @@ bool Mob::HarmonySpellLevelCheck(int32 spell_id, Mob *target)
 	return true;
 }
 
+bool Mob::PassCharmTargetRestriction(Mob *target) {
```
mackal, line 8422: Can we add logs here?

zone/spells.cpp
```diff
@@ -229,6 +229,16 @@ bool Mob::CastSpell(uint16 spell_id, uint16 target_id, CastingSlot slot,
 		return false;
 	}
 
+	if (IsEffectInSpell(spell_id, SE_Charm) && !PassCharmTargetRestriction(entity_list.GetMobID(target_id))) {
+		if (IsClient()) {
+			CastToClient()->SendSpellBarEnable(spell_id);
```
mackal, line 234: We don't need to cast to client here. We should also make sure the charm is a casted spell before calling SendSpellBarEnable.

zone/spell_effects.cpp
KayenEQ: What would you like in your logs?

mackal: at least something to indicate we failed because of this check, could help with potential bug reports (people can miss messages ..)

KayenEQ: Done

Kinglykrab (approved):

PR #1671, opened 2021-11-04
[Bug Fix] Fix for dual wield animation when same delay weapons.

KayenEQ:
Resolves an issue where if you start auto attacking with two weapons with the same delay the offhand animation will never fire due to the throttle timer, the primary would always go first. This solution checks when the timers are set if both attack and dw timers are the same, if so it will ensure an equal chance of primary and secondary animations.

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -138,17 +138,23 @@ EQ::skills::SkillType Mob::AttackAnimation(int Hand, const EQ::ItemInstance* wea
 		skillinuse = EQ::skills::SkillHandtoHand;
 		type = animHand2Hand;
 	}
-
+	Shout("PRE Hand %i Type %i", Hand, type);
 	// If we're attacking with the secondary hand, play the dual wield anim
 	if (Hand == EQ::invslot::slotSecondary) {// DW anim
 		type = animDualWield;
+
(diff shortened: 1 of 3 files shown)


zone/attack.cpp
```diff
@@ -5477,6 +5489,21 @@ void Client::SetAttackTimer()
...
+			primary_speed = speed;
+		}
+		else if (i == EQ::invslot::slotSecondary) {
+			secondary_speed = speed;
+		}
+	}
+
+	//To allow for duel wield animation to display correctly if both weapons have same delay
+	if (primary_speed == secondary_speed) {
+		SetDualWeildingSameDelayWeapons(1);
```
Akkadius, line 5503: Weild vs Wield

Akkadius:
I'll look at this more later, my eyes are going cross eyed :sleeping:

zone/attack.cpp
KayenEQ: I REFUSE to learn how to spell DUELWEILD

Natedog2012 (approved):
The visuals for dual wield working much better when both weapons are the same speed

mackal (approved):

mackal:
I feel like there has to be a better way to do this, but oh well. As long as this improves things it should be fine.

PR #1678, opened 2021-11-05
Client Checksum Verification

cybernine186:
Tested and implemented for RoF2 client, untested with other clients but if the correct OP codes are identified it should work. Files that the RoF2 client performs a crc64 checksum on is eqgame.exe, SkillCaps.txt, and BaseData.txt. Requires updating the patch_RoF2.conf on your server side because of a new identified OPCode and previously wrong OPCodes was listed in the patch_RoF2.conf. Potentially still missing an OPCode for spells_us.txt, or I just overlooked it. The OPCode packet size is 2056 exactly for each of the packets and the first 8 bytes is the CRC64 checksum value.

How it works
-------------
Once the client connects to the eqemu server it sends several OP codes that contain a crc64 checksum value along with other random data inside the corresponding file for additional verification. The server will check the checksum values against what the server allows and if enabled forces the client to use the server provided files or prevents access to login to the account.

How to use
-------------
Login and use GM command #updatechecksum, this will store the CRC64 values of the client the GM has logged in with. Enable rule World:EnableChecksumVerification, by default its set to false to enforce server side client verification. For GM's by default with account status of >= 100 will bypass the checksum authentication, if you wish to modify the status refer to rule GM:MinStatusToBypassCheckSumVerification.

New Server Rules
-----------------
World:EnableChecksumVerification, default false
GM:MinStatusToBypassCheckSumVerification, default 100

SQL Changes
-------------
Additional 3 fields to `accounts` table: checksum_crc1_eqgame, checksum_crc2_skillcaps, and checksum_crc3_basedata.
New records in the `variables` table: checksum_crc1_eqgame, checksum_crc2_skillcaps, and checksum_crc3_basedata.
Source 2021_11_05_checksum_verification.sql to add these fields to your database.

GM Commands
-------------
New GM command: #updatechecksum
This command takes the CRC64 checksum values stored in the GM's `account` table and store those values into the `variables` table, checksum_crc1_eqgame, checksum_crc2_skillcaps, and checksum_crc3_basedata.


mackal:
The new rules related to game play should be pulled out into their own PR.

cybernine186:
I didn't mean for that other commit to be included in this pull request.  Would you rather I just close this request and resubmit?

Kinglykrab:
> I didn't mean for that other commit to be included in this pull request. Would you rather I just close this request and resubmit?

Yes.

PR #1679, opened 2021-11-05
[Rules] Gate /tgb, /autofire and /melody

cybernine186:
Created new rules to negate server and client side effects of commands: /tgb, /autofire, and /melody. These commands are enabled by default and can be disabled to enforce a classic EQ experience if using progression style play for example.

Rules
--------------
RULE_BOOL(Character, EnableBardMelody, true, "Enable Bard /melody by default, to disable change to false for a classic experience.")
RULE_BOOL(Character, EnableRangerAutoFire, true, "Enable Ranger /autofire by default, to disable change to false for a classic experience.")
RULE_BOOL(Character, EnableTGB, true, "Enable /tgb (Target Group Buff) by default, to disable change to false for a classic experience.")


mackal:
You don't need to include the SQL for rules, the server will generate them when new ones are added.

cybernine186:
SQL has been removed.

mackal (approved):

PR #1682, opened 2021-11-06
[Bug Fix] SOF+ clients item click recast timer not met check

KayenEQ:
SOF+ client's were not checking Item Click Recast timers until after the click spell had completed casting. It should be checked upon initial click and not start casting if the recast timer is active. Titanium does this by default with no code checks needed., though we enforce it server side at end of cast for Titanium. Keeping the end of cast check for both, Titanium doesn't play as nice when I tried to move it up to be done on click.

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -8959,7 +8959,7 @@ void Client::Handle_OP_ItemVerifyRequest(const EQApplicationPacket *app)
 				}
 				if (GetLevel() >= item->Click.Level2)
 				{
-					if (item && item->RecastDelay > 0)
+					if (item->RecastDelay > 0)
 					{
 						if (!GetPTimers().Expired(&database, (pTimerItemStart + item->RecastType), false)) {
 							LogSpells("Casting of [{}] canceled: item spell reuse timer not expired", spell_id);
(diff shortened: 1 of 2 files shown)


Kinglykrab (approved):

zone/client_packet.cpp
```diff
@@ -8959,6 +8959,14 @@ void Client::Handle_OP_ItemVerifyRequest(const EQApplicationPacket *app)
 				}
 				if (GetLevel() >= item->Click.Level2)
 				{
+					if (item->RecastDelay > 0)
+					{
+						if (!GetPTimers().Expired(&database, (pTimerItemStart + item->RecastType), false)) {
+							LogSpells("Casting of [{}] canceled: item spell reuse timer not expired", spell_id);
```
Kinglykrab, line 8965: Do we not want a message here similar to the others?

hgtw (approved):

zone/client_packet.cpp
KayenEQ: Client gives the message for regular item clicks. Would just be a double message. For clicks from augments it is needed. Good question though.

Kinglykrab: Wasn't aware, wanted to make sure. 👍

PR #1683, opened 2021-11-06
[Quest API] Added NPC special ability to modify Riposte/Dodge/Parry/Block chance

KayenEQ:
Added NPC special ability 51 that will allow you to modify by percent your NPC's Riposte/Dodge/Parry/Block chance individually or all skill types.

Special Ability ID = 51  MODIFY_AVOID_MELEE
Param 1 = ALL SKILLS
Param 2 = Riposte
Param 3 = Block
Param 4 = Parry
Param 5 = Dodge

Example if set from perl quests.
$npc->SetSpecialAbility(51, 1); #This has to be set first, like any other special ability.
$npc->SetSpecialAbilityParam(51, 0, 50); #Will increase all skills chance by 50 pct
$npc->SetSpecialAbilityParam(51, 1, 25); #Will increase Riposte chance by 25 pct

Note: It will combine all skills and an individual skill chance if both are set. So in this case Riposte gets 75 pct and other
skills get 50 pct.

For reference: NPC innate chances for these skills are rather static based on the npcs skill. On average a lv 60ish NPC will have around a 6 pct chance to succeed at any of these individually. Thus if you want to make it a 100 pct chance to Riposte for quest script ect, you need to set the value to be about ~1550. Obviously higher will work too.

Reminder: special ability Special Ability ID = 44 'COUNTER_AVOID_MELEE' already exists which allows you to modify the ATTACKERS chance on the NPC for the above skills.  Uses same system as above.
Example. You want your NPC to reduce the chance of your player dodging it's attacks by 50 pct. 
$npc->SetSpecialAbilityParam(44, 5, 50);


zone/common.h
```diff
@@ -201,14 +201,15 @@ enum {
 	ALLOW_TO_TANK = 41,
 	IGNORE_ROOT_AGGRO_RULES = 42,
 	CASTING_RESIST_DIFF = 43,
-	COUNTER_AVOID_DAMAGE = 44,
+	COUNTER_AVOID_DAMAGE = 44,                   //Modify by percent NPC's opponents chance to riposte, block, parry or dodge indivually, or for all skills
```
Kinglykrab, line 204: Individually?

zone/common.h
```diff
@@ -201,14 +201,15 @@ enum {
...
 	PROX_AGGRO = 45,
 	IMMUNE_RANGED_ATTACKS = 46,
 	IMMUNE_DAMAGE_CLIENT = 47,
 	IMMUNE_DAMAGE_NPC = 48,
 	IMMUNE_AGGRO_CLIENT = 49,
 	IMMUNE_AGGRO_NPC = 50,
-	MAX_SPECIAL_ATTACK = 51
+	MODIFY_AVOID_DAMAGE = 51,                    //Modify by percent the NPCs chance to riposte, block, parry or dodge indivually, or for all skills
```
Kinglykrab, line 211: Here too.

Kinglykrab (approved):

Kinglykrab:
Looks good, please remember to add to documentation here: https://docs.eqemu.io/server/categories/npc/special-attacks

hgtw (approved):

zone/common.h
KayenEQ: thanks my spelling is on point as usual

KayenEQ:
updated gitbook

PR #1688, opened 2021-11-07
[Spells]  Implemented SPA 415 SE_FFItemClass

KayenEQ:
Implemented SPA 415 SE_FFItemClass **Limits focuses to check if cast from item clicks. Can be used to INCLUDE or EXCLUDE items by ItemType and/or SubType and/or Slots**

This is a very powerful effect that can have a lot of creative uses. Most simply, if you don't want a specific focus effect being applied to items you can add this to it. But, interestingly if you want a focus effect that say ONLY applies to Shield slot click effect, or weapon slot click effects. For example you have a Custom warrior Breastplate with a click effect that gives 90 pct damage mitigation for 3 tics. You can make a custom focus on another item that extends ONLY that breastplates click focus to last 6 tics, by setting the SubType on your item to say the Item ID, and then setting focus limit to that same Item ID so the match. Now the focus will only effect the click on that specific breastplate.

Full description and examples
```
Limits focuses to check if cast from item clicks. Can be used to INCLUDE or EXCLUDE items by ItemType and/or SubType and/or Slots
	
Not used on live, going on information we have plus implemented as broadly as possible to allow all possible options.

base = item table field 'ItemType' Limit = item table field 'SubType' Max = item table field 'Slots' (this is slot bitmask)
					
When including: Setting base, limit, max respectively to -1 will cause it to ignore that check, letting any type or slot ect be used.
					
Special rules for excluding. base value needs to be negative < -1, if excluding all ItemTypes set to -1000. 
For SubType and Slots set using same rules above as for includes. Ie. -1 for all, positive for specifics
To exclude a specific ItemType we have to do some math. The exclude value will be the negative value of (ItemType + 100).
If ItemType = 10, then SET ItemType= -110 to exclude. If its ItemType 0, then SET ItemType= -100 to exclude ect. Not ideal but it works.
					
Usage example: [INCLUDE] Only focus spell if from click cast and is a 'defense armor' item type=10 [base= 10, limit= -1, max= -1]
Usage example: [INCLUDE] Only focus spell if from click cast and is from helmet slot' slots= 4     [base= -1, limit= -1, max= 4]
Usage example: [EXCLUDE] Do not focus spell if it is from an item click. [base= -1000, limit= -1, max= -1]
Usage example: [EXCLUDE] Do not focus spell if it is from an item click from a helmet slot. [base= -1000, limit= -1, max= 4]
Usage example: [EXCLUDE] Do not focus spell if it is from an item click and is a 'defense armor' item type=10. [base= -110, limit= -1, max= -1]

Note: You can apply multiple includes or excludes to a single focus spell,  using multiple SPA 415 limits in the spell. Ie. Check for clicks from ItemType 10 or 11.
```
			

```

### DIFF
--- a/zone/spell_effects.cpp
+++ b/zone/spell_effects.cpp
@@ -5270,8 +5270,10 @@ int32 Mob::CalcFocusEffect(focusType type, uint16 focus_id, uint16 spell_id, boo
 
 	for (int i = 0; i < EFFECT_COUNT; i++) {
 
-		switch (focus_spell.effect_id[i]) {
+		Shout("Check SPA %i", focus_spell.effect_id[i]);
 
+		switch (focus_spell.effect_id[i]) {
+			
 			case SE_Blank:
 				break;
 
@@ -5594,11 +5596,13 @@ int32 Mob::CalcFocusEffect(focusType type, uint16 focus_id, uint16 spell_id, boo
 				break;
 
 			case SE_FFItemClass:
+				Shout("1 Limit Check ItemClass");
 				if (casting_spell_inventory_slot && casting_spell_inventory_slot != -1) {
 					if (IsClient() && casting_spell_slot == EQ::spells::CastingSlot::Item && casting_spell_inventory_slot != 0xFFFFFFFF) {
 						auto item = CastToClient()->GetInv().GetItem(casting_spell_inventory_slot);
 						if (item && item->GetItem()) {
-							
+							Shout("2 Limit Check ItemClass FROM ITEM %i %i %i", focus_spell.base_value[i], focus_spell.limit_value[i], focus_spell.max_value[i]);
+							Shout("2 Limit Check ItemClass FROM FOCUS %i %i %i", item->GetItem()->ItemType, item->GetItem()->SubType, item->GetItem()->Slots);
 							//If ItemType set to -2, then we will exclude either all items, or specific items by SubType or Slot.
 							if (focus_spell.base_value[i] == -2) { //Excludes
 								bool exclude_this_item = true;
@@ -5614,8 +5618,8 @@ int32 Mob::CalcFocusEffect(focusType type, uint16 focus_id, uint16 spell_id, boo
 										exclude_this_item = false;
 									}
 								}
-								
 								if (exclude_this_item) {
+									Shout("Excluded Fail");
 									return 0;
 								}
 							}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: zone/spell_effects.cpp [hunk 1]
```diff
@@ -5239,13 +5289,16 @@ int32 Mob::CalcFocusEffect(focusType type, uint16 focus_id, uint16 spell_id, boo
...
 	12/13 SE_LimitSpellClass:
 	14/15 SE_LimitSpellSubClass:
+	16/17 SE_FFItemCLass:
 	Remember: Update MaxLimitInclude in spdat.h if adding new limits that require Includes
 	*/
 
 	for (int i = 0; i < EFFECT_COUNT; i++) {
 
-		switch (focus_spell.effect_id[i]) {
+		Shout("Check SPA %i", focus_spell.effect_id[i]);
```
Kinglykrab, line 5298: Debug message here.

zone/spell_effects.cpp
```diff
@@ -5567,6 +5620,91 @@ int32 Mob::CalcFocusEffect(focusType type, uint16 focus_id, uint16 spell_id, boo
...
+								//item slot bitmask (if set to -1, ignore and include any slot)
+								if (focus_spell.max_value[i] >= 0) {
+									if (focus_spell.max_value[i] != item->GetItem()->Slots) {
+										include_this_item = false;
+									}
+								}
+
+								if (include_this_item) {
+									LimitInclude[IncludeFoundSEFFItemClass] = true;
+								}
```
Kinglykrab, line 5701: Why do we need to do this if this is already set to true at the beginning? Should we be setting this to true/false based on if item is included?

zone/spell_effects.cpp
KayenEQ: oops thought I fixed that, must have forgot to hit push. resolved

zone/spell_effects.cpp
```diff
@@ -4891,6 +4893,54 @@ int32 Client::CalcAAFocus(focusType type, const AA::Rank &rank, uint16 spell_id)
...
+									if (base_value != item->GetItem()->ItemType) {
+										include_this_item = false;
+									}
+								}
+								//SubType (if set to -1, ignore and include any SubType)
+								if (base_value >= 0) {
+									if (limit_value != item->GetItem()->SubType) {
+										include_this_item = false;
+									}
+								}
```
Kinglykrab, line 4935: Can this all be in the same condition? They're both checking base_value >= 0.

Kinglykrab: Looking at it closer, maybe the second one should be checking limit_value instead?

Kinglykrab (approved):

KayenEQ:
Hold on merge. Testing a few more situations that came up.

PR #1689, opened 2021-11-07
[Rules] Add optional rule for lifetap heals

mmcgarvey:
What:
	Add toggle for compounding bonuses for lifetap heals.

Why:
	When spell damage and heal amount bonuses are scaled to ludicrous
	levels, this double dip results in very high heals from
	relatively weak lifetaps.

Created new rule:  Spells:CompoundLifetapHeals

If true (default):
	Apply spell damage bonuses to lifetap damage
	Pass that amount through heal bonuses
	Heal for this resulting amount
If false:
	Apply spell damage bonuses to lifetap damage
	Heal for this resulting amount


Kinglykrab (approved):
Looks good to me, appreciate the contribution, looking forward to more. ❤️

KayenEQ:
Curious, do you know what the actual mechanic on live is for this?

mmcgarvey:
> Curious, do you know what the actual mechanic on live is for this?

I actually do not.  This was functionality requested by a server I play on, and this seemed like the least disruptive way to accomplish this.

KayenEQ:
Yup it is a solution. Maybe I will see if I can figure out the correct behavior so we won't need a rule.

mmcgarvey:
> Yup it is a solution. Maybe I will see if I can figure out the correct behavior so we won't need a rule. In the mean time, looks good.

The thing I wanted to do was use the spell effect's base damage and run that through the GetActSpellHealing method.  However, that proved to be a bit beyond me.  This was a compromise.

KayenEQ:
Unfortunately I can't confirm the behavior on live because there are no heal focuses that are not limited to only beneficial spells. Hence, this isn't an issue on a standard server. However, what I can confirm is other heal bonuses, like item stat amount added healing are applied to the life tap.  So anyone with the rule on would not receive that either.

After further review, this rule is a hammer to a problem that has an easier solution. I can only assume the problem is occurring due to custom healing focus that are not limited to only beneficial effects. If that is the case, my real answer to the server owner would be to fix your focus spells to only have them focus beneficial spells.

KayenEQ:
Just for reference. Life taps do you use the critical damage amount of the nuke portion when multiplying for critical heals. Thus it likely would be the same for any other heal related modifier.

Normal Vampiric Strike II does 171 damage and heals 179 (due to +healing mod items)

[Sun Nov 07 18:36:15 2021] You hit Test Seventy for 171 points of magic damage by Vampiric Strike II.
[Sun Nov 07 18:36:15 2021] Test Seventy is withered by a vampiric strike.
[Sun Nov 07 18:36:15 2021] You healed Dyri for 179 hit points by Vampiric Strike II.

[Sun Nov 07 18:12:38 2021] You hit Gonobn for 359 points of magic damage by Vampiric Strike II. (Critical Twincast)
[Sun Nov 07 18:12:38 2021] Gonobn is withered by a vampiric strike.
[Sun Nov 07 18:12:38 2021] You healed Dyri for 734 hit points by Vampiric Strike II. (Critical)

[Sun Nov 07 18:27:24 2021] You hit Gonobn for 171 points of magic damage by Vampiric Strike II.
[Sun Nov 07 18:27:24 2021] Gonobn is withered by a vampiric strike.
[Sun Nov 07 18:27:24 2021] You healed Dyri for 358 hit points by Vampiric Strike II. (Critical)

[Sun Nov 07 18:27:36 2021] You hit Gonobn for 359 points of magic damage by Vampiric Strike II. (Critical)
[Sun Nov 07 18:27:36 2021] Gonobn is withered by a vampiric strike.
[Sun Nov 07 18:27:36 2021] You healed Dyri for 367 hit points by Vampiric Strike II.

mmcgarvey:
> After further review, this rule is a hammer to a problem that has an easier solution. I can only assume the problem is occurring due to custom healing focus that are not limited to only beneficial effects.

I can research this.

mackal (approved):

mackal:
I'm fine with this, but Kayen brings up good points and avoiding branching is always nice.

mmcgarvey:
After doing some research on my end, I feel that this rule is still useful, but it is probably misnamed.  Kayen is correct that limiting focus effects by beneficial only is the proper way to avoid focus effects from impacting lifetap heals.  That said, this rule still toggles whether or not the Heal Amount stat impacts lifetap heals.

Would it make sense for me to change this rule name to "Spells:IgnoreHealAmountForLifetaps"?
Or is the implementation still too heavy handed even for that rule name?  If so, I have some ideas for how to fine tune the implementation a bit.

KayenEQ:
The rule is fine if the goal is to apply no heal modifiers to life taps, which will cover all critical heals, incoming heal modifier, item heal mods, and all the various focus effects.

Therefore, rule should be named something like Spells:DoNotApplyHealingModifiersForLifeTap

mmcgarvey:
> Therefore, rule should be named something like Spells:DoNotApplyHealingModifiersForLifeTap

Thank you for the feedback.  I will get this name changed and request new approvals.

PR #1698, opened 2021-11-08
[Commands] Cleanup #showskills Command.

Kinglykrab:
- Cleanup display and use GetSkillName() helper method.
- Only shows skills Client can have and that also have a current max of greater than 0. 
image

### DIFF
--- a/zone/command.cpp
+++ b/zone/command.cpp
@@ -3627,14 +3627,38 @@ void command_flymode(Client *c, const Seperator *sep)
 
 void command_showskills(Client *c, const Seperator *sep)
 {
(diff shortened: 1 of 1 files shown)


mackal:
Maybe we should add an option so it doesn't filter stuff out?

Kinglykrab:
> Maybe we should add an option so it doesn't filter stuff out? Default should probably be like you coded it though, because that's usually want the GM would want to look at :P

I’ll do that this afternoon. #showskills with an optional “all” parameter.

mackal (approved):

PR #1711, opened 2021-11-11
[Commands] Cleanup #zonelock Command.

Kinglykrab:
- Add support for Zone IDs.
- Cleanup messages and display.
- Fix dangling pointer in ZoneLongName() helper method so name is displayed properly.
image
image

### DIFF
--- a/world/zoneserver.cpp
+++ b/world/zoneserver.cpp
@@ -1035,7 +1035,7 @@ void ZoneServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p) {
 				zoneserver_list.SendEmoteMessage(
 					0,
 					0,
-					80,
+					EQ::constants::AccountStatus::QuestTroupe,
 					Chat::White,
 					fmt::format(
 						"Zone {} | Name: {} ({}) ID: {}",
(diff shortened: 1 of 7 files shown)


world/zoneserver.cpp
```diff
@@ -1022,22 +1022,44 @@ void ZoneServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p) {
...
-				this->SendEmoteMessageRaw(s->adminname, 0, 0, 0, "Failed to change lock");
+		} else if (
+			lock_zone->op == EQ::constants::ServerLockType::Lock ||
+			lock_zone->op == EQ::constants::ServerLockType::Unlock
+		) {
+			if (zoneserver_list.SetLockedZone(lock_zone->zoneID, lock_zone->op == EQ::constants::ServerLockType::Lock)) {
+				zoneserver_list.SendEmoteMessage(
+					0,
+					0,
+					80,
```
joligario, line 1038: Did they ever come up with a common GM status enum? If so, we should switch to it.

joligario (approved):

world/zoneserver.cpp
Kinglykrab: If not I can add one.

Akkadius:
Could add some Locked / Unlocked saylinks here while you're at it

Akkadius (approved):

Kinglykrab:
Adding saylinks to the world stuff requires a change in more than just the include to get them to work and seems kind of out of scope for this, maybe they can be added in the future at some point once that change is made?

PR #1713, opened 2021-11-11
[Commands] Cleanup #viewnpctype Command.

Kinglykrab:
- Create a temporary NPC to use ShowStats() instead.
- Cleanup message.
- Cleanup spawn/emote/textures logic in ShowStats() when unused.


joligario (approved):
Using OR, you aren't guaranteed to have all items, so some output may still be blank.

Kinglykrab:
That’s fine as long as they have one it doesn’t matter. Don’t want to display if they have none.

PR #1716, opened 2021-11-12
[Commands] Cleanup #faction Command.

Kinglykrab:
- Remove find subcommand as we have #findfaction now.
- Cleanup message.
image
image

### DIFF
--- a/zone/command.cpp
+++ b/zone/command.cpp
@@ -5167,20 +5167,11 @@ void command_findnpctype(Client *c, const Seperator *sep)
 
 void command_faction(Client *c, const Seperator *sep)
 {
-	if (sep->arg[1][0] == 0) {
-		c->Message(Chat::White, "Usage: #faction -- Displays Target NPC's Primary faction");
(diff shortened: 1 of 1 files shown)


zone/command.cpp
```diff
@@ -5167,116 +5167,155 @@ void command_findnpctype(Client *c, const Seperator *sep)
...
+				Chat::White,
+				fmt::format(
+					"{} ({}) has a Primary Faction  of {} ({}).",
+					npc_name,
+					npc_id,
+					content_db.GetFactionName(npc_faction_id),
+					npc_faction_id
+				).c_str()
+			);
+			c->Message(Chat::White, "Use: #setfaction [Faction ID] - Set an NPC's Primary Faction ID");
```
joligario, line 5317: Not sure of the need for #setfaction message here

joligario (approved):

zone/command.cpp
Kinglykrab: Yeah, it was there before so I just left it. I can remove it.

PR #1734, opened 2021-11-12
[Commands] Remove #crashtest Command.

Kinglykrab:
- Remove unused/deprecated command.


joligario (approved):
Yeah, not sure why we would ever want to crash from inside

PR #1739, opened 2021-11-12
[Loginserver] Worldserver Name Sanitization

Akkadius:
**Description**

* Implements a bad-word filter where servers with bad words in their names will be replaced with `*` in the name
* Adds config options that list standard world servers at the bottom of the list if the name contains `dev` or `test` or `installer`
* Adds config options that list standard world servers at the bottom of the list if the name starts with a special character

**Top**

image

**Test / Dev / Installer / Special Character Start at Bottom**

* `|D|` Dev
* `|T|` Test
* `|I|` Installer
* `|*|` Special character start

image


loginserver/world_server.cpp
```diff
@@ -1348,3 +1354,36 @@ void WorldServer::OnKeepAlive(EQ::Timer *t)
...
+	server_long_name = trim(server_long_name);
+
+	bool name_set_to_bottom = false;
+	if (server.options.IsWorldDevTestServersListBottom() && server_list_type == LS::ServerType::Standard) {
+		std::string s = str_tolower(server_long_name);
+		if (s.find("dev") != std::string::npos) {
+			server_long_name   = fmt::format("|D| {}", server_long_name);
+			name_set_to_bottom = true;
+		}
+		else if (s.find("test") != std::string::npos) {
```
hgtw, line 1370: Should this look for whole words instead? It could catch false positives if a server name has `test` as part of another word. `The greatest server` for example

Akkadius: It could, but then we'd be missing things like `(Test)` or `[Test]` and I don't want to overly science this. You could strip special characters before doing the match but then someone else prepends a letter to bypass it. You're going to have issues one way or another

loginserver/world_server.cpp
```diff
@@ -1348,3 +1354,36 @@ void WorldServer::OnKeepAlive(EQ::Timer *t)
...
+	if (server.options.IsWorldSpecialCharacterStartListBottom() && server_list_type == LS::ServerType::Standard &&
+		!name_set_to_bottom) {
+		auto first_char = server_long_name.c_str()[0];
+		if (IsAllowedWorldServerCharacterList(first_char) && first_char != '|') {
+			server_long_name   = fmt::format("|*| {}", server_long_name);
+			name_set_to_bottom = true;
+		}
+	}
+
+	strcpy(name, server_long_name.c_str());
```
hgtw, line 1388: This could be unsafe if a server long name is right at the max buffer length since this function can grow the string. May want to pass the max buffer length to the function as another arg and use `strn0cpy(name, server_long_name.c_str(), length)`

Akkadius: Good call and good eye; changed

hgtw (approved):

PR #1740, opened 2021-11-12
[Shared Tasks] Cross Zone Remove Fix

mmcgarvey:
Why:
	The cross_zone_remove_task quest methods were not removing from
	shared_task_members database table and were not clearing shared task
	cache.  This resulted in a situation where a character could not
	request other shared tasks.

What:
	Shamelessly copied shared task logic from ClientTaskState::CancelTask
	into ClientTaskState::RemoveTaskByTaskID


Kinglykrab (approved):
Looks good to me, @hgtw should look it over though.

Akkadius:
I’ll take a look at this one later

hgtw:
This would work but looking at this method it seems to just mostly duplicate what `CancelTask` and `RemoveTask` already do. Could you maybe instead rewrite the method to just forward to `CancelTask` after finding the task id?

Something like:
```cpp
if (m_active_task.task_id == task_id)
  CancelTask(client, TASKSLOTTASK, TaskType::Task);
else if (m_active_shared_task.task_id == task_id)
  CancelTask(client, TASKSLOTSHAREDTASK, TaskType::Shared);
else if (m_active_task_count > 0)
  loop m_active_quests to find task_id, if found CancelTask(client, i, TaskType::Quest);
```

mmcgarvey:
> Could you maybe instead rewrite the method to just forward to `CancelTask` after finding the task id?

Great idea.  I will work on that.  Thank you.

mackal (approved):

Akkadius (changes requested):
@hgtw had a suggestion around re-use

zone/task_client_state.cpp
```diff
@@ -2212,42 +2212,23 @@ void ClientTaskState::RemoveTaskByTaskID(Client *client, uint32 task_id)
 	switch (task_type) {
```
hgtw, line 2212: You can probably remove the switch/case entirely now and just check the task id without using the `task_type` variable

Remove the database calls above since that's handled by the `remove_from_db` passed to CancelTask now. Also remove the now unused `character_id` var.

mmcgarvey: I agree with the DB calls and character_id.

I think we still need the switch statement, because the logic to get the sequence number passed to CancelTask changes based on task_type.

hgtw (approved):

hgtw:
Thanks for the extra work. I know it went a little beyond the scope of the original pr fix but getting this method refactored and cleaned up is appreciated

mmcgarvey:
> Thanks for the extra work. I know it went a little beyond the scope of the original pr fix but getting this method refactored and cleaned up is appreciated

Happy to help.  Thank you for your feedback.  It will help me in future refactors.

Kinglykrab (approved):

Kinglykrab:
Any reason this shouldn’t be merged?

mmcgarvey:
> Any reason this shouldn’t be merged?

I think I was just waiting on @Akkadius to verify that I have addressed his concerns.

Akkadius (approved):

Akkadius:
@mmcgarvey sorry for the lag time on this one. Appreciate you following up on feedback

Akkadius:
Also appreciate your PR descriptions and the context you provide

PR #1741, opened 2021-11-12
[Spells] Update to IsCombatProc checks

KayenEQ:
Defensive procs and all AA procs added to IsCombatProc check to allow them to be exclude from focusing.
Forgot to add the new AA procs when recently implemented.

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -5680,7 +5680,7 @@ bool Mob::IsCombatProc(uint16 spell_id) {
 
 		for (int i = 0; i < MAX_PROCS; i++){
 			if (PermaProcs[i].spellID == spell_id || SpellProcs[i].spellID == spell_id
-				 || RangedProcs[i].spellID == spell_id){
+				 || RangedProcs[i].spellID == spell_id || DefensiveProcs[i].spellID == spell_id){
 				return true;
 			}
 		}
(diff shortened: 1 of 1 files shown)


Kinglykrab (approved):

zone/spells.cpp
```diff
@@ -5680,10 +5680,21 @@ bool Mob::IsCombatProc(uint16 spell_id) {
...
+		if (IsClient()) {
+			for (int i = 0; i < MAX_AA_PROCS; i += 4) {
+
+				if (aabonuses.SpellProc[i + 1] == spell_id ||
+					aabonuses.RangedProc[i + 1] == spell_id ||
+					aabonuses.DefensiveProc[i + 1] == spell_id) {
+					return true;
+				}
+			}
+		}
```
Kinglykrab, line 5697: Can format like this so it's easier to read them both.
```cpp
for (int i = 0; i < MAX_PROCS; i++){
	if (
		DefensiveProcs[i].spellID == spell_id ||
		PermaProcs[i].spellID == spell_id ||
		RangedProcs[i].spellID == spell_id ||
		SpellProcs[i].spellID == spell_id
	) {
		return true;
	}
}

if (IsClient()) {
	for (int i = 0; i < MAX_AA_PROCS; i += 4) {
		if (
			aabonuses.DefensiveProc[i + 1] == spell_id ||
			aabonuses.RangedProc[i + 1] == spell_id ||
			aabonuses.SpellProc[i + 1] == spell_id
		) {
			return true;
		}
	}
}
```

PR #1742, opened 2021-11-13
[Quest API] Add back removed lua class properties

hgtw:
Fixes regression from 7b6decae


Kinglykrab (approved):

Kinglykrab:
Thought I caught all these, good eye.

PR #1753, opened 2021-11-14
[Bug Fix] Make room for host names.

joligario:
Since m_remote_ip_address does not resolve IP address, world server may still use host name.


Kinglykrab:
May want to add an optional SQL or something so people with preexisting tables can update this.

Kinglykrab (approved):

Akkadius:
This shouldn’t be getting ninja merged

joligario:
Wasn’t the intent. Tried to push the SQL to this one and it didn’t work well.

PR #1762, opened 2021-11-15
[Config] Add some missing entries

joligario:


Akkadius:
I'm not sure how much this config still gets referenced these days. The installer pulls from https://github.com/Akkadius/eqemu-install-v2

joligario:
> 
> 
> I'm not sure how much this config still gets referenced these days. The installer pulls from https://github.com/Akkadius/eqemu-install-v2

Then we should delete these?

Akkadius:
Makes sense to avoid confusion of having multiple sources of truth and things going out dated

joligario:
> 
> 
> Makes sense to avoid confusion of having multiple sources of truth and things going out dated

Only problem is that you can't fill in

PR #1764, opened 2021-11-15
[Cleanup] Make use of AccountStatus constants wherever status is checked or used.

Kinglykrab:
- Cleanup all instances of SendEmoteMessage.
- Cleanup all instances of SendEmoteMessageRaw.
- Cleanup all instances of MessageStatus.
- Convert Quest API method defaults to use constants.
- Convert commands to use constants.


joligario:
Seen many with 80 admin to bypass locked, etc.

joligario:
So needed. Great work!

Akkadius:
Let's move `AccountStatus` top level and this looks great! 🚀

Akkadius (changes requested):
Change nesting

Akkadius (approved):
🔥

PR #1766, opened 2021-11-15
[GM Commands] Split GM Commands Into Separate Files

Akkadius:
When developing, working with command.cpp can lead to waiting up to a minute for linker times because of the massive amount of code and related commands in one file.

To make the development experience far better. This change splits all existing commands into separate files so they are easier to manage / find / organize and subsequent compilation times are far smaller even when making tiny changes

You can still add commands to `command.cpp` but it is recommended to add a separate file to contain the command logic

Command files need to be registered in CMake per usual in `./zone/CMakeLists.txt` under the `gm_commands` block

The one-time code generator script can be found in `./utils/gm_commands/main.go`


Kinglykrab (approved):

Akkadius:
# Full Build Test Compare

```time ninja -j$(expr $(nproc) - 2)

##################
# after change
##################

# run 1) 2m7.740s
# run 2) 2m10.803s

##################
# before change
##################

# run 1)  1m48.012s
# run 2)  1m48.728s
```


# Incremental Build Test Compare

Added the following to command_repop

```cpp
std::cout << "hello" << std::endl;
```

```##################
# after change
##################

# run 1) 3s

##################
# before change
##################

# run 1)  1m5s```

PR #1771, opened 2021-11-16
xackery/elixir

xackery:
NOTE: This is a Work In Progress and NOT yet ready to be pulled into main branch

This is a WIP of a new AI override system called Elixir. We can rename it to another name later if desired, but I made a unique identifier so folks know what to call it or how to debug it and such.


Akkadius:
Re-Open when its ready to be reviewed

PR #1775, opened 2021-11-18
[Bug Fix] - Monk AA Spirit of Master Wu

splose:
Monk AA Spirit of Master Wu sends message even if it grants 0 extra attacks.


splose:
Was setting up my authentication.. didn't know it was going to push all the test commits my bad.

Kinglykrab:
> Was setting up my authentication.. didn't know it was going to push all the test commits my bad.

Can squash and merge, it’ll only be 1 commit, no big deal.

Kinglykrab (approved):
Looks good, could maybe include the msg variable definition inside the check, being the message is only used if there are extra attacks, no need to initialize it otherwise.

PR #1784, opened 2021-11-19
[Skills] Make Tracking Skill Configurable

mmcgarvey:
Added 1 rule per class that defines tracking distance multiplier for that class
Kept the defaults of 12 for ranger, 10 for druid, and 7 for bard

Created 1 method for determining class tracking distance multiplier
Created 1 method for determining if a class can track, based on multiplier

Updated tracking logic to use these methods to determine whether a tracking
packet should and can be sent or not.


splose (approved):
looks good to me

Castious:
Should be noted that you would need to update the skillcaps.txt in whatever client is being used for the skill to function in-game.

Akkadius (approved):

Akkadius:
LGTM

PR #1785, opened 2021-11-19
[API] mob->SpellEffect can now properly display spell effect that are used on a target.

KayenEQ:
Usage: Mob::SpellEffect(THIS, uint32 effect, [uint32 duration = 5000], [uint32 finish_delay = 0], [bool zone_wide = false], [uint32 unk20 = 3000], [bool perm_effect = false], [Client* single_client]), [caster_id = 0], [target_id = 0]")

This function is used for affects like Nimbuses but it also has a large variety of spell casting graphics that are not self only. The way the function was coded it was always setting the target of the effect as the caster. So if you tried to use it to simulate say a "nuke" it would do the casting animation on the target and then spell landing animation. I added a way using optional parameters at the end to set the caster and target.

```
#put on an npc  and hail it
sub EVENT_SAY
{
	$client->SpellEffect(68, 5000,0, 0, 3000, 0, $client, $client->GetID(), $npc->GetID());# Client will cast, NPC will receive the spell
}
```

### DIFF
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -2910,13 +2910,21 @@ void Mob::CameraEffect(uint32 duration, uint32 intensity, Client *c, bool global
 	safe_delete(outapp);
 }
 
-void Mob::SendSpellEffect(uint32 effect_id, uint32 duration, uint32 finish_delay, bool zone_wide, uint32 unk020, bool perm_effect, Client *c) {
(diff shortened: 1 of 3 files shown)


Kinglykrab (approved):
Looks good, not sure if we need to do Lua or not, considering the Lua method is completely different.

PR #1786, opened 2021-11-19
[Bug Fix] Do not set teleport doors to Open

Natedog2012:


Kinglykrab (approved):

zone/doors.cpp
```diff
@@ -383,7 +383,9 @@ void Doors::HandleClick(Client* sender, uint8 trigger) {
 	if (!IsDoorOpen() || (open_type == 58)) {
 		if (!disable_timer)
 			close_timer.Start();
-		SetOpenState(true);
+		
+		if(strncmp(destination_zone_name, "NONE", strlen("NONE")) == 0)
```
Kinglykrab, line 387: Maybe use { } here like we are most other places now.

PR #1793, opened 2021-11-20
[Spells] Support for SPA 161 and 450 to give  percent spell or dot mitigation from Items or AA's.

KayenEQ:
Added support to SPA 161 and SPA 450 to provide spell or dot shielding respectively, when put on as an Item worn effect or as AA's. There was no easy way to provide percentage spell or dot mitigation via AA's so this solves that problem in a straightforward way. This will stack with item mod2 spellshielding or dotshielding.

SE_MitigateSpellDamage 161	// implemented, Runes, mitigate incoming spell damage by percentage until rune fades, base: percent mitigation, limit: max dmg absorbed per hit, max: rune amt, Note: If placed on item or AA, will provide stackable percent mitigation.

SE_MitigateDotDamage	450 // implemented, Runes, mitigate incoming dot damage by percentage until rune fades, base: percent mitigation, limit: max dmg absorbed per hit, max: rune amt, Note: If placed on item or AA, will provide stackable percent mitigation.

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -3298,7 +3298,8 @@ int32 Mob::AffectMagicalDamage(int32 damage, uint16 spell_id, const bool iBuffTi
 
 	// If this is a DoT, use DoT Shielding...
 	if (iBuffTic) {
-		damage -= (damage * itembonuses.DoTShielding / 100);
+		int total_dotshielding = itembonuses.DoTShielding + itembonuses.MitigateDotRune[SBIndex::MITIGATION_RUNE_PERCENT] + aabonuses.MitigateDotRune[SBIndex::MITIGATION_RUNE_PERCENT];
+		damage -= (damage * total_dotshielding / 100);
 
 		if (spellbonuses.MitigateDotRune[SBIndex::MITIGATION_RUNE_PERCENT]) {
 			slot = spellbonuses.MitigateDotRune[SBIndex::MITIGATION_RUNE_BUFFSLOT];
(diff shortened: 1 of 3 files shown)


zone/attack.cpp
```diff
@@ -3298,7 +3298,8 @@ int32 Mob::AffectMagicalDamage(int32 damage, uint16 spell_id, const bool iBuffTi
 
 	// If this is a DoT, use DoT Shielding...
 	if (iBuffTic) {
-		damage -= (damage * itembonuses.DoTShielding / 100);
+		int total_dotshielding = itembonuses.DoTShielding + itembonuses.MitigateDotRune[SBIndex::MITIGATION_RUNE_PERCENT] + aabonuses.MitigateDotRune[SBIndex::MITIGATION_RUNE_PERCENT];
+		damage -= (damage * total_dotshielding / 100);
```
Kinglykrab, line 3302: No possible change for weird overflows here, right?

zone/attack.cpp
```diff
@@ -3330,8 +3331,9 @@ int32 Mob::AffectMagicalDamage(int32 damage, uint16 spell_id, const bool iBuffTi
 	else
 	{
 		// Reduce damage by the Spell Shielding first so that the runes don't take the raw damage.
-		damage -= (damage * itembonuses.SpellShield / 100);
-
+		int total_spellshielding = itembonuses.SpellShield + itembonuses.MitigateSpellRune[SBIndex::MITIGATION_RUNE_PERCENT] + aabonuses.MitigateSpellRune[SBIndex::MITIGATION_RUNE_PERCENT];
+		damage -= (damage * total_spellshielding / 100);
```
Kinglykrab, line 3335: No possible change for weird overflows here, right?

Kinglykrab (approved):

zone/attack.cpp
KayenEQ: As discussed, same chance to overflow as the hundred of other calculations like it if you put in ridiculous numbers.

zone/attack.cpp
KayenEQ: see above

PR #1815, opened 2021-11-22
[Bug Fix] Pick Lock was allowing skillups on doors above player skill

noudess:
This PR fixes picking locks as follows:

- Skilling up class pick lock skill should only happen on a door that can be opened by the PC.  This is now true after this PR.
- Skilling up, as well as messages only happen when you open the door.  Closing the door is quiet.
- A small 1 second spam timer was added to prevent door/sync issues.

### DIFF
--- a/zone/doors.cpp
+++ b/zone/doors.cpp
@@ -278,8 +278,9 @@ void Doors::HandleClick(Client* sender, uint8 trigger) {
 
 		/**
 		 * Key required
+		 * If using a lock_pick_item leave messaging to that code below
 		 */
-		if (!IsDoorOpen()) {
+		if (lock_pick_item == nullptr && !IsDoorOpen()) {
 			sender->Message(Chat::LightBlue, "This is locked...");
 		}
 
--- a/zone/doors.cpp
+++ b/zone/doors.cpp
@@ -333,7 +333,7 @@ void Doors::HandleClick(Client* sender, uint8 trigger) {
 						// Stop full scale mad spamming
 						safe_delete(outapp);
 						return;
-						}
+					}
 
 					float player_pick_lock_skill = sender->GetSkill(EQ::skills::SkillPickLock);
 
--- a/zone/doors.cpp
+++ b/zone/doors.cpp
@@ -329,11 +329,11 @@ void Doors::HandleClick(Client* sender, uint8 trigger) {
 			if (sender->GetSkill(EQ::skills::SkillPickLock)) {
 				Timer* pick_lock_timer = sender->GetPickLockTimer();
 				if (lock_pick_item->GetItem()->ItemType == EQ::item::ItemTypeLockPick) {
-				if (!pick_lock_timer->Check()) {
-					// Stop full scale mad spamming
-					safe_delete(outapp);
-					return;
-					}
+					if (!pick_lock_timer->Check()) {
+						// Stop full scale mad spamming
+						safe_delete(outapp);
+						return;
+						}
 
 					float player_pick_lock_skill = sender->GetSkill(EQ::skills::SkillPickLock);
 
--- a/zone/doors.cpp
+++ b/zone/doors.cpp
@@ -279,7 +279,9 @@ void Doors::HandleClick(Client* sender, uint8 trigger) {
 		/**
 		 * Key required
 		 */
-		sender->Message(Chat::LightBlue, "This is locked...");
+		if (!IsDoorOpen()) {
+			sender->Message(Chat::LightBlue, "This is locked...");
+		}
 
 		/**
 		 * GM can always open locks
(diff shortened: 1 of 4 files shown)


zone/doors.cpp
```diff
@@ -325,19 +327,30 @@ void Doors::HandleClick(Client* sender, uint8 trigger) {
 		 */
 		else if (lock_pick_item != nullptr) {
 			if (sender->GetSkill(EQ::skills::SkillPickLock)) {
+				Timer* pick_lock_timer = sender->GetPickLockTimer();
 				if (lock_pick_item->GetItem()->ItemType == EQ::item::ItemTypeLockPick) {
+				if (!pick_lock_timer->Check()) {
```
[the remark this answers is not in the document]
noudess: Thanks.  Fixed.

zone/doors.cpp
```diff
@@ -329,11 +329,11 @@ void Doors::HandleClick(Client* sender, uint8 trigger) {
...
-				if (!pick_lock_timer->Check()) {
-					// Stop full scale mad spamming
-					safe_delete(outapp);
-					return;
-					}
+					if (!pick_lock_timer->Check()) {
+						// Stop full scale mad spamming
+						safe_delete(outapp);
+						return;
+						}
```
Kinglykrab, line 337: Now this line is messed up.

noudess: Thanks again!

zone/doors.cpp
```diff
@@ -278,8 +278,11 @@ void Doors::HandleClick(Client* sender, uint8 trigger) {
 
 		/**
 		 * Key required
+		 * If using a lock_pick_item leave messaging to that code below
 		 */
-		sender->Message(Chat::LightBlue, "This is locked...");
+		if (lock_pick_item == nullptr && !IsDoorOpen()) {
+			sender->Message(Chat::LightBlue, "This is locked...");
```
Kinglykrab, line 284: Can do if !lock_pick_item versus a direct compare to nullptr, unless 0 has a different meaning.

zone/doors.cpp
```diff
@@ -325,19 +328,30 @@ void Doors::HandleClick(Client* sender, uint8 trigger) {
 		 */
 		else if (lock_pick_item != nullptr) {
```
Kinglykrab, line 329: Can also change this to if lock_pick_item versus explicit nullptr check.

zone/doors.cpp
noudess: So I mimic'd what the old code did when checking this variable, assuming someone there had a preference for it.

Whatever is fine with me.  Let me know.  I just didn't want to go against what they were already doing.

zone/doors.cpp
noudess: See above - changed as little as I needed to in order to keep it simple.  Let me know what the consensus is.

martinlindhe:
Will this affect skilling up lock pick on the potranquility book?

noudess:
> Will this affect skilling up lock pick on the potranquility book?

Depends on the server setup.  If the book is set to be pickable at 1 in the server db, then  no.  From what I understand of it, that is the case.

martinlindhe:
@noudess thanks for looking into that. Yes, the book is a old skillup secret from live afaik.

Kinglykrab (approved):

Kinglykrab:
How do you want to move forward with this @noudess? Would you like to include an optional SQL update with this to keep that door from being abused for max skill?

noudess:
> How do you want to move forward with this @noudess? Would you like to include an optional SQL update with this to keep that door from being abused for max skill?

I'm open to what you guys want.  I think the PoT book would need to be changed to have a pick lock of 1 to satisfy being like live.  Everything else, as far as I am aware should work correctly after this code change.

Just waiting to hear in from at least projecteq on the SQL issue.

I'm happy to add the sql.  Either required or optional.

SecretsOTheP:
Needs to have a 'picklock_trivial' flag. The tranquility door skills up much faster than other doors in the game (ie; paineel entrance) due to its high trivial.

noudess:
> Needs to have a 'picklock_trivial' flag. The tranquility door skills up much faster than other doors in the game (ie; paineel entrance) due to its high trivial. I also think the tranquility door 'resets' instantly, so it may just be that it can be clicked quicker. Anyone able to hop on live and check?
> 
> I know they revamped it at one point.

Are you saying that existing code uses the picklock value of 335 to increase skill-up rate?  I didn't see that.  I'm all for making this better if door trivial change skill up rates, but I'd see that as a separate PR done by someone that understands that.

Currently I don't think the high picklock value for the PoT door has any impact on the code or skillup rate at all.   Am I wrong?

I did get on live.  The clicking was faster. but I think because they handle the spinning book better than we do (we just open/close).

For myself, I'd like to just add the SQL to drop the pick lock to 1 so people can skill up using that book.  If there is more to add, we can make that a next step.  I don't think dropping it to 1 will change anything at all, except make it work as it should (only skillup on successful open).

All that said, I'll do whatever is consensus.

noudess:
I'm ready to merge this.  Just waiting to hear that servers are OK with doing the optional SQL on their own.  The PoT book is pickable at 1, so it needs to be set to 1, now that this PR handles skillups correctly.

Kinglykrab:
I'm good with merging this.

Akkadius:
Reviewed this with @noudess 

I'm not seeing any doors that I would otherwise be concerned with regarding this change. I went through and spotchecked a handful of other non-zero lockpick.

```sql
MariaDB [peq]> select id, doorid, lockpick, count(*) from doors group by lockpick;
+-------+--------+----------+----------+
| id    | doorid | lockpick | count(*) |
+-------+--------+----------+----------+
|  5321 |     20 |       -1 |      122 |
|     9 |      8 |        0 |    18730 |
|  5387 |     52 |        1 |        1 |
|   131 |     10 |       11 |        1 |
|   124 |     17 |       25 |        3 |
|  1483 |     21 |       40 |        2 |
|  1487 |     17 |       45 |        1 |
|   136 |      5 |       50 |        3 |
|  7301 |    246 |       60 |        3 |
|  6452 |     38 |      100 |        7 |
|  5876 |     10 |      130 |        1 |
|  5440 |      6 |      150 |        5 |
|  8749 |     42 |      170 |        1 |
|  6446 |     13 |      200 |        8 |
|  6466 |     46 |      201 |       10 |
|  8324 |     61 |      214 |        2 |
| 11721 |      3 |      220 |        4 |
| 10972 |     46 |      335 |        1 |
+-------+--------+----------+----------+
18 rows in set (0.000 sec)
```

Akkadius (approved):

PR #1822, opened 2021-11-24
[Bug Fix] Bard instrument modifiers applying to effects when it shouldn't under certain conditions.

KayenEQ:
Fixed issue where targeted bard songs were able to bypass restricted effects check. This was making Harmony type spells not work properly.


noudess:
This is related/responding to the odd harmony think I posted in spells help for bard harmony?

Kinglykrab (approved):

PR #1825, opened 2021-11-25
[Bug Fix] Bind Sight will now function properly 

KayenEQ:
Fixed issue where Bind Sight effects would not display the buff on your character when initially cast
Fixed issue where you could not click off the buff on your character (preventing you from restoring your own vision)
Fixed issue where the effect was hard coded to only two spells

SE_BindSight	73 see through the eyes of your target, click off buff to end effect, base: 1, limit: none, max: none

### DIFF
--- a/common/spdat.cpp
+++ b/common/spdat.cpp
@@ -242,7 +242,7 @@ bool IsBeneficialSpell(uint16 spell_id)
 				    (sai == SAI_Calm || sai == SAI_Dispell_Sight || sai == SAI_Memory_Blur ||
 				     sai == SAI_Calm_Song))
 					return false;
-			} else {
+			} else {//KAYENFIX
 				// If the resisttype is not magic and spell is Bind Sight or Cast Sight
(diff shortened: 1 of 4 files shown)


common/spdat.cpp
```diff
@@ -242,7 +242,7 @@ bool IsBeneficialSpell(uint16 spell_id)
 				    (sai == SAI_Calm || sai == SAI_Dispell_Sight || sai == SAI_Memory_Blur ||
 				     sai == SAI_Calm_Song))
 					return false;
-			} else {
+			} else {//KAYENFIX
```
joligario, line 245: ?

joligario (approved):

Kinglykrab (approved):

common/spdat.cpp
KayenEQ: oops forgot to remove that one

noudess:
Excited to see this work!  Thanks!!

PR #1851, opened 2021-11-28
[Bug fix] SetPetID after we assign the new NPC an ID

Natedog2012:
Players / npcs need this set otherwise trying to scale NPCs AGI / AC will cause them to depop

Also any NPC pets with items will despawn if this is not set here as well.


Kinglykrab (approved):

splose:
Good catch on this one this has been a bug forever

splose (approved):

PR #1854, opened 2021-11-29
[Commands] Remove #shownumhits Command.

Kinglykrab:
- Remove unused and/or unnecessary command, buff hit number packets are now properly sent.


Kinglykrab:
Wasn't aware Titanium required this.

PR #1855, opened 2021-11-30
[Bug Fix] Charm Break Invisibility Fix.

Kinglykrab:
- Invisibility vs. Undead and Invisibility vs. Animals were not breaking charm.
- Add Invisibility enumerator.
- Add special identifier for Invisibility vs. Undead and Invisibility vs. Animals.

### DIFF
--- a/common/emu_constants.h
+++ b/common/emu_constants.h
@@ -371,4 +371,10 @@ enum AccountStatus : uint8 {
 	Max = 255
 };
 
+enum Invisibility : uint8 {
+	Visible,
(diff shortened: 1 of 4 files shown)


noudess:
I tested this locally.  It solves both problems:

1.  Calling SetInvisible(0) was stopping IVU from stacking on top of InV - they stack again if done in correct order
2. IVU and IVA now break charm again.

Thanks @Kinglykrab

joligario (approved):

PR #1868, opened 2021-12-06
[Combat] Allow npcs/pets to kick vs opponents requiring magic weapons if wearing magic booties.

noudess:
NPCs and Pets below the rule values allowing damage vs mobs requiring magic weapons can now kick successfully if wearing magical booties.


Kinglykrab (approved):

Akkadius (approved):
LGTM

PR #1869, opened 2021-12-07
[Doors] Add new rule enabling classic "key on cursor" for pre keyring keys

noudess:
Classic locked doors with keys that do not add to the key ring required the player to have the key on their cursor.

Added a new rule that allows this functionality to work correctly.  The default is the functionality before this patch, which is to just require it to be anywhere in their inventory.

### DIFF
--- a/zone/doors.cpp
+++ b/zone/doors.cpp
@@ -211,13 +211,23 @@ void Doors::HandleClick(Client* sender, uint8 trigger) {
 
 	uint32 required_key_item       = GetKeyItem();
 	uint8  disable_add_to_key_ring = GetNoKeyring();
-	uint32 player_has_key          = 0;
+	bool player_has_key            = false;
 	uint32 player_key              = 0;
 
 	const EQ::ItemInstance *lock_pick_item = sender->GetInv().GetItem(EQ::invslot::slotCursor);
-	player_has_key = static_cast<uint32>(sender->GetInv().HasItem(required_key_item, 1));
 
-	if (player_has_key != INVALID_INDEX) {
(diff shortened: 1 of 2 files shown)


zone/doors.cpp
```diff
@@ -211,13 +211,23 @@ void Doors::HandleClick(Client* sender, uint8 trigger) {
...
+	bool player_has_key            = false;
 	uint32 player_key              = 0;
 
 	const EQ::ItemInstance *lock_pick_item = sender->GetInv().GetItem(EQ::invslot::slotCursor);
-	player_has_key = static_cast<uint32>(sender->GetInv().HasItem(required_key_item, 1));
 
-	if (player_has_key != INVALID_INDEX) {
+	// If classic key on cursor rule, check for it, otherwise owning it ok.
+	if (RuleB(Doors, RequireKeyOnCursor)) {
+		if (lock_pick_item != nullptr &&
```
Kinglykrab, line 221: Can just do if (lock_pick_item && whatever. Checks for nullptr implicitly.

Kinglykrab (approved):
Looks good.

zone/doors.cpp
noudess: Yeah, once again since they used the check vs nullptr below a few lines  later, I chose to stay consistent with code around me.  If we have a standard to do it one way or the other globally, I'd be on board.

noudess:
Rule based and well tested.  Ok to merge?

Kinglykrab:
> Rule based and well tested. Ok to merge?

I’m okay with it, don’t see any reason not to. A fix is easily available and PEQ can fix it on their end for public distribution in their daily dump.

PR #1872, opened 2021-12-08
[XTarget] Add Character:XTargetAllowCorpses Rule

mmcgarvey:
Why:

A condition could occur where a member of a group or raid could FD and still
have mobs on his/her hate list due to feign memory.  If during this state a
mob in this client's XTarget window dies, the corpse of that mob would take
its place in the XTarget window.  This would not clear until the corpse rots.
This feels undesired.

What:

Added a rule (Character:XTargetAllowCorpses) to toggle the behavior of
allowing corpses to appear in the XTarget window.  I implemented this as a
condition within the IsXTarget method, as this seemed the cleanest way of
implementing this behavior across the board.  I defaulted the rule to true,
so server owners need to opt in to this rule by setting it to false.


Kinglykrab:
Maybe remove the rule if this fixes the issue with corpses? Seems like a bug.

mmcgarvey:
> Maybe remove the rule if this fixes the issue with corpses? Seems like a bug.

Seems reasonable to me, but I didn't want to make any assumptions.  If you feel it best to make this the default behavior, I am willing to remove the rule entirely and just add an IsCorpse check to IsXTarget.

mmcgarvey:
With some testing, at scale, my change wasn't a complete fix.  I am going to dig a little deeper.

PR #1821, opened 2021-11-24
[API] mob->AppearanceEffects improved functionality.

KayenEQ:
Credit to Natedog who figured out what extra parameters actually did!

SendAppearanceEffect  packet has 3 different parameters, with up to 5 different effects per packet 
parm1 = graphic animation
parm1slot (formerly parm1a) = location on body of the graphic animation
		0 = pelvis1
		1 = pelvis2
		2 = helm
		3 = Offhand
		4 = Mainhand
		5 = left foot
		6 = right foot
		9 = Face
parm1ground (formerly parm1b) = place graphic on the ground.  If this is set and and slot > 0, it will follow the mob and be removed upon zoning/death, if slot set to zero it will be perma set to be on ground in that location

Original Perl function Mob->SendAppearanceEffect has been expanded to accept slot and ground parameters

`Usage: Mob::SendAppearanceEffect(THIS, int32 param_1, [int32 param_2 = 0], [int32 param_3 = 0], [int32 param_4 = 0], [int32 param_5 = 0], [Client* single_client_to_send_to = null]), [uint32 value1slot = 1], [uint32 value1ground = 1], [uint32 value2slot = 1], [uint32 value2ground = 1], [uint32 value3slot = 1], [uint32 value3ground = 1], [uint32 value4slot = 1], [uint32 value4ground = 1], [uint32 value5slot = 1], [uint32 value5ground = 1]")`

Example: $client->SendAppearanceEffect(8,10,27,28,31,0, 1,0,2,0,3,0,4,0,5,0);

This will send 5 different graphics, with id 8 going to pelvis, id 10 going to helm, id 27 going to off hand, id 28 going to main hand, id 31 going to left foot.

New Perl functions added for simple use cases for just sending effects that go on the mob, and just sending effects that go on the ground.

`Usage: Mob::SendAppearanceEffectActor(THIS, int32 param_1, uint32 value1slot = 0, [int32 param_2 = 0] [uint32 value2slot = 0], [int32 param_3 = 0] [uint32 value3slot = 0], [int32 param_4 = 0] [uint32 value4slot = 0], [int32 param_5 = 0], [uint32 value5slot = 0], [Client* single_client_to_send_to = null])");`

Example $client->SendAppearanceEffectActor(8, 9) ... can add up to 5 effects
Send effect id 8 to helm,

[...]

PR #1874, opened 2021-12-09
[Features] Appearance Effects will now be sent to clients upon zone in. GM commands.

KayenEQ:
This builds on https://github.com/EQEmu/Server/pull/1821 which improved functionality of appearance effects.

This updates fixes issue where appearance effects set on a client or NPC would not be visible to a new client that enters zone after the appearance effects were originally sent. We now can store up to 20 appearance effects on a mob, which will be sent to clients upon zone in. Also fixes issue where appearance effects are lost if an illusion is used.

New GM Command 
```
#appearanceeffects 
[view] Display all appearance effects saved to your target. #appearanceffects view
[set] Set an appearance effects saved to your target. #appearanceffects set [app_effectid] [slotid]
[remove] Remove all appearance effects saved to your target. #appearanceffects remove
```
Update to perl function SendAppearanceEffectGround, simplified this since was mistaken when originally implemented, there is no way to remove the ground appearance effect thus no need to set slots, will only need to set appearance effect id now.
example usage, SendAppearanceEffectGround(441,442,443,445,446) 

`Usage: Mob::SendAppearanceEffectGround(THIS, int32 effect1, [int32 effect2 = 0], [int32 effect3 = 0], [int32 effect4 = 0], [int32 effect5 = 0], [Client* single_client_to_send_to = null])"); // @categories Script Utility`


zone/CMakeLists.txt
```diff
@@ -294,6 +294,7 @@ SET(gm_commands
     gm_commands/aggrozone.cpp
     gm_commands/ai.cpp
     gm_commands/appearance.cpp
+	gm_commands/appearanceeffects.cpp
```
Kinglykrab, line 297: Formatting. File uses 4 spaces for some reason.

zone/entity.cpp
```diff
@@ -4582,6 +4582,26 @@ void EntityList::SendUntargetable(Client *c)
 	}
 }
 
+void EntityList::SendAppearanceEffects(Client *c)
+{
+	if (!c)
+		return;
+
```
Kinglykrab, line 4589: for (const auto& mob : mob_list) that way you don’t have to increment manually.

zone/mob.cpp
```diff
@@ -2920,6 +2946,62 @@ void Mob::SendAppearanceEffect(uint32 parm1, uint32 parm2, uint32 parm3, uint32
 	safe_delete(outapp);
 }
 
+void Mob::SetAppearenceEffects(int32 slot, int32 value) 
+{
+	for (int i = 0; i < MAX_APPEARANCE_EFFECTS; i++) {
```
Kinglykrab, line 2951: You’re using < max +1 in other places, not sure your intentions there or here.

zone/gm_commands/appearanceeffects.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+void command_appearanceeffects(Client *c, const Seperator *sep)
+{
+	if (sep->arg[1][0] == '\0' || !strcasecmp(sep->arg[1], "help")) {
+		c->Message(Chat::White, "Syntax: #appearanceeffects [subcommand].");
+		c->Message(Chat::White, "[view] Display all appearance effects saved to your target. #appearanceffects view");
+		c->Message(Chat::White, "[set] Set an appearance effects saved to your target. #appearanceffects set [app_effectid] [slotid]");
+		c->Message(Chat::White, "[remove] Remove all appearance effects saved to your target. #appearanceffects remove");
+	}
+
```
Kinglykrab, line 11: Help is not listed as a sub command, also could just initialize target after the initial check if every sub command uses the target. Take a look at all the command stuff I did for a good reference.

noudess:
I've had a local hack in place for years for this.  Looking forward to testing this.

zone/CMakeLists.txt
KayenEQ: resolved

zone/entity.cpp
KayenEQ: I am fine with how I wrote this.

zone/mob.cpp
KayenEQ: Fixed the place where it did +1, forgot to change it.

zone/gm_commands/appearanceeffects.cpp
KayenEQ: This is written as intended. You can type nothing or type help and it will provide you with the commands description.
I don't see the value of initializing the target unless its actually going to be used in the correct command.

zone/mob.cpp
KayenEQ: also fixed the <= issue. Missed a few things when I rewrote it.

Kinglykrab (approved):

PR #1876, opened 2021-12-10
[Spells] Fixed issue with permanent Illusions not being consistent when zoning.

KayenEQ:
We were using old calculation for reapplying permanent illusions on zone change. Updated the code so that it uses same formula as the first time the illusion was cast. Bottom line, the way you looked before you zoned, should be same as when you zone.


zone/spell_effects.cpp
```diff
@@ -8917,3 +8810,119 @@ void Mob::SetProcLimitTimer(int32 base_spell_id, uint32 proc_reuse_time, int pro
...
+				(
+					caster->GetTarget()->GetRace(),
+					opposite_gender,
+					caster->GetTarget()->GetTexture()
+				);
+			}
+		}
+	}
+	// Racial Illusions
+	else {
```
Kinglykrab, line 8854: Formatting here.

Kinglykrab (approved):

PR #1887, opened 2021-12-14
[Skills] Configurable Exponential Decay Formula for Skill Up

### LINKED ISSUE #834: Skill Up rule Character:SkillUpModifier has no effect
The rule Character:SkillUpModifier for modifying skill up rate has no effect on some skills (weapon & defense for example) after skill level 150.

In zone/Client.cpp function CheckIncreaseSkill:

`	int32 Chance = 10 + chancemodi + ((252 - skillval) / 20);

	Chance = (Chance * RuleI(Character, SkillUpModifier) / 100);

	Chance = mod_increase_skill_chance(Chance, against_who);

	if(Chance < 1)
		Chance = 1; // Make it always possible

	if(zone->random.Real(0, 99) < Chance)`

The calculation Chance = 10 + chancemodi + ((252 - skillval) / 20) will yield negative chance from skill value 153 for skills that use -15 as chancemodi (weapon, offense and defense).

And since the rule Character:SkillUpModifier is applied before the check that chance is at least 1 the rule will then have no effect at all.

Suggested change.

Just move the check that Chance is at least 1 before the rule Character:SkillUpModifier is applied.

`	int32 Chance = 10 + chancemodi + ((252 - skillval) / 20);

	if(Chance < 1)
		Chance = 1; // Make it always possible

	Chance = (Chance * RuleI(Character, SkillUpModifier) / 100);

	Chance = mod_increase_skill_chance(Chance, against_who);

	if(zone->random.Real(0, 99) < Chance)`

mmcgarvey:
What this fixes:
The existing formula for determining whether or not to skill up could result
in negative chances, and made an assumption around the number 252.
This would ultimately result in an override that would set the chance to 1.
Fixes #834

My fix:
I created 2 new rules:
Character:SkillUpMaximumChancePercentage
Character:SkillUpMinimumChancePercentage

I changed the forumla to:
chance = ((max - min + skill_modification) * (.99^skill)) + min

This results in an exponential decay that starts at skill-modified maximum
and approaches minimum.

I decided that max-min+skill_modification should never be less than min
I also decided to continue to apply the Character:SkillUpModifier rule
post-calculation.  I do not really think this is necessary anymore, given
this new formula, but we can discuss removing it.
I chose 25 and 2 as default maximum and minimum based on feel.

Related method signature fix:

Client::mod_increase_skill_chance was changed to return a double and
accept a double as an input for chance.  This matches the actual data types
provided while calling the method and eliminates some type coersion and
resultant truncation.  Right now, this method doesn't do anything, but in the
future we could implement skill-specific training dummies that accelerate
skill ups.  I deduce that this is the purpose of this method call.


Akkadius:
Merge conflicts will need to be resolved. Still need other eyes to take a look over this.

Thanks for taking the time to contribute @mmcgarvey

mmcgarvey:
I went back and forth on implementing this behind a rule.  My reasoning for not doing that is that the old formula broke for certain skills at around 140 skill.  Since this felt like a bugfix, I opted to replace the old code.  If anyone has heartburn about that, I would love to hear about it.

Akkadius:
I don't have any notable issues with this although I can't speak to skills as much as others. Would like someone who has some knowledge around this to also take a look

noudess:
I like the changes.  Can we move this forward or do we still need someone with more skills knowledge?

PR #1888, opened 2021-12-14
[Spells] Instant Heals honor IgnoreSpellDmgLvlRestriction

mmcgarvey:
Why:
Heal Over Time spells honor the Spells:IgnoreSpellDmgLvlRestriction rule,
shouldn't instant heals honor this rule too?

The fix:
Added a check for Spells:IgnoreSpellDmgLvlRestriction in the GetActSpellHealing
method.

### DIFF
--- a/zone/effects.cpp
+++ b/zone/effects.cpp
@@ -362,8 +362,12 @@ int32 Mob::GetActSpellHealing(uint16 spell_id, int32 value, Mob* target) {
 
 		value += GetFocusEffect(focusFcHealAmtCrit, spell_id); //SPA 396 Add before critical
 
-		if (!spells[spell_id].no_heal_damage_item_mod && itembonuses.HealAmt && spells[spell_id].classes[(GetClass() % 17) - 1] >= GetLevel() - 5) {
-			value += GetExtraSpellAmt(spell_id, itembonuses.HealAmt, base_value); //Item Heal Amt Add before critical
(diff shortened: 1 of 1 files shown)


KayenEQ:
So I am almost certain the reason that exists in HOT spells is because that code was copy pasted from damage spells when it was recently implemented and requires rule if (RuleB(Spells, HOTsScaleWithHealAmt)) since live doesn't naturally apply that stat to HOTs.

Regardless, it does make sense if you are using the rule to ignore the restriction on the damage amount, then you probably would want to ignore it on the heal amount. So makes sense to have it be consistent for all use cases.

mmcgarvey:
@KayenEQ I wonder if, in this moment, we should create individual "IgnoreLevelRestriction" rules for each of these 4 use cases:  Direct Damage, DOT, Instant Heal, and HOT.  This would give us more fine-grained control.  Just a thought.

KayenEQ:
It would certainly give more flexibility, though I think it is quite unlikely anyone would ever be fine tuning that. If you probably want the level restriction or you don't.

mmcgarvey:
> It would certainly give more flexibility, though I think it is quite unlikely anyone would ever be fine tuning that. You probably want the level restriction or you don't.

Then I will leave it as a single rule.  We can always change it later, if desired.

Kinglykrab (approved):

KayenEQ (approved):

PR #1890, opened 2021-12-15
[Spells] Update to SPA 440 SE_FinishingBlowMaxLevel limit value sets HP ratio for FB

KayenEQ:
SPA 440 sets what the max level a target needs to be for it to be affected by a finishing blow. The limit value determines what the maximum hp ratio required for it to be affected. Was hardcoded at 10 pct, which is what the AA sets it at. You can now edit the limit value to change this.

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -4758,14 +4758,12 @@ void Mob::TryCriticalHit(Mob *defender, DamageHitInfo &hit, ExtraAttackOptions *
 
 bool Mob::TryFinishingBlow(Mob *defender, int &damage)
 {
-	// base2 of FinishingBlowLvl is the HP limit (cur / max) * 1000, 10% is listed as 100
 	float hp_limit = 10.0f;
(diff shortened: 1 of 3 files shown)


zone/attack.cpp
```diff
@@ -4759,7 +4759,14 @@ void Mob::TryCriticalHit(Mob *defender, DamageHitInfo &hit, ExtraAttackOptions *
...
 {
 	// base2 of FinishingBlowLvl is the HP limit (cur / max) * 1000, 10% is listed as 100
-	if (defender && !defender->IsClient() && defender->GetHPRatio() < 10) {
+	float hp_limit = 10.0f;
+	auto fb_hp_limit = std::max({ aabonuses.FinishingBlowLvl[SBIndex::FINISHING_BLOW_LEVEL_HP_RATIO], spellbonuses.FinishingBlowLvl[SBIndex::FINISHING_BLOW_LEVEL_HP_RATIO], itembonuses.FinishingBlowLvl[SBIndex::FINISHING_BLOW_LEVEL_HP_RATIO] });
+
+	if (fb_hp_limit) {
+		hp_limit = fb_hp_limit/10.0f;
+	}
+	Shout("hp limit %.2f hpratio %.2f", defender->GetHPRatio(), hp_limit);
```
Kinglykrab, line 4768: Debug message.

PR #1903, opened 2021-12-29
Added KickLinkDead

xackery:
Kick: Character goes to character select, instantly exits zone. (This means you can instant exit with no /camp timer and no LD entity in game)
WorldKick: Character goes to server select, instantly exits zone. (same symptom as above).
KickLinkDead: Character goes to server select, leaves LD character in zone (If World:EnforceCharacterLimitAtLogin is on, a message of a character already logged in will appear for Zone:ClientLinkdeadMS milliseconds)

This PR changes detected hacks or other methods a hacker may use intentionally or unintentionally to not get the advantage of what Kick originally did, by soft punishing them with the login return wait time based on if the server configures rules appropiately.

The only symptom for servers that have World:EnforceCharacterLimitAtLogin  false (the default PEQ setting) is that before you'd get put to character select (via Kick), you'd now get put to server select (via KickLinkDead)

Things needing testing:
- [ ] Ensure all clients honor this packet
- [ ] Verify what happens if the client ignores the GM Kick request. We may need to do a session close as well after other methods, to sever the connection.
- [ ] After about 3 KickLinkDead() packets, a client has a high chance of getting a black screen on RoF2. Perhaps debug the dbg.txt to see why that symptom arises.

### DIFF
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -861,7 +861,7 @@ void Client::ChannelMessageReceived(uint8 chan_num, uint8 language, uint8 lang_s
 				{
 					if(AttemptedMessages > RuleI(Chat, MaxMessagesBeforeKick))
 					{
-						Kick("Sent too many chat messages at once.");
+						KickLinkDead("Sent too many chat messages at once.");
 						return;
 					}
 					if(GlobalChatLimiterTimer)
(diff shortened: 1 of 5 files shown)


splose (approved):
Nice

Akkadius:
So this change seems welcome and I agree with the premise as currently people can exploit by fast exiting the game.

I'm worried that we rely on the hard Kick in some circumstances (e.g. a fast following save) to mitigate exploit attempts to duplicate things which to me is a much more harmful exploit surface area than fast exiting.

Without more intimate testing around this I wouldn't feel comfortable merging this as is.

Akkadius:
Who is willing to take next steps on testing this one and making sure this doesn't open up more issues? Otherwise I can see this going stale.

MidnightMusicClub:
Looks like it's going stale! Wish I could contribute more...

xackery:
Yeah.. I am caught on client side edits, but worst case, it'll be a stale PR we can reintroduce later, when I can put more time (or someone else) to test.

Akkadius:
Again, totally not opposed to this PR. 

We just need to be careful about side effects. 

I've pissed away over 20 hours playing "find and fix the regression" over the past few weeks because small changes like this become needles in the haystack after we've had 100 PR's go in over a period of a month or two and problems can be tough to pin down.

Going to close it in the meantime, if someone wants to pick it up and ensure it doesn't open up exploits by all means.

PR #1909, opened 2022-01-02
[Expansion] Change default expansion values for ALL to -1 from 0

Akkadius:
**Why**

While expansion filtering works great in most cases; 0 as a default value collides with `Classic` and while you shouldn't need to filter on classic in most cases there are still cases where its warranted.

This is primarily a database migration change to shift the default value of `min_expansion` and `max_expansion` to `-1` which is equivalent to **ALL**

```cpp
	enum ExpansionNumber {
		Classic = 0,
		TheRuinsOfKunark,
		TheScarsOfVelious,
		TheShadowsOfLuclin,
// ... truncated
	};
```

### DIFF
--- a/common/repositories/criteria/content_filter_criteria.h
+++ b/common/repositories/criteria/content_filter_criteria.h
@@ -40,14 +40,14 @@ namespace ContentFilterCriteria {
 		}
 
 		criteria += fmt::format(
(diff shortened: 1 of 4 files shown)


Akkadius:
Going to add more things and re-cut PR

PR #1912, opened 2022-01-03
[Bug Fix] Add range check to OP_PickPocket

Natedog2012:


zone/client_packet.cpp
```diff
@@ -11066,48 +11066,35 @@ void Client::Handle_OP_PickPocket(const EQApplicationPacket *app)
...
-		pick_out->coin = 0;
-		pick_out->from = victim->GetID();
-		pick_out->to = GetID();
-		pick_out->myskill = GetSkill(EQ::skills::SkillPickPockets);
-		pick_out->type = 0;
-		//if we do not send this packet the client will lock up and require the player to relog.
-		QueuePacket(outapp);
-		safe_delete(outapp);
+	}
+	else if (!CombatRange(victim, 3.0)) { //Give benefit of the doubt due to pathing NPCs
```
mackal, line 11084: Distance is actually just 15.

mackal (changes requested):

zone/client_packet.cpp
Natedog2012: You can sometimes send it from a longer distance when the mob is pathing and the server gets confused and will refuse it

mackal: the distance is still just flat 15, not Combat Range

Kinglykrab (approved):

mackal (approved):

PR #1915, opened 2022-01-03
[Bug Fix] XTarget Changes Causing Crashes

mmcgarvey:
Added nullptr check when processing XTargetAutoHaters for pet owner.

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -2571,7 +2571,7 @@ bool NPC::Death(Mob* killer_mob, int32 damage, uint16 spell, EQ::skills::SkillTy
 
 		entity_list.RemoveFromAutoXTargets(this);
 
-		if (killer && killer->GetUltimateOwner() && killer->GetUltimateOwner()->IsClient()) {
+		if (killer && killer != 0 && killer->GetUltimateOwner() && killer->GetUltimateOwner()->IsClient()) {
 			killer->GetUltimateOwner()->CastToClient()->ProcessXTargetAutoHaters();
 		}
 		uint16 emoteid = this->GetEmoteID();
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -2571,7 +2571,7 @@ bool NPC::Death(Mob* killer_mob, int32 damage, uint16 spell, EQ::skills::SkillTy
 
 		entity_list.RemoveFromAutoXTargets(this);
 
-		if (killer && killer != nullptr && killer != 0 && killer->GetUltimateOwner() && killer->GetUltimateOwner()->IsClient()) {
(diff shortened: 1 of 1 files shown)


mmcgarvey:
@Akkadius @mackal 

I believe that these two commits could resolve the crashing issues.

zone/attack.cpp
```diff
@@ -2571,7 +2571,7 @@ bool NPC::Death(Mob* killer_mob, int32 damage, uint16 spell, EQ::skills::SkillTy
 
 		entity_list.RemoveFromAutoXTargets(this);
 
-		if (killer->GetUltimateOwner() && killer->GetUltimateOwner()->IsClient()) {
+		if (killer && killer != 0 && killer->GetUltimateOwner() && killer->GetUltimateOwner()->IsClient()) {
```
mackal, line 2574: Should be just `killer != nullptr` (yes I know the rest of the code is gross)

mackal (changes requested):

zone/attack.cpp
mmcgarvey: So, do you want the line to be:

if (killer != nullptr && killer != 0 && killer->GetUltimateOwner() && killer->GetUltimateOwner()->IsClient()) {
if (killer && killer != nullptr && killer->GetUltimateOwner() && killer->GetUltimateOwner()->IsClient()) {
if (killer != nullptr && killer->GetUltimateOwner() && killer->GetUltimateOwner()->IsClient()) {

Akkadius: Guys lets close the loop on this

mackal: The last one, they all do the same and likely compile down to the same thing. Someday I'll tackle rewriting the section of code, it's ugly and likely doesn't need that many checks on the killer pointer, but that first if is too gross to unpack atm

mmcgarvey: I will just add what was requested to what I already had.  It will be overkill, but I think it will get us beyond the issue tonight.

mmcgarvey: I must have been typing out my response at the same time as you.  I will get your requested change in now.

mmcgarvey: @mackal @Akkadius 
Latest commit reflects the requested changes.

mackal (approved):

Kinglykrab (approved):

Akkadius:
PEQ feedback since latest changes

“ lots of lag tonight, had issues in Anguish with OMM falling off extended target window, and multiple crashes to char select during the fight. Issues with following / chars needing to be camped and relogged and connected.”

Akkadius:
I called this out in the previous PR as I’m concerned about the frequency of entity list calls and hot paths can create nasty performance issues especially in raids. Not sure if you’re in discord but would be easier to communicate with PEQ players on the issue

PR #1917, opened 2022-01-05
[Bug Fix] Fix issue where you can set your title to titles you don't have.

Kinglykrab:
- Fixes exploit where you can use titles that are in the database regardless of if you have it.
- Add TitleManager::HasTitle(client, title_id) help method.
- Cleanup entire titles.cpp/titles.h logic, loops, and formatting.

### DIFF
--- a/zone/titles.cpp
+++ b/zone/titles.cpp
@@ -19,6 +19,7 @@
 #include "../common/eq_packet_structs.h"
 #include "../common/string_util.h"
 #include "../common/misc_functions.h"
+#include "../common/repositories/titles_repository.h"
 
 #include "client.h"
 #include "entity.h"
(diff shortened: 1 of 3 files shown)


zone/titles.cpp
```diff
@@ -34,349 +35,330 @@ TitleManager::TitleManager() {
...
+		title.suffix = row[12];
+		title.titleset = std::stoi(row[13]);
+		titles.push_back(title);
 	}
 
 	return true;
 }
 
-EQApplicationPacket *TitleManager::MakeTitlesPacket(Client *c)
+EQApplicationPacket *TitleManager::MakeTitlesPacket(Client *client)
```
mackal, line 70: We could rewrite this to use [SerializeBuffer](https://github.com/EQEmu/Server/blob/ffa968f64fca86f106bf502e1be2444749a1b675/common/serialize_buffer.h#L9) instead of having to calculate the size.

Natedog2012 (approved):

splose (approved):

PR #1918, opened 2022-01-05
[XTarget] Performance Improvement After Corpse Change

mmcgarvey:
Removed a conditional that was rendered obsolete by moving the addition of a
mob to the auto haters list to fire after an IsValidXTarget check.  This
made an entity_list call unnecessary.  [zone/client.cpp]

Removed said unnecessary entity_list call.  [zone/client.cpp]

Removed a superfluous call to ProcessXTargetAutoHaters [zone/attack.cpp]

I spent a few hours today, testing to ensure that corpses still won't get stuck in the XTarget window, with these modifications.  I was unable to get a corpse to stay in the XTarget window, which is good.  This is the desired behavior.


mmcgarvey:
@Akkadius @mackal @Kinglykrab 

This is the best I can come up with to clean up the performance mess that my last XTarget PR introduced.

Akkadius (approved):

PR #1921, opened 2022-01-10
[Bug Fix] Fix issue with mobs summoning PCs into ceilings

noudess:
This has been in production for a few weeks on my server now and has completely eliminated the summon issue that cropped up when the summon code was changed a couple of months ago.

Talking with @mackal I believe he feels we need something more comprehensive than this, ala checking ceiling or something similar.

However, I believe this correct the issue for all cases I have seen.  I can't say if it will solve everyone's issue, but the change is small and makes sense.


fryguy503 (approved):
LGTM

Kinglykrab (approved):

PR #1934, opened 2022-01-19
Remove content flag filters so all loot gets added to shared db.

randy-girard:
This line (https://github.com/EQEmu/Server/blob/master/zone/loottables.cpp#L53) checks the shared database, but if any global loot has a content flag set, that loot does not get loaded, and this line returns NULL. The content flag values are empty at the shared DB level, so the where clause essentially is only loading loot where the flag is NULL or empty string.


randy-girard:
Turns out it looks like there are more content flag related issues beyond this. I will open a new PR when i correct those as well.

PR #1936, opened 2022-01-19
[Spells] Support for bards using Disciplines while casting or /melody.

KayenEQ:
Bards on live can use disciplines while casting a song or using /melody. This update provides support for that.

Also added  live like support for spell field 'cast not standing' to allowing allow when divine aura is present on the character.


Kinglykrab (approved):

KayenEQ:
Also added live like support for spell field 'cast not standing' to allowing allow when divine aura is present on the character.

PR #1940, opened 2022-01-20
[Content Filtering] Updates contents flags to be checked at runtime.

randy-girard:
This helps remove the need to restart server when changing content flags. If you change a flag from enabled to disabled or back, all it requires is a `#reloadcontentflags` and then a `#repop`. If you change the flags on a loottable or lootdrop, then you have to also do a `#hotfix` before `#repop`.

### DIFF
--- a/common/content/world_content_service.cpp
+++ b/common/content/world_content_service.cpp
@@ -157,11 +157,11 @@ bool WorldContentService::IsContentFlagDisabled(const std::string &content_flag)
 
 bool WorldContentService::DoesPassContentFiltering(const ContentFlags &content_flags)
 {
-	if (content_flags.min_expansion >= 0 && current_expansion < content_flags.min_expansion) {
+	if (content_flags.min_expansion > Expansion::EXPANSION_ALL && current_expansion <= content_flags.min_expansion) {
 		return false;
 	}
 
-	if (content_flags.max_expansion >= 0 && current_expansion > content_flags.max_expansion) {
+	if (content_flags.max_expansion > Expansion::EXPANSION_ALL && current_expansion >= content_flags.max_expansion) {
 		return false;
 	}
 
--- a/common/content/world_content_service.cpp
+++ b/common/content/world_content_service.cpp
@@ -155,23 +155,24 @@ bool WorldContentService::IsContentFlagDisabled(const std::string &content_flag)
 	return false;
 }
 
-bool WorldContentService::DoesPassContentFiltering(const ContentFlags_Struct& content_flags) {
-	if(content_flags.min_expansion >= 0 && current_expansion < content_flags.min_expansion) {
+bool WorldContentService::DoesPassContentFiltering(const ContentFlags &content_flags)
+{
+	if (content_flags.min_expansion >= 0 && current_expansion < content_flags.min_expansion) {
 		return false;
 	}
 
-	if(content_flags.max_expansion >= 0 && current_expansion > content_flags.max_expansion) {
+	if (content_flags.max_expansion >= 0 && current_expansion > content_flags.max_expansion) {
 		return false;
 	}
 
-	for(auto content_flag : SplitString(content_flags.content_flags)) {
-		if(!contains(GetContentFlagsEnabled(), content_flag)) {
+	for (const auto& content_flag: SplitString(content_flags.content_flags)) {
+		if (!contains(GetContentFlagsEnabled(), content_flag)) {
 			return false;
 		}
 	}
 
-	for(auto content_flag : SplitString(content_flags.content_flags_disabled)) {
-		if(!contains(GetContentFlagsDisabled(), content_flag)) {
+	for (const auto& content_flag: SplitString(content_flags.content_flags_disabled)) {
+		if (!contains(GetContentFlagsDisabled(), content_flag)) {
 			return false;
 		}
 	}
--- a/common/content/world_content_service.cpp
+++ b/common/content/world_content_service.cpp
@@ -157,11 +157,11 @@ bool WorldContentService::IsContentFlagDisabled(const std::string &content_flag)
 
 bool WorldContentService::DoesPassContentFiltering(const ContentFlags &content_flags)
 {
-	if (content_flags.min_expansion > Expansion::EXPANSION_ALL && current_expansion <= content_flags.min_expansion) {
+	if (content_flags.min_expansion > Expansion::EXPANSION_ALL && current_expansion < content_flags.min_expansion) {
 		return false;
 	}
 
-	if (content_flags.max_expansion > Expansion::EXPANSION_ALL && current_expansion >= content_flags.max_expansion) {
+	if (content_flags.max_expansion > Expansion::EXPANSION_ALL && current_expansion > content_flags.max_expansion) {
 		return false;
 	}
 
--- a/common/content/world_content_service.cpp
+++ b/common/content/world_content_service.cpp
@@ -139,6 +139,47 @@ bool WorldContentService::IsContentFlagEnabled(const std::string &content_flag)
 	return false;
 }
 
+/**
+ * @param content_flag
+ * @return
+ */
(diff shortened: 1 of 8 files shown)


common/content/world_content_service.cpp
```diff
@@ -139,6 +139,47 @@ bool WorldContentService::IsContentFlagEnabled(const std::string &content_flag)
...
+
+bool WorldContentService::DoEnabledFlagsPass(const char *content_flags) {
+	if(strlen(content_flags) == 0)
+		return true;
+
+	auto split_content_flags = SplitString(std::string(content_flags));
+
+	for (auto content_flag : split_content_flags)
+		if(this->IsContentFlagEnabled(content_flag))
+			return true;
```
Kinglykrab, line 165: Brackets. `this` is always implicit.

common/content/world_content_service.cpp
```diff
@@ -139,6 +139,47 @@ bool WorldContentService::IsContentFlagEnabled(const std::string &content_flag)
...
+}
+
+bool WorldContentService::DoDisabledFlagsPass(const char *content_flags) {
+	if(strlen(content_flags) == 0)
+		return true;
+
+	auto split_content_flags = SplitString(std::string(content_flags));
+
+	for (const auto& content_flag : split_content_flags)
+		if(this->IsContentFlagDisabled(content_flag))
```
Kinglykrab, line 177: Same comment as above.

Kinglykrab (commented):
Needs some formatting, we prefer opening and closing curlies for explicit scoping on most every condition.

randy-girard:
@Kinglykrab Is this look good?

randy-girard:
@Akkadius thoughts when you get a chance, based on our discord convos.

Kinglykrab:
Looks good. As long as no one else cares about the implicitly scoped conditions in here, should be good.

common/shareddb.cpp
```diff
@@ -2075,6 +2073,8 @@ void SharedDatabase::LoadLootTables(void *data, uint32 size) {
 			  loottable.mincash,
 			  loottable.maxcash,
 			  loottable.avgcoin,
+			  COALESCE(loottable.content_flags, ''),
```
Akkadius, line 2076: This should be added at the end, not the middle

Also should not be transforming the content; just safely check for it in the row iterator and return an empty string if its null or empty. Content flags can be comma separated in the fields to have multiple so we need to be able to pass all of them

common/content/world_content_service.cpp
```diff
@@ -139,6 +139,51 @@ bool WorldContentService::IsContentFlagEnabled(const std::string &content_flag)
...
+	for (auto content_flag : split_content_flags) {
+		if(IsContentFlagEnabled(content_flag)) {
+			return true;
+		}
+	}
+
+	return false;
+}
+
+bool WorldContentService::DoDisabledFlagsPass(const char *content_flags) {
```
Akkadius, line 172: We should use std::string passed in to this function

common/content/world_content_service.cpp
```diff
@@ -139,6 +139,51 @@ bool WorldContentService::IsContentFlagEnabled(const std::string &content_flag)
...
+	if(strlen(content_flags) == 0)
+		return true;
+
+	auto split_content_flags = SplitString(std::string(content_flags));
```
Akkadius, line 176: Inline the split string in the for loop so we're not allocating extra here

common/content/world_content_service.cpp
```diff
@@ -139,6 +139,51 @@ bool WorldContentService::IsContentFlagEnabled(const std::string &content_flag)
...
+	for (auto &f: GetContentFlags()) {
+		if (f.flag_name == content_flag && f.enabled == false) {
+			return true;
+		}
+	}
+
+	return false;
+}
+
+bool WorldContentService::DoEnabledFlagsPass(const char *content_flags) {
```
Akkadius, line 157: Also pass `std::string` into this

Name should be a bit more consistent with the rest of the class but just imply "many"

```cpp
IsAnyContentFlagEnabled(std::string content_flags)
IsAnyContentFlagDisabled(std::string content_flags)
```

Also; it would be better to split the string into a vector before entering this function and just pass a vector of strings into the function which makes your code easier to work with in the function itself

```cpp
IsAnyContentFlagEnabled(std::vector<std::string> content_flags)
IsAnyContentFlagDisabled(std::vector<std::string> content_flags)
```

Then you can do 

```cpp
if (content_flags.empty()) {
    return true;
}
```

Akkadius:
> This helps remove the need to restart server when changing content flags. If you change a flag from enabled to disabled or back, all it requires is a `#reloadcontentflags` and then a `#repop`. If you change the flags on a loottable or lootdrop, then you have to also do a `#hotfix` before `#repop`.

First, thank you for taking the effort to put up a pull request and be receptive to feedback, not everyone is capable of handling that.

Your code looks sound overall, left some minor ergonomic comments around naming and interfacing with the class primitives.

Taking a step back, this doesn't quite save you much in the way of what your original comment was meant to solve. You can still hotfix without rebooting the server and even if you are adding new loot entries you're still stuck rebooting the server because data sizes change. Regardless, its still good to get a runtime implementation of filtering in place because later down the road we will need it when expansions gets off the ground again.

Thank you

common/content/world_content_service.cpp
Akkadius: The other thing I'm not seeing included in the runtime filtering here is that expansions is not included whatsoever in the filtering logic at runtime. That was being enforced at the query level

shared_memory/main.cpp
```diff
@@ -167,9 +167,6 @@ int main(int argc, char **argv)
 
 
 	content_service.SetCurrentExpansion(RuleI(Expansion, CurrentExpansion));
-	content_service.SetDatabase(&database)
```
Akkadius, line 170: This shouldn't have been removed

randy-girard: I figured this wasn't needed anymore since the passing in of the content flags were no longer used.

common/content/world_content_service.cpp
randy-girard: I will work on changing to this. I tried this originally and was running into issues with it. I will take another stab.

shared_memory/main.cpp
Akkadius: It should still be there because that is the uniform initializer logic that loads both the flags and sets the expansion context.

If we change the source of the current expansion to be pulled from somewhere it will be handled by the loader functions in the content service rather than being passed in 

Doesn't hurt to keep it

zone/loottables.cpp
```diff
@@ -54,6 +54,13 @@ void ZoneDatabase::AddLootTableToNPC(NPC* npc,uint32 loottable_id, ItemList* ite
 	if (!lts)
 		return;
 
+
+	if(!content_service.DoEnabledFlagsPass(lts->content_flags))
```
Akkadius, line 58: Per my comments earlier around having one filtering function, we should have a function here that passes in the 4 filtering columns from the table

```
min_expansion
max_expansion
content_flags
content_flags_disabled
```

Pass all 4 to a superset filtering function that can determine whether or not we should return

eg


```cpp
!content_service.DoesPassContentFiltering(min_expansion, max_expansion, content_flags, content_flags_disabled)
```

shared_memory/main.cpp
randy-girard: So this would end up making it always need to do a hotfix, which I assume is ok?

zone/loottables.cpp
randy-girard: Yes, I will update to do this

Akkadius:
This is still on my list. Haven't had an abundance of time. Will follow up soon.

Akkadius:
Made a handful of tweaks. Largely looked good. Nice work!

Thanks for making the adjustments after the feedback

Akkadius (approved):

PR #1945, opened 2022-01-25
[Bug Fix (faction)] Do not award faction if NPC is charmed.

noudess:
@nilbog told me that faction isn't awarded for quests if the quest npc is charmed.  He knew this was classic but wasn't sure about live.  I logged onto live and tested.  Sure enough, no faction for bone chips turn in to Lashun when she was charmed.  The rest worked (she cast cure disease) and I got a ding.

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -2411,7 +2411,7 @@ bool NPC::Death(Mob* killer_mob, int32 damage, uint16 spell, EQ::skills::SkillTy
 		give_exp_client = give_exp->CastToClient();
 
 	//do faction hits even if we are a merchant, so long as a player killed us
-	if (give_exp_client && !RuleB(NPC, EnableMeritBasedFaction))
+	if (!IsCharmed() && give_exp_client && !RuleB(NPC, EnableMeritBasedFaction))
 		hate_list.DoFactionHits(GetNPCFactionID());
 
 	bool IsLdonTreasure = (this->GetClass() == LDON_TREASURE);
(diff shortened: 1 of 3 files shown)


mackal:
What about killing a charmed NPC? This might need to be moved deeper. If not, we need this guard added to more places as well.

noudess:
> What about killing a charmed NPC? This might need to be moved deeper. If not, we need this guard added to more places as well.

I can't attack a charmed NPC.  Is there a way?  If you identify other areas, I can add this to those spots.

mackal:
> > What about killing a charmed NPC? This might need to be moved deeper. If not, we need this guard added to more places as well.
> 
> I can't attack a charmed NPC. Is there a way? If you identify other areas, I can add this to those spots.

/duel the charmer?

noudess:
> > > What about killing a charmed NPC? This might need to be moved deeper. If not, we need this guard added to more places as well.
> > 
> > 
> > I can't attack a charmed NPC. Is there a way? If you identify other areas, I can add this to those spots.
> 
> /duel the charmer?

I'll test this tonight.

mackal:
https://github.com/EQEmu/Server/blob/e4f2aec11eafaa96a84d1e098cb88de37a72b5ff/zone/client.cpp#L8658

Needs something as well.

noudess:
> https://github.com/EQEmu/Server/blob/e4f2aec11eafaa96a84d1e098cb88de37a72b5ff/zone/client.cpp#L8658
> 
> Needs something as well.

Ok, I can do that.  What about Quests calling SetFactionLevel and SetFactionLevel2?  I don't see a way to fix those.  Obscure enough not to care?

mackal (approved):

mackal:
We should probably not merge this for a day or two to give people time to comment on maybe adding a rule (I think that would be a bit much, but who knows!)

Kinglykrab (approved):

PR #1946, opened 2022-01-26
[Bug Fix] Any use of TempName left old clean_name.

noudess:
clean_name is cached by the 1st use of GetCleanName().  Any use of TempName() did not cause that cache to be invalidated.

One resulting bug is that #petname would change the pets name, but would not alter some messages that used GetCleanName() until a zone or a camp.

Simply setting *clean_name to NULL forces a recache on the next call to GetCleanName().

### DIFF
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -3094,6 +3094,7 @@ void Mob::TempName(const char *newname)
 	char temp_name[64];
 	char old_name[64];
 	strn0cpy(old_name, GetName(), 64);
+	*clean_name = (char) NULL;
 
 	if(newname)
 		strn0cpy(temp_name, newname, 64);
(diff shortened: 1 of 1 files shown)


Kinglykrab (approved):
Nice. Think I’ve seen this with TempName() on pets and stuff where they show the wrong name in the chat due to this.

zone/mob.cpp
```diff
@@ -3094,6 +3094,7 @@ void Mob::TempName(const char *newname)
 	char temp_name[64];
 	char old_name[64];
 	strn0cpy(old_name, GetName(), 64);
+	*clean_name = (char) NULL;
```
mackal, line 3097: This should just be

```cpp
clean_name[0] = 0;
```

mackal (changes requested):

zone/mob.cpp
noudess: Why?

mackal (approved):

PR #1954, opened 2022-01-31
[Spells] Major update to Bard song pulsing, Bard item clicks while singing, and spell casting restriction logic.

KayenEQ:
This is a large update to the spell casting logic to facilitate fixes for bards and other interrelated casting related issues.

Update to bard song pulsing logic to make it consistent to how live bards have always worked and standardize it with the general eqemu casting logic. Should resolve many outlying bard issues.

- Bard when pulsing will recast the song every 6 seconds, buff duration will be refreshed.

-  Pulse restrictions will be accurate to live in regards to if pulsing while DA, Feared, Silenced. 

Bards on ROF2 client can now use item clickies while singing or using /melody with live like mechanics.

-  If you use a clicky that has a cast time while your actively casting a song, then the song will immediately cancel and be   replaced by the clickies spell bar and any pulsing bard songs will be stopped. 
- If you are actively pulsing a song and use a clicky with a cast time, then the pulsing will be stopped.
- Instant cast clickies can be used while pulsing or actively casting and will NOT stop the pulse or cast.

- Bard Disciplines can also now be used while casting a clicky item.

Major updates to how and when spell casting can fail due to restrictions on the caster, target or zone to more consistent to live. We were a bit all over the place with this. In most instances now casts will immediately fail if casting requirements are not met. Checks are also improved down stream of initial casting. Overall everything should just be tighter and more consistent in regards to if a spell can be cast or affect a target. Should also see less instances of recast timers being set even though cast failed.

Added live like messages for when a cast fails due to target or caster restrictions using spell file fields 'CastRestriction' or 'caster_requirment_id'. These messages will give you the details of WHY your spell failed. Which is this case greatly enhances the usability of these fields, which offer a huge amount of flexibility when designing spells. See spdat.h for list of all associated ids and what they restrict. These descriptions are found in dbstr_us of modern clients, but not for ROF2.

Bindsight when cast as GM will now function properly

Buffs will numhits counter will now display instantly in all cases.

Other minor optimizations to spell casting logic.

* I know this is a lot and I am in no rush to merge, but I am at the point where I can't find anything not working with it. Any suggestions or recommendations are welcome.


zone/spell_effects.cpp
```diff
@@ -8189,6 +8193,1098 @@ bool Mob::PassCastRestriction(int value)
...
+		If target_requirement is false then use the caster requirement message.
+		Added support for different messages for certain high yield restrictions based on if
+		target or caster requirements.
+
+	*/
+
+	std::string msg = "";
+
+	if (target_requirement) {
+		msg = "Your target does not meet the spell requirements. ";
```
Kinglykrab, line 8213: Extra space here? The messages below add a space after these.

KayenEQ: Intentional, because that is combined with the specific message string. That is how live, displays it.

Kinglykrab: Ah okay, so the space here and the additional space added in the Message is intended, got it.

common/spdat.cpp
```diff
@@ -1462,6 +1462,19 @@ bool IsInstrumentModAppliedToSpellEffect(int32 spell_id, int effect)
 	//Allowing anything not confirmed to be restricted / allowed to receive modifiers, as to not inhbit anyone making custom bard songs.
 }
 
+bool IsPulsingBardSong(int32 spell_id)
+{
```
mackal, line 1466: Should add an IsValidSpell() check

KayenEQ: sure, done.

common/spdat.h
```diff
@@ -193,6 +191,15 @@
 #define MAX_APPEARANCE_EFFECTS 20 //Up to 20 Appearance Effects can be saved to a mobs appearance effect array, these will be sent to other clients when they enter a zone (This is arbitrary)
 #define MAX_CAST_ON_SKILL_USE 36 //Actual amount is MAX/3
 
+#define ZONE_TUTORIAL 183
+#define ZONE_LOAD 184
```
Kinglykrab, line 195: We have constants for all zones already.

zone/spells.cpp
```diff
@@ -611,77 +521,376 @@ void Mob::SendBeginCast(uint16 spell_id, uint32 casttime)
...
+				MessageString(Chat::Red, CAST_OUTDOORS);
+				LogSpells("Spell casting canceled [{}] : can not cast outdoors.", spell_id);
+				return false;
+			}
+		}
+		/*
+			Zones where you can not gate.
+		*/
+		if (IsClient() &&
+			(zone->GetZoneID() == ZONE_TUTORIAL || zone->GetZoneID() == ZONE_LOAD) &&
```
Kinglykrab, line 713: Zones::TUTORIAL and Zones::LOAD.

zone/spells.cpp
```diff
@@ -611,77 +521,376 @@ void Mob::SendBeginCast(uint16 spell_id, uint32 casttime)
...
+			CastToClient()->Admin() < AccountStatus::QuestTroupe) {
+			if (IsEffectInSpell(spell_id, SE_Gate) ||
+				IsEffectInSpell(spell_id, SE_Translocate) ||
+				IsEffectInSpell(spell_id, SE_Teleport)) {
+				Message(0, "The Gods brought you here, only they can send you away.");
```
Kinglykrab, line 718: Chat::White

Kinglykrab (approved):

common/spdat.h
KayenEQ: Ok, was not aware. Removed.

zone/spells.cpp
KayenEQ: replaced

zone/spells.cpp
KayenEQ: why not

PR #1958, opened 2022-02-03
[Bug Fix] Illusions will now properly display armor to other clients when they zone in.

KayenEQ:
This update resolves an issue where if a mob already has an illusion on them, when a new client zones in, the armor would not display properly on the mob with the illusion to that client who just zoned in.


noudess:
Awesome.

noudess (commented):
LGTM

noudess (approved):

Kinglykrab (approved):

PR #1975, opened 2022-02-08
[API] Perl functions added to apply spell effects directly to NPCs without requiring buffs.

KayenEQ:
These functions are extension of the 'npc_spells_effects' table which allow you to directly apply spell effect bonuses to NPC's without requiring actual spell buffs. You apply the buff as if you were designing a custom spell with a single effect. These buffs work just like regular spell buffs, can be overwritten by more powerful real buffs, or block weaker ones.

```
$npc->AddAISpellEffect(spell_effect id , base_value, limit_value, max_value)
```
Example: $npc->AddAISpellEffect(5 , 100, 0, 0);
This uses SPA 5 that increase strength stat, so this will increase your npcs STR by 100

```
$npc->RemoveAISpellEffect(spelleffect_id)
```
Example: $npc->RemoveAISpellEffect(5);
This will remove all quest or table applied STR buffs.

Using this obviously requires knowledge of how to make spell effects. This works with any spell effect that typically is found on buffs. Very useful in adding modifiers that otherwise would require a custom buff to make, like critical hits, damage shields ect.

Usual disclaimer, I don't code in Lua, anyone is welcome to port these over.

### DIFF
--- a/zone/perl_npc.cpp
+++ b/zone/perl_npc.cpp
@@ -1849,7 +1849,7 @@ XS(XS_NPC_AddAISpellEffect); /* prototype to pass -Wmissing-prototypes */
 XS(XS_NPC_AddAISpellEffect) {
 	dXSARGS;
 	if (items != 5)
-		Perl_croak(aTHX_ "Usage: NPC::AddAISpellEffect(THIS, spell effect id, base value, limit value, max value)"); // @categories Spells and Disciplines
+		Perl_croak(aTHX_ "Usage: NPC::AddAISpellEffect(THIS, spell_effect id, base_value, limit_value, max_value)"); // @categories Spells and Disciplines
 	{
 		NPC *THIS;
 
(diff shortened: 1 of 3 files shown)


Kinglykrab (approved):

zone/perl_npc.cpp
```diff
@@ -1969,6 +2003,8 @@ XS(boot_NPC) {
 	newXSproto(strcpy(buf, "StartSwarmTimer"), XS_NPC_StartSwarmTimer, file, "$$");
 	newXSproto(strcpy(buf, "StopWandering"), XS_NPC_StopWandering, file, "$");
 	newXSproto(strcpy(buf, "UpdateWaypoint"), XS_NPC_UpdateWaypoint, file, "$$");
+	newXSproto(strcpy(buf, "AddAISpellEffect"), XS_NPC_AddAISpellEffect, file, "$$$$$");
+	newXSproto(strcpy(buf, "RemoveAISpellEffect"), XS_NPC_RemoveAISpellEffect, file, "$");
```
Kinglykrab, line 2007: Please alphabetize.

KayenEQ: oops meant to go back and do that.

PR #1984, opened 2022-02-10
[Bots] Fix bot spawn when bot id = char_id

neckkola:
When a bot owner id is equal to a bot id this bot will be spawned when the character zones.   To test; bot owner spawns a few bots and creates a group.  Zone.  When appearing, a new bot will be spawned.  This bot has a bot id that is equal to the char id of the client.  This is an edge case for when the bot owner's id matches a bot id.


Akkadius:
Thanks for submitting this pull request @neckkola and taking the time to do so

I see you're querying a view; is it possible to make an equivalent fix not using a view or are bots still heavily coupled to using views?

neckkola:
Thanks @Akkadius for the question and reviewing the PR.  I very much enjoying contributing back to the community.  The bot subsystem uses views a handful of times throughout.  As such my approach was to follow along and not deviate.  That said, I am happy to update to avoid the use of the view and stay true to the design patterns used, just let me know.  Thank you.

neckkola:
Good morning.  Apologies for the delay.  I have looked at the request and have modified to not use the view, and instead just a slight modification to the original query.  For your review and consideration and thank you.

Akkadius:
> Good morning. Apologies for the delay. I have looked at the request and have modified to not use the view, and instead just a slight modification to the original query. For your review and consideration and thank you.

Thank you for taking the time to make this tweak. Looks good to me assuming you've tested

Akkadius (approved):

PR #2000, opened 2022-02-14
[Spells] AE Duration effect (Rains) will now work with Target Ring and PBAE spells.

KayenEQ:
Spells with AE Duration effects can be applied to Target Ring cast spells and PBAE spells. This update provides support for both of those.

### DIFF
--- a/zone/beacon.cpp
+++ b/zone/beacon.cpp
@@ -52,10 +52,10 @@ extern Zone* zone;
 
 // if lifetime is 0 this is a permanent beacon.. not sure if that'll be
 // useful for anything
-Beacon::Beacon(Mob *at_mob, int lifetime)
+Beacon::Beacon(const glm::vec4 &in_pos, int lifetime)
 :Mob
 (
-	nullptr, nullptr, 0, 0, 0, INVISIBLE_MAN, 0, BT_NoTarget, 0, 0, 0, 0, 0, at_mob->GetPosition(), 0, 0, 0,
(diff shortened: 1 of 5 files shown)


zone/beacon.cpp
```diff
@@ -121,16 +121,23 @@ bool Beacon::Process()
...
 	spell_id = cast_spell_id;
 	this->resist_adjust = resist_adjust;
 	spell_iterations = spells[spell_id].aoe_duration / 2500;
 	spell_iterations = spell_iterations < 1 ? 1 : spell_iterations;	// at least 1
-	if (spells[spell_id].aoe_max_targets)
+	if (spells[spell_id].aoe_max_targets) {
 		max_targets = spells[spell_id].aoe_max_targets;
+	}
+
+	if (spells[cast_spell_id].target_type == ST_Ring) {
```
mackal, line 137: I think I would rather switch `Beacon::Beacon` to take a position rather than a Mob and set the location that way (that's all it's used for in the ctor)

mackal: And if you're wondering if we should take a vec3 or vec4, live always has their entities facing north :P (which means ones with push always push north on live, we currently don't have that issue)

KayenEQ: But it's so easy and it just works.

mackal (approved):

Kinglykrab (approved):

PR #2001, opened 2022-02-15
[Bug Fix] Memory Corruption issue with SE_LimitToSkill when setting bonus.

KayenEQ:
Memory Corruption issue with SE_LimitToSkill when setting bonus. This will fix issues with skill procs.


mackal (approved):

zone/bonuses.cpp
```diff
@@ -5568,6 +5569,7 @@ void Mob::NegateSpellEffectBonuses(uint16 spell_id)
 				case SE_SkillProcAttempt: {
 					for (int e = 0; e < MAX_SKILL_PROCS; e++)
 					{
+						Shout("Fail NEGATE");
```
martinlindhe, line 5572: Debug text

PR #2003, opened 2022-02-15
[Bug] Fixed trade items record log

cybernine186:
LogPlayerHandin was not recording the item details of the trade. It should be inserting the trade details into `qs_player_handin_record_entries` but it was not running the query because of missing QueryDatabase function.


jmkca (approved):

Akkadius:
Thanks @cybernine186 for submitting this!

Akkadius (approved):

PR #2007, opened 2022-02-17
[Quest API] Add IsTargetClient() and IsTargetNPC() to Perl/Lua.

Kinglykrab:
- Add $mob->IsTargetClient() to Perl.
- Add $mob->IsTargetNPC() to Perl.
- Add mob:IsTargetClient() to Lua.
- Add mob:IsTargetNPC() to Lua.
- Allows for shorthand checking of if mob's current target is client or NPC.

### DIFF
--- a/zone/lua_mob.cpp
+++ b/zone/lua_mob.cpp
@@ -2478,6 +2478,17 @@ void Lua_Mob::SetBuffDuration(int spell_id, int duration) {
 	self->SetBuffDuration(spell_id, duration);
 }
 
+
+bool Lua_Mob::IsTargetClient() {
+	Lua_Safe_Call_Bool();
+	return self->IsTargetClient();
+}
+
+bool Lua_Mob::IsTargetNPC() {
+	Lua_Safe_Call_Bool();
+	return self->IsTargetNPC();
+}
+
 luabind::scope lua_register_mob() {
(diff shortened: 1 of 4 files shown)


Akkadius:
This looks good; however I do have to ask about the use-case for something like this when you can already chain this already flexibly

We already have checks for all primitive types that can be used

eg

```
IsBot
IsClient
IsNPC
IsHorse
IsPet
IsObject
IsDoor
IsMerc
IsTrap

.. etc.
```

While the usage is verbose; it forces the user to check for validity through the chain

```
c && c->GetTarget() && c->GetTarget()->IsNPC
```

Whereas `c->IsTargetNPC` makes assumptions about having a target to begin with and you will need to duplicate checks like this for all primitive types. It seems a bit redundant even if it does save a tiny bit of work

Curious your thoughts

Kinglykrab:
Just wanted an easy shorthand to do something like this.
```pl
my $var = $client->IsTargetClient() ? “Client” : $client->IsTargetNPC() ? “NPC” : “Other”;
```
Versus
```pl
my $var = $client->GetTarget() && $client->GetTarget()->IsClient() ? “Client” : $client->GetTarget() && $client->GetTarget()->IsNPC() ? “NPC” : “Other”;
```

PR #2008, opened 2022-02-17
[Commands] Bug fix for #logs command.

Kinglykrab:


Akkadius:
Tested, works well. Thank you for addressing.

Some things I had in my mind to improve this long before you refactored it. Not suggesting you should even pursue them here but it would be nice quality of life.

Add clickable saylinks to the commands in the listing so you can just click `#logs list` or `#logs reload`

Similarly, adding a text table where you can click saylinks log levels 0 - 1 - 2 - 3 for each category, for each log pipe. The list would need clickable saylink pagination as well.

Akkadius (approved):

PR #2012, opened 2022-02-18
[API] perl added GetNPCStat(identifier)

KayenEQ:
This returns the stat value for everything you can set using $npc->ModifyNPCStat(string key, string value)
Beats having to look up every single individual npc stat return quest function for simple use cases.

Will likely expand this in the future to get more values from table.


Kinglykrab (approved):
Looks good.

PR #2017, opened 2022-02-22
Rule to apply Spell Dmg and Heal Amount stats as a percentage instead of flat value.

catapultam-habeo:
Rule to apply Spell Dmg and Heal Amount stats as a percentage instead of flat value.

This was useful for my server. Each point of these stats is reflected as a 1% increase in spell effectiveness, subject to the 5-level spread and recalculation around casting time, etc (unless the rules to disable those are in effect).


catapultam-habeo:
I actually see a mistake that I introduced trying to accommodate the other rules surrounding these stats at the last minute. I will need to update this.

KayenEQ:
Name of the rule was pretty ambiguous. Try something like.

RULE_BOOL(Spells, ItemExtraSpellAmtCalcAsPercent, false, "Apply Item stat Spell Dmg and Heal Amount as Percentage-based modifiers instead of flat additions")

catapultam-habeo:
Hi Kayen,

Thank you for the feedback. I've made the requested adjustment.

KayenEQ (approved):

PR #2021, opened 2022-02-27
[Commands] Cleanup #worldwide command.

Kinglykrab:
- Cleanup messages and logic.
- Command usages would not display when using commands improperly, this will fix that.
- Added a return for invalid spells on cast and remove so they cannot be accidentally used.

### DIFF
--- a/zone/gm_commands/worldwide.cpp
+++ b/zone/gm_commands/worldwide.cpp
@@ -2,18 +2,55 @@
 
 void command_worldwide(Client *c, const Seperator *sep)
 {
-	std::string sub_command;
-	if (sep->arg[1]) {
-		sub_command = sep->arg[1];
+	int arguments = sep->argnum;
+	if (!arguments) {
+		c->Message(Chat::White, "Usage: #worldwide cast [Spell ID] [Disable Message] - Cast a spell worldwide, use 1 for Disable Message to not send a global message");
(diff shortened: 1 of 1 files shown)


noudess:
Thanks for this.

zone/gm_commands/worldwide.cpp
```diff
@@ -2,20 +2,57 @@
...
+		c->Message(Chat::White, "Usage: #worldwide cast [Spell ID] [Disable Message] - Cast a spell worldwide, use 1 for Disable Message to not send a global message");
+		c->Message(Chat::White, "Usage: #worldwide message [Message] - Send a message worldwide");
+		c->Message(Chat::White, "Usage: #worldwide move [Zone ID|Zone Short Name] - Move all players worldwide to specified zone");
+		c->Message(Chat::White, "Usage: #worldwide moveinstance [Instance ID] - Move all players worldwide to specified instance");
+		c->Message(Chat::White, "Usage: #worldwide remove [Spell ID] - Remove a spell worldwide");
+		return;
+	}
+
+	if (is_cast) {
+		if (sep->IsNumber(2)) {
```
joligario, line 37: Should these be arg[2] ? A couple of those in here.

Kinglykrab: The method checks if sep->arg[2] is a number.

joligario (approved):

PR #2026, opened 2022-03-03
[Bug Fix] Bandolier didn't recognize source weapon on cursor

noudess:
If you activated a bandolier set and one of the source weapons was on your cursor, the code didn't see it.

For whatever reason, m_inv.GetItem(EQ::invslot::slotCursor)->GetCharges() was returning 0 when the source weapon was on the cursor, bypassing the code that was written to handle this case.

### DIFF
--- a/zone/inventory.cpp
+++ b/zone/inventory.cpp
@@ -3332,8 +3332,7 @@ void Client::SetBandolier(const EQApplicationPacket *app)
 			// removed 'invWhereCursor' argument from above and implemented slots 30, 331-340 checks here
 			if (slot == INVALID_INDEX) {
 				if (m_inv.GetItem(EQ::invslot::slotCursor)) {
-					if (m_inv.GetItem(EQ::invslot::slotCursor)->GetItem()->ID == m_pp.bandoliers[bss->Number].Items[BandolierSlot].ID &&
-						m_inv.GetItem(EQ::invslot::slotCursor)->GetCharges() >= 1) { // '> 0' the same, but this matches Inventory::_HasItem conditional check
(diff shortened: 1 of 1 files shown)


Kinglykrab (approved):

mackal (approved):

mackal:
I guess we should figure out why cursor slot returns 0 here, I would assume that's not right :P

noudess:
> I guess we should figure out why cursor slot returns 0 here, I would assume that's not right :P

I agree.  I don't know what's going on there.  Should we not merge this?  I feel like maybe this spot doesn't need the check since we know the item is there, so maybe this is ok anyway.

mackal:
> > I guess we should figure out why cursor slot returns 0 here, I would assume that's not right :P
> 
> I agree. I don't know what's going on there. Should we not merge this? I feel like maybe this spot doesn't need the check since we know the item is there, so maybe this is ok anyway.

no, but maybe we should add a TODO comment noting that we had to remove it, just so someone doesn't have to look through commit history/PRs to find discussion about it

noudess:
> > > I guess we should figure out why cursor slot returns 0 here, I would assume that's not right :P
> > 
> > 
> > I agree. I don't know what's going on there. Should we not merge this? I feel like maybe this spot doesn't need the check since we know the item is there, so maybe this is ok anyway.
> 
> no, but maybe we should add a TODO comment noting that we had to remove it, just so someone doesn't have to look through commit history/PRs to find discussion about it

Normal comment or do we have  an agreed upon keyword to key on for TODO items?

mackal (approved):

Akkadius (approved):

PR #2027, opened 2022-03-04
[Bug Fix] Doors::GetSize() Perl Croak Typo.

Kinglykrab:
- Typo was causing GetIncline() to show twice on Spire and causing GetSize() not to show at all on Spire.


fryguy503 (approved):
LGTM

PR #2028, opened 2022-03-04
[Bug Fix] Objects::GetTiltX() and Objects::GetTiltY() Perl Croak Typos.

Kinglykrab:
- Typo was causing GetIncline() to show three times on Spire and causing GetTiltX() and GetTiltY() not to show at all on Spire.


fryguy503 (approved):
LGTM

PR #2032, opened 2022-03-06
[Feature] EQ2-style implied targeting for spells.

catapultam-habeo:
I always thought that this was one of the better features of EQ2. I think implied targeting melee would be more tricky to implement, and is pretty opaque to me (not a C++ guru by any stretch of the imagination), so skipping that for now, but spells are pretty easy and a pretty nice QOL feature.


Akkadius (approved):

Akkadius:
Optional feature; LGTM

PR #2035, opened 2022-03-06
[Bug Fix] Preserve Permanent pet buffs across zone transitions

catapultam-habeo:


mackal:
Can you provide a list of spells that this is needed for?

catapultam-habeo:
> Can you provide a list of spells that this is needed for?

There aren't any in PEQ or live-like content. Honestly I don't know of anything in PEQ or live-like content which uses buffdurationformula 50, pet or no.

That said, -1 is a valid buff duration, while 0 is not. The methods that account for saving and loading pet buffs cannot account for -1 as a valid buff duration without this change.

Akkadius:
This is less of a bug and more of an optional feature if server admins want it

Akkadius (approved):

PR #2044, opened 2022-03-08
[API] Perl functions to set invulnerable to and modify environmental damage.

KayenEQ:
Functions to allow environmental damage to be altered from scripts.

Set client to mitigate or increase environmental damage taken. Negative value will reduce damage.
$client->GetEnvironmentDamageModifier()
$client->SetEnvironmentDamageModifier(int32 modifier)

Set client to be invulnerable to environmental damage
$client->InvulnerableEnvironmentDamage()
$client->SetInvulnerableEnvironmentDamage(bool invulnerable)

### DIFF
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -353,6 +353,7 @@ Client::Client(EQStreamInterface* ieqs)
 	temp_pvp = false;
 	is_client_moving = false;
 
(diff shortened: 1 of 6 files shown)


Kinglykrab (approved):
I’ll add to Lua for you later.

PR #2045, opened 2022-03-08
[Bots] Update Bot Heal & Damage methods to more closely match Clients + Bugfixes

catapultam-habeo:
- Fixed Bots Spell Damage being negative under some circumstances (heal target)
- Allow Bots to send non-melee damage reports in the same manner as Clients
- Refactor\Update Bots Spell Damage & Heal calculations to match current state of Clients.
- Allow Bots to actually utilize Spell Damage and Heal Amount stats

This is mostly just copy\paste from the methods in effects.cpp, with IsClient() checks and sections that only apply to NPCs removed. I tried simply calling the Mob methods, but there are some incompatibilities in regard to determining spell Foci that will require some more significant refactoring than I'm capable of right now.

A few foci that I don't believe apply to any relevant vanilla content (boost received healing) are not handled by the Heal method, because resolving the access problems are beyond my ability.

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -3932,10 +3932,9 @@ void Mob::CommonDamage(Mob* attacker, int &damage, const uint16 spell_id, const 
 				}
 			}
 			skip = owner;
-		}
+		}	
 		else {
 			//attacker is not a pet, send to the attacker
-
 			//if the attacker is a client, try them with the correct filter
 			if (attacker && (attacker->IsClient() || attacker->IsBot())) {
 				if ((spell_id != SPELL_UNKNOWN || FromDamageShield) && damage > 0) {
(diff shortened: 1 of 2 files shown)


KayenEQ (approved):

catapultam-habeo:
This may be causing zone crashes. I'm testing to rule it out or not.

zone/attack.cpp
```diff
@@ -3932,12 +3932,11 @@ void Mob::CommonDamage(Mob* attacker, int &damage, const uint16 spell_id, const
 				}
 			}
 			skip = owner;
-		}
+		}
```
Kinglykrab, line 3935: Random tab here.

catapultam-habeo:
idk if this actually fixed the crashes. I could never replicate them, just folks on my server having them and the stack trace pointed there, this seems like the only thing it could conceivably be,

Hold off on merge until I can run this on my server a little while and see if we stop seeing it.

Kinglykrab (approved):

catapultam-habeo:
I'm confident in these changes at this point, for what that is worth. The zone crashes which were occurring at high frequency no longer occur.

PR #2057, opened 2022-03-14
[Bug Fix] Restore missing messages for lifetap and dmg spells.

noudess:
I noticed that when an npc (spectre) cast lifetap on a PC that the "feel your life drain away" message was missing.

This PR fixes this, and some other missing messages.  It also uses the proper message Filters.  This function had conflicting filters.

Been tested on Rolath for about a week.  Have not seen any extra messages and missing messages are restored.

### DIFF
--- a/common/spdat.cpp
+++ b/common/spdat.cpp
@@ -139,11 +139,13 @@ bool IsEvacSpell(uint16 spellid)
 
 bool IsDamageSpell(uint16 spellid)
 {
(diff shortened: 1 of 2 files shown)


noudess:
I'd like to ask that this be held open.  While it does make things better, I have found some outstanding issues and I think I can make it better.

Kinglykrab:
Assuming this is ready.

noudess:
> Assuming this is ready.

Yep.

PR #2059, opened 2022-03-15
[Bug Fix] Force NPCs to respect special ability 24 and 50 when set on player pets

Natedog2012:


zone/aggro.cpp
```diff
@@ -436,11 +436,16 @@ bool Mob::CheckWillAggro(Mob *mob) {
...
 			!RuleB(Aggro, AggroPlayerPets) ||
-			pet_owner->CastToClient()->GetGM()
+			pet_owner->CastToClient()->GetGM() ||
+			mob->GetSpecialAbility(IMMUNE_AGGRO)
 		)
 	) {
 		return false;
 	}
+	
+	if (this->IsNPC() && mob->IsNPC() && mob->GetSpecialAbility(IMMUNE_AGGRO_NPC)) {
```
Kinglykrab, line 446: `this` is implied and can be removed here.

Kinglykrab (approved):

zone/aggro.cpp
Natedog2012: Oh derp, removed

PR #2060, opened 2022-03-17
[API] Methods for getting more information on quest timers.

KayenEQ:
Following methods allow you to obtain more information about the status of your quest timers.

PERL
```
quest::hastimer(timer name) //bool for if a quest timer is currently active
quest::getremainingtimeMS(timer name) //Returns how much time is remaining on a quest timer in MS
quest::gettimerdurationMS(timer name) //Returns total duration of a quest timer in MS
```

LUA
```
bool lua_has_timer(const char *timer) {
	return quest_manager.hastimer(timer);
}

uint32 lua_get_remaining_time(const char *timer) {
	return quest_manager.getremainingtimeMS(timer);
}

uint32 lua_get_timer_duration(const char *timer) {
	return quest_manager.gettimerdurationMS(timer);
}
```


zone/lua_general.cpp
```diff
@@ -361,6 +361,18 @@ bool lua_is_paused_timer(const char *timer) {
...
+}
+
+uint32 lua_get_remaining_time(const char *timer) {
+	return quest_manager.getremainingtimeMS(timer);
+}
+
+uint32 lua_get_timer_duration(const char *timer) {
+	return quest_manager.gettimerdurationMS(timer);
+}
+
```
Kinglykrab, line 375: Missing the luabind::def for all of these.

KayenEQ: ok think I added it correctly now.

Kinglykrab: The ones you added are copy and paste of void methods. Have to set as the proper type and number of arguments.

Kinglykrab:
```cpp
luabind::def("has_timer", (bool(*)(const char*))&lua_has_timer),
luabind::def("get_remaining_time", (uint32(*)(const char*))&lua_get_remaining_time),
luabind::def("get_timer_duration", (uint32(*)(const char*))&lua_get_timer_duration),
```

KayenEQ: ok

zone/questmgr.cpp
```diff
@@ -738,6 +738,66 @@ bool QuestManager::ispausedtimer(const char *timer_name) {
 	return false;
 }
 
+bool QuestManager::hastimer(const char *timer_name) {
+	QuestManagerCurrentQuestVars();
+
+	std::list<QuestTimer>::iterator cur = QTimerList.begin(), end;
```
Akkadius, line 744: The verbosity on this code can be cut down rather significantly and use references

```
	for (auto & cur : QTimerList) {
		if (cur.mob && cur.mob == owner && cur.name == timer_name) {
			if (cur.Timer_.Enabled()) {
				return true;
			}
		}
	}
```

Akkadius:
Overall looks good outside of KK's comments being addressed.

I would have used `std::string` C++ strings versus old C style pointers, makes things a bit easier to work in general with but its fine for this change

PR #2063, opened 2022-03-19
[Bug Fix] #peqzone no longer bypass Handle_OP_ZoneChange

Natedog2012:
The original #peqzone would bypass EVENT_ZONE this will make it properly fire


Kinglykrab (approved):
Awesome.

PR #2073, opened 2022-04-13
[Bug Fix] Blocked spells max spell id increased

KayenEQ:
Blocked spell id should be checking int32 not int16


Kinglykrab:
> I'm not usually a fan of changing more than the bug you're addressing, but in this file, this could/should also be updated?
> 
> _SetPendingRezzData(int XP, uint32 DBID, **uint16 SpellID**, const char *CorpseName)_

Might not _need_ to change that as uint16 is as high as any client allows us to ago anyway. Up to @KayenEQ though.

PR #2076, opened 2022-04-21
[Quest API] Add EVENT_SKILL_UP & EVENT_LANGUAGE_SKILL_UP to Perl/Lua

nytmyr:
Adds sub EVENT_SKILL_UP output for use.
Exports:
$skill_id - ID of the skill being exported. Will export skill or language as the same so check below.
$skill_value - New skill level
$skill_max - Maximum value of skill
$is_tradeskill - 0 for non-tradeskill, 1 for tradeskill

Example usage:
sub EVENT_SKILL_UP {
if($is_tradeskill == 0) {
quest::shout("Skill Increase! " . $client->GetCleanName() . " has increased their " . quest::getskillname($skill_id) . " to " . $skill_value . " of " . $skill_max . "!"); #deleteme
}
if ($is_tradeskill == 1) {
quest::shout("Tradeskill Increase! " . $client->GetCleanName() . " has increased their " . quest::getskillname($skill_id) . " to " . $skill_value . " of " . $skill_max . "!"); #deleteme
}
}

Adds sub EVENT_LANGUAGE_SKILL_UP output for use.
Exports:
$skill_id - ID of the skill being exported. Will export skill or language as the same so check below.
$skill_value - New skill level
$skill_max - Maximum value of skill

Example usage:
sub EVENT_LANGUAGE_SKILL_UP {
quest::shout("Language Increase! " . $client->GetCleanName() . " has increased their " . quest::getlanguagename($skill_id) . " to " . $skill_value . " of " . $skill_max . "!"); #deleteme
}


zone/embparser.cpp
```diff
@@ -1648,6 +1649,7 @@ void PerlembParser::ExportEventVariables(
 			Seperator sep(data);
 			ExportVar(package_name.c_str(), "skill_id", sep.arg[0]);
 			ExportVar(package_name.c_str(), "skill_level", sep.arg[1]);
+			ExportVar(package_name.c_str(), "skill_max", sep.arg[2]);
```
Kinglykrab, line 1652: Where is the code for this change to export?

Kinglykrab:
Don’t think having is language or is tradeskill is the cleanest way to handle the different types. Would make more sense to just have 3 different events. One for skill ups, one for language level ups, and one for tradeskill level ups so you don’t have to have extra knowledge of the event to just use it. Someone else might think different though, so I’ll leave it up to them.

zone/embparser.cpp
nytmyr: Oops, that wasn't suppose to be there lol. Removed.

nytmyr:
> Don’t think having is language or is tradeskill is the cleanest way to handle the different types. Would make more sense to just have 3 different events. One for skill ups, one for language level ups, and one for tradeskill level ups so you don’t have to have extra knowledge of the event to just use it. Someone else might think different though, so I’ll leave it up to them.
> 
> Edit: To clarify, the skill ID value exported represents 3 different things based on if it’s a skill, tradeskill, or language, so it just makes for possible headache. Could maybe do 2 events, one for skills/tradeskills, one for languages. That way the skill ID isn’t sometimes actually a language ID.

Split.

zone/embparser.cpp
```diff
@@ -1709,6 +1711,23 @@ void PerlembParser::ExportEventVariables(
 			break;
 		}
 
+		case EVENT_SKILL_UP: {
+			Seperator sep(data);
+			ExportVar(package_name.c_str(), "skill_id", sep.arg[0]);
+			ExportVar(package_name.c_str(), "skill_value", sep.arg[1]);
+			ExportVar(package_name.c_str(), "skill_max", sep.arg[2]);
+			ExportVar(package_name.c_str(), "is_tradeskill", sep.arg[3]);
```
Kinglykrab, line 1719: Don’t need is tradeskill flag anymore. Can tell that by skill ID.

Kinglykrab:
Looks good. No need to export the is tradeskill flag though since we can infer based on skill ID that’s already exported whether or not it’s a tradeskill.

Kinglykrab (approved):

nytmyr:
> 

I left it just under the idea that it's there to split stuff off if you wanted. For example, tracking skills vs tradeskills you can just split the variable rather than individual IDs. Otherwise it's mostly going to be unused and everything passes through. I can remove or split off to another event as well.

zone/embparser.cpp
Akkadius: Agreed

Akkadius:
> > 
> 
> I left it just under the idea that it's there to split stuff off if you wanted. For example, tracking skills vs tradeskills you can just split the variable rather than individual IDs. Otherwise it's mostly going to be unused and everything passes through. I can remove or split off to another event as well.

It's fine as is given regular skills are contiguous and contain tradeskills https://docs.eqemu.io/server/player/skills/ and languages have their own range https://docs.eqemu.io/server/player/languages/

The two sub events are appropriate

Akkadius (approved):

PR #2080, opened 2022-04-28
[Bug Fix] Update to target types Beam and Cone to ignore invalid targets.

KayenEQ:
AOE target types for Beam and Cone type spells were attempting to hit targets incorrectly like pets and aura's.


mackal:
Are you sure this is wrong? I know auras eat my heal rains on my SHM (not those type of spells, but still)

KayenEQ:
Can find no evidence of this applying to beams.

mackal (approved):

PR #2081, opened 2022-04-29
[Bots] Fix ^dyearmor command math.

Kinglykrab:
The math made items blueish purple no matter what values you gave as Red.


joligario (approved):
Not a bots guy, but lgtm

PR #2084, opened 2022-04-30
[Bots] Bot::PerformTradeWithClient Cleanup.

Kinglykrab:
- Adds saylinks to return/accept messages for bot trades instead of just displaying the item name.
- Cleanups message and logic.
- Initial cleanup to eventually allow easy use with Perl/Lua quest API.


Akkadius (approved):
🤖

PR #2089, opened 2022-05-01
[Compile] Decrease build times using unity build strategy

Akkadius:
A while back I introduced https://github.com/EQEmu/Server/pull/1766 which had the benefit of making our code easier to maintain in terms of splitting commands up, but actually added a significant penalty of compile time - 20-30s to 5m or more depending on hardware.

This is because each cpp becomes its own compilation unit and all of the headers get compiled, parsed for each cpp file and an object get created and it all gets individually linked at the end as well.

This change utilizes a common practice in industry known as [Unity Builds](https://en.wikipedia.org/wiki/Unity_build)

Until C++20 and modules and investing into our project utilizing them, this is a cheap lift to get some development iteration time back.


Akkadius:
Dropped literally 5 minutes on Drone builds

image

mackal:
I know Google ended up abandoning jumbo builds (their name) for Chrome because of issues they ran into.

Akkadius:
> I know Google ended up abandoning jumbo builds (their name) for Chrome because of issues they ran into.

You can certainly shoot yourself in the foot if you abuse them poorly. I don't think we should go crazy with them and use them sparingly if at all

KayenEQ (approved):

PR #2090, opened 2022-05-01
[Combat] Basic Combat Recording

Akkadius:
This PR implements a basic combat recording mechanism for operators to see very simple DPS / Heal/s stats at the end of a combat encounter. 

This helps when making assessments on scaling content at a very basic level and can help with pulling analytics on the kind of DPS players are pushing in content by also logging to a file.

This can eventually extend out to include more complicated metrics, realtime popup window watching, tracking multiple entities but only gets a basic implementation for this PR.

image

**Changes**

* Adds class `CombatRecord`
* Adds logging category `CombatRecord`
* Hooks SetHP for increment / decrement events
* Hooks engaged and death events to properly start and stop Combat Record

### DIFF
--- a/zone/combat_record.h
+++ b/zone/combat_record.h
@@ -3,6 +3,7 @@
 
 #include <ctime>
 #include <string>
+#include "../common/types.h"
 
 class CombatRecord {
 public:
@@ -15,8 +16,8 @@ private:
 	std::string mob_name;
 	time_t      start_time      = 0;
 	time_t      end_time        = 0;
-	__int64_t   damage_received = 0;
-	__int64_t   heal_received   = 0;
+	int64       damage_received = 0;
+	int64       heal_received   = 0;
 };
 
 #endif //EQEMU_COMBAT_RECORD_H
--- a/zone/combat_record.h
+++ b/zone/combat_record.h
@@ -1,0 +1,22 @@
+#ifndef EQEMU_COMBAT_RECORD_H
+#define EQEMU_COMBAT_RECORD_H
+
+#include <ctime>
+#include <string>
+
+class CombatRecord {
+public:
+	void Start(std::string in_mob_name);
+	void Stop();
+	bool InCombat();
+	void ProcessHPEvent(int hp, int current_hp);
+	double TimeInCombat() const;
+private:
+	std::string mob_name;
+	time_t      start_time      = 0;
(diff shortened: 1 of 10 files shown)


zone/combat_record.h
```diff
@@ -0,0 +1,22 @@
...
+	void Start(std::string in_mob_name);
+	void Stop();
+	bool InCombat();
+	void ProcessHPEvent(int hp, int current_hp);
+	double TimeInCombat() const;
+private:
+	std::string mob_name;
+	time_t      start_time      = 0;
+	time_t      end_time        = 0;
+	__int64_t   damage_received = 0;
```
mackal, line 18: we have int64

zone/combat_record.h
```diff
@@ -0,0 +1,22 @@
...
+class CombatRecord {
+public:
+	void Start(std::string in_mob_name);
+	void Stop();
+	bool InCombat();
+	void ProcessHPEvent(int hp, int current_hp);
+	double TimeInCombat() const;
+private:
+	std::string mob_name;
+	time_t      start_time      = 0;
```
mackal, line 16: I would also think about std::chrono instead of old C time API, but this works so whatever

zone/combat_record.h
Akkadius: That's fine but its the same thing under the hood

```
typedef __int64_t int64_t;
```

```
typedef int64_t int64;
```

It's just aliases references

mackal: Consistency with the rest of the code base is good though :P

Akkadius: The rest of the code uses this same type under the hood. Pointless to pull in a header file `types.h` for what is essentially the same thing for one field on one struct

zone/combat_record.h
Akkadius: `time_t` is incredibly lightweight and serves the simple needs of calculating start / end time here

zone/combat_record.h
mackal: https://en.cppreference.com/w/cpp/header/cstdint Is what I would actually prefer we use, but that's a lot of changes and never really brought it up

Akkadius: That's fine

KayenEQ (approved):

PR #2095, opened 2022-05-02
[Combat] Fix Frenzy vs opponents immune to non-magic

noudess:
I believe that frenzy should fail against opponents that are immune to non-magical weapons, when not wielding one. This PR fixes that.

NOTE: It seems live has removed this invulnerability from wisps and ghouls...  Not sure if it's just gone, or only on selected mobs.

Also note, this fixes the issue for clients.  Very similar code exists in Client::DoClassAttacks for Frenzy.  Does this need to be changed also?  I don't know where that code is called from??  Changing that code had no effect on this bug.  Maybe I need to put the same fix there for cases when client is AI controlled?

### DIFF
--- a/zone/special_attacks.cpp
+++ b/zone/special_attacks.cpp
@@ -340,6 +340,11 @@ void Client::OPCombatAbility(const CombatAbility_Struct *ca_atk)
 		ReuseTime = FrenzyReuseTime - 1 - skill_reduction;
 		ReuseTime = (ReuseTime * HasteMod) / 100;
 
+		auto primaryInUse = GetInv().GetItem(EQ::invslot::slotPrimary);
+		if (primaryInUse && GetWeaponDamage(GetTarget(), primaryInUse) <= 0) {
+			max_dmg = DMG_INVULNERABLE;
+		}
+
 		while (AtkRounds > 0) {
 			if (GetTarget())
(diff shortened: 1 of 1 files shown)


zone/special_attacks.cpp
```diff
@@ -340,6 +340,11 @@ void Client::OPCombatAbility(const CombatAbility_Struct *ca_atk)
 		ReuseTime = FrenzyReuseTime - 1 - skill_reduction;
 		ReuseTime = (ReuseTime * HasteMod) / 100;
 
+		auto primaryInUse = GetInv().GetItem(EQ::invslot::slotPrimary);
```
Akkadius, line 343: snake_case

noudess: Fixed.

Akkadius:
Still not snake_case for the local variable

noudess:
> Still not snake_case for the local variable

Fixing this.  But I have tended to match "local style" rather than have mixed stuff.

Akkadius:
> > Still not snake_case for the local variable
> 
> Fixing this. But I have tended to match "local style" rather than have mixed stuff.

I feel ya, best opportunity to make changes towards desired state is when we’re in the code making changes

Kinglykrab (approved):

PR #2091, opened 2022-05-01
[int64] Support for HP / Mana / End / Damage / Hate

Akkadius:
Need to do additional testing, also looking for others to do testing as well

**Implemented**

* HP
* Mana
* Endurance
* Damage
* Hate

**What's Verified Working**

* HP
* Mana
* Damage
* Hate
* Spell Damage

Damage works but clients only display int32 so we'll have to come up with some sort of way to emulate messages that overflow. The client just displays it as a "miss" even though the server still registers it. That will be outside of the scope of this PR though

image

PR #2097, opened 2022-05-03
[Tasks] Implement Task Goal Match List

Akkadius:
This PR implements a separate, far more ergonomic way of driving task activity goal list update targets.

For example when using a goal list, instead of maintaining individual ID's in a goal list table for every single NPC type ID that you want to be in a task, its far easier to just grep by the name of the NPC's within the context of the zone that you want to provide your update for.

For example; in Tutorial there is a Task called **Arachnia**

**Kill 12 Gloom Spiders 0/12 The Mines of Gloomingdeep**

You can provide a `goal_match_list` of `gloomingdeep grunt|rat|goblin` and you will get activity updates from pipe delimited partial matches. This would give updates from any NPC with the partial matches within the pipes. This works together with other filters, such as zoneids to filter updates only to a specific zone. 

Most servers handle this in script and this would allow for more simplicity server side.

Task editor in Spire will inform users what kind of match it is depending on the activity. This will also allow me to finish the Spire task editor since there's very little left.

### DIFF
--- a/zone/task_goal_list_manager.cpp
+++ b/zone/task_goal_list_manager.cpp
@@ -163,7 +163,7 @@ bool TaskGoalListManager::IsInMatchListPartial(const std::string &match_list, co
 {
 	std::string entry_match = str_tolower(entry);
 	for (auto &s: SplitString(match_list, '|')) {
-		if (str_tolower(entry_match).find(s) != std::string::npos) {
+		if (entry_match.find(str_tolower(s)) != std::string::npos) {
 			return true;
 		}
 	}
--- a/zone/task_goal_list_manager.cpp
+++ b/zone/task_goal_list_manager.cpp
@@ -147,3 +147,26 @@ bool TaskGoalListManager::IsInList(int list_id, int entry)
 
 	return true;
 }
+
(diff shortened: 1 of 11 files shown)


Akkadius:
Migration set to merge after https://github.com/EQEmu/Server/pull/2091

Order of merge TBD

jmkca (approved):

Kinglykrab (approved):

zone/task_goal_list_manager.cpp
```diff
@@ -147,3 +147,26 @@ bool TaskGoalListManager::IsInList(int list_id, int entry)
...
+	}
+
+	return false;
+}
+
+bool TaskGoalListManager::IsInMatchListPartial(const std::string &match_list, const std::string &entry)
+{
+	std::string entry_match = str_tolower(entry);
+	for (auto &s: SplitString(match_list, '|')) {
+		if (str_tolower(entry_match).find(s) != std::string::npos) {
```
hgtw, line 166: `entry_match` already had `str_tolower` used on it

Akkadius: Good catch, originally had this flipped

hgtw (approved):

PR #2099, opened 2022-05-03
[Quest API] Add commify to Perl/Lua.

Kinglykrab:
- Add quest::commify(number) to Perl.
- Add eq.commify(number) to Lua.

### DIFF
--- a/common/string_util.cpp
+++ b/common/string_util.cpp
@@ -1265,3 +1265,24 @@ std::string ConvertMoneyToString(uint32 platinum, uint32 gold, uint32 silver, ui
 	}
 	return money_string;
 }
+
+std::string commify(const std::string &number) {
+	std::string temp_string;
+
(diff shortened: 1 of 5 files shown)


Akkadius:
lgtm outside of small comment

Kinglykrab:
> The rename is a bit needless. You can overload types in C so we could have two `commify` with different type inputs for both string and numbers

I only renamed it because when I had the questmgr and string_util method named the same it immediately crashed and I got an access violation.

Akkadius:
You can access commify string util directly from within the lua / perl calls and skip creating a quest manager wrapper function

Kinglykrab:
> You can access commify string util directly from within the lua / perl calls and skip creating a quest manager wrapper function

Will do.

Kinglykrab:
Done.

Akkadius (approved):

PR #2108, opened 2022-05-04
[Commands] Cleanup #npcspecialattk Command.

Kinglykrab:
- Cleanup messages and logic.


Akkadius (approved):

zone/gm_commands/npcspecialattk.cpp
```diff
@@ -2,15 +2,97 @@
...
 		);
+		c->Message(
+			Chat::White,
+			"Usage: #npcspecialattk help - List all the special attack identifiers"
+		);
+		return;
+	} else if (arguments == 1) {
+		if (!strcasecmp(sep->arg[1], "help")) {		
+			c->SendPopupToClient(
+				"Special Attack Flags",
```
Akkadius, line 24: Using flags like this is deprecated, not sure we want to even encourage the use at this point

Kinglykrab: Can remove the command altogether if you'd prefer and just add to a Wiki page for legacy purposes.

PR #2109, opened 2022-05-04
[Commands] Cleanup #npctype_cache Command.

Kinglykrab:
- Cleanup messages and logic.
- Loop was returning as soon as it found first ID, so you couldn't use the spaced ID list functionality at all.


zone/gm_commands/npctype_cache.cpp
```diff
@@ -2,26 +2,31 @@
...
-				zone->ClearNPCTypeCache(-1);
-			}
-			else {
-				int id = atoi(sep->arg[i + 1]);
-				if (id > 0) {
-					c->Message(Chat::White, "Clearing npc type %d from the cache.", id);
-					zone->ClearNPCTypeCache(id);
-					return;
-				}
+	int arguments = sep->argnum;
```
Akkadius, line 5: Not sure we need this command anymore to be honest

Akkadius (approved):

zone/gm_commands/npctype_cache.cpp
Kinglykrab: Wonderful, I'll delete it.

PR #2120, opened 2022-05-04
[Commands] Cleanup #reloadaa Command.

Kinglykrab:
- Cleanup messages.
- Remove unncessary file_exists command from file.


zone/worldserver.cpp
```diff
@@ -2013,6 +2013,35 @@ void WorldServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p)
 		parse->LoadPerlEventExportSettings(parse->perl_event_export_settings);
 		break;
 	}
+	case ServerOP_ReloadAAData: {
+		if (zone) {
+			worldserver.SendEmoteMessage(
```

Akkadius (approved):

PR #2122, opened 2022-05-04
[Commands] Cleanup #reloadlevelmods Command.

Kinglykrab:
- Cleanup messages and logic.
- Make the reloading of modifiers global instead of zone-specific.


zone/worldserver.cpp
```diff
@@ -2013,6 +2013,34 @@ void WorldServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p)
 		parse->LoadPerlEventExportSettings(parse->perl_event_export_settings);
 		break;
 	}
+	case ServerOP_ReloadLevelEXPMods: {
+		if (zone) {
+			worldserver.SendEmoteMessage(
```

world/zoneserver.cpp
```diff
@@ -889,6 +889,11 @@ void ZoneServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p) {
 		zoneserver_list.SendPacket(pack);
 		break;
 	}
+	case ServerOP_ReloadLevelEXPMods:
```
Akkadius, line 892: Whenever these are all merged, all of these reloads could be stacked ontop of eachother as cases
eg

```cpp
case1:
case2:
case3:
    zoneserver_list.SendPacket(pack);
```

Akkadius (approved):

world/zoneserver.cpp
Kinglykrab: Yeah, absolutely. 👍

PR #2123, opened 2022-05-04
[Commands] Cleanup #reloadmerchants Command.

Kinglykrab:
- Cleanup messages and logic.
- Make the reloading of merchants global instead of zone specific.


zone/worldserver.cpp
```diff
@@ -2013,6 +2013,34 @@ void WorldServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p)
 		parse->LoadPerlEventExportSettings(parse->perl_event_export_settings);
 		break;
 	}
+	case ServerOP_ReloadMerchants: {
+		if (zone) {
+			worldserver.SendEmoteMessage(
```

Akkadius (approved):

zone/worldserver.cpp
Kinglykrab: :+1:

Akkadius (approved):

PR #2124, opened 2022-05-04
[Commands] Cleanup #reloadperlexportsettings Command.

Kinglykrab:
- Cleanup messages and logic.

### DIFF
--- a/zone/worldserver.cpp
+++ b/zone/worldserver.cpp
@@ -2017,7 +2017,7 @@ void WorldServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p)
 				AccountStatus::GMAdmin,
 				Chat::Yellow,
 				fmt::format(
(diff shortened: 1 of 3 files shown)


zone/worldserver.cpp
```diff
@@ -2010,6 +2010,30 @@ void WorldServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p)
 		break;
 	}
 	case ServerOP_ReloadPerlExportSettings: {
+		if (zone) {
+			worldserver.SendEmoteMessage(
```
Akkadius, line 2014: Same as other PR comment

Akkadius (approved):

zone/worldserver.cpp
Kinglykrab: :+1:

PR #2125, opened 2022-05-04
[Commands] Cleanup #reloadtitles Command.

Kinglykrab:
- Cleanup messages.

### DIFF
--- a/zone/worldserver.cpp
+++ b/zone/worldserver.cpp
@@ -1717,7 +1717,7 @@ void WorldServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p)
 				AccountStatus::GMAdmin,
 				Chat::Yellow,
 				fmt::format(
-					"Titles reloaded for {}.",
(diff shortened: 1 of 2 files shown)


zone/worldserver.cpp
```diff
@@ -1710,6 +1710,31 @@ void WorldServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p)
 
 	case ServerOP_ReloadTitles:
 	{
+		if (zone) {
+			worldserver.SendEmoteMessage(
```
Akkadius, line 1714: Lots of messages like this across a lot of the PR's. could be simplified into a method and save 23 lines each call.

eg `zone->NotifyWorldReloaded(std::string reload_label);`

Akkadius (approved):

zone/worldserver.cpp
Kinglykrab: 👍

PR #2129, opened 2022-05-05
[Commands] Cleanup #reloadzps Command.

Kinglykrab:
- Cleanup messages and logic.
- Make reloading of zone points global instead of zone specific.
- Add zone->GetZoneDescription().
- Add mob->GetTargetDescription(mob).
- Cleanup all zoneserver packets and make use of auto.
- Fix all spots where GetTargetDescription()/GetZoneDescription() would fit.


zone/worldserver.cpp
```diff
@@ -2013,6 +2013,34 @@ void WorldServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p)
...
+	case ServerOP_ReloadZonePoints: {
+		if (zone) {
+			worldserver.SendEmoteMessage(
+				0,
+				0,
+				AccountStatus::GMAdmin,
+				Chat::Yellow,
+				fmt::format(
+					"Zone points reloaded for {}{}.",
+					fmt::format(
```
Akkadius, line 2025: Similar recommendation as in other PR; `zone->DescribeZoneLabel()`

Akkadius (approved):

zone/worldserver.cpp
Kinglykrab: 👍

zone/gm_commands/task.cpp
```diff
@@ -359,21 +326,12 @@ void command_task(Client *c, const Seperator *sep)
...
-							client_target->GetCleanName(),
-							client_target->GetID()
-						)
-					)
+					c->GetTargetDescription(target)
 				).c_str()
 			);
 
-			client_target->UpdateTaskActivity(task_id, activity_id, count);
-			c->ShowClientTasks(client_target);
```
Akkadius, line 375: Why is this flipped

zone/gm_commands/task.cpp
```diff
@@ -282,62 +266,45 @@ void command_task(Client *c, const Seperator *sep)
 
 		return;
 	} else if (is_show) {
-		c->ShowClientTasks(client_target);
+		target->ShowClientTasks(c);
 		return;
 	} else if (is_uncomplete) {
 		if (sep->IsNumber(2)) {
 			auto task_id = std::stoul(sep->arg[2]);
-			if (task_id && task_id < MAXTASKS) {
```
Akkadius, line 290: Not sure why this changed

Akkadius:
There are a ton of changes in this PR that happened after approval. Not ideal especially when it can be the source of issues, it can be hard to weed out the problem

zone/gm_commands/task.cpp
Kinglykrab: The logic flow changed, the return comes first so we fail out instead of it coming after.

PR #2133, opened 2022-05-05
[Commands] Cleanup #resetdisc_timer Command.

Kinglykrab:
- Cleanup messages and logic.

### DIFF
--- a/zone/gm_commands/resetdisc_timer.cpp
+++ b/zone/gm_commands/resetdisc_timer.cpp
@@ -2,22 +2,59 @@
 
 void command_resetdisc_timer(Client *c, const Seperator *sep)
 {
-	Client *target = c->GetTarget()->CastToClient();
-	if (!c->GetTarget() || !c->GetTarget()->IsClient()) {
-		target = c;
+	int arguments = sep->argnum;
+	if (!arguments) {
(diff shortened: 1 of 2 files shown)


zone/gm_commands/resetdisc_timer.cpp
```diff
@@ -2,22 +2,59 @@
...
-		c->ResetAllDisciplineTimers();
+
+	bool is_all = !strcasecmp(sep->arg[1], "all");
+
+	if (is_all) {
+		c->Message(
+			Chat::White,
+			fmt::format(
+				"Reset all Discipline timers for {}.",
+				(
```
Akkadius, line 24: Been seeing this ternary in a few PR's now. Could be simplified to a helper

eg something to the extent of `Mob::` `m->DescribeTarget(Mob *t)`

Maybe for another PR

Akkadius (approved):

zone/gm_commands/resetdisc_timer.cpp
Kinglykrab: Yeah, I was thinking that earlier that I'm using it almost everywhere. 👍

Akkadius: `m->DescribeTargetLabel(Mob *t)` maybe

PR #2137, opened 2022-05-05
[Commands] Cleanup #sensetrap Command.

Kinglykrab:
- Cleanup messages and logic.

### DIFF
--- a/zone/gm_commands/sensetrap.cpp
+++ b/zone/gm_commands/sensetrap.cpp
@@ -2,23 +2,28 @@
 
 void command_sensetrap(Client *c, const Seperator *sep)
 {
-	Mob *target = c->GetTarget();
-	if (!target) {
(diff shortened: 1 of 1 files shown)


zone/gm_commands/sensetrap.cpp
```diff
@@ -2,23 +2,28 @@
...
+				Chat::White,
+				fmt::format(
+					"{} ({}) is too far away.",
+					target->GetCleanName(),
+					target->GetID()
+				).c_str()
+			);
+			return;
 		}
+		c->HandleLDoNSenseTraps(target, c->GetSkill(EQ::skills::SkillSenseTraps), LDoNTypeMechanical);
```
Akkadius, line 24: Opportunity to `return;` here and drop the `else`

Akkadius (approved):

PR #2138, opened 2022-05-05
[Commands] Cleanup #serverinfo Command.

Kinglykrab:
- Cleanup message and logic.
- Use popup instead of messages.
image

### DIFF
--- a/zone/gm_commands/serverinfo.cpp
+++ b/zone/gm_commands/serverinfo.cpp
@@ -51,8 +51,6 @@ void command_serverinfo(Client *c, const Seperator *sep)
 	);
 
 	popup_text.append("</table>");
-		
-	popup_text.append("<br>");
 
 	popup_text.append("<c \"#00FF00\">CPU Information</c>");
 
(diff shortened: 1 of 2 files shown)


zone/gm_commands/serverinfo.cpp
```diff
@@ -3,31 +3,100 @@
...
+	auto os = EQ::GetOS();
+	auto cpus = EQ::GetCPUs();
+	auto process_id = EQ::GetPID();
+	auto rss = EQ::GetRSS() / 1048576.0;
+	auto uptime = static_cast<uint32>(EQ::GetUptime());
+
+	std::string popup_text;
+
+	popup_text.append("<c \"#00FF00\">Operating System Information</c>");
+
```
Akkadius, line 15: Could probably kill all the newlines between each of the `.append` calls

Akkadius (approved):

Akkadius:
Thanks for dropping a visual of the change in the description

zone/gm_commands/serverinfo.cpp
Kinglykrab: 👍

PR #2140, opened 2022-05-06
[Discord Integration] Native Discord Integration

Akkadius:
### Discord Webhook Logging Integration

This is meant to be purely a one-way, highly performing implementation of publishing data to Discord.

* Strictly one-way
* Highly performing
* Extends the EQEmu Logging system to include another category called 
* Messages get batched in multiples of the same type to prevent from sending many individual messages
* Runs through UCS. If you do not use UCS; this integration will not send messages to Discord
* Runs a queue listener in its own thread on UCS so even there is rate limiting congestion, it will sleep and retry on its own thread not to interrupt the main UCS thread or networking threads
* Has new Quest API methods to leverage this functionality for logging events to player facing channels or administrative facing channels. For example maybe you want to have the hackers logging category be send to an admin channel and you want an in game world wide emote to broadcast to a player announcement channel when a certain event happens
* Optimizations to EQEmu Core Logging system

image

### New Table (discord_webhooks)

This table pertains to webhook mappings that will correspond to `discord_webhook_id` in the `logsys_categories` table

```MariaDB [peq]> describe discord_webhooks;
+--------------+--------------+------+-----+---------+----------------+
| Field        | Type         | Null | Key | Default | Extra          |
+--------------+--------------+------+-----+---------+----------------+
| id           | int(11)      | NO   | PRI | NULL    | auto_increment |
| webhook_name | varchar(100) | YES  |     | NULL    |                |
| webhook_url  | varchar(255) | YES  |     | NULL    |                |
| created_at   | datetime     | YES  |     | NULL    |                |
| deleted_at   | datetime     | YES  |     | NULL    |                |
+--------------+--------------+------+-----+---------+----------------+
```

### New Fields (logsys_categories)

```
MariaDB [peq]> describe logsys_categories;
+--------------------------+--------------+------+-----+---------+-------+
| Field                    | Type         | Null | Key | Default | Extra |
+--------------------------+--------------+------+-----+---------+-------+
| log_category_id          | int(11)      | NO   | PRI | NULL    |       |
| log_category_description | varchar(150) | YES  |     | NULL    |       |
| log_to_console           | smallint(11) | YES  |     | 0       |       |
| log_to_file              | smallint(11) | YES  |     | 0       |       |
| log_to_gmsay             | smallint(11) | YES  |     | 0       |       |
| log_to_discord           | smallint(11) | YES  |     | 0       |       |
| discord_webhook_id       | int(11)      | YES  |     | 0       |       |
+--------------------------+--------------+------+-----+---------+-------+
7 rows in set (0.003 sec)
```

### Quest API

**Lua**

```lua
eq.discord_send(std::string webhook_name, std::string message);
```

**Perl**

```perl
quest::discordsend(string webhook_name, string message);
```

### How to Use 

You first need will need a webhook to send messages to your Discord Channel. Assuming you are an admin of a Discord and/or ability to make Webhook integrations. You can follow this Discord guide https://support.discord.com/hc/en-us/articles/228383668-Intro-to-Webhooks

Once you have the webhook you can install it in your `discord_webhooks` table like so. You will want to give you webhook a name if you want to later call it by the quest api

```
MariaDB [peq]> select * from discord_webhooks;
+----+--------------+--------------------------------------------------------------------------------------------------------------------------+------------+------------+
| id | webhook_name | webhook_url                                                                                                              | created_at | deleted_at |
+----+--------------+--------------------------------------------------------------------------------------------------------------------------+------------+------------+
|  1 | test         | https://discord.com/api/webhooks/971964565959114793/xxx-xxx                                                              | NULL       | NULL       |
|  2 | quest_api    | https://discord.com/api/webhooks/971964720372396062/xxx-xxx                                                              | NULL       | NULL       |
+----+--------------+--------------------------------------------------------------------------------------------------------------------------+------------+------------+
2 rows in set (0.001 sec)

```

Once you have a webhook created, you can use it in either a Quest API method or hook it into any of your logging categories per normal.

For example, you can see I have logging categories setup just to test the integration

```
MariaDB [peq]> select log_category_id, log_category_description, log_to_discord, discord_webhook_id from logsys_categories where discord_webhook_id > 0;
+-----------------+--------------------------+----------------+--------------------+
| log_category_id | log_category_description | log_to_discord | discord_webhook_id |
+-----------------+--------------------------+----------------+--------------------+
|              36 | MySQL Query              |              1 |                  1 |
|              53 | Info                     |              1 |                  1 |
+-----------------+--------------------------+----------------+--------------------+
2 rows in set (0.002 sec)
```

And then in my test channel you can see both Info and MySQL messages being broadcasted

image


common/discord/discord.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include "../string_util.h"
+#include "../eqemu_logsys.h"
+
+constexpr int MAX_RETRIES = 10;
+
+void Discord::SendWebhookMessage(const std::string &message, const std::string &webhook_url)
+{
+	// validate
+	if (webhook_url.empty()) {
+		LogDiscord("[webhook_url] [{}] is empty", webhook_url);
```
hgtw, line 13: No reason to log `webhook_url` itself if it's empty (unless this is an intentional log output thing)

Akkadius:
I'll wrap this one up within the next week or so. Covid has had me up and down with energy levels

Kinglykrab (approved):
Tested with some different webhooks back when code was initially implemented, nothing has changed since then functionally, so looks good to me. 👍

nytmyr:
Have been using this since you pushed the last commit. Using 11 different hooks ATM without issue. Can even use some formatting on the output to Discord which is nice. Haven't seen any crashes or errors yet and I have it pushing up to 4 different things that hits on about 50% of NPC kills so it's running a lot. 

This is just a nitoicky thing for me but I may end up trying to put an extra column or something to toggle the throttling check that combines messages on certain hooks. For example, I have it pushing achievements and if 2 achievements hit at once it'll combine them but it's only a couple messages.

But this is fantastic, so happy to have this working and for what you put into it.

Akkadius:
> Have been using this since you pushed the last commit. Using 11 different hooks ATM without issue. Can even use some formatting on the output to Discord which is nice. Haven't seen any crashes or errors yet and I have it pushing up to 4 different things that hits on about 50% of NPC kills so it's running a lot.
> 
> This is just a nitoicky thing for me but I may end up trying to put an extra column or something to toggle the throttling check that combines messages on certain hooks. For example, I have it pushing achievements and if 2 achievements hit at once it'll combine them but it's only a couple messages.
> 
> But this is fantastic, so happy to have this working and for what you put into it.

That’s great. What do you mean on your message combine comment, what is the issue ?

nytmyr:
> 

Just the way it combines similar messages to prevent the throttling to Discord. For example:
Screenshot_20220605-135925_Discord
I could also try finding a way to line break on Discord though and just include it at the end of each message.

Akkadius:
Ah I will get that fixed. That must be only happening with quests

nytmyr:
> Ah I will get that fixed. That must be only happening with quests

Ahh, now that you mention it, it does only happen to the quest methods. Appreciate you.

Akkadius:
> > Ah I will get that fixed. That must be only happening with quests
> 
> Ahh, now that you mention it, it does only happen to the quest methods. Appreciate you.

Just pushed up the fix

nytmyr:
> > > Ah I will get that fixed. That must be only happening with quests
> > 
> > 
> > Ahh, now that you mention it, it does only happen to the quest methods. Appreciate you.
> 
> Just pushed up the fix

Works perfect. Thank you!

PR #2144, opened 2022-05-06
[Commands] Cleanup #spawnstatus Command.

Kinglykrab:
- Cleanup messages and logic.


zone/gm_commands/spawnstatus.cpp
```diff
@@ -2,27 +2,41 @@
 
 void command_spawnstatus(Client *c, const Seperator *sep)
 {
-	if ((sep->arg[1][0] == 'e') | (sep->arg[1][0] == 'E')) {
-		// show only enabled spawns
-		zone->ShowEnabledSpawnStatus(c);
```
Akkadius, line 7: If the goal is to truly clean code up. All of these methods that are referenced as `Zone` scoped methods are only referenced by this GM command file. They could be inlined within the GM command file and removed from Zone.

* ShowEnabledSpawnStatus
* ShowDisabledSpawnStatus
* SpawnStatus
* ShowSpawnStatusByID

This should reduce code bloat in Zone and keep it specific to this command

zone/zone.cpp
```diff
@@ -2043,100 +2043,168 @@ bool ZoneDatabase::LoadStaticZonePoints(LinkedList<ZonePoint *> *zone_point_list
...
+				"{} Spawn{} listed.",
+				spawn_count,
+				spawn_count != 1 ? "s" : ""
+			).c_str()
+		);
+	} else {
+		client->Message(
+			Chat::White,
+			fmt::format(
+				"{} Of {} spawn{} listed.",
```
Akkadius, line 2133: Of intentionally capitalized? vs `of`

zone/zone.cpp
```diff
@@ -2043,100 +2043,168 @@ bool ZoneDatabase::LoadStaticZonePoints(LinkedList<ZonePoint *> *zone_point_list
...
+			).c_str()
+		);
+		return;
+	}
+	
+	if (status_type == "All") {
+		client->Message(
+			Chat::White,
+			fmt::format(
+				"{} Spawn{} listed.",
```
Akkadius, line 2124: Lowercase `spawn(s)` ?

zone/zone.cpp
```diff
@@ -2043,100 +2043,168 @@ bool ZoneDatabase::LoadStaticZonePoints(LinkedList<ZonePoint *> *zone_point_list
...
-		if (iterator.GetData()->GetID() == spawnid)
-		{
-			if (iterator.GetData()->timer.GetRemainingTime() == 0xFFFFFFFF)
-				client->Message(Chat::White, "  %d: %1.1f, %1.1f, %1.1f: disabled", iterator.GetData()->GetID(), iterator.GetData()->GetX(), iterator.GetData()->GetY(), iterator.GetData()->GetZ());
-			else
-				client->Message(Chat::White, "  %d: %1.1f, %1.1f, %1.1f: %1.2f", iterator.GetData()->GetID(), iterator.GetData()->GetX(), iterator.GetData()->GetY(), iterator.GetData()->GetZ(), (float)iterator.GetData()->timer.GetRemainingTime() / 1000);
-
-			iSpawnIDCount++;
+	while (iterator.MoreElements()) {
+		if (iterator.GetData()->GetID() == spawn_id) {
```
Akkadius, line 2151: Could shorten things up a bit by putting just below the while

```cpp
auto &e = iterator.GetData();
```

Then `e->GetID()` etc.

Kinglykrab: Done.

zone/zone.cpp
Kinglykrab: Will lowercase.

zone/zone.cpp
Kinglykrab: Will lowercase.

zone/gm_commands/spawnstatus.cpp
Kinglykrab: Will consolidate them all in to the GM file and remove all the methods from zone.cpp/zone.h. Will just use || inside the loop to determine how to handle each type.

Akkadius (approved):

PR #2145, opened 2022-05-06
[Commands] Cleanup #summon Command.

Kinglykrab:
- Cleanup messages and logic.
- Command will now tell you if the specified character name does not exist instead of incorrectly telling you it's summoning a non-existent character.
- Add glm::vec4 overload for GMMove.
- Remove unused parameter from GMMove.
- Remove unnecessary Lua GMMove now that parameter is gone.

### DIFF
--- a/zone/gm_commands/summon.cpp
+++ b/zone/gm_commands/summon.cpp
@@ -58,55 +58,27 @@ void command_summon(Client *c, const Seperator *sep)
 	} else if (c->GetTarget()) {
 		target = c->GetTarget();
 	}
+	
+	if (c == target) {
+		c->Message(Chat::White, "You cannot summon yourself.");
+		return;
+	}
+	
+	c->Message(
+		Chat::White,
+		fmt::format(
+			"Summoning {} ({}) to {:.2f}, {:.2f}, {:.2f} in {} ({}).",
+			target->GetCleanName(),
+			target->GetID(),
+			c->GetX(),
+			c->GetY(),
+			c->GetZ(),
+			zone->GetLongName(),
+			zone->GetZoneID()
+		).c_str()
+	);
 
-	if (target->IsNPC()) {
-		c->Message(
-			Chat::White,
-			fmt::format(
-				"Summoning NPC {} ({}) to {:.2f}, {:.2f}, {:.2f} in {} ({}).",
-				target->GetCleanName(),
-				target->GetID(),
-				c->GetX(),
-				c->GetY(),
-				c->GetZ(),
-				zone->GetLongName(),
-				zone->GetZoneID()
-			).c_str()
-		);
-
-		target->GMMove(c->GetPosition());
-		target->CastToNPC()->SaveGuardSpot(glm::vec4(0.0f));
-	} else if (target->IsCorpse()) {
-		c->Message(
-			Chat::White,
-			fmt::format(
-				"Summoning Corpse {} ({}) to {:.2f}, {:.2f}, {:.2f} in {} ({}).",
-				target->GetCleanName(),
-				target->GetID(),
-				c->GetX(),
-				c->GetY(),
-				c->GetZ(),
-				zone->GetLongName(),
-				zone->GetZoneID()
-			).c_str()
-		);
-
-		target->GMMove(c->GetPosition());
-	} else if (target->IsClient()) {
-		c->Message(
-			Chat::White,
-			fmt::format(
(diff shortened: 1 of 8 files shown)


zone/gm_commands/summon.cpp
```diff
@@ -7,82 +7,79 @@ extern WorldServer worldserver;
 
 void command_summon(Client *c, const Seperator *sep)
 {
-	Mob *t;
+	int arguments = sep->argnum;
+	if (!arguments && !c->GetTarget()) {
```
Akkadius, line 11: May not want to require a target to get the help menu.

You could have a target and only type `#summon` but then not get help menu

zone/gm_commands/summon.cpp
```diff
@@ -7,82 +7,79 @@ extern WorldServer worldserver;
...
-				strcpy(szp->zone, zone->GetShortName());
-				szp->x_pos       = c->GetX(); // May need to add a factor of 8 in here..
-				szp->y_pos       = c->GetY();
-				szp->z_pos       = c->GetZ();
-				szp->instance_id = zone->GetInstanceID();
-				worldserver.SendPacket(pack);
-				safe_delete(pack);
-			}
+			auto pack = new ServerPacket(ServerOP_ZonePlayer, sizeof(ServerZonePlayer_Struct));
+			ServerZonePlayer_Struct *szp = (ServerZonePlayer_Struct *) pack->pBuffer;
```
Akkadius, line 44: ServerZonePlayer_Struct can be `auto`

zone/gm_commands/summon.cpp
```diff
@@ -7,82 +7,79 @@ extern WorldServer worldserver;
...
-	Mob *t;
+	int arguments = sep->argnum;
+	if (!arguments && !c->GetTarget()) {
+		c->Message(Chat::White, "Usage: #summon - Summon your target, if you have one, to your position");
+		c->Message(Chat::White, "Usage: #summon [Character Name] - Summon a character by name to your position");
+		c->Message(Chat::White, "Note: You may also summon your target if you have one.");
+		return;
+	}
+
+	Mob* target;
```
Akkadius, line 18: Totally ok and actually preferred to single letter variables of complex type receivers.

Complex types meaning structs, classes.

Example, `Client *c` vs `Client *client` is not any less understandable or readable in code. It actually does make things more readable and less verbose as you access it many times in following code.

Common practice in Go: https://github.com/golang/go/wiki/CodeReviewComments#receiver-names

zone/gm_commands/summon.cpp
```diff
@@ -92,6 +89,12 @@ void command_summon(Client *c, const Seperator *sep)
 			2,
 			GMSummon
 		);
+	} else {
```
Akkadius, line 92: Can we simply return inside `if (target->IsClient()) {` and not rely on the `else` block here?

If `if (target->IsClient()) {` is true we should never reach this code which would have the same effect.

Materially this doesn't change the end result just is a bit cleaner to read

```cpp
// example
if (target->IsClient()) {
	// do stuff
	return
}

// old else code
target->GMMove(c->GetPosition());
if (target->IsNPC()) {
	target->CastToNPC()->SaveGuardSpot(glm::vec4(0.0f));
}
```

zone/mob.cpp
```diff
@@ -2433,7 +2433,7 @@ void Mob::ShowBuffList(Client* client) {
 	}
 }
 
-void Mob::GMMove(float x, float y, float z, float heading, bool SendUpdate) {
```
Akkadius, line 2436: Good catch with `SendUpdate` not being used anymore

zone/gm_commands/summon.cpp
Kinglykrab: Well, if you have a target and use #summon, it will summon your target.

The idea is you would only ever need the help menu if you don't currently have a target, otherwise I'd have to add a parameter to tell it when to summon when you have a target and when not to summon when you have a target.

zone/gm_commands/summon.cpp
Kinglykrab: True enough.

zone/gm_commands/summon.cpp
Kinglykrab: Yeah no problem, I do that everywhere else, just didn't here.

zone/gm_commands/summon.cpp
Akkadius: That's fair

Akkadius (approved):

PR #2146, opened 2022-05-06
[Commands] Cleanup #summonburiedplayercorpse Command.

Kinglykrab:
- Cleanup messages and logic.

### DIFF
--- a/zone/gm_commands/summonburiedplayercorpse.cpp
+++ b/zone/gm_commands/summonburiedplayercorpse.cpp
@@ -3,26 +3,34 @@
 
 void command_summonburiedplayercorpse(Client *c, const Seperator *sep)
 {
-	Client *t = c;
-
+	auto target = c;
 	if (c->GetTarget() && c->GetTarget()->IsClient() && c->GetGM()) {
-		t = c->GetTarget()->CastToClient();
-	}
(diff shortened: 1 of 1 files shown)


zone/gm_commands/summonburiedplayercorpse.cpp
```diff
@@ -3,26 +3,34 @@
...
-	else {
-		c->Message(Chat::White, "You must first select a target!");
-		return;
+		target = c->GetTarget()->CastToClient();
 	}
 
-	Corpse *PlayerCorpse = database.SummonBuriedCharacterCorpses(
-		t->CharacterID(),
-		t->GetZoneID(),
+	auto corpse = database.SummonBuriedCharacterCorpses(
```
Akkadius, line 11: Small thing, but doesn't hurt to keep the pointer `auto *corpse` because it prevents a copy. Doesn't make a huge material impact here but good practice to not copy when you don't have to especially in hotter paths

Akkadius (approved):

zone/gm_commands/summonburiedplayercorpse.cpp
Kinglykrab: 👍

PR #2147, opened 2022-05-06
[Commands] Cleanup #time and #timezone Command.

Kinglykrab:
- Cleanup messages and logic.

### DIFF
--- a/zone/gm_commands/time.cpp
+++ b/zone/gm_commands/time.cpp
@@ -2,31 +2,144 @@
 
 void command_time(Client *c, const Seperator *sep)
 {
-	char timeMessage[255];
-	int  minutes = 0;
-	if (sep->IsNumber(1)) {
-		if (sep->IsNumber(2)) {
-			minutes = atoi(sep->arg[2]);
-		}
-		c->Message(Chat::Red, "Setting world time to %s:%i (Timezone: 0)...", sep->arg[1], minutes);
-		zone->SetTime(atoi(sep->arg[1]) + 1, minutes);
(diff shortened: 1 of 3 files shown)


zone/gm_commands/time.cpp
```diff
@@ -2,31 +2,144 @@
...
+			fmt::format(
+				"Setting world time to {}:{}{} {} (Timezone: {}:{}{} {}).",
+				(
+					((hours - 1) % 12) == 0 ?
+					12 :
+					((hours - 1) % 12)
+				),
+				(
+					minutes < 10 ?
+					"0" :
```
Akkadius, line 77: Might be a little cleaner to pad the `0` to read `00:00` vs `0:0` like clock time

Example: https://github.com/EQEmu/Server/blob/master/common/server_event_scheduler.cpp#L36

Also I see this ternary is used in a lot of places in this PR; might be worth extracting to a helper method under the `EQTime` class to read back the time in string form

zone/gm_commands/time.cpp
```diff
@@ -2,31 +2,144 @@
...
+					"PM" :
+					"AM"
+				)
+			).c_str()
+		);
+
+		zone->SetTime(hours, minutes);
+
+		LogInfo(
+			"{} :: Setting world time to {}:{} {} (Timezone: {}:{}{} {})",
```
Akkadius, line 108: Reading 8 placeholder positions here and 9 arguments

Akkadius (approved):

PR #2148, opened 2022-05-06
[Commands] Cleanup #trapinfo Command.

Kinglykrab:
- Cleanup messages and logic.
- Does not modify command file, just the function called by the command.

### DIFF
--- a/zone/trap.cpp
+++ b/zone/trap.cpp
@@ -421,7 +421,7 @@ void EntityList::GetTrapInfo(Client* client)
 	client->Message(
 		Chat::White,
 		fmt::format(
-			"{} Trap{} found.",
+			"{} trap{} found.",
 			trap_count,
 			trap_count != 1 ? "s" : ""
 		).c_str()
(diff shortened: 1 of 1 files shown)


zone/trap.cpp
```diff
@@ -366,21 +366,66 @@ void EntityList::UpdateAllTraps(bool respawn, bool repopnow)
...
-		if (cur->IsTrap())
-		{
-			bool isset = (cur->chkarea_timer.Enabled() && !cur->reset_timer.Enabled());
-			client->Message(Chat::Default, " Trap: (%d) found at %0.2f,%0.2f,%0.2f. Times Triggered: %d Is Active: %d Group: %d Message: %s", cur->trap_id, cur->m_Position.x, cur->m_Position.y, cur->m_Position.z, cur->times_triggered, isset, cur->group, cur->message.c_str());
-			++count;
+	uint32 trap_count = 0;
+	uint32 trap_number = 1;
+
+	for (const auto& trap : trap_list) {
+		auto current_trap = trap.second;
```
Akkadius, line 373: Can just use `t` in place of `current_trap` to simplify the receiver 

There are 13 calls to the receiver below should shorten things up

Akkadius (approved):
lgtm

zone/trap.cpp
Kinglykrab: 👍

PR #2149, opened 2022-05-06
[Commands] Cleanup #ucs Command.

Kinglykrab:
- Cleanup messages and logic.


zone/gm_commands/ucs.cpp
```diff
@@ -2,88 +2,87 @@
 
 void command_ucs(Client *c, const Seperator *sep)
 {
-	if (!c) {
+	LogInfo(
```
Akkadius, line 5: Not sure how useful this is as this doesn't actually trigger a reconnect

Akkadius (approved):

zone/gm_commands/ucs.cpp
Kinglykrab: Yeah, me either, it literally does nothing from my testing, but I figured I'd clean it up.

Kinglykrab:
Would you prefer this be removed? Not sure of proper way to reconnect to UCS, just cleaned up what was here.

PR #2152, opened 2022-05-07
[Bug Fix] restore some messages lost due to dup fix 2 years ago

noudess:
Some messages for spells were not getting sent anymore because of a 2 year old fix for duplicate messages.

Old Fix that fixed dups but lost some messages:  https://github.com/EQEmu/Server/commit/60e194e32bb5af2049ff1900c408fa663e60ffef

This is tested but lets get feedback and let me test for another week.

### DIFF
--- a/common/spdat.cpp
+++ b/common/spdat.cpp
@@ -139,7 +139,7 @@ bool IsEvacSpell(uint16 spellid)
 
 bool IsDamageSpell(uint16 spellid)
 {
-	if (spells[spellid].target_type == ST_Tap)
+	if (IsLifetapSpell(spellid))
 		return false;
 
 	for (int o = 0; o < EFFECT_COUNT; o++) {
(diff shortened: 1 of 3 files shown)


noudess:
Updated after finding more duplicates of DD messages cast by NPCs.

noudess:
Lifetap spells cast as spells rather than procs have an issue of (2) xyz Staggers to close clients other than the target.  Lifetap cast as a proc do NOT.  I believe this is a bug in CommonDamage and as such I will address it separately.

nytmyr:
Do you know if this resolves the double messages for Canni for Shaman?
For the casting player it shows the "You body aches as your mind clears. You have taken x points of damage." But then repeats a "Your body aches as your mind clears."
When another casts it you see a double "xxx winces."

noudess:
> Do you know if this resolves the double messages for Canni for Shaman? For the casting player it shows the "Your body aches as your mind clears. You have taken x points of damage." But then repeats a "Your body aches as your mind clears." When another casts it you see a double "xxx winces."

Yes.  This removes the 2nd message.

noudess:
Made alot of progress today.  Lifetap is the really odd duck.  When cast by NPCs it all looked good, but in PVP it was  hosed by my fix.  

We have a new fix for this and tested it about 75%.  Gonna test more tomorrow and push changes.,

PR #2153, opened 2022-05-07
[Commands] Cleanup #cvs Command.

Kinglykrab:
- Cleanup messages and logic.
- std::find was using .begin() in both comparisons, so it wasn't working and unique IPS always returned 1.


noudess:
Helped test.  Looks good to me.

Akkadius (approved):

PR #2160, opened 2022-05-08
[Merchants] Add Merchant Data Bucket capability.

Kinglykrab:
- Allows server operators to limit merchant items based on data bucket values and comparisons.
- Adds 3 columns, bucket_name, bucket_value, and bucket_comparison to merchantlist table.
- Bucket is checked based on GetBucketKey()-bucket_name.
- Buckets are mass loaded when using the merchant so it's not a database call per item, just a grouping of all their buckets from the start.
- This is a nearly year old pull request redone for master.
- bucket_comparison Values are as follows:
```
0: bucket_name == bucket_value
1: bucket_name != bucket_value
2: bucket_name >= bucket_value
3: bucket_name <= bucket_value
4: bucket_name > bucket_value
5: bucket_name < bucket_value
6: bucket_name is any of pipe(|)-separated bucket_value
7: bucket_name is not any of pipe(|)-separated bucket_value
8: bucket_name is between first and second value of pipe(|)-separated bucket_value
9: bucket_name is not between first and second value of pipe(|)-separated bucket_value
```

Example logs when I was testing.
image

### DIFF
--- a/zone/client_process.cpp
+++ b/zone/client_process.cpp
@@ -813,152 +813,168 @@ void Client::BulkSendInventoryItems()
 }
 
 void Client::BulkSendMerchantInventory(int merchant_id, int npcid) {
-	const EQ::ItemData* handyitem = nullptr;
-	uint32 numItemSlots = 80; //The max number of items passed in the transaction.
+	const EQ::ItemData* handy_item = nullptr;
+
+	uint32 merchant_slots = 80; //The max number of items passed in the transaction.
 	if (m_ClientVersionBit & EQ::versions::maskRoFAndLater) { // RoF+ can send 200 items
-		numItemSlots = 200;
+		merchant_slots = 200;
 	}
+
 	const EQ::ItemData *item = nullptr;
-	std::list<MerchantList> merlist = zone->merchanttable[merchant_id];
-	std::list<MerchantList>::const_iterator itr;
-	Mob* merch = entity_list.GetMobByNpcTypeID(npcid);
-	if (merlist.size() == 0) { //Attempt to load the data, it might have been missed if someone spawned the merchant after the zone was loaded
+	auto merchant_list = zone->merchanttable[merchant_id];
+	auto npc = entity_list.GetMobByNpcTypeID(npcid);
+	if (!merchant_list.size() == 0) {
 		zone->LoadNewMerchantData(merchant_id);
-		merlist = zone->merchanttable[merchant_id];
-		if (merlist.size() == 0)
+		merchant_list = zone->merchanttable[merchant_id];
+		if (!merchant_list.size()) {
 			return;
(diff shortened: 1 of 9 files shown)


zone/zone.cpp
```diff
@@ -634,25 +621,28 @@ void Zone::LoadNewMerchantData(uint32 merchantid) {
         return;
 	}
 
-	for (auto row = results.begin(); row != results.end(); ++row) {
+	for (auto row : results) {
```
Akkadius, line 624: `&row` to avoid unnecessary copy

zone/zone.cpp
```diff
@@ -605,26 +605,13 @@ void Zone::LoadTempMerchantData()
 
 void Zone::LoadNewMerchantData(uint32 merchantid) {
 
-	std::list<MerchantList> merlist;
+	std::list<MerchantList> merchant_list;
 
-	std::string query = fmt::format(
-		SQL(
```
Akkadius, line 611: Any reason we scrunched up the query (readability) ? Repository usage would also clean this up

zone/zone.cpp
Akkadius: And const

zone/client_process.cpp
```diff
@@ -813,152 +813,168 @@ void Client::BulkSendInventoryItems()
...
-	std::list<TempMerchantList> tmp_merlist = zone->tmpmerchanttable[npcid];
-	std::list<TempMerchantList>::iterator tmp_itr;
 
-	uint32 i = 1;
-	uint8 handychance = 0;
-	for (itr = merlist.begin(); itr != merlist.end() && i <= numItemSlots; ++itr) {
-		MerchantList ml = *itr;
-		if (ml.probability != 100 && zone->random.Int(1, 100) > ml.probability)
-			continue;
+	auto client_data_buckets = GetMerchantDataBuckets();
```
Akkadius, line 834: Thanks for turning this into a single DB call from the last time

Akkadius (approved):
LGTM.

Really make sure there's docs around this so users know how to find it (eqemu.docs.io) and more importantly know how to use it.

zone/zone.cpp
Kinglykrab: I can put it back the way it was, I figured it was more readable this way.

Kinglykrab:
Documented and available already: https://docs.eqemu.io/server/scripting/merchant-data-buckets/

Going to wait on people to put some eyes back on this so if there's something more I need to do here, I can before merge.

zone/zone.cpp
Akkadius: Yeah the underlying doesn't exactly play nicely

Akkadius:
Docs look good

Issue #1905, opened 2021-12-29
#reloadstatic does not work properly on ground spawns


noudess:
Using #reloadstatic

- Eliminates groundspawns from #list objects
- Does not seem to impact the client's ground spawns themselves at all

Natedog2012:
https://github.com/EQEmu/Server/pull/2815

Fixes the issue listed above... however the individual commands #reload doors / #reload objects ect will still cause issues... I just only using #reload static when trying to reload doors/objects/groundspawns

PR #2162, opened 2022-05-09
[Commands] #reload Command Overhaul.

Kinglykrab:
- Consolidated #reloadaa, #reloadallrules, #reloadcontentflags, #reloademote, #reloadlevelmods, #reloadmerchants, #reloadperlexportsettings, #reloadqst, #reloadstatic, #reloadtitles, #reloadtraps, #reloadworld, and #reloadzps in to one command.
- #reload has 15 different sub commands you may use, including Log Settings and Tasks reloading.
- All the reload commands are a part of the Developer Tools Menu messages now, as well as part of the documentation.
- Fixes the commands that weren't actually sending their packet to zone server to globally reload stuff.
- Reorder packets a bit (alphabetically) and move Reload packets to their own ID range.
- Added Variables table reloading to command.

**Pending Documentation**
https://github.com/EQEmu/eqemu-docs-v2/pull/35

**Developer Tools Menu**
image

**Reload Command Menu**
image


Kinglykrab:
There are some other things I'd like to add to this.

Kinglykrab:
Ready for review.

joligario:
May be worth testing against #1905

zone/zone.cpp
```diff
@@ -919,21 +919,19 @@ void Zone::Shutdown(bool quiet)
...
 		LogInfo("No doors loaded");
 		return;
 	}
 
-	for (const auto& entry : door_entries)
-	{
-		auto newdoor = new Doors(entry);
-		entity_list.AddDoor(newdoor);
+	for (const auto &entry : door_entries) {
+		auto d = new Doors(entry);
```
Akkadius, line 933: Yay single letter receiver

Kinglykrab: 👍

zone/client.cpp
```diff
@@ -9514,63 +9514,202 @@ void Client::CheckVirtualZoneLines()
 
 void Client::ShowDevToolsMenu()
 {
-	std::string menu_commands_search;
-	std::string menu_commands_show;
-	std::string reload_commands_show;
-	std::string devtools_toggle;
+	std::string menu_search;
+	std::string menu_show;
+	std::string menu_reload_one;
```
Akkadius, line 9519: Not in love with the one, two etc.

We could vector string build and then build each display line dynamically based on number of elements

Fine for now

Akkadius (approved):
Looks great

zone/client.cpp
Kinglykrab: That sounds good to me, only reason I did it this way was because I manually built the menu to where the stuff would fit well.

Kinglykrab:
> May be worth testing against #1905

This doesn't fix that issue, I just added the code for the command made sure it at least sent the packet.

Fixing ground spawns will require changing the way they're loaded/reloaded.

Kinglykrab:
Not sure if we want to wait on merge for ground spawns fix or not. Seems like the code for them is just really old and would need a revamp.

Akkadius:
> Not sure if we want to wait on merge for ground spawns fix or not. Seems like the code for them is just really old and would need a revamp.

Separate scope imo

PR #2167, opened 2022-05-11
[Opcode] Implement SetFace opcode 

hgtw:
This implements the opcode that's used to update zone clients when a client updates their face and uses it for the #feature eyes command which isn't in the illusion struct


Kinglykrab:
Do the separate eye colors not work?

hgtw:
Client overrides both eyes to the first one

Kinglykrab (approved):

PR #2170, opened 2022-05-13
[Bug Fix] Fix duplicate and missing messages due to innate in spells

noudess:
The fact that some spells use skill_id 52 (innate) has caused confusion and multiple attempts at understanding why some messages are missing and some are duplicated.

This PR fixes spells that use skill_id 52 to send (2) messages to the person being damaged (attack.cpp)

1) The strike message
2) The actual spell effect (life drains away, you have been poisoned, etc).

Some of the above were being send by spells.cpp because attack.cpp was missing them.

This has been tested using lifetaps, DD, DD+DOT, DOT, root w/damage,etc.

It has been tested in PVP,, and npcs procing as well as casting,

I think more messages come out now (that we were missing) and all duplicates are gone.

Fingers crossed.

@nytmyr This fixes canni also..

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -4021,6 +4021,24 @@ void Mob::CommonDamage(Mob* attacker, int64 &damage, const uint16 spell_id, cons
 		// we don't send them here.
 		if (!FromDamageShield) {
 
+			int range;
+			if (IsValidSpell(spell_id)) {
+				range = RuleI(Range, SpellMessages);
+			}
+			else {
+				range = RuleI(Range, DamageMessages);
+			}
+
+			entity_list.QueueCloseClients(
+				this, /* Sender */
+				outapp, /* packet */
+				true, /* Skip Sender */
+				range, /* distance packet travels at the speed of sound */
+				(IsValidSpell(spell_id)) ? 0 : skip, /* Skip if not spell */
+				true, /* Packet ACK */
+				filter /* eqFilterType filter */
+				);
+
 			// Send damage to client (including inates skill_id)
 			if (IsClient()) {
 				//I dont think any filters apply to damage affecting us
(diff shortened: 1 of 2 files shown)


noudess:
Added code to use correct Range: rule based on type of message.

noudess:
Refined innate spell damage messages to only come out when "non-melee" damage wasn't coming out so we don't throw off parsers.  Still preserving all new innate spell based messages (xyz was poisoned" etc.)

zone/attack.cpp
```diff
@@ -4029,27 +4030,38 @@ void Mob::CommonDamage(Mob* attacker, int64 &damage, const uint16 spell_id, cons
...
-
-			// Send damage to client (including inates skill_id)
-			if (IsClient()) {
-				//I dont think any filters apply to damage affecting us
-				CastToClient()->QueuePacket(outapp);
+			// If an "innate" spell, change to spell type to
+			// produce a spell message.  Send to everyone.
+			// This fixes issues with npc-procs like 1002 and 918 which
+			// need to spit out extra spell color.
+			if (IsValidSpell(spell_id) && skill_used == 52) {
```
Kinglykrab, line 4037: EQ::skills:SkillTigerClaw

zone/attack.cpp
```diff
@@ -4029,27 +4030,38 @@ void Mob::CommonDamage(Mob* attacker, int64 &damage, const uint16 spell_id, cons
...
-			if (IsClient() && IsValidSpell(spell_id) && skill_used == 52) {
-				a->type = DamageTypeSpell;
-				CastToClient()->QueuePacket(outapp);
+				// Otherwise, send normal spell or melee message to observers.
+				entity_list.QueueCloseClients(
+					this, /* Sender */
+					outapp, /* packet */
+					true, /* Skip Sender */
+					range, /* distance packet travels at the speed of sound */
+					(IsValidSpell(spell_id) && skill_used != 52) ? 0 : skip,
```
Kinglykrab, line 4061: EQ::skills::SkillTigerClaw

noudess: Yeah - weird right.  This is what is in skill field in spells_new for spells like 1002....

noudess: Do we really want to use that constant or something that desribes it diff?

Kinglykrab: The constant is descriptive enough, it's just a skill ID and that's all we need to know versus a magic "52".

noudess: Done!

Kinglykrab (approved):

noudess:
This one looks good.  Testing finding no bugs for several days.

PR #2174, opened 2022-05-14
[Titles] Cleanup titles, title suffix, and last name methods.

Kinglykrab:
- Use strings instead of const chars*.
- Add optional parameter to SetAATitle in Lua so you can save to the database similar to Perl.
- Cleanup #lastname command.
- Cleanup #title command.
- Cleanup #titlesuffix command.

### DIFF
--- a/zone/npc.cpp
+++ b/zone/npc.cpp
@@ -3253,7 +3253,6 @@ void NPC::ChangeLastName(std::string last_name)
 void NPC::ClearLastName()
 {
 	std::string empty;
-	empty = '\0'; //Clear Last Name
 	ChangeLastName(empty);
 }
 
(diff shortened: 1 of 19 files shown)


zone/npc.cpp
```diff
@@ -3231,27 +3231,30 @@ void NPC::DoQuestPause(Mob *other) {
...
 	safe_delete(outapp);
 }
 
 void NPC::ClearLastName()
 {
-	std::string WT;
-	WT = '\0'; //Clear Last Name
-	ChangeLastName( WT.c_str());
+	std::string empty;
+	empty = '\0'; //Clear Last Name
```
hgtw, line 3256: Can remove this assignment since the string is already empty by default

hgtw (approved):

zone/npc.cpp
Kinglykrab: True enough.

PR #2177, opened 2022-05-15
[Quest API] Add TaskSelector to Perl/Lua.

Kinglykrab:
- Add $client->TaskSelector(task_list) to Perl.
- Add client:TaskSelector({task_list}) to Lua.
- Allow $client->AssignTask(task_id) in Perl.
- Allow client:AssignTask(task_id) in Lua.
- Can now assign tasks in scripts without the NPC ID parameter.

### DIFF
--- a/zone/client.h
+++ b/zone/client.h
@@ -1236,7 +1236,7 @@ public:
 	}
 	inline void AssignTask(
 		int task_id,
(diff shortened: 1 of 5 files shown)


Akkadius (approved):
LFGTM!

PR #2183, opened 2022-05-20
[Rules] Remove unused avoidance cap rules.

Kinglykrab:
These rules are no longer used since the combat overhaul.


Kinglykrab:
Closed because there is a lot more unused rules, so I made another removing them all: https://github.com/EQEmu/Server/pull/2184

PR #2186, opened 2022-05-20
[Login] Added OP_ExpansionPacketData for RoF2 and update payload for Titanium

neckkola:
Good afternoon.

I have added the ability to update the expansion info on the Server Select screen for Titanium and Rof2.  Titanium uses the encrypted payload while RoF2 uses a new packet structure.

I have based it on using the rule World:ExpansionSettings and force the client to stay within it's max as it is based on data within eqlsstr_us.txt.

Open to suggestions on the implementation.


loginserver/client.cpp
```diff
@@ -583,6 +588,73 @@ void Client::DoSuccessfulLogin(
...
+								3045, 3046, 3047, 3514, 3516,
+								3518, 3520, 3522, 3524 };
+
+
+	//Get the active expansion from the database Rule:WorldExpansionSettings.  Do this only once.  Requires restart to effect change
+	if (!owned_expansion) {
+		std::string r_name = RuleManager::Instance()->GetRulesetName(server.db, default_ruleset);
+		if (r_name.size() > 0) {
+			RuleManager::Instance()->LoadRules(server.db, r_name.c_str(), false);
+			expansion = RuleManager::Instance()->GetIntRule(RuleManager::Int__ExpansionSettings);
```
hgtw, line 607: A config option might make more sense for this since it's loaded once anyway, but I'll defer to others for that

hgtw: Also are rules a standard table for the login server database? Mine doesn't have it but I don't remember the last time I sourced anything so it might just be really out of date

Akkadius: Loginserver does not use rules. Configuration should be driven from the config file

loginserver/client.cpp
```diff
@@ -583,6 +588,73 @@ void Client::DoSuccessfulLogin(
...
+			buf.WriteInt32(0xFFFFFFFF);
+		}
+
+		auto out = std::make_unique<EQApplicationPacket>(OP_LoginExpansionPacketData, buf);
+		m_connection->QueuePacket(out.get());
+
+	}
+	else if (m_client_version == cv_titanium)
+	{
+		if (expansion > 1023)
```
hgtw, line 642: Changing this to `>= EQ::expansions::bitPoR` would make this a little clearer

loginserver/client.cpp
```diff
@@ -583,6 +588,73 @@ void Client::DoSuccessfulLogin(
...
+		{
+			// Titanium shipped with 10 expansions.  Set owned expansions to be max 10.
+			plrs.offer_min_days = (1023 << 1) | 1;
```
hgtw, line 645: Same as above, maybe using `bitTSS -1` or adding a couple titanium specific consts to clear this up

loginserver/client.cpp
```diff
@@ -583,6 +588,73 @@ void Client::DoSuccessfulLogin(
...
+		}
+		else
+		{
+			plrs.offer_min_days = owned_expansion;
```
hgtw, line 649: Were the fields here misidentified or are they just completely different for Titanium? A separate Titanium specific struct in `login_types.h` would be useful just for documentation purposes if you have it mapped out

neckkola:
Thanks for the feedback.  I will make some updates for consideration.

neckkola:
Good evening.  Updated the implementation to use login.json -> client_configuration.  Another update will come for the struct info.  Please let me know if there is any feedback/suggestions.  Thank you.

loginserver/client.cpp
```diff
@@ -5,8 +5,11 @@
 #include "../common/string_util.h"
 #include "encryption.h"
 #include "account_management.h"
+#include "../common/rulesys.h"
```
hgtw, line 8: This include should no longer be needed

loginserver/client.cpp
```diff
@@ -5,8 +5,11 @@
...
 
 extern LoginServer server;
+int32_t expansion = 0;
+int32_t owned_expansion = 0;
```
hgtw, line 12: Both of these can be moved to local variables in `SendExpansionPacketData`

loginserver/client.cpp
```diff
@@ -583,6 +588,70 @@ void Client::DoSuccessfulLogin(
 	m_client_status = cs_logged_in;
 }
 
+void Client::SendExpansionPacketData(PlayerLoginReply_Struct& plrs)
+{
+	int default_ruleset = 1;
```
hgtw, line 593: This is unused now

loginserver/login_util/login.json
```diff
@@ -36,5 +36,7 @@
     "titanium_opcodes": "login_opcodes.conf",
     "sod_port": 5999,
     "sod_opcodes": "login_opcodes_sod.conf"
+    "display_expansions" : true,
+	"max_expansions" : 524287
```
hgtw, line 40: Broken indent

loginserver/options.h
```diff
@@ -142,6 +152,7 @@ class Options {
 	bool        world_trace;
 	bool        dump_in_packets;
 	bool        dump_out_packets;
+	bool		display_expansions;
```
hgtw, line 155: Replace the tab with spaces between the `bool` and `display_expansions` here for alignment

loginserver/options.h
```diff
@@ -152,6 +163,7 @@ class Options {
 	bool        auto_link_accounts;
 	bool        update_insecure_passwords;
 	int         encryption_mode;
+	int			max_expansions;
```
hgtw, line 166: Replace the tab with spaces between the `int` and `max_expansions` here for alignment

Akkadius:
Still pending re-visitation by @neckkola

neckkola:
Sorry for my tardiness.  It has been busy with some work things.  I will clean this up and get it ready for further review.

loginserver/client.cpp
neckkola: updated to use login.json instead of rules table.

loginserver/client.cpp
neckkola: updated to use EQ::expansions

loginserver/client.cpp
neckkola: updated as noted above

loginserver/client.cpp
neckkola: Some of the fields are missed labelled though I have not confirmed the impact on other clients so left them the same.  I am reviewing this now though I do not have access to other clients so I can only test with Titanium and RoF2.  Happy to share my work on the struct and adjust as results are shown.  Could also add some functionality to lock a login server so clients get a soft landing (a client message) if the server is down for maintenance and add the 'offers' functionality.   Open to thoughts and feedback.

loginserver/client.cpp
neckkola: removed.

loginserver/client.cpp
neckkola: Done.

loginserver/client.cpp
neckkola: Done

loginserver/login_util/login.json
neckkola: Done

loginserver/options.h
neckkola: Done

loginserver/options.h
neckkola: Done

neckkola:
Good evening.  Updates have been made to resolve the comments (save for the full struct work).  Would appreciate any further comments/suggestions and testing.  I have tested with Titanium and RoF2.  Please let me know and again, sorry for my tardiness.

loginserver/login_util/login.json
```diff
@@ -36,5 +36,7 @@
     "titanium_opcodes": "login_opcodes.conf",
     "sod_port": 5999,
     "sod_opcodes": "login_opcodes_sod.conf"
+    "display_expansions" : true,
+    "max_expansions" : 524287
```
hgtw, line 40: Let's rename this to `max_expansions_mask` since it might be confusing

loginserver/client.cpp
```diff
@@ -583,6 +585,70 @@ void Client::DoSuccessfulLogin(
...
+			auto out = std::make_unique<EQApplicationPacket>(OP_LoginExpansionPacketData, buf);
+			m_connection->QueuePacket(out.get());
+
+		}
+		else if (m_client_version == cv_titanium)
+		{
+			if (expansion >= EQ::expansions::bitPoR)
+			{
+				// Titanium shipped with 10 expansions.  Set owned expansions to be max 10.
+				plrs.offer_min_days = ((EQ::expansions::bitDoD << 2) | 1) - 2;
```
hgtw, line 638: Sorry I didn't notice this before but the expansions enum also has masks that you can set this directly to.

Since titanium has a max of 10 expansions should it be `EQ::expansions::maskDoD`  (1023)? or am I misunderstanding why it's also including bitPoR (if so this can be set directly to `maskPoR`)

loginserver/client.cpp
hgtw: The additional lock and offers features can be done in another pr (even if they're not used it'll be nice to have the functionality documented)

loginserver/login_util/login.json
```diff
@@ -36,5 +36,7 @@
     "titanium_opcodes": "login_opcodes.conf",
     "sod_port": 5999,
     "sod_opcodes": "login_opcodes_sod.conf"
```
hgtw, line 38: Need a comma at the end of this line

loginserver/client.cpp
neckkola: My logic is based on the fact that you can force Titanium to display beyond its limit of 10.  For example, You own 18 of 27 expansions.  However, the equivalent strings are not in the eqls file and therefore the english does not work.  Therefore thought it best to add the limit.  Further, I wanted to use the same masking throughout for consistency and then create the formula to populate the bits to the right of the max.  Hope that makes sense.  :-)

loginserver/login_util/login.json
neckkola: updated

loginserver/login_util/login.json
```diff
@@ -35,6 +35,8 @@
     "titanium_port": 5998,
     "titanium_opcodes": "login_opcodes.conf",
     "sod_port": 5999,
-    "sod_opcodes": "login_opcodes_sod.conf"
+    "sod_opcodes": "login_opcodes_sod.conf",
+    "display_expansions" : true,
+    "max_expansions" : 524287
```
hgtw, line 40: Missed renaming this one

neckkola: Thanks for the catch.  Updated.  Also thanks for the comments, advice and second set of eyes.  I really appreciate it.

hgtw (approved):
This works in sod and underfoot as well. SoF didn't work but it didn't prevent the login and that client probably isn't used much if at all anymore (at a glance it looked like it uses the new struct but maybe not the separate opcode). Pending the optional struct work I think this is ok

neckkola:
Glad it works on SoD and Underfoot as well!  If I can find/be pointed to SoF I would be happy to test it as well.  The recent clients use a different opcode and a slightly different struct.  Thanks again and I hope someone finds this 'vanity' update useful.  :-)

PR #2187, opened 2022-05-20
[Commands] Cleanup #lastname, #npcedit, #title, and #titlesuffix Commands.

Kinglykrab:
- Cleanup messages and logic.
- Commifies messages where commification would make reading easier.
- Requires parameters be set for most subcommands to avoid undefined behavior and to make sure of intent when using subcommands.
- Help messages when incorrectly using any subcommand where not providing a parameter is not a valid use case.


common/emu_constants.cpp
```diff
@@ -380,5 +391,51 @@ std::string EQ::constants::GetEnvironmentalDamageName(uint8 damage_type)
...
+		{ SpawnAnimations::Looting, "Looting" }
+	};
+
+	return spawn_animation_map;
+}
+
+std::string EQ::constants::GetSpawnAnimationName(uint8 animation_id)
+{
+	if (EQ::ValueWithin(animation_id, SpawnAnimations::Standing, SpawnAnimations::Looting)) {
+		auto spawn_animations = EQ::constants::GetSpawnAnimationMap();
```
Akkadius, line 436: Any reason we didn't just inline this with the return

Kinglykrab: Not sure what you mean here. Glad to do whatever cleans it up, just need clarification.

Akkadius: `return EQ::constants::GetSpawnAnimationMap()[animation_id]`

Haven't tested, but just inlines without declaring

Kinglykrab: Makes sense, guess I could clean up the others to do that, assuming it works.

Kinglykrab: Converted them to use .find on the map itself.
Example:
```cpp
return EQ::constants::GetLDoNThemeMap().find(theme_id)->second;
```

PR #2189, opened 2022-05-22
[Bug Fix] Fix IP Exemptions.

Kinglykrab:
- IP Exemptions were broken due to GetAccountID() returning 0 in logic somehow, resolved this by setting a variable to GetAccountID() at the beginning of the method.
- Fixed weird IP messages where the long form of IP was displayed instead of the string form.
- Fixes edge case where IP rule may be set to -1 and this will make anyone get instantly kicked if IP Exemptions were enabled as their IP Count would always be greater than -1.

### DIFF
--- a/world/client.cpp
+++ b/world/client.cpp
@@ -721,38 +721,40 @@ bool Client::HandleCharacterCreatePacket(const EQApplicationPacket *app) {
 }
 
 bool Client::HandleEnterWorldPacket(const EQApplicationPacket *app) {
-	if (GetAccountID() == 0) {
-		LogInfo("Enter world with no logged in account");
+	auto account_id = GetAccountID();
+	if (!account_id) {
+		LogInfo("Enter world with no logged in account.");
 		eqs->Close();
 		return true;
 	}
 
-	if(GetAdmin() < 0)
-	{
-		LogInfo("Account banned or suspended");
+	if (GetAdmin() < 0) {
+		LogInfo("Account [{}] is banned or suspended.", account_id);
 		eqs->Close();
 		return true;
 	}
 
-	if (RuleB(World, EnableIPExemptions) || RuleI(World, MaxClientsPerIP) >= 0) {
+	if (
+		RuleB(World, EnableIPExemptions) ||
+		RuleI(World, MaxClientsPerIP) > 0
+	) {
 		client_list.GetCLEIP(GetIP()); //Check current CLE Entry IPs against incoming connection
 	}
 
-	EnterWorld_Struct *ew=(EnterWorld_Struct *)app->pBuffer;
-	strn0cpy(char_name, ew->name, 64);
+	auto ew = (EnterWorld_Struct *) app->pBuffer;
+	strn0cpy(char_name, ew->name, sizeof(char_name));
 
-	EQApplicationPacket *outapp;
-	uint32 tmpaccid = 0;
-	charid = database.GetCharacterInfo(char_name, &tmpaccid, &zone_id, &instance_id);
-	if (charid == 0) {
+	uint32 temporary_account_id = 0;
+	charid = database.GetCharacterInfo(char_name, &temporary_account_id, &zone_id, &instance_id);
+	if (!charid) {
 		LogInfo("Could not get CharInfo for [{}]", char_name);
 		eqs->Close();
 		return true;
 	}
 
 	// Make sure this account owns this character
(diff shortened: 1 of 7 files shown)


world/client.cpp
```diff
@@ -816,27 +821,26 @@ bool Client::HandleEnterWorldPacket(const EQApplicationPacket *app) {
...
 
-	if(instance_id > 0)
-	{
-		if (!database.VerifyInstanceAlive(instance_id, GetCharID()) ||
-		    !database.VerifyZoneInstance(zone_id, instance_id))
-		{
+	if (instance_id) {
+		if (
+			!database.VerifyInstanceAlive(instance_id, GetCharID()) ||
+		    !database.VerifyZoneInstance(zone_id, instance_id)
```
hgtw, line 842: This broke indent due to the space alignment before the second condition. (I also find this style a bit hard to read with the if scope parentheses newlines but that's just personal preference)

world/client.cpp
```diff
@@ -845,83 +849,80 @@ bool Client::HandleEnterWorldPacket(const EQApplicationPacket *app) {
...
+			default:
+				connection_type = EQ::versions::ucsUnknown;
+				break;
 		}
 
-		buffer = StringFormat("%s,%i,%s.%s,%c%08X",
-			Config->ChatHost.c_str(),
-			Config->ChatPort,
-			Config->ShortName.c_str(),
+		buffer = fmt::format("{},{},{}.{},{}{}",
```
hgtw, line 919: This will break the formatting of the buffer since it was using a char code for connection type and padded hex for the mail key.

It will need to be `fmt::format("{},{},{}.{},{}{:08X}`

The fmt version we use doesn't like using `{:c}` for the connection type formatting for some reason (a newer version worked) so the connection type will need to be explicitly cast to format as a char instead:

`static_cast<char>(connection_type)`

The newer version of fmt doesn't even need the explicit cast so I think it just detects the enum base now

world/client.cpp
```diff
@@ -933,21 +934,21 @@ bool Client::HandleEnterWorldPacket(const EQApplicationPacket *app) {
...
+			default:
+				// retain value from previous switch
+				break;
 		}
 
-		buffer = StringFormat("%s,%i,%s.%s,%c%08X",
-			Config->MailHost.c_str(),
-			Config->MailPort,
-			Config->ShortName.c_str(),
+		buffer = fmt::format("{},{},{}.{},{}{}",
```
hgtw, line 945: same as above

world/clientlist.cpp
```diff
@@ -100,114 +100,156 @@ ClientListEntry* ClientList::GetCLE(uint32 iID) {
 
 //Check current CLE Entry IPs against incoming connection
 
-void ClientList::GetCLEIP(uint32 iIP) {
-	ClientListEntry* countCLEIPs = 0;
+void ClientList::GetCLEIP(uint32 in_ip) {
+	ClientListEntry* cle = 0;
```
hgtw, line 104: `= nullptr`

world/client.cpp
Akkadius: Mail keys will definitely need to be tested here, I broke this one myself converting to a string 

Mail is used so infrequently the last thing we want out of this is it being broken for over a year

hgtw (approved):

world/clientlist.cpp
```diff
@@ -100,114 +100,156 @@ ClientListEntry* ClientList::GetCLE(uint32 iID) {
...
 		}
+
 		iterator.Advance();
 	}
 }
 
-uint32 ClientList::GetCLEIPCount(uint32 iIP) {
-	ClientListEntry* countCLEIPs = 0;
+uint32 ClientList::GetCLEIPCount(uint32 in_ip) {
+	ClientListEntry* cle = 0;
```
hgtw, line 211: Oops, had another tab open for some reason with this highlighted. `= nullptr` here

world/clientlist.cpp
```diff
@@ -100,114 +100,156 @@ ClientListEntry* ClientList::GetCLE(uint32 iID) {
...
 	}
 
-	return IPInstances;
+	return count;
 }
 
-void ClientList::DisconnectByIP(uint32 iIP) {
-	ClientListEntry* countCLEIPs = 0;
+void ClientList::DisconnectByIP(uint32 in_ip) {
+	ClientListEntry* cle = 0;
```
hgtw, line 236: `= nullptr`

PR #2200, opened 2022-05-23
[Rules] Add Spells:BuffsFadeOnDeath.

Kinglykrab:
- Allows you to disable buffs fading on death.


hgtw (approved):
Wonder if this should be a `Character` scoped rule like `UnmemSpellsOnDeath`, but there's also some buff rules in `Spells` so probably doesn't matter

Akkadius (approved):

Akkadius:
> Wonder if this should be a `Character` scoped rule like `UnmemSpellsOnDeath`, but there's also some buff rules in `Spells` so probably doesn't matter

Agree but probably fine as is

PR #2204, opened 2022-05-23
[Commands] Cleanup #findaliases and #help Commands.

Kinglykrab:
- Cleanup messages and logic.
- Add saylinks to messages for ease of use when searching for commands as most have help messages when using them incorrectly or require no arguments and can be used immediately from the say link.
image
image
image
image
image


zone/command.cpp
```diff
@@ -44,18 +44,15 @@ extern FastMath g_Math;
 void CatchSignal(int sig_num);
 
 
-int commandcount;					// how many commands we have
+int command_count; // how many commands we have
 
 // this is the pointer to the dispatch function, updated once
 // init has been performed to point at the real function
-int (*command_dispatch)(Client *,char const *)=command_notavail;
```
Akkadius, line 51: There a reason this was changed ? This might have unforeseen implications

Akkadius: Nevermind I see it below

Akkadius (approved):
Assuming tested, LGTM. Nice work. Was wondering when someone would saylink #help

PR #2205, opened 2022-05-23
[Money Messages] Cleanup quest::givecash(), split, and task reward messages.

Kinglykrab:
- Cleans up all the money messages using ConvertMoneyToString().
- Allows quest::givecash() to have optional parameters other than copper.

### DIFF
--- a/zone/groups.cpp
+++ b/zone/groups.cpp
@@ -116,91 +116,65 @@ Group::~Group()
 //Split money used in OP_Split (/split and /autosplit).
 void Group::SplitMoney(uint32 copper, uint32 silver, uint32 gold, uint32 platinum, Client *splitter) {
 	//avoid unneeded work
-	if(copper == 0 && silver == 0 && gold == 0 && platinum == 0)
+	if (
+		!copper &&
+		!silver &&
+		!gold &&
+		!platinum
+	) {
 		return;
+	}
 
-	uint32 i;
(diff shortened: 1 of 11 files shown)


Akkadius (approved):
Assuming thoroughly tested, LGTM

zone/groups.cpp
```diff
@@ -116,91 +116,65 @@ Group::~Group()
...
+		}
+	}
+
+	auto copper_split = copper / member_count;
+	auto silver_split = silver / member_count;
+	auto gold_split = gold / member_count;
+	auto platinum_split = platinum / member_count;
+
+	auto cash_message = fmt::format(
+		"You receive {} as your split.",
```
hgtw, line 170: Think there's an eqstr id that can be used for this instead and just the money part can be sent

zone/questmgr.cpp
```diff
@@ -1170,52 +1170,35 @@ void QuestManager::untraindiscs() {
...
-			else
-				tmp += ",";
 
-			tmp += itoa(copper);
-			tmp += " copper";
+		if (initiator) {
+			initiator->Message(
+				Chat::Green,
+				fmt::format(
+					"You receive {}.",
```
hgtw, line 1197: Probably the same here, maybe eqstr 5941

hgtw (approved):

PR #2208, opened 2022-05-25
[Bug Fix] non-trivial types aren't supported with variadic methods

mackal:


Kinglykrab:
Real fix here: https://github.com/EQEmu/Server/pull/2209

PR #2214, opened 2022-05-27
Revert "[Aggro] Rooted mobs will add other hated targets to Hate list"

mackal:
Reverts EQEmu/Server#2180


noudess:
Agreed to revert based on @mackal testing conflicts with mine.

PR #2215, opened 2022-05-28
[Commands] Cleanup #npcedit, #lastname, #title, and #titlesuffix Commands.

Kinglykrab:


common/emu_constants.cpp
```diff
@@ -288,15 +290,16 @@ const std::map<bodyType, std::string>& EQ::constants::GetBodyTypeMap()
...
 	return bodytype_map;
 }
 
 std::string EQ::constants::GetBodyTypeName(bodyType bodytype_id)
 {
-	auto bodytypes = EQ::constants::GetBodyTypeMap();
-	if (!bodytypes[bodytype_id].empty()) {
-		return bodytypes[bodytype_id];
+	if (!EQ::constants::GetBodyTypeMap().find(bodytype_id)->second.empty()) {
+		return EQ::constants::GetBodyTypeMap().find(bodytype_id)->second;
```
hgtw, line 300: Can just return the result instead of checking if it's empty since it returns an empty string in the fall through case anyway. If there's worry that the function will be called with an invalid bodyType it should check something like `find(bodytype) != map.end()` instead

common/emu_constants.cpp
```diff
@@ -351,15 +356,16 @@ const std::map<uint8, std::string>& EQ::constants::GetConsiderLevelMap()
...
 	return consider_level_map;
 }
 
 std::string EQ::constants::GetConsiderLevelName(uint8 faction_consider_level)
 {
-	auto consider_levels = EQ::constants::GetConsiderLevelMap();
-	if (!consider_levels[faction_consider_level].empty()) {
-		return consider_levels[faction_consider_level];
+	if (!EQ::constants::GetConsiderLevelMap().find(faction_consider_level)->second.empty()) {
+		return EQ::constants::GetConsiderLevelMap().find(faction_consider_level)->second;
```
hgtw, line 366: Same as `GetBodyTypeName` above

zone/gm_commands/npcedit.cpp
```diff
@@ -1,552 +1,847 @@
...
-		c->Message(Chat::White, "#npcedit armortint_id - Sets an NPC's Armor Tint ID");
-		c->Message(Chat::White, "#npcedit color [red] [green] [blue] - Sets an NPC's Red, Green, and Blue armor tint");
-		c->Message(Chat::White, "#npcedit ammoidfile - Sets an NPC's Ammo ID File");
+		c->Message(Chat::White, "You must target an NPC to use this command.");
+		return;
+	}
+
+	int arguments = sep->argnum;
+
+	bool is_help = arguments >= 1 ? !strcasecmp(sep->arg[1], "help") : true;
```
hgtw, line 15: This may be more clear as something like
`bool is_help = (arguments == 0 || !strcasecmp(sep->arg[1], "help"))` instead of using a ternary

hgtw (approved):

PR #2216, opened 2022-05-28
[Quest API] Use binding library for perl apis

hgtw:
This started as a half-serious pet project experiment that became viable.

It replaces the XS apis with something that is more similar to Lua binding libraries like luabind or sol2 and will hopefully be more maintainable.

The only libraries I found to do similar were libperl++ and SWIG but I couldn't get libperl++ to compile without significant changes (at least with msvc) so I decided to try rolling my own. It was developed as an independent library with the very creative name of perlbind.

The library with test coverage can be found in its repo:
https://github.com/hgtw/perlbind

Since this is a new library this will need a lot of testing on perl heavy servers to make sure nothing is broken. Especially in long running zones that see a lot of quest activity to be sure that there's no memory leaks.

It was designed to be a drop-in replacement so quests should not need to be changed, but mistakes could have been made while converting the XS apis to use it. I audited the quest apis while converting but there was also lots of liberal regex usage.

If any servers test it and find issues report the OS and perl version the server uses in addition to the bug.


Akkadius:
This is currently baking on EZ Production - smooth so far. Spire Quest API explorer will need to be updated to handle all of the new formatting

Akkadius (approved):
Really nice work sir.

Kinglykrab (approved):

Kinglykrab:
Looks good, not sure if it's missing any API changes since this was made or not.

Akkadius:
> Looks good, not sure if it's missing any API changes since this was made or not.

I mentioned to HG there was quite a few that went in during his rebases so I think he’s got them covered.

hgtw:
Yea I added the new apis when I rebased

PR #2217, opened 2022-05-28
[Rules] Add Frontal Stun Immunity Rules.

Kinglykrab:
- Add Combat:FrontalStunImmunityClasses rule for determining Frontal Stun Immunity by class bitmasks
- Add Combat:FrontalStunImmunityRaces rule for determining Frontal Stun Immunity by race bitmasks
- Add Combat:NPCsUseFrontalStunImmunityClasses rule for determining if NPCs use Frontal Stun Immunity by class bitmasks
- Add Combat:NPCsUseFrontalStunImmunityRaces rule for determining if NPCs uses Frontal Stun Immunity by race bitmasks
- Cleanup GetDefaultGender() as best as possible with the globalload I have.
- Combat:NPCsUseFrontalStunImmunityRaces defaults to true as NPCs currently use Frontal Stun Immunity.
- Added **all** Ogre races to the check for Frontal Stun Immunity.

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -3832,11 +3832,63 @@ void Mob::CommonDamage(Mob* attacker, int64 &damage, const uint16 spell_id, cons
 				can_stun = true;
 			}
 
-			if ((GetBaseRace() == OGRE || GetBaseRace() == OGGOK_CITIZEN) &&
-				!attacker->BehindMob(this, attacker->GetX(), attacker->GetY()))
+			bool is_immune_to_frontal_stun = false;
+
+			if (IsBot() || IsClient() || IsMerc()) {
+				if (
+					IsPlayerClass(GetClass()) &&
+					RuleI(Combat, FrontalStunImmunityClasses) & GetPlayerClassBit(GetClass())
+				) {
+					is_immune_to_frontal_stun = true;
+				}
+
+
+				if (
+					(
+						IsPlayerRace(GetBaseRace()) &&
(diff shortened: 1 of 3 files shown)


mackal:
The Ogre and Ogre Citizen were the only hardcoded ones because that's what the client had as hardcoded ones.

Kinglykrab:
> The Ogre and Ogre Citizen were the only hardcoded ones because that's what the client had as hardcoded ones.

Should I just leave those hard coded and have the option for everything else?

mackal:
> > The Ogre and Ogre Citizen were the only hardcoded ones because that's what the client had as hardcoded ones.
> 
> Should I just leave those hard coded and have the option for everything else?

Options should by default match hardcoded values in client. This is luckily one of the hardcoded bits we can find in the client and do whatever we want.

zone/attack.cpp
```diff
@@ -3832,11 +3832,63 @@ void Mob::CommonDamage(Mob* attacker, int64 &damage, const uint16 spell_id, cons
...
+				}
+
+
+				if (
+					(
+						IsPlayerRace(GetBaseRace()) &&
+						RuleI(Combat, FrontalStunImmunityRaces) & GetPlayerRaceBit(GetBaseRace())
+					) ||
+					GetBaseRace() == RACE_OGGOK_CITIZEN_93 ||
+					GetBaseRace() == RACE_RALLOS_OGRE_325 ||
```
mackal, line 3852: I do not see any evidence these extra ogre races should get this ability.

joligario (approved):
No obvious issues from my perspective

Akkadius (approved):

PR #2221, opened 2022-05-29
Enable database dump of bot data

titanium-forever:


Akkadius:
Is this actually ready to be reviewed? It looks like its still been updated frequently

titanium-forever:
Yes, ready for review. 

The scope changed slightly after I first raised the PR so there were an extra couple of commits afterwards, but I think it's ready now.

Akkadius (approved):
Looks good to me. 

Thanks for taking the initiative to add this for other bot users!

PR #2223, opened 2022-05-29
[Hot Fix] Fix Linux compile due to missing include.

Kinglykrab:
- Not sure how Windows compiles, but Linux fails.

Not sure if we need this, as AppVeyor build succeeded this time, it had just said YOU_RECEIVE and YOU_RECEIVE_AS_SPLIT were undefined before, so I assumed it was this.


nolangreene (approved):

nolangreene:
Adding a +1 here. These includes are required to build on Debian 10

joligario (approved):

PR #2231, opened 2022-06-01
[Crash] Fix Null Pointer Crash via GetZoneDescription

Akkadius:
Fix null pointer crash via GetZoneDescription when #reload commands are calls and zones aren't actually booted


Kinglykrab (approved):

Akkadius:
This wasn't ready to fully merge yet but it "should be good" having it tested currently

PR #2234, opened 2022-06-01
Fix shield calculation

Quintinon:
Correct shield calculation and incorrect variables provided in a few places related to it.


Akkadius:
Thanks for PRing! What evidence is the change based on?

Akkadius:
cc @KayenEQ for review

Quintinon:
The basis for the change was as mackal said. I noticed an unused assignment from std::min, and some unused parameters and looked at the git blame to see the original PR. I then looked up what /shield did online and found two different sources stating that the calculation maxes at 50% damage reduction for the caster if wearing a shield, starting at 25% reduction if no shield AC.

KayenEQ:
Thanks good catch!

KayenEQ (approved):

Kinglykrab (approved):

PR #2238, opened 2022-06-02
[Bug Fix] Fix two invalid data accesses in zone/client.cpp

Quintinon:
Static code analysis found two instances of read out-of-bounds of an array issues in zone/client.cpp. This PR is to correct this potential error.

### DIFF
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -6519,8 +6519,7 @@ void Client::Doppelganger(uint16 spell_id, Mob *target, const char *name_overrid
 
 	npc_type = made_npc;
 
-	int summon_count = 0;
-	summon_count = pet.count;
+	int summon_count = pet.count;
 
 	if(summon_count > MAX_SWARM_PETS)
 		summon_count = MAX_SWARM_PETS;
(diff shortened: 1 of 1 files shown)


zone/client.cpp
```diff
@@ -6541,7 +6540,7 @@ void Client::Doppelganger(uint16 spell_id, Mob *target, const char *name_overrid
 		NPC* swarm_pet_npc = new NPC(
 				(npc_dup!=nullptr)?npc_dup:npc_type,	//make sure we give the NPC the correct data pointer
 				0,
-				GetPosition() + glm::vec4(swarmPetLocations[summon_count], 0.0f, 0.0f),
+				GetPosition() + glm::vec4(swarmPetLocations[summon_count - 1], 0.0f, 0.0f),
```
Kinglykrab, line 6543: So if we're doing `>= 0` for the loop here, when we get to `0` we'd be accessing index `-1` here, right? Or am I reading this incorrectly.

Quintinon: Correct, I meant to only do one of those two changes (I was playing with doing >= or -1, it shouldn't have been both)

Kinglykrab: 👍

Kinglykrab (approved):
Looks good to me, would like some other 👀 on this though before merging. 👍

Issue #837, opened 2019-03-30
Focus effect messages sent with wrong message type.


wmrojer:
Focus effect text is sent to the client as MT_Spells instead of MT_FocusEffect. 
In method Client::GetFocusEffect in spell_effects.cpp and method Mob::DoCastSpell in Spells.cpp.
There could be more places, but that's the ones I've found.

This makes it impossible to filter out those message on the client.

Aeadoin:
I believe this can be closed, I searched for any other instances where Focus effect messages sent with wrong message type... and couldn't find any.

Focus Effects can be filtered out now also per #2447

PR #2243, opened 2022-06-04
[Messages] Convert messages from Spells to FocusEffect where necessary.

Kinglykrab:
https://github.com/EQEmu/Server/issues/837 from 2019 notes a couple places that should use Focus Effect chat type instead of Spells chat type.

### DIFF
--- a/zone/spell_effects.cpp
+++ b/zone/spell_effects.cpp
@@ -6497,7 +6497,7 @@ int64 Client::GetFocusEffect(focusType type, uint16 spell_id, Mob *caster, bool 
 			default:
 				break;
 			}
-			MessageString(Chat::Spells, string_id, UsedItem->Name);
+			MessageString(Chat::FocusEffect, string_id, UsedItem->Name);
 		}
 	}
 
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -478,7 +478,7 @@ bool Mob::DoCastSpell(uint16 spell_id, uint16 target_id, CastingSlot slot,
 	if (IsClient() && slot == CastingSlot::Item && item_slot != 0xFFFFFFFF) {
 		auto item = CastToClient()->GetInv().GetItem(item_slot);
 		if (item && item->GetItem())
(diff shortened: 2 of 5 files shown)


nolangreene (approved):

zone/spells.cpp
```diff
@@ -478,7 +478,7 @@ bool Mob::DoCastSpell(uint16 spell_id, uint16 target_id, CastingSlot slot,
 	if (IsClient() && slot == CastingSlot::Item && item_slot != 0xFFFFFFFF) {
 		auto item = CastToClient()->GetInv().GetItem(item_slot);
 		if (item && item->GetItem())
-			MessageString(Chat::Spells, BEGINS_TO_GLOW, item->GetItem()->Name);
+			MessageString(Chat::FocusEffect, BEGINS_TO_GLOW, item->GetItem()->Name);
```
hgtw, line 481: Live doesn't have this message (anymore?) when casting items so it should probably be removed or rule gated. Tested with a couple random items (Oil of Fennin Ro and Elder Spiritist Boots)

hgtw (approved):

zone/spells.cpp
Kinglykrab: Is this the only message they removed?

hgtw: Furious bash and the Client::GetFocusEffect messages still occur. I dunno what merc focus effects are but the function is also `GetFocusEffect` so it's probably ok

hgtw (approved):

zone/spell_effects.cpp
```diff
@@ -6497,7 +6497,7 @@ int64 Client::GetFocusEffect(focusType type, uint16 spell_id, Mob *caster, bool
 			default:
 				break;
 			}
-			MessageString(Chat::Spells, string_id, UsedItem->Name);
+			MessageString(Chat::FocusEffect, string_id, UsedItem->Name);
```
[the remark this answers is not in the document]
hgtw: Ignore that, it looks like emu already correctly handles focus orders. I must have been testing a focus that occured on cast start

PR #2244, opened 2022-06-05
[Memory Leak] Fix leak of CommandRecords in commandlist

mackal:
Quicker than porting to std::unique_ptr since we do fun stuff with
commandlist ...


mackal:
Wait, this is gonna double free with aliases I think ...

mackal:
The LinkedList that was removed in d7f38361f2 was doing this automatically.

Kinglykrab (approved):

PR #2246, opened 2022-06-05
[Fix] Boats should never get FixZ'd

noudess:
If a boat spawn point or waypoint had a Z value that caused the code to no longer see the boat in water, even though a great deal of the boat was below water. FixZ would kick in.  This can cause massive issues to clients viewpoints of the boat (other than the client riding the boat) - causing some people that zone in after the FixZ to see the boat and rider underwater.

This is a tiny patch to avoid the issue in general - but I'll be looking to touch up my Boat navigation in movement_manager to be more in sync with how the client deals with Z for boats.

### DIFF
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -512,6 +512,18 @@ Mob::Mob(
 	mob_close_scan_timer.Trigger();
 
 	SetCanOpenDoors(true);
+
+	is_boat = (
+        race == RACE_SHIP_72 ||
+        race == RACE_LAUNCH_73 ||
+        race == RACE_GHOST_SHIP_114 ||
+        race == RACE_SHIP_404 ||
+        race == RACE_MERCHANT_SHIP_550 ||
+        race == RACE_PIRATE_SHIP_551 ||
+        race == RACE_GHOST_SHIP_552 ||
+        race == RACE_BOAT_533
+    );
+
 }
 
 Mob::~Mob()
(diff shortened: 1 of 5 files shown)


Kinglykrab (approved):
👍

zone/waypoints.cpp
```diff
@@ -795,6 +795,10 @@ void Mob::FixZ(int32 z_find_offset /*= 5*/, bool fix_client_z /*= false*/) {
 		return;
 	}
 
+	if (IsBoat()) {
```
Akkadius, line 798: Since this is running through a significant hot path, I'd like this to reference a mob scoped private member variable boolean that is loaded on spawn and we don't have to run OR's in this check repeatedly

Otherwise I'm good with the overall premise and change of this PR

noudess: Ok, family here today.  Will make those changes when I get back to it.

noudess: Ok, is this what you had in mind @Akkadius?  Still compiling.

zone/mob.cpp
```diff
@@ -512,6 +512,18 @@ Mob::Mob(
...
+	is_boat = (
+        race == RACE_SHIP_72 ||
+        race == RACE_LAUNCH_73 ||
+        race == RACE_GHOST_SHIP_114 ||
+        race == RACE_SHIP_404 ||
+        race == RACE_MERCHANT_SHIP_550 ||
+        race == RACE_PIRATE_SHIP_551 ||
+        race == RACE_GHOST_SHIP_552 ||
+        race == RACE_BOAT_533
+    );
```
Kinglykrab, line 525: Formatting.

Akkadius: I think we still keep `IsBoat()` with the or checks, it served a purpose that's worth keeping

The getter to the bool can just be `GetIsBoat()` with `SetIsBoat(bool)`

noudess: Uncalled?  Or leave all the other spots using that instead of the variable?

noudess: Let me know if this is what you wanted.

Akkadius (approved):

PR #2253, opened 2022-06-07
[Bug Fix] Add required distance to CoTH before aggro wipe

fryguy503:
This should fix eqemu/master#2205

### DIFF
--- a/zone/spell_effects.cpp
+++ b/zone/spell_effects.cpp
@@ -2136,7 +2136,7 @@ bool Mob::SpellEffect(Mob* caster, uint16 spell_id, float partial, int level_ove
 
 				if (IsClient()) {
 					if (caster->IsClient()) {
-						if (!entity_list.IsInSameGroupOrRaidGroup(caster->CastToClient(), this->CastToClient())) {
+						if (!entity_list.IsInSameGroupOrRaidGroup(caster->CastToClient(), CastToClient())) {
 							caster->Message(Chat::Red, "Your target must be a group member for this spell.");
 							break;
 						}
--- a/zone/spell_effects.cpp
+++ b/zone/spell_effects.cpp
@@ -2130,20 +2130,30 @@ bool Mob::SpellEffect(Mob* caster, uint16 spell_id, float partial, int level_ove
 
 			case SE_SummonPC:
 			{
-				if (!caster)
+				if (!caster) {
 					break;
+				}
+
 				if (IsClient()) {
+					if (caster->IsClient()) {
+						if (!entity_list.IsInSameGroupOrRaidGroup(caster->CastToClient(), this->CastToClient())) {
+							caster->Message(Chat::Red, "Your target must be a group member for this spell.");
+							break;
+						}
+
+						// clear aggro when summoned in zone and further than aggro clear distance rule.
+						if (caster->CalculateDistance(GetX(), GetY(), GetZ()) >= RuleR(Spells, CallOfTheHeroAggroClearDist)) {
+							entity_list.ClearAggro(this);
+						}
+					}
+
 					CastToClient()->MovePC(zone->GetZoneID(), zone->GetInstanceID(), caster->GetX(),
 							       caster->GetY(), caster->GetZ(), caster->GetHeading(), 2,
 							       SummonPC);
 					Message(Chat::Yellow, "You have been summoned!");
-					// only for beneficial spells like Call of the Hero
-					// This clear probably isn't actually needed, but need to investigate more
-					if (IsBeneficialSpell(spell_id))
-						entity_list.ClearAggro(this);
-				} else
+				} else {
 					caster->Message(Chat::Red, "This spell can only be cast on players.");
-
+				}
 				break;
 			}
 
--- a/zone/spell_effects.cpp
+++ b/zone/spell_effects.cpp
@@ -2142,7 +2142,7 @@ bool Mob::SpellEffect(Mob* caster, uint16 spell_id, float partial, int level_ove
 						}
 
 						// clear aggro when summoned in zone and further than aggro clear distance rule.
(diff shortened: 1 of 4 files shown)


zone/spell_effects.cpp
```diff
@@ -2130,20 +2130,30 @@ bool Mob::SpellEffect(Mob* caster, uint16 spell_id, float partial, int level_ove
...
 			case SE_SummonPC:
 			{
-				if (!caster)
+				if (!caster) {
 					break;
+				}
+
 				if (IsClient()) {
+					if (caster->IsClient()) {
+						if (!entity_list.IsInSameGroupOrRaidGroup(caster->CastToClient(), this->CastToClient())) {
```
Kinglykrab, line 2139: `this` is implied.

Kinglykrab:
The original logic has changed a bit here, we're no longer checking `IsBeneficialSpell` for the aggro clearing and we're now requiring same group/same raid group for any `SummonPC` spell effect. Not sure intention with the logic changes, code looks fine though.

Valorith:
Any chance this can be put behind a rule? Right or wrong, this is a pretty significant change that will effect content balanced around its current state. Just a thought...

fryguy503:
> 

The wipe itself is behind a rule, can set it to 0 or 99999 to force enable/disable, no reason to multiple rules to enable then set distance. @Kinglykrab what are your thoughts? Have a bool to enable then the distance rule, or just set the default distance to 0 so it always wipes like it currently does?

Kinglykrab:
> The wipe itself is behind a rule, can set it to 0 or 99999 to force enable/disable, no reason to multiple rules to enable then set distance. @Kinglykrab what are your thoughts? Have a bool to enable then the distance rule, or just set the default distance to 0 so it always wipes like it currently does?

If the change is consistent with Live, our defaults should be Live-like and just be based upon operator discretion to change. Don’t want the default Live-like behavior to have to be enabled, would prefer you have to disable it for custom behavior like @Valorith mentioned. We could use -1 as a disabled value in the current rule so it doesn’t run the distance check if the rule isn’t >= 0.

Valorith:
> > The wipe itself is behind a rule, can set it to 0 or 99999 to force enable/disable, no reason to multiple rules to enable then set distance. @Kinglykrab what are your thoughts? Have a bool to enable then the distance rule, or just set the default distance to 0 so it always wipes like it currently does?
> 
> 
> 
> If the change is consistent with Live, our defaults should be Live-like and just be based upon operator discretion to change. Don’t want the default Live-like behavior to have to be enabled, would prefer you have to disable it for custom behavior like @Valorith mentioned. We could use -1 as a disabled value in the current rule so it doesn’t run the distance check if the rule isn’t >= 0.

Ya, that makes sense. I was just requesting to have a rule to toggle the distance check off if desired. Totally fine to have the distance check as default if that's the live like behavior. Thanks :)

Kinglykrab:
> > > The wipe itself is behind a rule, can set it to 0 or 99999 to force enable/disable, no reason to multiple rules to enable then set distance. @Kinglykrab what are your thoughts? Have a bool to enable then the distance rule, or just set the default distance to 0 so it always wipes like it currently does?
> > 
> > 
> > If the change is consistent with Live, our defaults should be Live-like and just be based upon operator discretion to change. Don’t want the default Live-like behavior to have to be enabled, would prefer you have to disable it for custom behavior like @Valorith mentioned. We could use -1 as a disabled value in the current rule so it doesn’t run the distance check if the rule isn’t >= 0.
> 
> Ya, that makes sense. I was just requesting to have a rule to toggle the distance check off if desired. Totally fine to have the distance check as default if that's the live like behavior. Thanks :)

Yeah, changed my mind after I commented as 0 makes more sense than -1. The default would be on,  just change to 0 to disable. So we’d need to change distance check to `if (rule == 0 || rule >= 1 && distance > rule)`.

fryguy503:
> > > > The wipe itself is behind a rule, can set it to 0 or 99999 to force enable/disable, no reason to multiple rules to enable then set distance. @Kinglykrab what are your thoughts? Have a bool to enable then the distance rule, or just set the default distance to 0 so it always wipes like it currently does?
> > > 
> > > 
> > > If the change is consistent with Live, our defaults should be Live-like and just be based upon operator discretion to change. Don’t want the default Live-like behavior to have to be enabled, would prefer you have to disable it for custom behavior like @Valorith mentioned. We could use -1 as a disabled value in the current rule so it doesn’t run the distance check if the rule isn’t >= 0.
> > 
> > 
> > Ya, that makes sense. I was just requesting to have a rule to toggle the distance check off if desired. Totally fine to have the distance check as default if that's the live like behavior. Thanks :)
> 
> Yeah, changed my mind after I commented as 0 makes more sense than -1. The default would be on, just change to 0 to disable. So we’d need to change distance check to `if (rule == 0 || rule >= 1 && distance > rule)`.

Yea, this was the path I was thinking for this so there is not too many rules. Adjustments incoming.

Kinglykrab (approved):

MidnightMusicClub:
Here's the Patch Note that changed how CotH aggro wipe works. 
ea6796aa1399611e7326483a0f690e09

zone/spell_effects.cpp
```diff
@@ -2130,20 +2130,30 @@ bool Mob::SpellEffect(Mob* caster, uint16 spell_id, float partial, int level_ove
...
 			{
-				if (!caster)
+				if (!caster) {
 					break;
+				}
+
 				if (IsClient()) {
+					if (caster->IsClient()) {
+						if (!entity_list.IsInSameGroupOrRaidGroup(caster->CastToClient(), CastToClient())) {
+							caster->Message(Chat::Red, "Your target must be a group member for this spell.");
```
hgtw, line 2140: Looks like live uses eqstr 196 in the spell failure color for this
```cpp
caster->MessageString(Chat::SpellFailure, TARGET_GROUP_MEMBER)
```

hgtw (approved):

PR #2254, opened 2022-06-08
[Bug Fix] Fix stack leaks in Lua events

hgtw:
EventItem was leaving the package table on the stack in successful
(non-encounter) calls and only popping the return. All events were
also leaking the full stack on lua_pcall errors. Lua mod callbacks
were popping error returns but leaking the table.


Kinglykrab (approved):

Quintinon:
Is there a different coding pattern or some generic checks that could be put in place to prevent this from happening again?

Akkadius (approved):

PR #2255, opened 2022-06-08
[Bash] Uses shoulders when using 2 hand weapon

fryguy503:
Skill Bash will now correctly use the Shoulders when using a 2 hand weapon.

### DIFF
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -441,6 +441,7 @@ RULE_REAL(Combat, ProcPerMinDexContrib, 0.075, "Increases the probability of a p
 RULE_REAL(Combat, BaseProcChance, 0.035, "Base chance for procs")
 RULE_REAL(Combat, ProcDexDivideBy, 11000, "Divisor for the probability of a proc increased by dexterity")
 RULE_INT(Combat, MinRangedAttackDist, 25, "Minimum Distance to use Ranged Attacks")
+RULE_REAL(Combat, BashACBonusDivisor, 25.0f, "this divides the AC value contribution to bash damage, lower to increase damage")
 RULE_BOOL(Combat, ArcheryBonusRequiresStationary, true, "does the 2x archery bonus chance require a stationary npc")
 RULE_REAL(Combat, ArcheryNPCMultiplier, 1.0, "Value is multiplied by the regular dmg to get the archery dmg")
 RULE_BOOL(Combat, AssistNoTargetSelf, true, "When assisting a target that does not have a target: true = target self, false = leave target as was before assist (false = live like)")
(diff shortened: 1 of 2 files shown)


common/ruletypes.h
```diff
@@ -441,6 +441,7 @@ RULE_REAL(Combat, ProcPerMinDexContrib, 0.075, "Increases the probability of a p
 RULE_REAL(Combat, BaseProcChance, 0.035, "Base chance for procs")
 RULE_REAL(Combat, ProcDexDivideBy, 11000, "Divisor for the probability of a proc increased by dexterity")
 RULE_INT(Combat, MinRangedAttackDist, 25, "Minimum Distance to use Ranged Attacks")
+RULE_REAL(Combat, BashACBonusDivisor, 25.0f, "this divides the AC value contribution to bash damage, lower to increase damage")
```
Kinglykrab, line 444: Don’t think you need the f for float here. None of the others use it.

mackal:
I do not believe this is live like, unless they changed it recently.

fryguy503:
> I do not believe this is live like, unless they changed it recently.

There are comments in 2003 that backup shoulders being used. I do not see any patch notes that say it was changed to secondary when 2 handing. I can set this as a rule if preferred.

mackal:
> > I do not believe this is live like, unless they changed it recently.
> 
> There are comments in 2003 that backup shoulders being used. I do not see any patch notes that say it was changed to secondary when 2 handing. I can set this as a rule if preferred.

dev comments or just comments? This function is in the client and can be reversed engineered.

fryguy503:
> 

100% player comments.

`If you do not hold an item in your offhand, such as when weilding a 2 handed weapon, then your SHOULDER item is the item used to consider whether your slam is magic or not (again tested with troll war and bar shm by taking off hand, arm and shoulder items one at a time while trying to slam wisps.) Characters with racial slam and a non magic shield can quickly unequip this item and magic bash using their magic shoulder item and also small warriors can quickly replace their offhand weapon for a shield if they need to bash.`

Pretty sure this is how TAKP is rolling, as I had implemented this based on that. I will add a rule to this for era based servers.

mackal:
> > 
> 
> 100% player comments.
> 
> `If you do not hold an item in your offhand, such as when weilding a 2 handed weapon, then your SHOULDER item is the item used to consider whether your slam is magic or not (again tested with troll war and bar shm by taking off hand, arm and shoulder items one at a time while trying to slam wisps.) Characters with racial slam and a non magic shield can quickly unequip this item and magic bash using their magic shoulder item and also small warriors can quickly replace their offhand weapon for a shield if they need to bash.`
> 
> Pretty sure this is how TAKP is rolling, as I had implemented this based on that. I will add a rule to this for era based servers.

TAKP should just be using a bonus if they have a shield. 2H doesn't get a bonus from what I can tell.

fryguy503:
Per feedback and client decompiles, I will close this. Thank you Mackal!

PR #2256, opened 2022-06-08
[Bug Fix] Tradeskill Item 0 Error

fryguy503:
When trying to fully script a combine so the tradeskill itself does not return an item (Item 0) and handling all items through quests, the source will provide an error regardless that "Item 0 does not exist". This change will move the item summon after the validity check thus preventing the error to the client.


Kinglykrab (approved):
Looks good to me. :+1:

PR #2258, opened 2022-06-08
[Quest API] Expand SaveGuardSpot

fryguy503:
- Adds
-- mob:saveguardspot()
-- mob:saveguardspot(true)

Pending perl implementation


Kinglykrab (approved):
:+1:

PR #2261, opened 2022-06-08
[Loot] Remove unnecessary loot error messages.

Kinglykrab:
- These messages are only sent when you loot a corpse that is mid-decay (i.e. no items and you go to loot it immediately) or if you try to loot a non-corpse entity.
- The ent != 0 shows up a lot if you're trying to loot too quickly and really isn't an error, the server just hasn't caught up to decay the corpse before you try to loot.
- Add LootItem_Struct to give more verbosity to Handle_OP_LootItem.

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -9531,8 +9531,8 @@ void Client::Handle_OP_LootItem(const EQApplicationPacket *app)
 		return;
 	}
 
-	auto* l = (LootItem_Struct*) app->pBuffer;
-	auto entity = entity_list.GetID(l->entity_id);
+	auto* l = (LootingItem_Struct*) app->pBuffer;
+	auto entity = entity_list.GetID(static_cast<uint16>(l->lootee));
 	if (!entity) {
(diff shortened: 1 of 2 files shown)


zone/client_packet.cpp
```diff
@@ -9531,26 +9531,21 @@ void Client::Handle_OP_LootItem(const EQApplicationPacket *app)
 		return;
 	}
 
-	EQApplicationPacket* outapp = nullptr;
-	Entity* entity = entity_list.GetID(*((uint16*)app->pBuffer));
-	if (entity == 0) {
-		Message(Chat::Red, "Error: OP_LootItem: Corpse not found (ent = 0)");
-		outapp = new EQApplicationPacket(OP_LootComplete, 0);
+	auto* l = (LootItem_Struct*) app->pBuffer;
```
hgtw, line 9534: Can just use the current `LootingItem_Struct` for this (and cast the `lootee` field for the entity list api)

hgtw (approved):

PR #2262, opened 2022-06-09
[Bug Fix] Fix miscellaneous memory leaks related to EQApplicationPacket and it's pBuffer

Quintinon:
Have not had a chance to even compile this let alone run it, hence the draft PR status.

### DIFF
--- a/zone/gm_commands/reload.cpp
+++ b/zone/gm_commands/reload.cpp
@@ -61,7 +61,7 @@ void command_reload(Client *c, const Seperator *sep)
 		return;
 	}
 
-	ServerPacket pack;
+	ServerPacket* pack;
 
 	if (is_aa) {
 		c->Message(Chat::White, "Attempting to reload Alternate Advancement Data globally.");	
--- a/zone/gm_commands/reload.cpp
+++ b/zone/gm_commands/reload.cpp
@@ -61,7 +61,7 @@ void command_reload(Client *c, const Seperator *sep)
 		return;
 	}
 
-	ServerPacket* pack;
+	ServerPacket* pack = nullptr;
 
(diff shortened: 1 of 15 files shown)


Quintinon:
GitHub compiled it for me, so that is passing at least. Still have not had a chance to run this on my server, however I've finished my initial pass through the code base (I looked at every instance of the word new).

I still have some notes on rabbit trails for follow up on but any findings from those will be in a separate PR.

Feedback is welcome at this point on anything. I tried to be as minimally invasive as possible on the changes since there were a lot in a single PR.

zone/gm_commands/reload.cpp
```diff
@@ -61,7 +61,7 @@ void command_reload(Client *c, const Seperator *sep)
 		return;
 	}
 
-	auto pack = new ServerPacket;
+	ServerPacket* pack;
```
hgtw, line 64: Initialize this with `= nullptr` since it might contain a non-null garbage value

Quintinon: Done

hgtw (approved):

zone/trading.cpp
```diff
@@ -1151,8 +1152,8 @@ void Client::Trader_EndTrader() {
 			}
 
 			safe_delete(outapp);
-			safe_delete(gis);
 		}
+		safe_delete(gis);
```
hgtw, line 1156: Just undo this change and push a new commit to this pr's branch to resolve the #2266 duplicate

Quintinon: Done

Akkadius:
Conflicts

Akkadius:
Handled conflicts, awaiting build

PR #2263, opened 2022-06-11
[Telnet] Add guildsay to console commands and Guild Channel to QueueMessage.

Kinglykrab:
- Will allow you to send guild-specific messages from things like Discord EQ.
- Add Guild support in `EQ::Net::ConsoleServerConnection::SendChannelMessage` so guild ID can be parsed out.
- Fix Auction messages not sending to telnet (thanks @xackery).
image
image

### DIFF
--- a/world/console.cpp
+++ b/world/console.cpp
@@ -411,6 +411,40 @@ void ConsoleGMSay(
  * @param command
  * @param args
  */
+void ConsoleGuildSay(
+	EQ::Net::ConsoleServerConnection *connection,
+	const std::string &command,
+	const std::vector<std::string> &args
+)
+{
+	if (args.size() < 1) {
+		return;
+	}
+
+	char tmpname[64];
+	tmpname[0] = '*';
+	strcpy(&tmpname[1], connection->UserName().c_str());
+
+	auto from = args[0];
+	auto guild_id = std::stoul(args[1]);
+
+	auto join_args = args;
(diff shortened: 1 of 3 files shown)


world/console.cpp
```diff
@@ -406,6 +406,36 @@ void ConsoleGMSay(
...
+	const std::string &command,
+	const std::vector<std::string> &args
+)
+{
+	if (args.size() < 1) {
+		return;
+	}
+
+	auto from = args[0];
+	auto guild_id = std::stoul(args[1]);
```
hgtw, line 425: Since `stoul` can throw you might want to just use `strtoul` here and abort if 0 to avoid crashing world with a bad input

world/console.cpp
```diff
@@ -406,6 +406,39 @@ void ConsoleGMSay(
...
+		return;
+	}
+
+	auto join_args = args;
+	join_args.erase(join_args.begin(), join_args.begin() + 2);
+
+	auto message = fmt::format(
+		"{} tells the guild, '{}'",
+		from,
+		JoinString(join_args, " ")
```
hgtw, line 436: Just a note for future use, fmt has a join method that you can also use without creating the `join_args` temporary
`fmt::join(args.begin()+2, args.end(), " ")`

hgtw (approved):

Akkadius (approved):

PR #2265, opened 2022-06-11
[Bug Fix] Correct memory leaks from SharedDatabase return values

Quintinon:
SharedDatabse::CreateItem() returns an instance that in most cases needs to be deleted after use (Most inventory functions create a Clone before storing it).
I went through all the usages of this function and added any missing deletes.


hgtw (approved):

Quintinon:
These changes are fully encompassed by PR #2266

PR #2270, opened 2022-06-15
[Cleanup] Cleanup code smells and compiler warnings in common/shareddb

Quintinon:
Some warnings were from VS2022 in windows, and some were from Resharper.

### DIFF
--- a/common/shareddb.cpp
+++ b/common/shareddb.cpp
@@ -929,7 +929,7 @@ bool SharedDatabase::LoadItems(const std::string &prefix) {
 		std::string file_name = Config->SharedMemDir + prefix + std::string("items");
 		LogInfo("[Shared Memory] Attempting to load file [{}]", file_name);
 		items_mmf = std::make_unique<EQ::MemoryMappedFile>(file_name);
-		items_hash = std::make_unique<EQ::FixedMemoryHashSet<EQ::ItemData>>(reinterpret_cast<uint8*>(items_mmf->Get()), items_mmf->Size());
+		items_hash = std::make_unique<EQ::FixedMemoryHashSet<EQ::ItemData>>(static_cast<uint8*>(items_mmf->Get()), items_mmf->Size());
 		mutex.Unlock();
 	} catch(std::exception& ex) {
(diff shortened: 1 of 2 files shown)


Kinglykrab (approved):

Quintinon:
I've run these changes on my solo server for a few hours of play time now while I was grinding some levels out. I haven't noticed any issues.

common/shareddb.cpp
```diff
@@ -774,27 +769,26 @@ bool SharedDatabase::GetInventory(uint32 account_id, char *name, EQ::InventoryPr
...
-
-		bool instnodrop = (row[10] && (uint16)atoi(row[10])) ? true : false;
-		uint32 ornament_icon = std::stoul(row[12]);
-		uint32 ornament_idfile = std::stoul(row[13]);
+		aug[0] = static_cast<uint32>(atoi(row[4]));
+		aug[1] = static_cast<uint32>(atoi(row[5]));
+		aug[2] = static_cast<uint32>(atoi(row[6]));
+		aug[3] = static_cast<uint32>(atoi(row[7]));
+		aug[4] = static_cast<uint32>(atoi(row[8]));
+		aug[5] = static_cast<uint32>(atoi(row[9]));
```
Kinglykrab, line 784: With casts like this, should we just use std::stoul? Versus casting to int then back to unsigned.

Quintinon: I had thought about that, but functionality would technically change since stoul will error if input is invalid where atoi will not and will return 0 instead.

Kinglykrab: Could use `atoul` at that point if you want the default 0. Just doesn't make sense to pull as signed then cast to unsigned.

Quintinon: atoul sounds like a good solution to me, doesn't change functionality but fixes the code smells. I'll make the change when I have time at home.

Kinglykrab: Sounds good. There were more places doing this than just right here I believe.

Akkadius (approved):

PR #2272, opened 2022-06-19
[Quest API] Add TrackNPC to Perl/Lua.

Kinglykrab:
- Add quest::tracknpc(entity_id) to Perl.
- Add eq.track_npc(entity_id) to Lua.
- This will allow server operators to arbitrarily Track NPCs for clients with scripts.
- Modified tracking to auto turn off and tell you that you discovered your target if within 10 units, this is to allow scripted Tracking to turn off without the need for the Tracking skill.
- @fryguy503 mentioned a Live quest that uses this methodology to track an NPC on a Warrior, a class that doesn't have Tracking.


fryguy503 (approved):
LGTM Testing worked well

PR #2276, opened 2022-06-29
[Feature] Add humanoid and non-wielded restrictions to pick pocket

noudess:
The pick pocket skill should only work on certain body types.

The pick pocket skill should not allow picking items that are wielded.

Verified on live (though some mobs are marked as humanoid on live that are questionable - this will be a db issue).

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -11059,14 +11059,6 @@ void Client::Handle_OP_PickPocket(const EQApplicationPacket *app)
 		return;
 
 	p_timers.Start(pTimerBeggingPickPocket, 8);
-	auto outapp = new EQApplicationPacket(OP_PickPocket, sizeof(sPickPocket_Struct));
-	sPickPocket_Struct* pick_out = (sPickPocket_Struct*)outapp->pBuffer;
-	pick_out->coin = 0;
-	pick_out->from = victim->GetID();
-	pick_out->to = GetID();
-	pick_out->myskill = GetSkill(EQ::skills::SkillPickPockets);
-	pick_out->type = 0;
-	//if we do not send this packet the client will lock up and require the player to relog.
 
 	if (victim == this) {
 		Message(Chat::White, "You catch yourself red-handed.");
@@ -11085,14 +11077,12 @@ void Client::Handle_OP_PickPocket(const EQApplicationPacket *app)
 		auto body = victim->GetBodyType();
 		if (body == BT_Humanoid || body == BT_Monster || body == BT_Giant ||
 			body == BT_Lycanthrope) {
-			safe_delete(outapp);
 			victim->CastToNPC()->PickPocket(this);
 			return;
 		}
 	}
 
-	QueuePacket(outapp);
-	safe_delete(outapp);
+	SendPickPocketResponse(victim, 0, PickPocketFailed);
 }
 
 void Client::Handle_OP_PopupResponse(const EQApplicationPacket *app)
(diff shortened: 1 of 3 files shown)


zone/client_packet.cpp
```diff
@@ -11074,18 +11074,22 @@ void Client::Handle_OP_PickPocket(const EQApplicationPacket *app)
...
 	}
-	else if (victim->IsNPC()) {
+	else if (victim->IsNPC() && victim->GetBodyType() == BT_Humanoid) {
 		safe_delete(outapp);
 		victim->CastToNPC()->PickPocket(this);
 		return;
 	}
 	else {
-		Message(0, "Stealing from clients not yet supported.");
+		Message(0, "Your attempt at stealing was unsuccessful.");
```
fryguy503, line 11090: Could be a Client String:
12898 Your attempt at stealing was unsuccessful.

zone/client_packet.cpp
```diff
@@ -11074,18 +11074,22 @@ void Client::Handle_OP_PickPocket(const EQApplicationPacket *app)
 	else if (victim->GetOwnerID()) {
 		Message(0, "You cannot steal from pets!");
 	}
+	else if (victim->IsClient()) {
+		Message(0, "Stealing from clients not yet supported.");
```
fryguy503, line 11078: Message Color can be changed from magic number

zone/client_packet.cpp
```diff
@@ -11069,23 +11069,27 @@ void Client::Handle_OP_PickPocket(const EQApplicationPacket *app)
 	//if we do not send this packet the client will lock up and require the player to relog.
 
 	if (victim == this) {
-		Message(0, "You catch yourself red-handed.");
+		Message(Chat::White, "You catch yourself red-handed.");
 	}
 	else if (victim->GetOwnerID()) {
-		Message(0, "You cannot steal from pets!");
+		Message(Chat::White, "You cannot steal from pets!");
```
fryguy503, line 11075: I see these available DB Strings. I assume if its not a valid target, ID 12405 should be used. This can be cleaned up later, don't need to hold this PR up for that.


12405 You must target a player to steal from first.
12406 You must target a player to steal from first.  You may not steal from corpses.
12407 You can't try to steal while STUNNED!
12408 You can't try to steal while attacking.
12409 You can't seem to steal from yourself for some reason...

fryguy503 (approved):
LGTM with comments

zone/client_packet.cpp
noudess: 12405 sounds like a client to client PP?

mackal: It's likely just default failed PP message.

noudess:
Gonna look at this more tomorrow.  I see the client is handing some rejections on it's own without sending a packet.  I may not need the check for bodytype.

mackal:
> Gonna look at this more tomorrow. I see the client is handing some rejections on it's own without sending a packet. I may not need the check for bodytype.

it makes sense to check everything the client does just in case someone tries to send a forged packet when the client would otherwise block it.

noudess:
My extra rejection message was useless.  The existing code either handles everything.  I moved the Client check up in the sequence to avoid some unneeded checks, and ended up tossing in a catch all at the end - but that was an error.  The existing reply packet, or the result of a call to Pick() handles it all.

noudess:
Ok, verified my suspicions.  There is already a function that handles the reply packet for PickPocket.  Removed the redundant code to build the packet and used the existing function.  Reduces Handle_OP_PickPocket to a simpler form.

noudess:
As a side note, several of the MT_XXXX constants in the code are diff from the eqemu documentation page.

zone/client_packet.cpp
noudess: Gonna just stick with what I have now, if no objections.

PR #2277, opened 2022-06-29
[Bug Fix] NPCs were getting weapon proc added twice

noudess:
NPCs and pets that wield a weapon with a proc were getting the proc added twice to the in memory tables.  

This also caused pets to sometimes proc weapons that should not proc for their level, while sometimes yielding the "not able to command this weapon" other times.

Tested thoroughly with the fix.

### DIFF
--- a/zone/loottables.cpp
+++ b/zone/loottables.cpp
@@ -473,8 +473,6 @@ void NPC::AddLootDrop(
 		}
 
 		if (foundslot == EQ::invslot::slotPrimary) {
-			if (item2->Proc.Effect != 0)
-				CastToMob()->AddProcToWeapon(item2->Proc.Effect, true);
 
 			eslot = EQ::textures::weaponPrimary;
 			if (item2->Damage > 0) {
@@ -489,8 +487,6 @@ void NPC::AddLootDrop(
 			&& (GetOwner() != nullptr || (CanThisClassDualWield() && zone->random.Roll(NPC_DW_CHANCE)) || (item2->Damage==0)) &&
 			(item2->IsType1HWeapon() || item2->ItemType == EQ::item::ItemTypeShield || item2->ItemType ==  EQ::item::ItemTypeLight))
 		{
-			if (item2->Proc.Effect!=0)
-				CastToMob()->AddProcToWeapon(item2->Proc.Effect, true);
 
 			eslot = EQ::textures::weaponSecondary;
 			if (item2->Damage > 0)


Kinglykrab (approved):

Akkadius:
Kind of baffled this has been a thing for how long?

Akkadius:
Also where else is it getting added?

noudess:
> Kind of baffled this has been a thing for how long?

I've seen pets proc weapons they should not be able to for a very long time (years?).  I never took the time to look into it until recently.

noudess:
> Also where else is it getting added?

I don't recall 100%.  I logged all the places where procs were happening and I found (2) copies of the proc. One I think was labelled perma and the other one not.  Then I tracked down the lines I deleted.

I can say that 100% pre-change my pet was doing the proc two different ways.  One would allow him to proc above his level which is the one removed.

I have 2+ weeks of testing and mobs are procing their weapons just fine, and pets only proc correctly now.  I tested after a trade to a pet, an NPC having a proc weapon in their loottable, and a char logging back on with a pet already having the weapon.  All work fine now.


If you need me to do the research again to refresh my memory I can, should only be 2-3 hours to track it down again.

But testing does look very good with those lines gone.

Akkadius:
I believe you on your testing, just a bit baffled that this is a thing for so long and want to make sure we don't cram this one through. Would like to have a few other folks test / verify even though this one looks straight forward

noudess:
> I believe you on your testing, just a bit baffled that this is a thing for so long and want to make sure we don't cram this one through. Would like to have a few other folks test / verify even though this one looks straight forward

Sounds good.  @Kinglykrab  or @TurmoilToad maybe?

Kinglykrab:
> > I believe you on your testing, just a bit baffled that this is a thing for so long and want to make sure we don't cram this one through. Would like to have a few other folks test / verify even though this one looks straight forward
> 
> Sounds good. @Kinglykrab or @TurmoilToad maybe?

I could take a look, what were you using to easily test that? Some sort of debug message, or what?

noudess:
I traded a pet anything with a proc - fought a bit to show it still procs correctly.  In current code, if you trade a procing weapon too high, it still procs about half the time, the other half you get the "too high to use" message.

For NPCs I hunted paw and waited for one to have the gnoll hide lariat.   Or traded one to a mob as GM.

Kinglykrab:
Definitely seeing the issue, will test with your change and see if it resolves it.
Pet is level 60, using a weapon with a level 70 proc requirement.
Pet
image
Item
image
Messages
image

Kinglykrab:
Fix seems to work, the only spell I get now is the one in the NPC's spell list, no more weapon procs. Insane this has existed for so long.
image

Akkadius (approved):

PR #2289, opened 2022-07-06
[DoT Messages] Add DoT messages for mob->PC casts, fixed others to use correct strings

noudess:
DoT's cast by mobs on PCs did not produce the messages.  Fixed this, and fixed duel situations as well.

Needed to extend FilteredMessageCloseString to be able to ignore one client in addition to the sender.

Fixed DoT to use the injured party as center for the CloseClients part instead of the caster as far as broadcasting.

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -4121,19 +4121,29 @@ void Mob::CommonDamage(Mob* attacker, int64 &damage, const uint16 spell_id, cons
 	else {
 		//else, it is a buff tic...
 		// So we can see our dot dmg like live shows it.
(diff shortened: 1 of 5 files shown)


noudess:
Additional info.  

Dot Messages on live:

NPC cast on me: [Thu Apr 02 09:05:52 2015] You have taken 1 damage from Feeble Poison
NPC cast on other: [Mon Apr 27 11:01:14 2015] A gelatinous cube has taken 11 damage from Humbertnec by Poison Bolt.
NPC from my dot: [Sun Dec 08 09:23:00 2019] A young kodiak has taken 10 damage from your Poison Bolt

These are what we see now with my patch.  An example of the new NPC cast on me in a screenshot below:

EQ000177

Kinglykrab (approved):
Looks good to me. 👍

Issue #2280, opened 2022-07-03
Zone table expansion and min_expansion


xackery:
For the zone table:
![image](https://user-images.githubusercontent.com/845670/177054178-75c94bf9-2859-4c0c-a5dd-16a924b78d73.png)
These two fields are honored and use two different methods to do zone checks:
https://github.com/EQEmu/Server/blob/master/zone/zoning.cpp#L361

Do we want to maybe merge them? Are we planning to maintain both in future? Just wanted to discuss and see if we can maybe simplify the process

Kinglykrab:
Looks to be resolved by #2297?

xackery:
Yep.

PR #2297, opened 2022-07-09
[Zone] Deprecate Zone `expansion` Field

Akkadius:
We historically had a cosmetic `expansion` field in the `zone` table used to show what expansion zones were associated with. This had been used in some tools like the PEQ editor and was used very early in the source for expansion zoning validation before we had introduced content filter fields `min_expansion` `max_expansion` `content_flags` `content_flags_disabled` at the `zone` level.

This PR deprecates `expansion` and sets all values of `expansion` to `min_expansion` and drops the `expansion` field to reduce confusion and duplication.

Tested cross-expansion zoning after changes.

image

Related issue https://github.com/EQEmu/Server/issues/2280

### DIFF
--- a/common/repositories/base/base_zone_repository.h
+++ b/common/repositories/base/base_zone_repository.h
@@ -80,7 +80,6 @@ public:
 		int         insttype;
 		int64       shutdowndelay;
 		int         peqzone;
-		int         expansion;
 		int         suspendbuffs;
 		int         rain_chance1;
 		int         rain_chance2;
@@ -184,7 +183,6 @@ public:
 			"insttype",
 			"shutdowndelay",
 			"peqzone",
-			"expansion",
 			"suspendbuffs",
 			"rain_chance1",
 			"rain_chance2",
@@ -284,7 +282,6 @@ public:
 			"insttype",
 			"shutdowndelay",
 			"peqzone",
-			"expansion",
 			"suspendbuffs",
 			"rain_chance1",
 			"rain_chance2",
@@ -418,7 +415,6 @@ public:
 		entry.insttype                  = 0;
 		entry.shutdowndelay             = 5000;
 		entry.peqzone                   = 1;
-		entry.expansion                 = 0;
 		entry.suspendbuffs              = 0;
 		entry.rain_chance1              = 0;
 		entry.rain_chance2              = 0;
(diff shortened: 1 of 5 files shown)


xackery:
There's also a 99 check on expansion.. I .. don't think that'll have an issue for locked out zones, but just food for thought.

I think each check is going to be ok with min_expansion -1 when we do conversion.

Akkadius:
> There's also a 99 check on expansion.. I .. don't think that'll have an issue for locked out zones, but just food for thought.
> 
> I think each check is going to be ok with min_expansion -1 when we do conversion.

Is a good call out.

These zones have expansion 99

```
 select zoneidnumber, short_name, long_name, min_expansion from zone where min_expansion > 30; 
+--------------+------------+---------------------+---------------+
| zoneidnumber | short_name | long_name           | min_expansion |
+--------------+------------+---------------------+---------------+
|          999 | apprentice | Designer Apprentice |            99 |
|          180 | arena2     | The Arena Two       |            99 |
|          996 | arttest    | Art Testing Domain  |            99 |
|           53 | aviak      | Aviak Village       |            99 |
|          190 | clz        | Loading             |            99 |
|          130 | erudsxing2 | Marauders Mire      |            99 |
|          184 | load       | Loading Zone        |            99 |
|          185 | load2      | New Loading Zone    |            99 |
|           28 | nektropos  | Nektropos           |            99 |
|           43 | neriakd    | Neriak Palace       |            99 |
|          183 | tutorial   | EverQuest Tutorial  |            99 |
+--------------+------------+---------------------+---------------+
11 rows in set (0.00 sec)
```

Our filter logic

```cpp
		int current_expansion_filter_criteria = content_service.GetCurrentExpansion();
		if (current_expansion_filter_criteria == Expansion::EXPANSION_ALL) {
			current_expansion_filter_criteria = Expansion::EXPANSION_FILTER_MAX;
		}

		criteria += fmt::format(
			" AND ({}min_expansion <= {} OR {}min_expansion = -1)",
			table_prefix,
			current_expansion_filter_criteria,
			table_prefix
		);
```

These still get loaded even when expansions set to `-1` with `min_expansion` <= 99

hgtw (approved):

mackal (approved):

PR #2300, opened 2022-07-10
[Tasks] Place task item rewards in free slots

hgtw:
Live places item rewards into the first free inventory slot and only
places it on the cursor if inventory is full

This was more of a pet peeve than a critical fix, but it's parity with live

### DIFF
--- a/zone/task_client_state.cpp
+++ b/zone/task_client_state.cpp
@@ -1291,7 +1291,8 @@ void ClientTaskState::RewardTask(Client *client, TaskInformation *task_informati
 	switch (task_information->reward_method) {
 		case METHODSINGLEID: {
 			if (task_information->reward_id) {
-				client->SummonItem(task_information->reward_id);
+				int16_t slot = client->GetInv().FindFreeSlot(true, true);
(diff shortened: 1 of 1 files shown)


Akkadius (approved):

mackal:
Tasks should actually be using the Client::QuestReward stuff.

hgtw:
Live doesn't use the QuestReward packet for task rewards. It does use the separate CashReward packet for money rewards though (I don't think emu has this implemented right now)

mackal:
> Live doesn't use the QuestReward packet for task rewards. It does use the separate CashReward packet for money rewards though (I don't think emu has this implemented right now)

Did they change it on live? I know live has 2 different reward systems hooked up into the task system and completing a task always did the quest ding (which is why we have the SendSound call at the end, which is just QuestReward ...)

hgtw:
Possibly, doing a DoN solo progression quest and a shared task mission (old style rewards) both didn't play a sound and didn't use the quest reward packet though.

mackal:
Talking in private with @hgtw it does appear they sometimes do use QuestRewards and sometimes don't. Thanks SOE :P 

They also use OP_CashRewards for cash rewards.

PR #2304, opened 2022-07-10
[Validation] Add Size Validation to #hotfix.

Kinglykrab:
- Validates size of shared memory pool versus current count of database so people don't accidentally #hotfix and mess something up.
**Items**
image
**Spells**
image

### DIFF
--- a/zone/command.cpp
+++ b/zone/command.cpp
@@ -766,7 +766,7 @@ void command_hotfix(Client *c, const Seperator *sep)
 			).c_str()
 		);
 
-		c->Message(Chat::Yellow, "You can avoid this message by creating placeholder items so you can reload them dynamically.");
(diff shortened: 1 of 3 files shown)


Akkadius:
Very nice

zone/command.cpp
```diff
@@ -752,6 +752,46 @@ void command_findaliases(Client *c, const Seperator *sep)
...
+		c->Message(
+			Chat::Yellow,
+			fmt::format(
+				"Database Count: {} Shared Memory Count: {}",
+				items_count,
+				shared_items_count
+			).c_str()
+		);
+
+		c->Message(Chat::Yellow, "You can avoid this message by creating placeholder items so you can reload them dynamically.");
```
Akkadius, line 769: Maybe something like

"If you want to be able to add new items to your server while it is online, you need to create placeholder entries in the database ahead of time and do not add or remove rows / entries. Only modify the existing placeholders rows / entries to safely use #hotfix"

Kinglykrab: 👍

Akkadius (approved):

mackal (approved):

PR #2305, opened 2022-07-10
[Strings] Refactor Strings Usage

Akkadius:
Historically, in `C` manipulating strings have not the most user friendly. Over time CPP standards made things easier by introducing `std::string` and gradually we've sprinkled in string utilities into our own code-base (string_util.h) with many of miscellaneous global functions. 

This PR

* Consolidates string functions (Mainly those that interact with std::string) into a `Strings` class.
* `string_util.cpp/h` was changed to `strings.cpp/h`
* Renames any `snake_case` methods to `PascalCase`
* String methods that had `String` in the name redact the word `string` for obvious reasons as the `Strings` class clearly describes its type before the method
* Legacy C global functions were moved to `strings_legacy.cpp` (included by `strings.cpp`) 
* Miscellaneous string utility functions that should probably be refactored into a different class because they're not generic enough to fit the `Strings` class were moved to `strings_misc.cpp` for now (included by `strings.cpp`)
* Added `Saylink::Create` (15 characters) to simplify / alias `EQ::SayLinkEngine::GenerateQuestSaylink` (39 characters)
* Improves readability of all references to strings related code by making it very clear functions are (`Strings`) functions that manipulate strings

**Strings Class**

Below is the new strings class with all `std::string` methods already converted in the utility class that were generic. C string functions were left alone

```cpp
class Strings {
public:
	static bool Contains(std::vector<std::string> container, std::string element);
	static bool IsNumber(const std::string &s);
	static const std::string ToLower(std::string s);
	static const std::string ToUpper(std::string s);
	static const std::string UcFirst(std::string s);
	static std::string &LTrim(std::string &str, const std::string &chars = "\t\n\v\f\r ");
	static std::string &RTrim(std::string &str, const std::string &chars = "\t\n\v\f\r ");
	static std::string &Trim(std::string &str, const std::string &chars = "\t\n\v\f\r ");
	static std::string Commify(const std::string &number);
	static std::string ConvertToDigit(int n, std::string suffix);
	static std::string Escape(const std::string &s);
	static std::string GetBetween(const std::string &s, std::string start_delim, std::string stop_delim);
	static std::string Implode(std::string glue, std::vector<std::string> src);
	static std::string Join(const std::vector<std::string> &ar, const std::string &delim);
	static std::string MillisecondsToTime(int duration);
	static std::string Money(uint32 platinum, uint32 gold = 0, uint32 silver = 0, uint32 copper = 0);
	static std::string NumberToWords(unsigned long long int n);
	static std::string Repeat(std::string s, int n);
	static std::string Replace(std::string subject, const std::string &search, const std::string &replace);
	static std::string SecondsToTime(int duration, bool is_milliseconds = false);
	static std::string::size_type SearchDelim(const std::string &haystack, const std::string &needle, const char deliminator = ',');
	static std::vector<std::string> Split(const std::string &s, const char delim = ',');
	static std::vector<std::string> Split(std::string s, std::string delimiter);
	static std::vector<std::string> Wrap(std::vector<std::string> &src, std::string character);
	static void FindReplace(std::string &string_subject, const std::string &search_string, const std::string &replace_string);
	template<typename T> static std::string ImplodePair(const std::string &glue,const std::pair<char, char> &encapsulation,const std::vector<T> &src);
};
```


zone/embparser_api.cpp
```diff
@@ -3744,7 +3744,7 @@ void perl_register_quest()
 	package.add("clear_zone_flag", &Perl__clear_zone_flag);
 	package.add("clearspawntimers", &Perl__clearspawntimers);
 	package.add("collectitems", &Perl__collectitems);
-	package.add("commify", &Perl__commify);
+	package.add("Commify", &Perl__commify);
```
hgtw, line 3747: This is case sensitive so need to undo this change to prevent api breakage

zone/lua_general.cpp
```diff
@@ -3856,7 +3856,7 @@ luabind::scope lua_register_general() {
 		luabind::def("get_body_type_name", &lua_get_body_type_name),
 		luabind::def("get_consider_level_name", &lua_get_consider_level_name),
 		luabind::def("get_environmental_damage_name", &lua_get_environmental_damage_name),
-		luabind::def("commify", &lua_commify),
+		luabind::def("Commify", &lua_commify),
```
hgtw, line 3859: same here

zone/embparser_api.cpp
Akkadius: Good eye, I had to revert a handful of these already as the IDE mucked up one of my refactors

hgtw (approved):

PR #2308, opened 2022-07-12
[CPP] Update C++ standard to C++17

mackal:


Akkadius:
Was this tested with MySQL being touched?

Akkadius (changes requested):
Was this tested with MySQL being touched?

mackal:
> Was this tested with MySQL being touched?

I ran around in game, did combat etc

Akkadius (approved):
Did some testing of my own and tested code that touches the async scheduler logic

hgtw (approved):

PR #2309, opened 2022-07-13
[CPP] Update submodules

mackal:
fmtlib has some breaking changes, do later


Akkadius (changes requested):
Was this tested thoroughly in game?

Akkadius:
I'm hesitant of just version bumping all of these without testing their related surface areas more closely. It seems like a simple change to bump but for some its been a while and we can definitely be burned

Kinglykrab:
Is this something we’re going to move forward with, or is it something we should wait until the future for? Can close and reopen in the future maybe?

mackal:
I think it would be best to just update everything and keep on track better making sure we don't go years out of date.

Akkadius:
> I think it would be best to just update everything and keep on track better making sure we don't go years out of date.

If you're willing to be ready to deal with unexpected aftermath, we can get it staged on PEQ

PR #2311, opened 2022-07-15
[Expansions] Correct Zone Expansion Gating Logic & Data

Akkadius:
This PR corrects zone expansion gating logic since it used the "off by 1" logic of the cosmetic previous `expansion` field.

This corrects expansions to be inline with https://docs.eqemu.io/server/operation/expansion-list/

This also runs a one time SQL migration to adjust expansion values

Iteration on this will be likely as we further v1 / v2 zone transitioning and expansion project furthers.


xackery (approved):
Looks good

mackal (approved):

Akkadius:
Per discussion in #server-support we are going to actually revert for now until we can more thoroughly sit down and make sure cases are covered

PR #2314, opened 2022-07-16
This mem leak was missed due to merge issues in the previous PRs.

Quintinon:
https://github.com/EQEmu/Server/commit/b5c4357de2e14a1c97abbab26c60eee8ff308331#r77584958

### DIFF
--- a/zone/trading.cpp
+++ b/zone/trading.cpp
@@ -1152,8 +1152,8 @@ void Client::Trader_EndTrader() {
 			}
 
 			safe_delete(outapp);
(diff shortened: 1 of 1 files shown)


mackal (approved):

mackal:
I don't think this can actually ever leak. It should take that branch as long as it's not null and if the new returns a null, we're gonna crash :P

Quintinon:
It's really only if Consumer is null that it matters to have it outside. It would probably never leak, but this guarantees it.

PR #2317, opened 2022-07-16
[Tasks] Add Task Reward Points Field

hgtw:
This replaces the separate DoN crystal reward fields with points and
point_type fields. This will make it easier to import data from
packet/client captures and possibly better support any future clients
or tasks that don't reward through the newer reward window.

### DIFF
--- a/common/tasks.h
+++ b/common/tasks.h
@@ -50,6 +50,20 @@ enum class TaskTimerType
 	Request
 };
 
+// live alt currency types, may not match current server alternate currency ids
+enum class AltCurrencyType
+{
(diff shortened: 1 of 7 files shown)


mackal (approved):

common/tasks.h
```diff
@@ -50,6 +50,20 @@ enum class TaskTimerType
 	Request
 };
 
+// live alt currency types, may not match current server alternate currency ids
```
Akkadius, line 53: I believe our IDs matched whatever was in the DB strings file

Akkadius (approved):

PR #2318, opened 2022-07-17
[Client] Fix IsMoving for Client

Akkadius:
Consolidates "moving" tracker member variables and fixes `IsMoving` Mob export for Client instead of just NPC

**Testing**

image


mackal:
What's the point of these 2 different variables?

Akkadius:
> What's the point of these 2 different variables?

They should probably be consolidated as they serve the same purpose, just different class member vars

mackal (approved):

PR #2319, opened 2022-07-18
[Rules] Add Keep Level on Death

trentdm:
I still want to do some testing around this, but am putting this out as a draft PR to get feedback on whether this is something that would be desired or not.

### DIFF
--- a/utils/sql/db_update_manifest.txt
+++ b/utils/sql/db_update_manifest.txt
@@ -443,6 +443,7 @@
 9187|2022_07_09_task_zone_version_matching.sql|SHOW COLUMNS FROM `task_activities` LIKE 'zone_version'|empty|
 9188|2022_07_14_zone_expansion_revert.sql|SHOW COLUMNS FROM `zone` LIKE 'expansion'|empty|
 9189|2022_07_10_character_task_rewarded.sql|SHOW COLUMNS FROM `character_tasks` LIKE 'was_rewarded'|empty|
+9190|2022_07_18_rule_death_keep_level.sql|SELECT * FROM `rule_values` WHERE `rule_name` LIKE '%Character:DeathKeepLevel%'|empty|
 
 # Upgrade conditions:
 # 	This won't be needed after this system is implemented, but it is used database that are not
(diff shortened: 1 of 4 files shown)


utils/sql/db_update_manifest.txt
```diff
@@ -443,6 +443,7 @@
 9187|2022_07_09_task_zone_version_matching.sql|SHOW COLUMNS FROM `task_activities` LIKE 'zone_version'|empty|
 9188|2022_07_14_zone_expansion_revert.sql|SHOW COLUMNS FROM `zone` LIKE 'expansion'|empty|
 9189|2022_07_10_character_task_rewarded.sql|SHOW COLUMNS FROM `character_tasks` LIKE 'was_rewarded'|empty|
+9190|2022_07_18_rule_death_keep_level.sql|SELECT * FROM `rule_values` WHERE `rule_name` LIKE '%Character:DeathKeepLevel%'|empty|
```
mackal, line 446: Rules are automatically inserted, this is unneeded, should be removed

mackal (changes requested):

nytmyr:
Way I look at it is, someone somewhere will use it and anything custom hid behind a rule just adds to the simple customization of a server.

Akkadius:
If you are ready for review, this is still in draft

utils/sql/db_update_manifest.txt
trentdm: @mackal , I also removed the migration file, https://github.com/EQEmu/Server/pull/2319/commits/de93b545fab621623aa3d97aba7f98481404ecd9, but should I have kept that in?

I'm trying to test locally but not seeing the rule added in the database, but very well am doing something else wrong in getting this set up.

trentdm:
> If you are ready for review, this is still in draft

Sorry for the long delay, I've been meaning to test locally first and haven't had time. I'm planning to work on it this weekend.

zone/attack.cpp
```diff
@@ -1843,6 +1843,15 @@ bool Client::Death(Mob* killerMob, int64 damage, uint16 spell, EQ::skills::Skill
 		exploss *= zone->level_exp_mod[GetLevel()].ExpMod;
 	}
 
+	if (exploss > 0 && RuleB(Character, DeathKeepLevel)) {
+		int32 totalExp = GetEXP();
```
Akkadius, line 1847: `snake_case` eg `total_exp` for locally scoped variables

Same for below variables

Akkadius (approved):

PR #2324, opened 2022-07-25
[Roambox] Improve Path Finding

noudess:
While debugging roam boxes under water, I found the log messages were not populating the data fields.  Fixed this.

I also found that the calls to FindPath were very often not finding tiles from start/end when mob started in a water position, causing them to never move and just spew logging errors.

Looking into this, the calls in the mob_ai roambox call to FindPath() are really unneeded and do not take into account terrain.  Simply allowing the Movement Manager code to handle this fixed all my issues, and movement manager has the stuck behavior code as well, which was being bypassed by this pre-emptive check in the roambox code.

### DIFF
--- a/zone/mob_ai.cpp
+++ b/zone/mob_ai.cpp
@@ -1600,9 +1600,8 @@ void NPC::AI_DoMovement() {
 				move_delay_max
 			);
 
-			Log(
-				Logs::Detail,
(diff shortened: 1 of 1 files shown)


noudess:
As a followup, I've written some instrumentation to log how many roam issues this fixes, and to log any results that produce a later problem, if any.  I'll post results here.  So far no known issues.

Akkadius:
Think that code was more relevant prior to movement manager. I’ll look closer this weekend.

Slightly unrelated but roamboxes should probably be migrated from group level to entry level

Akkadius (approved):

PR #2327, opened 2022-07-26
Allow scripts to prevent door click

hgtw:
Returning non-zero from EVENT_CLICK_DOOR will prevent the default
handler.

This should have been implemented when expeditions were put in to
allow scripts to prevent zoning out of vxed and tipt without flags.
Some upcoming changes may need this to allow scripts to prevent
automatic dz entry.


Kinglykrab:
Very nice, extremely useful.

Kinglykrab (approved):

mackal (approved):

Akkadius (approved):

PR #2330, opened 2022-07-29
[Rules] Add adjustment for zone forage.

fryguy503:
- Added rule to allow the adjustment of zone foraging %

### DIFF
--- a/zone/forage.cpp
+++ b/zone/forage.cpp
@@ -443,7 +443,7 @@ void Client::ForageItem(bool guarantee) {
 		}
 
 		//not an else in case theres no DB food
-		if(foragedfood == 0) {
+		if (foragedfood == 0) {
 			uint8 index = 0;
 			index = zone->random.Int(0, MAX_COMMON_FOOD_IDS-1);
 			foragedfood = common_food_ids[index];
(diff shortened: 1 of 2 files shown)


zone/forage.cpp
```diff
@@ -457,37 +456,38 @@ void Client::ForageItem(bool guarantee) {
 			return;
 		}
 
-		if(foragedfood == 13106)
+		if(foragedfood == 13106) {
```
Kinglykrab, line 459: Can we make this a constant?

Kinglykrab:
`if (` over `if(` maybe as well/

Kinglykrab (approved):
Looks good, minor comments.

zone/forage.cpp
Kinglykrab: `FOOD_FISHING_GRUBS` or something.

mackal:
Our forage and fishing tables are so fucking wrong.

https://github.com/EQEmu/Server/blob/4002d6a0835f55f470981b17cc2e62aa1ae05976/common/patches/rof2_structs.h#L596 there is this data in that struct ...

mackal (approved):

Akkadius (approved):

PR #2339, opened 2022-07-31
[Shared Tasks] Implement Activity Locking

hgtw:
This adds the `lock_activity_id` field to the tasks table which will
automatically lock a shared task when that element becomes active.

A method was added to world analogous to zone's UnlockActivities to
determine when an activity is active with respect to task steps.

Also adds quest apis to manually lock or unlock a client's shared task

### DIFF
--- a/world/shared_task_manager.cpp
+++ b/world/shared_task_manager.cpp
@@ -1824,6 +1824,7 @@ bool SharedTaskManager::HandleCompletedActivities(SharedTask* s)
 	std::array<bool, MAXACTIVITIESPERTASK> completed_steps;
 	completed_steps.fill(true);
 
+	// multiple activity ids may share a step, sort so previous step completions can be checked
 	auto activity_states = s->GetActivityState();
 	std::sort(activity_states.begin(), activity_states.end(),
 		[](const auto& lhs, const auto& rhs) { return lhs.step < rhs.step; });
--- a/world/shared_task_manager.cpp
+++ b/world/shared_task_manager.cpp
@@ -14,6 +14,7 @@
 #include "../common/repositories/completed_shared_task_members_repository.h"
 #include "../common/repositories/completed_shared_task_activity_state_repository.h"
 #include "../common/repositories/shared_task_dynamic_zones_repository.h"
+#include <array>
 #include <ctime>
 
 extern ClientList client_list;
@@ -153,6 +154,8 @@ void SharedTaskManager::AttemptSharedTaskCreation(
 		e.activity_id    = a.activityid;
 		e.done_count     = 0;
 		e.max_done_count = a.goalcount;
+		e.step           = a.step;
+		e.optional       = a.optional;
 
 		shared_task_activity_state.emplace_back(e);
 	}
@@ -182,7 +185,7 @@ void SharedTaskManager::AttemptSharedTaskCreation(
 	new_shared_task.SetMembers(request_members);
 
 	// add to shared tasks list
-	m_shared_tasks.emplace_back(new_shared_task);
+	auto& inserted = m_shared_tasks.emplace_back(new_shared_task);
 
 	// send accept to members
 	for (auto &m: request_members) {
@@ -196,6 +199,9 @@ void SharedTaskManager::AttemptSharedTaskCreation(
 		);
 	}
 	SendSharedTaskMemberListToAllMembers(&new_shared_task);
+
+	// check if task should immediately lock
+	HandleCompletedActivities(&inserted);
 
 	LogTasks(
 		"[AttemptSharedTaskCreation] shared_task_id [{}] created successfully | task_id [{}] member_count [{}] activity_count [{}] current tasks in state [{}]",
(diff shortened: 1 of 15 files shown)


world/shared_task_manager.cpp
```diff
@@ -1815,3 +1798,74 @@ SharedTaskManager *SharedTaskManager::PurgeExpiredSharedTasks()
...
+bool SharedTaskManager::HandleCompletedActivities(SharedTask* s)
+{
+	bool is_task_complete = true;
+	bool lock_task = false;
+
+	std::array<bool, MAXACTIVITIESPERTASK> completed_steps;
+	completed_steps.fill(true);
+
+	auto activity_states = s->GetActivityState();
+	std::sort(activity_states.begin(), activity_states.end(),
```
Akkadius, line 1828: Would be good to add simple comments to new viewers why some of these steps are happening. Why are we sorting here for example (I'm sure there's plenty good reason, but good to know)

world/shared_task_manager.cpp
```diff
@@ -1815,3 +1798,74 @@ SharedTaskManager *SharedTaskManager::PurgeExpiredSharedTasks()
...
+		}
+	}
+}
+
+bool SharedTaskManager::HandleCompletedActivities(SharedTask* s)
+{
+	bool is_task_complete = true;
+	bool lock_task = false;
+
+	std::array<bool, MAXACTIVITIESPERTASK> completed_steps;
```
Akkadius, line 1824: Comments: What is this doing and why

world/shared_task_manager.cpp
```diff
@@ -1815,3 +1798,74 @@ SharedTaskManager *SharedTaskManager::PurgeExpiredSharedTasks()
...
+		{
+			is_task_complete = false;
+			if (a.step > 0 && a.step <= MAXACTIVITIESPERTASK)
+			{
+				completed_steps[a.step - 1] = false;
+			}
+		}
+
+		int lock_index = s->GetTaskData().lock_activity_id;
+		if (a.activity_id == lock_index && a.step > 0 && a.step <= MAXACTIVITIESPERTASK)
```
Akkadius, line 1843: Comments: What is this doing and why?

(Hard for me to interpret this at a glance)

Akkadius: Makes sense to me, but to new travelers might feel magical

world/shared_task_manager.cpp
```diff
@@ -1815,3 +1798,74 @@ SharedTaskManager *SharedTaskManager::PurgeExpiredSharedTasks()
...
+		{
+			// lock if element is active (on first step or previous step completed)
+			lock_task = (a.step == 1 || completed_steps[a.step - 2]);
+		}
+	}
+
+	// completion locks are silent
```
Akkadius, line 1850: Silent meaning we just don't broadcast to clients?

world/shared_task_manager.cpp
hgtw: It's just iterating to determine completed steps (activities have separate step grouping)

world/shared_task_manager.cpp
```diff
@@ -1815,3 +1798,74 @@ SharedTaskManager *SharedTaskManager::PurgeExpiredSharedTasks()
 
 	return this;
 }
+
+void SharedTaskManager::LockTask(SharedTask* s, bool lock)
```
Akkadius, line 1802: Beautiful

world/shared_task_manager.cpp
hgtw: Which part? It's just checking the activity id is the lock index (and some index bounds)

world/shared_task_manager.cpp
hgtw: Correct

world/shared_task_manager.cpp
```diff
@@ -182,7 +185,7 @@ void SharedTaskManager::AttemptSharedTaskCreation(
 	new_shared_task.SetMembers(request_members);
 
 	// add to shared tasks list
-	m_shared_tasks.emplace_back(new_shared_task);
+	auto& inserted = m_shared_tasks.emplace_back(new_shared_task);
```
Akkadius, line 188: Curious why you use inserted this way instead of just passing `new_shared_task` to `HandleCompletedActivities` ?

Akkadius (approved):
Some extra comments to help describe what's happening could be helpful, not required.

Looks great per usual, clean

world/shared_task_manager.cpp
hgtw: LockTask modifies the shared task with the `GetDbSharedTask` and `SetDbSharedTask` calls. I would have had to call it before it was inserted into m_shared_tasks to do that otherwise it would modify the local copy. (I originally had it like this but was playing with triggering completion after for tasks with no elements, but ended up removing that and leaving it after the emplace)

Akkadius: Ahh interesting

PR #2340, opened 2022-07-31
[Shared Tasks] Implement task timer groups

hgtw:
This adds task replay and request timer groups (an arbitrary id) which
allows for different tasks to share lockouts

### DIFF
--- a/world/shared_task_manager.cpp
+++ b/world/shared_task_manager.cpp
@@ -132,6 +132,7 @@ void SharedTaskManager::AttemptSharedTaskCreation(
 			timer.character_id = m.character_id;
 			timer.task_id      = task.id;
 			timer.timer_type   = static_cast<int>(TaskTimerType::Request);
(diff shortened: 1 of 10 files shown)


world/shared_task_manager.cpp
```diff
@@ -1300,6 +1302,29 @@ void SharedTaskManager::SendMembersMessageID(
 	}
 }
 
+std::vector<CharacterTaskTimersRepository::CharacterTaskTimers> SharedTaskManager::GetCharacterTimers(
```
Akkadius, line 1305: Why do we vector this when we only return 1?

Akkadius (approved):
Minor comment, LGTM

world/shared_task_manager.cpp
hgtw: It was just because the previous code used the vector repository result and I didn't want to add more noise modifying code that already relied on it being a vector when I broke it out to a new function

PR #2343, opened 2022-07-31
[Dynamic Zones] Implement dz switch id

hgtw:
This adds the `dz_switch_id` field to doors and dynamic_zones. It will
allow for compasses to be automatically added to dz entrances and will
support moving clients to the dz on use without needing to script it.
These can be imported for switches/doors from live packet dumps.

Also removes compass packet encoders (same struct in all clients)


Akkadius (approved):
Clean

PR #2351, opened 2022-07-31
[Tasks] Make #task reloadall not quit shared tasks

hgtw:
This was causing #task reloadall to desync state for shared tasks.
Characters were removed from the shared task but remained in the
character_tasks table. This CancelTask is cosmetic only to reload
the client window.


mackal (approved):

Kinglykrab (approved):
:+1:

Akkadius (approved):

PR #2355, opened 2022-08-01
[Feature] Splash Factions

mackal:
This is based on theories that have been formulated by several people about how factions are actually linked.

You get hit in one faction and you "splash" the others.

From the QuestReward packet we know they just use the faction ID and the value that faction is hit with, the rest follow, ALWAYS in the same ratio (sometimes hard to see with lower values since they min cap to abs(1) for a change)

This should be 100% compatible with current system and just be something to opt in.

This will allow us to use QuestReward to handle pretty much everything, exactly like live does :)

I will eventually work through what I can on PEQ and add an optional SQL that's a dump of the ratios I've been able to figure out.


mackal:
The trigger for using the newer system is if faction amount is set in the various places, otherwise old system is still used.

mackal:
The design of the splash_faction table is kind of meh, but it gets the job done and isn't needlessly complex for emulating live

mackal:
https://github.com/ProjectEQ/projecteqquests/blob/a3f6750bc075c674eb35d2b2f6bb17efcb274859/barindu/%23Njanu_Dinari.lua#L7 example of a quest we currently are calling QuestReward and then HAVE to manually edit factions, this allow us to handle it all in QuestReward

https://github.com/ProjectEQ/projecteqquests/blob/a3f6750bc075c674eb35d2b2f6bb17efcb274859/akanon/Baxok_Curhunter.lua#L40 Another example where we are currently making 8 quest calls that should just be a single QuestReward. (Currently that example we could make 3 of them into a single QuestReward)

mackal:
Figured I should give an example of a `splash_faction` entry, using Baxok Curhunter as an example:

```sql
INSERT INTO `splash_faction` (id, splash_id1, splash_mod1, splash_id2, splash_mod2, splash_id3, splash_mod3, splash_id4, splash_mod4) VALUES (255, 288, 0.25, 333, 0.25, 238, -0.25, 1604, -0.05);
```

Then for the Recruitment Summons QuestReward with provide faction 255 and 100. The other quest he has if we look on alla, the faction reward should actually be 255 with 20

common/database_schema.h
```diff
@@ -232,6 +232,7 @@ namespace DatabaseSchema {
 			"spawnentry",
 			"spawngroup",
 			"spells_new",
+			"splash_faction",
```
hgtw, line 235: Maybe this should be called something like `faction_sets` or `faction_set_ratios`? At the very least `faction` as the prefix for the name I think would be better

common/shareddb.cpp
```diff
@@ -1448,6 +1448,105 @@ bool SharedDatabase::LoadNPCFactionLists(const std::string &prefix) {
...
+{
+	EQ::FixedMemoryHashSet<SplashFactions> hash(reinterpret_cast<uint8 *>(data), size, list_count, max_lists);
+	SplashFactions faction;
+
+	// yes, this table is kind of shit, but it works and given how VI wrote somethings *cough* spells *cough* it's
+	// probably what they do :P
+	const std::string query =
+	    "SELECT id, splash_id1, splash_mod1, splash_id2, splash_mod2, splash_id3, splash_mod3, splash_id4, "
+	    "splash_mod4, splash_id5, splash_mod5, splash_id6, splash_mod6, splash_id7, splash_mod7, splash_id8, "
+	    "splash_mod8, splash_id9, splash_mod9, splash_id10, splash_mod10 FROM splash_faction";
```
hgtw, line 1494: Should be able to just do `auto results = SplashFactionRepository::All(*this)` and loop the returned vector instead of manually constructing the query

shared_memory/main.cpp
```diff
@@ -298,7 +304,17 @@ int main(int argc, char **argv)
 		}
 	}
 
-	if (load_all || load_bd) {
+	if(load_all || load_splash_factions) {
+		LogInfo("Loading splash factions");
+		try {
+			LoadSplashFaction(&database, hotfix_name);
```
hgtw, line 310: Should this be `&content_db`?

utils/sql/git/required/2022_08_01_splash_faction.sql
```diff
@@ -0,0 +1,27 @@
...
+	`splash_id7` INT(10) NOT NULL DEFAULT 0,
+	`splash_mod7` FLOAT NOT NULL DEFAULT 0,
+	`splash_id8` INT(10) NOT NULL DEFAULT 0,
+	`splash_mod8` FLOAT NOT NULL DEFAULT 0,
+	`splash_id9` INT(10) NOT NULL DEFAULT 0,
+	`splash_mod9` FLOAT NOT NULL DEFAULT 0,
+	`splash_id10` INT(10) NOT NULL DEFAULT 0,
+	`splash_mod10` FLOAT NOT NULL DEFAULT 0,
+	PRIMARY KEY(`id`)
+);
```
hgtw, line 24: Could probably make this into a table that just links a primary faction with a sub faction id and a multiplier,
something like:
`primary key id autoincrement, primary_faction_id, sub_faction_id, mod/multiplier`

Not sure if that would make things more complicated or not, but wouldn't need to worry about reaching any limit (though 10 seems high already)

zone/client.cpp
```diff
@@ -8814,6 +8820,37 @@ void Client::CashReward(uint32 copper, uint32 silver, uint32 gold, uint32 platin
 	QueuePacket(outapp.get());
 }
 
+void Client::RewardFaction(int id, int amount)
+{
+	auto splash_hits = database.GetSplashFactionHit(id);
```
hgtw, line 8825: Should this be `content_db` or does it have to be `database` for shared memory?

zone/lua_client.cpp
```diff
@@ -1761,9 +1766,27 @@ void Lua_Client::QuestReward(Lua_Mob target, luabind::adl::object reward) {
...
 	cur = reward["faction"];
 	if (luabind::type(cur) != LUA_TNIL) {
-		try {
-			faction = luabind::object_cast<bool>(cur);
-		} catch (luabind::cast_failed &) {
+		// if it's a table it will be {faction, faction_mod}
+		if (luabind::type(cur) == LUA_TTABLE) {
+			auto item = cur[1];
+			if (luabind::type(item) != LUA_TNIL) {
+				try {
```
hgtw, line 1773: Should be able to just check if type is `LUA_TNUMBER` here and do the cast without the try/catch. The parser will catch any exceptions as quest errors anyway for bad inputs which might be more desirable than silently failing

zone/main.cpp
```diff
@@ -316,6 +316,11 @@ int main(int argc, char** argv) {
 		LogError("Loading npcs faction lists failed!");
 		return 1;
 	}
+	LogInfo("Loading splash faction hits");
+	if (!database.LoadSplashFaction(hotfix_name)) {
```
hgtw, line 320: Same here with `content_db`, though maybe there's something with shared memory I don't understand because I see others using `database` that are listed as content tables too

zone/worldserver.cpp
```diff
@@ -3184,6 +3184,11 @@ void WorldServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p)
 			LogError("Loading npcs faction lists failed!");
 		}
 
+		LogInfo("Loading splash faction lists");
+		if (!database.LoadSplashFaction(hotfix_name)) {
```
hgtw, line 3188: Possibly `content_db` like the others

common/database_schema.h
mackal: splash faction is the name I've seen people use for this concept so I went with splash faction. It was already established so I don't think there is a good reason not to use it.

zone/lua_client.cpp
mackal: I was copying the pattern we were already doing.

shared_memory/main.cpp
mackal: I suppose so, I did write this before repositories existed.

zone/worldserver.cpp
hgtw: I'm guessing for shared memory the database doesn't matter outside of (re)loading from the real database since it's just local, just some of the inconsistencies are confusing

utils/sql/git/required/2022_08_01_splash_faction.sql
mackal: I would rather keep it simple. This isn't about giving us the ability to do anything more than what live does. And I would rather increase the limit than making it arbitrary if we come across a case that live needs.

zone/lua_client.cpp
mackal: I think it might be scope creep to rewrite this, the function does this with other stuff and rewriting the whole function is probably a good idea, but that's not implementing this feature.

utils/sql/git/required/2022_08_01_splash_faction.sql
brainiac: curious why these wouldn't just be additional columns on the faction table?

common/database_schema.h
brainiac: i vote for faction associations

utils/sql/git/required/2022_08_01_splash_faction.sql
mackal: I didn't really want to touch the current stuff and risk breaking anything. Plus joins are involved in loading faction stuff currently and I didn't wanna deal with that :D

mackal: Okay, looked into how `npc_faction` and `npc_faction_entries` works ... it doesn't really fit in there :P

brainiac: If it were in an existing table I would expect it in faction_base_data or faction_list tbh

mackal: faction_base_data could probably be expanded for this.

noudess: > faction_base_data could probably be expanded for this.

One argument against is that faction_base_data is a direct import from client files, or was anyway.

noudess:
Overall , I like @mackal 's approach to keep things separate rather than merging into an existing table.  At one point, I also did a redesign of the tables, but it would have had such a large impact on existing tools, code, even the quest API that I let it be.

mackal:
I'm fine with renaming it to faction associations or something better if anyone can think of something else  :P

Akkadius:
> I'm fine with renaming it to faction associations or something better if anyone can think of something else :P

Would certainly prefer to have it prefixed to its functionor category per https://docs.eqemu.io/server/database/conventions-guidelines/#table-names

common/shareddb.cpp
Akkadius: Agreed

common/database_schema.h
Akkadius: I agree, at least `faction_` something

Akkadius:
Closing until comments addressed, Mackal mentioned making a new PR anyhow

PR #2356, opened 2022-08-01
[WIP] [Feature] Corpse Overhaul

fryguy503:
- Player corpses now have two timers, one specific to the rezability of the corpse and the other to cover the overall rot timer of the player corpse.
- The rezability timer is based on the online presence of the player and is not affected by being offline.
- The rot timer is not affected by offline/online status and will count to the rot status of the corpse.
- Corpses can be rezzed multiple times, however only the first rez that yeilds returned xp will be counted. Not other rez will return any xp. This allows for a "Poor mans COTH" as was used many times in the early eras.

This PR is a Draft PR as I have not yet tested this on a stock build of eqemu and it will likely need lots of adjustments and fixes. Prior to digging too much further, I figured I should get some feedback to the overall direction of this change.


zone/client.h
```diff
@@ -170,6 +170,14 @@ struct XTarget_Struct
 	char Name[65];
 };
 
+typedef enum {
```
mackal, line 173: I think an `enum class` would probably work better here, we can always safely cast in places need be

Akkadius:
I’ll take a look at this one in the next day or so, getting caught up from travel still

zone/corpse.cpp
```diff
@@ -854,12 +883,18 @@ bool Corpse::IsEmpty() const {
 }
 
 bool Corpse::Process() {
-	if (player_corpse_depop)
+	if (player_corpse_depop) {
 		return false;
+	}
+
+	if (owner_online_timer.Check() || rezzable) {
```
hgtw, line 890: Is the `|| rezzable` here intentional? This could be pretty expensive polling for online characters from Corpse::Process. Maybe something can be done in world for client connect/disconnect to tell zones (and new booted zones) a player's online state instead of polling.

world/zoneserver.cpp
```diff
@@ -901,6 +901,20 @@ void ZoneServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p) {
...
+				break;
+			}
+			ServerIsOwnerOnline_Struct* online = (ServerIsOwnerOnline_Struct*) pack->pBuffer;
+			ClientListEntry* cle = client_list.FindCharacter(online->name);
+			if (cle) {
+				online->online = 1;
+			} else {
+				online->online = 0;
+			}
+			zoneserver_list.FindByZoneID(online->zoneid)->SendPacket(pack);
```
hgtw, line 915: Need to check if `FindByZoneID` is nullptr

zone/client_packet.cpp
```diff
@@ -4889,13 +4889,19 @@ void Client::Handle_OP_ConsiderCorpse(const EQApplicationPacket *app)
...
 				Chat::NPCQuestSay,
 				fmt::format(
-					"This corpse {} be resurrected.",
-					target->IsRezzed() ? "cannot" : "can"
+					"This corpse's resurrection time will expire in {}",
+					rez_time_string
 				).c_str()
 			);
+			} else {
+				Message(Chat::NPCQuestSay, "This corpse is too old to be resurrected.");
```
hgtw, line 4903: We should be able to switch this and the ressurection time message for eqstrs. Would have to confirm with live but probably these
```
231 This player cannot be resurrected. The corpse is too old.
491 This corpse's resurrection time will expire in %1 hour(s) %2 minute(s) %3 seconds.
492 This corpse's resurrection time will expire in %1 minute(s) %2 seconds.
```

zone/corpse.cpp
```diff
@@ -310,6 +315,15 @@ Corpse::Corpse(Client* client, int32 in_rezexp) : Mob (
 	silver			= 0;
 	gold			= 0;
 	platinum		= 0;
+	killedby		= in_killedby;
+	rezzable		= true;
+	rez_time		= 0;
+	is_owner_online = false; // We can't assume they are online just because they just died. Perhaps they rage smashed their router.
```
Akkadius, line 321: LOL

zone/corpse.cpp
Akkadius: Agreed. This is being needlessly polled when we can hook into existing places in the code that clearly can inform that connected status of a client

For example, where these events are triggered we can send a packet to all zones to inform status: EVENT_CONNECT or EVENT_DISCONNECT

Akkadius:
Generally good with the changes. Need to rebase with master and address feedback

Akkadius:
Closing since been stale for 2 months now. Would love to take these changes with the suggestions made

PR #2328, opened 2022-07-28
[Feature] GM State Change Persistance

fryguy503:
- Flymode and Invulnerable will now persist over zoning.
- Appended GMSpeed, Flymode and Invulnerable to the hideme message GMs see when they first login.
- Added #godmode [on/off] command to turn on or off hideme, flymode, gmspeed and invulnerable all in one shot.
- /becomenpc will now disable tells to the target player. It will also automatically disable GM States that interfere with its functionality.
- GM Command /toggle will not properly turn tells on/off
- GMs will now be notified if they are ignoring tells when they first zone-in, provided their GM flag is up.
- Added TellsOff variable to the output to #showstats

PR #2358, opened 2022-08-04
[Rules] Add Rule to Enable Tells with #hideme

Kinglykrab:
With the recent changes to #hideme as of #2328, tells are disabled to hidden GMs, meaning a GM must show themselves in order to receive tells.

### DIFF
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -809,6 +809,7 @@ RULE_CATEGORY_END()
 
 RULE_CATEGORY(Command)
 RULE_BOOL(Command, DyeCommandRequiresDyes, false, "Enable this to require a Prismatic Dye (32557) each time someone uses #dye.")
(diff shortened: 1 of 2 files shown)


mackal:
I think this PR is implying before, #hideme on wouldn't auto /toggle off, which it is now? Is that correct?

Kinglykrab:
> I think this PR is implying before, #hideme on wouldn't auto /toggle off, which it is now? Is that correct?

Yes. It currently forces tells off in #hideme, it did not do that before.

Kinglykrab:
> `/toggle on/off` will enable/disable when in hideme i thought.

The idea is to not have to disable it for those who don’t know how.

fryguy503:
> > `/toggle on/off` will enable/disable when in hideme i thought.
> 
> The idea is to not have to disable it for those who don’t know how.

Makes Sense.

fryguy503 (approved):
lgtm

PR #2363, opened 2022-08-08
[Feature] Change casting_spell_mana, to uint32

Aeadoin:
Confirmed that Mana Costs above 65,535 are now functioning as expected.

### DIFF
--- a/zone/perl_mob.cpp
+++ b/zone/perl_mob.cpp
@@ -2447,6 +2447,11 @@ void Perl_Mob_ApplySpellBuff(Mob* self, int spell_id, int duration) // @categori
 	self->ApplySpellBuff(spell_id, duration);
 }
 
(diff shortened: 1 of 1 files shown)


mackal (approved):

mackal:
Pretty sure live uses signed ints here.

Aeadoin:
> Pretty sure live uses signed ints here.

I could change these to signed ints, and check other called functions if you'd prefer?

Kinglykrab:
> > Pretty sure live uses signed ints here.
> 
> I could change these to signed ints, and check other called functions to ensure they are signed ints if you'd prefer?

Sounds good to me.

Kinglykrab (approved):

Aeadoin:
Will make new commit to change values to Signed Int

PR #2368, opened 2022-08-12
[Bug] Loot Drop Randomization adjustment

fryguy503:
Loot Table currently favors the first item in the loot drop. This should help remove that condition.

Code by Ailia.

### DIFF
--- a/zone/loottables.cpp
+++ b/zone/loottables.cpp
@@ -165,11 +165,18 @@ void ZoneDatabase::AddLootDropToNPC(NPC *npc, uint32 lootdrop_id, ItemList *item
 	}
 
 	float       roll_t           = 0.0f;
+	float       no_loot_prob     = 1.0f;
+	bool        RollTableChanceBypass = false;
 	bool        active_item_list = false;
 	for (uint32 i                = 0; i < loot_drop->NumEntries; ++i) {
 		const EQ::ItemData *db_item = GetItem(loot_drop->Entries[i].item_id);
 		if (db_item && npc->MeetsLootDropLevelRequirements(loot_drop->Entries[i])) {
 			roll_t += loot_drop->Entries[i].chance;
+			if(loot_drop->Entries[i].chance >= 100) {
+				RollTableChanceBypass = true;
+			} else {
+				no_loot_prob *= (100-loot_drop->Entries[i].chance)/100.0f;
+			}
 			active_item_list = true;
 		}
 	}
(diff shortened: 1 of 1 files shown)


noudess:
I don't think this is true.  If other people do, I'll look again, but I spend a long time understanding this code long ago and while convoluted, I think it skips by stuff based on a random roll for the total of all entries.

fryguy503:
> 

Ailia outlined it the case as a few people had reported it about 6 months ago. He mentioned they had fixed it on classless and had quickly wrote up that example. I have been running it on my server for the last month and it so far seems good. it was most noticeable on equal percent drop items. You would see a heavier skew towards the lowest item id (first in the list) item drop rates. I do not believe this is the best fix or even a correct fix, but its a dirty band aide until the entire loot system can be re-written to properly emulate how live manages it.

noudess:
I'd love to know more, because I spent alot of time understanding this code a long while back, and had to fix some of it.  It's so convoluted I know I'd get lost in in a half dozen times again.

Has anyone run before after data collection on the failing case?

fryguy503:
Prior I ran small scale manual tests of the %'s with a 5 item spread at 20% for each item and the weight was around 40% to the first item. These are not great tests as they were only 100-200 permutations. Afterwards it was closer to +/-5% from the 20% I would love for us to get a #symloot count loottable at some point. With a system like that we could check on 1000's of attempts.

noudess:
I could write something and run it on the old code, then give it to you.  Provide me a loottable/lootdrop scenario and I'll write something I can share.

noudess:
I've gone over this in a chat session with the code's author.  I see his points and how the fix addresses them.

I stand corrected,  Earlier, I just read the part about 1st entry bias and didn't really understand.

noudess (approved):
Nicely consolidated.  Author talked me through how it created common "sets" of drops.  This is cleaner and applies the same idea to all types of tables.

Nice.

zone/loottables.cpp
```diff
@@ -165,11 +165,18 @@ void ZoneDatabase::AddLootDropToNPC(NPC *npc, uint32 lootdrop_id, ItemList *item
 	}
 
 	float       roll_t           = 0.0f;
+	float       no_loot_prob     = 1.0f;
+	bool        RollTableChanceBypass = false;
```
Akkadius, line 169: `snake_case`

Issue #2367, opened 2022-08-11
More precise door controls, not have the menu pop up every time I click hotlink


DeadZergling:
I would like to see the #door command have more exact increments, and larger size options mostly, something similar to this:
![doormenu](https://user-images.githubusercontent.com/18122846/184066073-c53b0210-ee43-4b72-bd29-9c169c719f69.png)
The -1000/+1000 are helpful when I make a new door and it ends up super large.

I would also like to see the door menu not appear every time I click one of the hotlinks, it makes it more tedious then it needs to be, with the text jumping about as I am trying to place a door, sometimes in such a way that I end up clicking the wrong hotlink and mess it up even more.

Hope this sees traction soon, thanks!

Kinglykrab:
Resolved by #2370.

PR #2370, opened 2022-08-13
[Doors] Improvements to door manipulation

Akkadius:
This PR is in response to https://github.com/EQEmu/Server/issues/2367

**Changes**

* Simplified the saylink menu building
* Fixed a bug where `Doors` scoped variables were the same as variables being passed in so member variables were not being set and creating uninitialized behavior - resulting in created objects being incredibly small or large
* Refactored private member variables to be `m_` prefixed for readability and addressing / preventing the above issue from happening again
* Silent saylinks now pipe to commands (So when a saylink is clicked it doesn't result in. "You say, "message"")
* NPC Scale Manager now skips the door helpers because they have no NPC type ID assigned
* The door menu no longer prints every time a command is clicked (because silent saylinks now work)
* All door edit commands were converted to silent saylinks
* Door target saylink was converted to be silent

PR needs additional testers to make sure nothing was broken around EVENT_SAY, EVENT_COMMAND and any other say behaviors because of code consolidation, cleanup

image

### DIFF
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -1111,8 +1111,8 @@ void Client::ChannelMessageReceived(uint8 chan_num, uint8 language, uint8 lang_s
 		break;
 	}
 	case ChatChannel_Say: { /* Say */
-		if(message[0] == COMMAND_CHAR) {
-			if(command_dispatch(this, message) == -2) {
+		if (message[0] == COMMAND_CHAR) {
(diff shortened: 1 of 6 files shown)


Kinglykrab:
Tested with Perl and Lua scripted `#` command and seemed to work fine even still. I'll need to update the `#help` saylinks to be silent.

Kinglykrab (approved):

PR #2372, opened 2022-08-13
[Saylinks] Add Silent helper

Akkadius:
Simply adds helper function

```cpp
std::string Silent(const std::string &saylink_text, const std::string &link_name);
```

To be used in later PRs

Also passes string by reference to avoid copying


Kinglykrab (approved):
👍

PR #2373, opened 2022-08-14
[Saylinks] Convert all GM Command Saylinks to Silent Saylinks.

Kinglykrab:
- This cleans up all non-silent GM Command Saylinks that we had before due to the way they worked before. All saylinks like this should be silent now.
- Add source short hand capability for say links with same link as text.

### DIFF
--- a/zone/worldserver.cpp
+++ b/zone/worldserver.cpp
@@ -3148,7 +3148,7 @@ void WorldServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p)
 			zone->SetQuestHotReloadQueued(true);
 		} else if (request_zone_short_name == "all") {
 			std::string reload_quest_saylink = Saylink::Silent("#reload quest", "Locally");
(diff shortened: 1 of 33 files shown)


zone/worldserver.cpp
```diff
@@ -3147,16 +3147,8 @@ void WorldServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p)
...
-				false,
-				"Locally"
-			);
-			std::string reload_world_saylink = Saylink::Create(
-				"#reload world",
-				false,
-				"Globally"
-			);
+			std::string reload_quest_saylink = Saylink::Silent("#reload quest", "Locally");
+			std::string reload_world_saylink = Saylink::Silent("#reload world",	"Globally");
```
Akkadius, line 3151: Extra spaces

Kinglykrab: Was a tab, good eye.

Akkadius (approved):

PR #2374, opened 2022-08-14
[Tasks] Implement Task Activity Prerequisites

hgtw:
Some live tasks make new elements available without requiring all
currently active ones to be completed first.

This adds the `req_activity_id` field to task activities which will mark
an element active if its required activity id is completed. If a valid
value is set then it's used instead of checking the current step.

The `step` field may still be set on rows with a valid `req_activity_id`
to specify its logical step and prevent later steps from becoming active
until completed. It's only ignored when deciding if the current element
is active.

The legacy task logic for unlocking activities was completely refactored
for this. A common method has been added so both zone and world can make
use of it to determine which elements are currently active. The previous
step system should remain unchanged.

The world logic for locking shared tasks when an element became active
did not account for "sequential" mode (all steps 0), unordered steps, or
gaps in step numbers. This also resolves that issue.

### DIFF
--- a/common/version.h
+++ b/common/version.h
@@ -34,7 +34,7 @@
  * Manifest: https://github.com/EQEmu/Server/blob/master/utils/sql/db_update_manifest.txt
  */
 
-#define CURRENT_BINARY_DATABASE_VERSION 9198
+#define CURRENT_BINARY_DATABASE_VERSION 9199
 
 #ifdef BOTS
 	#define CURRENT_BINARY_BOTS_DATABASE_VERSION 9029
--- a/tests/main.cpp
+++ b/tests/main.cpp
@@ -29,6 +29,7 @@
 #include "string_util_test.h"
 #include "data_verification_test.h"
 #include "skills_util_test.h"
+#include "task_state_test.h"
 #include "../common/eqemu_config.h"
 
 const EQEmuConfig *Config;
@@ -49,6 +50,7 @@ int main() {
 		tests.add(new StringUtilTest());
 		tests.add(new DataVerificationTest());
 		tests.add(new SkillsUtilsTest());
+		tests.add(new TaskStateTest());
 		tests.run(*output, true);
 	} catch(...) {
 		return -1;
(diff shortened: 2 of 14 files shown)


Kinglykrab (approved):
👍

Akkadius:
I'll look at this one tomorrow when my eyes aren't crossed

zone/task_client_state.cpp
```diff
@@ -366,218 +366,94 @@ static void DeleteCompletedTaskFromDatabase(int character_id, int task_id)
...
-					}
-					else {
-						++iterator;
-					}
-				}
+void ClientTaskState::RecordCompletedTask(uint32_t character_id, const TaskInformation& task, const ClientTaskInformation& client_task)
+{
+	// If we are only keeping one completed record per task, and the player has done
+	// the same task again, erase the previous completed entry for this task.
+	if (RuleB(TasksSystem, KeepOneRecordPerCompletedTask))
```
Akkadius, line 424: Do we need a rule for this ?

zone/task_client_state.cpp
```diff
@@ -366,218 +366,94 @@ static void DeleteCompletedTaskFromDatabase(int character_id, int task_id)
...
-		for (int i             = 0; i < p_task_data->activity_count; i++) {
-
-			if (p_task_data->activity_information[i].step_number == 0 &&
-				task_info.activity[i].activity_state == ActivityHidden) {
-				task_info.activity[i].activity_state = ActivityActive;
-				// task_info.activity_information[i].updated=true;
-			}
+	for (int activity_id : res.active)
+	{
+		ClientActivityInformation& client_activity = task_info.activity[activity_id];
```
Akkadius, line 403: This is a receiver of a complex type so can we just make it `a`

tests/main.cpp
```diff
@@ -49,6 +50,7 @@ int main() {
 		tests.add(new StringUtilTest());
 		tests.add(new DataVerificationTest());
 		tests.add(new SkillsUtilsTest());
+		tests.add(new TaskStateTest());
```
Akkadius, line 53: BEAUTIFUL

common/version.h
```diff
@@ -34,7 +34,7 @@
  * Manifest: https://github.com/EQEmu/Server/blob/master/utils/sql/db_update_manifest.txt
  */
 
-#define CURRENT_BINARY_DATABASE_VERSION 9197
+#define CURRENT_BINARY_DATABASE_VERSION 9198
```
Akkadius, line 37: Will need to be rebased

Akkadius (approved):

PR #2380, opened 2022-08-20
[Expansions] Zone expansion consistency changes

Akkadius:
* Removes the "off-by-1" issue the `expansion` field has which follows values in this list https://docs.eqemu.io/server/operation/expansion-list but + 1. This is confusing for operators and this PR eliminates that
* Adds `bypass_expansion_check` to the `zone` table particularly in cases where servers are using modern clients, using v2 versions of zones but want them available even when a server is set in a far earlier expansion

PR is complete pending finalization of getting the remaining zones to the bypass criteria


Akkadius:
This is now ready

Kinglykrab (approved):
👍

PR #2384, opened 2022-08-21
[Feature] Change Mana Costs to use Signed Int

Aeadoin:
Changed Spell Mana Costs from Unsigned Int16 to Signed Int32.

### DIFF
--- a/common/spdat.h
+++ b/common/spdat.h
@@ -1298,7 +1298,7 @@ struct SPDat_Spell_Struct
 /* 016 */	uint32 buff_duration_formula; // -- DURATIONBASE
 /* 017 */	uint32 buff_duration; // -- DURATIONCAP
 /* 018 */	uint32 aoe_duration;	// sentinel, rain of something -- IMPACTDURATION
-/* 019 */	uint16 mana; // Mana Used -- MANACOST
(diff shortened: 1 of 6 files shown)


common/spdat.h
```diff
@@ -1298,7 +1298,7 @@ struct SPDat_Spell_Struct
 /* 016 */	uint32 buff_duration_formula; // -- DURATIONBASE
 /* 017 */	uint32 buff_duration; // -- DURATIONCAP
 /* 018 */	uint32 aoe_duration;	// sentinel, rain of something -- IMPACTDURATION
-/* 019 */	uint16 mana; // Mana Used -- MANACOST
```
Kinglykrab, line 1301: May want to change the comments after this to indicate the proper byte size due to these change, just have to add 2 bytes to everything after this.

mackal (approved):

Kinglykrab (approved):

common/spdat.h
mackal: Comments are just a running column count of the CSV spells file.

Kinglykrab: Ah, didn't even read them, some places we have bytes.

Aeadoin: Good to know, if I do encounter that! @Kinglykrab

PR #2388, opened 2022-08-22
[Code Cleanup] Zone Data Loading Refactor

Akkadius:
This PR cleans up a variety of access patterns both for preparation of some expansion oriented work but to also simplify other zone data related access patterns throughout the source that make it painful for existing and new travelers to reason about.

**Testing**

This PR will need some extensive testing in the surface areas it touches so we can prevent unnecessary regressions.

**Notable Changes**

* `world_store` was consolidated into simply `zone_store` since it served almost identical purposes and were largely duplicated functions
* `zone_store` exploits globals that we've used historically everywhere for zone helper related functions
* `zone_store` holds zone data (The entire repository object) in memory for rapid access in plenty of other places in the code
* `zone_store` gets loaded initially on zone process bootup and can be reloaded through the `#reload` comand under the traditional `zone` (headers) reload, but additional things that were missed before in being reloaded are now being reloaded, such as zone safe coordinates amongst other data
* Zone safe coordinates / heading get properly used in #zone 
* Simplify `GetPEQZone` `getZoneShutDownDelay` and various other zone data access patterns throughout the source
* Deprecated `Database::GetSafePoints` since it was very overloaded and violated SRP
* Deprecated `Database::GetZoneLongName` since it was very overloaded and violated SRP
* Deprecated `ZoneDatabase::GetZoneCFG` since it was very overloaded and violated SRP
* Consolidated zone header variable names to use uniform `snake_casing` since the last handful of changes decided to `camelCase` 

**Zone Store Helpers**

```cpp
ZoneRepository::Zone *GetZone(uint32 zone_id, int version = 0);
ZoneRepository::Zone *GetZone(const char *in_zone_name);
uint32 GetZoneID(const char *in_zone_name);
uint32 GetZoneID(std::string zone_name);
std::string GetZoneName(uint32 zone_id);
std::string GetZoneLongName(uint32 zone_id);
const char *GetZoneName(uint32 zone_id, bool error_unknown = false);
const char *GetZoneLongName(uint32 zone_id, bool error_unknown = false);
ZoneRepository::Zone *GetZoneWithFallback(uint32 zone_id, int version = 0);
```

**Example(s) of Access Simplification**

**Before**

```cpp
uint8 Database::GetPEQZone(uint32 zone_id, uint32 version){
	std::string query = fmt::format(
		"SELECT peqzone FROM zone WHERE zoneidnumber = {} AND (version = {} OR version = 0) ORDER BY version DESC LIMIT 1",
		zone_id,
		version
	);
	auto results = QueryDatabase(query);
	if (!results.Success() || !results.RowCount()) {
		return 0;
	}
	auto row = results.begin();
	return static_cast<uint8>(std::stoi(row[0]));
}
```

**After**

```cpp
uint8 Database::GetPEQZone(uint32 zone_id, uint32 version){
	auto z = GetZoneVersionWithFallback(zone_id, version);
	return z ? z.peqzone : 0;
}
```

Another one pulling safe coordinates

```cpp
auto z = GetZone(pp.zone_id);
if (z) {
	pp.x = z->safe_x;
	pp.y = z->safe_y;
	pp.z = z->safe_z;
}
```


zone/client.cpp
```diff
@@ -5364,14 +5364,13 @@ void Client::SetStartZone(uint32 zoneid, float x, float y, float z, float headin
...
 	if (x == 0 && y == 0 && z == 0) {
-		content_db.GetSafePoints(
-			ZoneName(m_pp.binds[4].zone_id),
-			0,
-			&m_pp.binds[4].x,
-			&m_pp.binds[4].y,
-			&m_pp.binds[4].z,
-			&m_pp.binds[4].heading
-		);
+		auto zd = GetZoneVersionWithFallback(m_pp.binds[4].zone_id);
```
Kinglykrab, line 5367: GetZone() instead maybe since it's not using version here?

Akkadius: Good eye, missed one

common/zone_store.cpp
```diff
@@ -152,11 +154,35 @@ ZoneRepository::Zone ZoneStore::GetZone(uint32 zone_id, int version)
...
+	return ZoneRepository::NewEntity();
+}
+
+const std::vector<ZoneRepository::Zone> &ZoneStore::GetZones() const
+{
+	return m_zones;
+}
+
+// gets zone data by using explicit version and falling back to version 0 if not found
+ZoneRepository::Zone ZoneStore::GetZoneWithFallback(uint32 zone_id, int version)
```
brainiac, line 172: This function is returning a copy of Zone, which doesn't seem very efficient.

Akkadius: Agreed. Although - if it was easily able to be returned by reference or pointer we would here, but we are not always guaranteed we'll find an entity so we return an entirely new default entity so related object can determine it doesn't exist (z.id == 0) etc. 

Even with returning a memory copy this will be infinitely more efficient than all the silly database calls

brainiac: If you can return const ref, you can initialize a function level static and return that in the case a value doesn't exist.

brainiac: another option is to return a pointer, and return nullptr if it doesn't exist.

Akkadius: Could explore different models. I’m out for the night for now. Need to get ready for the week

world/worlddb.cpp
```diff
@@ -234,40 +234,35 @@ void WorldDatabase::GetCharSelectInfo(uint32 account_id, EQApplicationPacket **o
...
-							&y,
-							&z,
-							&heading
-						);
+						auto zone = GetZone(pp.binds[4].zone_id);
+						if (zone) {
+							pp.binds[4].x       = zone->safe_x;
+							pp.binds[4].y       = zone->safe_y;
+							pp.binds[4].z       = zone->safe_z;
+							pp.binds[4].heading = zone->safe_heading;
```
hgtw, line 260: There's a logic change here I think might be unintended. The previous code only set `binds[4]` to the safe coordinates if x,y,z,h were all 0, otherwise it set it to the x,y,z,h values. The new code only ever sets it to the safe coordinates

Akkadius: Roger that, good eye, will change when I’m coding next. I’ll take another pass at touch points as well

Akkadius: Fixed

Akkadius:
Addressed PR comments, this should be good to go.

Would like others to at least run through some testing on this if possible before merge later this week

world/worlddb.cpp
```diff
@@ -234,40 +234,35 @@ void WorldDatabase::GetCharSelectInfo(uint32 account_id, EQApplicationPacket **o
...
+							x       = zone->safe_x;
+							y       = zone->safe_y;
+							z       = zone->safe_z;
+							heading = zone->safe_heading;
+						}
 					}
-					player_profile_struct.binds[4].x = x;
-					player_profile_struct.binds[4].y = y;
-					player_profile_struct.binds[4].z = z;
-					player_profile_struct.binds[4].heading = heading;
```
hgtw, line 267: These need to be added back I think

Akkadius: Done

hgtw (approved):

Kinglykrab (approved):
👍

PR #2390, opened 2022-08-22
[Zone] Fix and simplify zone shutdown logic

Akkadius:
Zone shutdown logic has been polluted with many hacks and race conditions over the years, this PR addresses most of them. 

**Changes**

* Eliminates race issues from being called all over the source as hacks in different places
* `StartShutdownTimer` is now only called during bootup and during instance shutdown timer routines. The instance shutdown handler should likely be further clarified because it makes assumptions about a few things
* In instances where the timer just needs to be reset (players zone) we have a new method called `ResetShutdownTimer` which simply restarts the original timer instead of hitting the database and recalling raced logic
* Improved logging to inform users what is happening with timers

**Logging Examples**

Initial loading

```
[Zone] [Info] [StartShutdownTimer] Reset to [1 Hour] (Loaded from [zone table]) from original remaining time [5 Seconds] duration [5 Seconds] zone [The Surefall Glade (3)]
```

When a timer is being reset (player zones or players still in zone and timer is reached)

```
[Zone] [Info] [ResetShutdownTimer] Reset to [6 Seconds] from original remaining time [6 Seconds] duration [6 Seconds] zone [The Surefall Glade (3)]
```


Akkadius:
@hgtw you've recent context on instance shutdowns. We will likely need to consider coverage over it here but this PR should still make things easy enough to reason about

hgtw (approved):

PR #2392, opened 2022-08-22
[Utility] Add std::string_view overloads for std::from_chars

mackal:
std::from_chars floating point support is still pour, so we provide some
fall backs. These fall backs currently lack error handling.

(since GCC 11 and MSVC 2019 16.4 FP support is added, clang with libc++
still doesn't support FP)

GCC's floating point support is still rather pour (performance) and
probably shouldn't be used, but we have support to if people want to
now.

Probably need more test cases ...

### DIFF
--- a/common/strings.cpp
+++ b/common/strings.cpp
@@ -1,3 +1,4 @@
+// for folly stuff
 /*
  * Copyright 2013 Facebook, Inc.
  *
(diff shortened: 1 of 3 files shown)


Akkadius:
I’ll take a look at this one in the next day or so, getting caught up from travel still

Akkadius (approved):

PR #2395, opened 2022-08-23
[Feature] Implement Heroic Strikethrough to NPCs

Aeadoin:
Heroic Strikethrough Implementation per Dev Quotes (2018):
https://forums.daybreakgames.com/eq/index.php?threads/illusions-benefit-neza-10-dodge.246757/#post-3622670

Heroic Strikethrough is set to 0 by default under each npc_types entries, each point of heroic strike-through negates 25 points of Heroic Dex/Agi for Avoidance based Skill checks (Dodge, Parry, Riposte, Block Skill)

If an NPC's Heroic Strikethrough is higher than your character's Heroic Agility bonus, you are disqualified from the "bonus" you would have gotten at the end. 

Can be used in Tandem with Special Attack 44 Counter Avoid Damage

*Updated Version to 9201, and updated manifest.


zone/mob.h
```diff
@@ -640,6 +641,7 @@ class Mob : public Entity {
 	void SetIsBoat(bool boat) { is_boat = boat; }
 	bool IsControllableBoat() const;
 	inline const bool AlwaysAggro() const { return always_aggro; }
+	inline virtual int32 GetHeroicStrikethrough() const  { return heroic_strikethrough; }
```
mackal, line 644: This does not need to be virtual

mackal (changes requested):

zone/mob.h
Aeadoin: @mackal fixed.

zone/npc.cpp
```diff
@@ -2785,6 +2791,8 @@ float NPC::GetNPCStat(const char *identifier)
 	}
 	else if (id == "npc_spells_effects_id") {
 		return npc_spells_effects_id;
+	}else if (id == "heroic_strikethrough") {
```
Kinglykrab, line 2794: Formatting. L

zone/npc.h
```diff
@@ -448,6 +448,7 @@ class NPC : public Mob
 	const char* GetRawNPCTypeName() const { return NPCTypedata_ours ? NPCTypedata_ours->name : NPCTypedata->name; }
 
 	virtual int GetKillExpMod() const { return NPCTypedata_ours ? NPCTypedata_ours->exp_mod : NPCTypedata->exp_mod; }
+	int GetHeroicStrikethrough() const { return NPCTypedata_ours ? NPCTypedata_ours->heroic_strikethrough : NPCTypedata->heroic_strikethrough; }
```
mackal, line 451: Some how I missed this when I initially reviewed it, but this function shouldn't be necessary since we will be using the duped/customized NPCType struct when we initialize the Mob and copy the data to Mob class. If we didn't do that, my initial comment about virtual would have been wrong too :P

mackal (changes requested):

zone/npc.h
Aeadoin: Thanks! taken care of.

mackal (approved):

Kinglykrab (approved):

PR #2397, opened 2022-08-26
[Quest API] Add caster_id and caster_level to Lua npc cast events

JasXSL:
First pull request, so I've probably messed some formatting up.

The client version of these events have caster_id and caster_level provided. I think being able to figure out who cast a spell on an NPC would be useful too, especially since the data provided from spells.cpp is the same for both player and npc. Tested on my server and seems to work well for both NPC and Client casters. 

Example:

```
function event_cast_on(e)
	local client = eq.get_entity_list():GetClientByID(e.caster_id);
	if client.valid then
		eq.debug("Spell "..e.spell:ID().." was cast by client "..client:GetCleanName())
	end
end
```

### DIFF
--- a/zone/client.h
+++ b/zone/client.h
@@ -1166,27 +1166,16 @@ public:
 	{
 		if (task_state) { task_state->UpdateTasksOnTouch(this, dz_switch_id); }
 	}
-	inline void TaskSetSelector(Mob *mob, int task_set_id)
+	inline void TaskSetSelector(Mob* mob, int task_set_id, bool ignore_cooldown)
 	{
-		if (task_manager) {
-			task_manager->TaskSetSelector(
-				this,
-				task_state,
-				mob,
-				task_set_id
-			);
+		if (task_manager && task_state) {
+			task_manager->TaskSetSelector(this, mob, task_set_id, ignore_cooldown);
 		}
 	}
-	inline void TaskQuestSetSelector(Mob *mob, int count, int *tasks)
+	inline void TaskQuestSetSelector(Mob* mob, const std::vector<int>& tasks, bool ignore_cooldown)
 	{
-		if (task_manager) {
-			task_manager->TaskQuestSetSelector(
-				this,
-				task_state,
-				mob,
-				count,
-				tasks
-			);
+		if (task_manager && task_state) {
+			task_manager->TaskQuestSetSelector(this, mob, tasks, ignore_cooldown);
 		}
 	}
 	inline void EnableTask(int task_count, int *task_list)
--- a/zone/embparser_api.cpp
+++ b/zone/embparser_api.cpp
@@ -1088,17 +1088,32 @@ void Perl__taskselector(perl::array task_ids)
 		throw std::runtime_error(fmt::format("Exceeded max number of task offers [{}]", MAXCHOOSERENTRIES));
 	}
 
-	int tasks[MAXCHOOSERENTRIES];
+	std::vector<int> tasks;
 	for (int i = 0; i < task_ids.size(); ++i)
 	{
-		tasks[i] = task_ids[i];
+		tasks.push_back(task_ids[i]);
 	}
-	quest_manager.taskselector(static_cast<int>(task_ids.size()), tasks);
+	quest_manager.taskselector(tasks);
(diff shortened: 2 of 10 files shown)


zone/lua_parser_events.cpp
```diff
@@ -210,18 +210,30 @@ void handle_npc_death(QuestInterface *parse, lua_State* L, NPC* npc, Mob *init,
...
-		l_spell_o.push(L);
-		lua_setfield(L, -2, "spell");
-	} else {
-		Lua_Spell l_spell(nullptr);
-		luabind::adl::object l_spell_o = luabind::adl::object(L, l_spell);
-		l_spell_o.push(L);
-		lua_setfield(L, -2, "spell");
-	}
+
+        Seperator sep(data.c_str());
```
hgtw, line 214: Need to use tabs for indent on all this instead of spaces

Kinglykrab: Agreed.

zone/lua_parser_events.cpp
```diff
@@ -210,18 +210,30 @@ void handle_npc_death(QuestInterface *parse, lua_State* L, NPC* npc, Mob *init,
...
+        }
+
+        lua_pushinteger(L, std::stoi(sep.arg[1]));
+        lua_setfield(L, -2, "caster_id");
+
+        lua_pushinteger(L, std::stoi(sep.arg[2]));
+        lua_setfield(L, -2, "caster_level");
+
+
+
```
Kinglykrab, line 236: Unnecessary white space.

hgtw:
Wonder if it would be better to just pass a caster object as "other" like other events. Since this handler also handles self-cast events nullptr is passed when there is no other Mob though, so would have to either pass a separate caster object that requires scripts to check validity or change the handlers to pass the same npc for both self and other.

JasXSL:
> Wonder if it would be better to just pass a caster object as "other" like other events. Since this handler also handles self-cast events nullptr is passed when there is no other Mob though, so would have to either pass a separate caster object that requires scripts to check validity or change the handlers to pass the same npc for both self and other.

This would be the preferred solution for me at least. So long as it's both for the pc and npc event. I have no idea how to do it though.

Akkadius:
Pending feedback

Kinglykrab:
> > Wonder if it would be better to just pass a caster object as "other" like other events. Since this handler also handles self-cast events nullptr is passed when there is no other Mob though, so would have to either pass a separate caster object that requires scripts to check validity or change the handlers to pass the same npc for both self and other.
> 
> This would be the preferred solution for me at least. So long as it's both for the pc and npc event. I have no idea how to do it though.

There are plenty of examples: https://github.com/EQEmu/Server/blob/6f7fa989969df8e74bca972d4c3fa0af1609845d/zone/lua_parser_events.cpp#L126

hgtw:
I haven't tested it but for players you'll probably have to copy some code that manually adds an `other` object to the export strings for everywhere `parse->EventPlayer(EVENT_CAST_BEGIN`, `parse->EventPlayer(EVENT_CAST`, and `parse->EventPlayer(EVENT_CAST_ON` are called since player events don't have the additional Mob argument.

For NPCs I think you'd want to change:
https://github.com/EQEmu/Server/blob/efe187911557c2bab07c626451994c14043b1523/zone/spells.cpp#L258
to `parse->EventNPC(EVENT_CAST_BEGIN, CastToNPC(), this, export_string, 0);`

and

https://github.com/EQEmu/Server/blob/efe187911557c2bab07c626451994c14043b1523/zone/spells.cpp#L1672
to `parse->EventNPC(EVENT_CAST, CastToNPC(), this, export_string, 0);`

Then in `handle_npc_cast` adding an `other` export based on the `Mob* init` argument like other events.

Again this is all untested

JasXSL:
I've gotten event.other to work well with EVENT_CAST_ON, but the are some issues with using it with EVENT_CAST_BEGIN and EVENT_CAST. It works great for targeted spells. But for self buffs, target_id will be your active target, so that'd need to be handled differently. And then there's AoE which I haven't got a clue how to handle since there can be multiple targets.

PR #2403, opened 2022-09-01
[Bug Fix] Resolve issues with Client::SetHideMe

mackal:
Arg was shadowing the member variable and we weren't updating the member
variable


Kinglykrab (approved):
👍 Looks like a ripe one to me.

PR #2404, opened 2022-09-01
[Bug Fix] Resolve logic error in Raid::QueueClients

mackal:
We're checking if they're groupless here, so this should be correct now


Kinglykrab (approved):
Spicy meatball. 👍

Akkadius (approved):

PR #2408, opened 2022-09-02
[Feature] Faction Association

mackal:
See #2355 for previous discussion

This is based on theories that have been formulated by several people about how factions are actually linked.

You get hit in one faction and you "splash" the others. Some "FactionQuesters" called this sort of thing splash faction, but I went with Faction Association so the tables can start with faction and these are associated factions.

From the QuestReward packet we know they just use the faction ID and the value that faction is hit with, the rest follow, ALWAYS in the same ratio (sometimes hard to see with lower values since they min cap to abs(1) for a change)

This should be 100% compatible with current system and just be something to opt in.

This will allow us to use QuestReward to handle pretty much everything, exactly like live does :)


hgtw (approved):

shared_memory/main.cpp
```diff
@@ -300,7 +306,17 @@ int main(int argc, char **argv)
 		}
 	}
 
-	if (load_all || load_bd) {
+	if(load_all || load_faction_assoc) {
```
Akkadius, line 309: Small:

`if(load_all` vs `if (load_all`

zone/lua_client.cpp
```diff
@@ -1761,9 +1766,27 @@ void Lua_Client::QuestReward(Lua_Mob target, luabind::adl::object reward) {
...
-		try {
-			faction = luabind::object_cast<bool>(cur);
-		} catch (luabind::cast_failed &) {
+		// if it's a table it will be {faction, faction_mod}
+		if (luabind::type(cur) == LUA_TTABLE) {
+			auto item = cur[1];
+			if (luabind::type(item) != LUA_TNIL) {
+				try {
+					quest_reward.faction = luabind::object_cast<uint32>(item);
+				} catch (luabind::cast_failed &) {
```
Akkadius, line 1775: Could pipe these errors to `LogQuests`

zone/gm_commands/faction_association.cpp
```diff
@@ -0,0 +1,17 @@
...
+{
+	if (sep->argnum < 2 || !sep->IsNumber(1) || !sep->IsNumber(2)) {
+		c->Message(Chat::White, "Usage: #factionassociation <factionid> <amount>");
+		c->Message(Chat::White, "  Defaults to self, unless target is also a client.");
+		return;
+	}
+
+	// default to self unless target is also a client
+	auto target = c;
+	if (c->GetTarget() && c->GetTarget()->IsClient())
```
Akkadius, line 13: Would really prefer if we just close these. There's a few other places of implied brackets

Akkadius (approved):
Looks great. Love the new log category, use of repositories, looks like you got it added to content database in schema and in connection pointer. Using repositories in the loader, provided data backfill and regened related repositories. A+

Few minor comments.

Also thanks for being receptive to feedback with the prior PR

shared_memory/main.cpp
mackal: yeah, I just copied load_bd branch, which doesn't match with the rest of the code :P

zone/lua_client.cpp
mackal: See @hgtw comment on this code in the previous PR.

PR #2412, opened 2022-09-03
[Zoning] Fix zoning logic issues

Akkadius:
This PR addresses both issues that I introduced from https://github.com/EQEmu/Server/pull/2388 and furthermore cleans up around the originally refactored areas

**Issue**

Zone flags were getting corrupted in c-string format, usage of those c-strings have been removed for the more modern std::string container and related checks have been adjusted accordingly. Because it was corrupted, zones always thought they had zone flags to check and clients always did not meet zone requirements. This coupled with another long time issue addressed in this PR was causing clients to go an infinite loop, looping between arena and the zone they didn't meet requirements for.

**Resolution**

This infinite loop was caused by a race condition of world not entirely being up to date on where the client has actually zoned so we end up in this waffling state. The condition for which a client does not meet requirements and is sent to arena is moved later in the client connect process so they can be properly and safely zoned out

**Other Changes**

* Added logging statements to all zone store getters in cases where we fail to find a zone (fall-through) indicating a problem that should never happen. This will make subsequent troubleshooting easier if we find edge cases that trigger these messages
* Added more comprehensive zoning logging in the `Info` category in many other places in the source
* Fixed regression in GMZone struct logic
* Simplify expansion gating zoning logic to leverage the in-memory zone store logic

### DIFF
--- a/common/zone_store.cpp
+++ b/common/zone_store.cpp
@@ -59,6 +59,8 @@ uint32 ZoneStore::GetZoneID(std::string zone_name)
 		}
 	}
 
+	LogInfo("[GetZoneID] Failed to get zone_name [{}]", zone_name);
+
 	return 0;
 }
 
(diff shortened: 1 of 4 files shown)


KimLS (approved):
Don't see anything wrong, getting rid of strcpys is always good.

PR #2413, opened 2022-09-03
[Character Starting Points] Fix headings data

Akkadius:
Fix data regressions caused by https://github.com/EQEmu/Server/commit/a32dedeb48b5f90e9f0d16e1a0c8e90ce0d2d933 that didn't fully and retroactively backfill existing data causing things to appear incorrectly in game.

This PR addresses a data backfill specifically around character starting points

For example when I create a Shaman I end up pointing towards a wall.

**Data Source**

Using a PEQ database backup from 2018 before the headings change to build and retroactively apply the correct math to our data and use data signatures to apply it precisely only to those entries in everyone's database.

**Data Impact**

This change updated 92 starting points out of 319

**Data Target (Signatures)**

The way the retroactive back-fill will run is by very specific keys, meaning you need to match exact x,y,z, player_choice, player_class, player_deity, player_race, start_zone for this content to update your database which should not touch any custom implemented starting points.

Below is an example of being pointed at the wall when starting with a Barbarian Shaman and correctly facing the guild masters afterwards

**Before**

image

**After**

image


Akkadius:
This PR will need to rebase database versioning along with others that are open depending on what order things get approved / merged

hgtw (approved):

mackal (approved):

Kinglykrab (approved):

PR #2415, opened 2022-09-03
[Zone Points] Fix zone point heading data

Akkadius:
Fix data regressions caused by https://github.com/EQEmu/Server/commit/a32dedeb48b5f90e9f0d16e1a0c8e90ce0d2d933 that didn't fully and retroactively backfill existing data causing things to appear incorrectly in game.

This PR addresses a data backfill specifically around `zone_points` table via `target_heading`

**PEQ Data from 2018**

This based off of ProjectEQ backup data from 2018 prior to the above PR, so we're not basing it off of data that was 1-off changed after the 2018 change.

**Data Impact**

This change 1265 zone points updated out of 1779

252 had 0 heading already, which means that 262 were updated or set manually since 2018

**Data Target (Signatures)**

Only zone_points that are the same `target_heading` as they were in 2018 are affected, as well as content signatures that show the entry is uniquely what its supposed to be without messing up custom or manually written data.

Below is an example of a wrong zone point heading coming from **everfrost** into **blackburrow**

**Before**

image

**After**

image


Akkadius:
This PR will need to rebase database versioning along with others that are open depending on what order things get approved / merged

hgtw (approved):

mackal (approved):

Kinglykrab (approved):

PR #2421, opened 2022-09-03
[Clean up] Tweak object name checking item's IDFile

mackal:
Clang was complaining about these

Checking if first char is null should be faster than strlen


Kinglykrab (approved):
👍

PR #2423, opened 2022-09-04
[Feature] AA Cap Limit

fryguy503:
Will force unused AA points to the cap if they exceed the cap. Will also force AA Percentage to 0%. NOTE: The variable, UnusedAAPointCap, should NOT be lowered once implemented without first checking the DB for how many Unused AAs people have.  The next time they gain any EXP, it WILL drop them to cap.  Also, PCs should be given warning prior to this patch going to Live, to ensure they have a chance to spend any AAs above the cap, else they will lose them.

### DIFF
--- a/zone/exp.cpp
+++ b/zone/exp.cpp
@@ -505,7 +505,7 @@ void Client::AddEXP(uint32 in_add_exp, uint8 conlevel, bool resexp) {
 	uint32 exp = 0;
 	uint32 aaexp = 0;
 
-	if (m_epp.perAA<0 || m_epp.perAA>100) {
+	if (m_epp.perAA < 0 || m_epp.perAA > 100) {
 		m_epp.perAA = 0;    // stop exploit with sanity check
 	}
 
@@ -546,7 +546,7 @@ void Client::AddEXP(uint32 in_add_exp, uint8 conlevel, bool resexp) {
 	// Doing this here means potentially one kill wasted worth of experience, but easiest to put it here than to rewrite this function.
 	int aa_cap = RuleI(AA, UnusedAAPointCap);
 
-	if (aa_cap > 0 && aaexp > 0) {
+	if (aa_cap >= 0 && aaexp > 0) {
 		if (m_pp.aapoints == aa_cap) {
 			MessageString(Chat::Red, AA_CAP);
 			aaexp = 0;
--- a/zone/exp.cpp
+++ b/zone/exp.cpp
@@ -544,14 +544,18 @@ void Client::AddEXP(uint32 in_add_exp, uint8 conlevel, bool resexp) {
 
 	// Check for Unused AA Cap.  If at or above cap, set AAs to cap, set aaexp to 0 and set aa percentage to 0.
 	// Doing this here means potentially one kill wasted worth of experience, but easiest to put it here than to rewrite this function.
-	if (aaexp > 0) {
-		int aa_cap = RuleI(AA, UnusedAAPointCap);
+	int aa_cap = RuleI(AA, UnusedAAPointCap);
+
+	if (aa_cap > 0 && aaexp > 0) {
+		char val1[20] = {0};
+		char val2[20] = {0};
+		
 		if (m_pp.aapoints == aa_cap) {
 			MessageString(Chat::Red, AA_CAP);
 			aaexp = 0;
 			m_epp.perAA = 0;
 		} else if (m_pp.aapoints > aa_cap) {
-			MessageString(Chat::Red, OVER_AA_CAP);
+			MessageString(Chat::Red, OVER_AA_CAP, ConvertArray(aa_cap, val1), ConvertArray(aa_cap, val2));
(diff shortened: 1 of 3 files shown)


Kinglykrab:
I would set the default to -1 and check if the rule value is greater than 0 when checking if you should use the cap. As-is, the cap is force enabled on any server that receives this update and has to be manually disabled.

zone/exp.cpp
```diff
@@ -541,6 +542,22 @@ void Client::AddEXP(uint32 in_add_exp, uint8 conlevel, bool resexp) {
...
+		if (m_pp.aapoints == aa_cap) {
+			MessageString(Chat::Red, AA_CAP);
+			aaexp = 0;
+			m_epp.perAA = 0;
+		} else if (m_pp.aapoints > aa_cap) {
+			MessageString(Chat::Red, OVER_AA_CAP);
+			m_pp.aapoints = aa_cap;
+			aaexp = 0;
+			m_epp.perAA = 0;
+		}
```
Kinglykrab, line 558: If both trees set these 2 variables to 0, can have the setters outside of either condition specifically.

zone/string_ids.h
```diff
@@ -295,6 +296,7 @@
 #define ALREADY_SHIELDING			3280	//Either you or your target is already shielding another.
 #define START_SHIELDING				3281	//%1 begins to use %2 as a living shield!
 #define END_SHIELDING				3282	//%1 ceases protecting %2.
+#define OVER_AA_CAP					3374	//Warning: You are currently over the earned Advancement point limit of %1. Please spend some of your stored AA points. The NEXT time you zone all of your AA points over %2 will be deleted. You MUST spend the extra points now.
```
Kinglykrab, line 299: String uses %1 and %2, but nothing is sent along with string when the message is sent using this.

mackal:
AAs gained via quests should ignore the cap (epic 1.5/2.0 come to mind)


Live uses a formula for max banked AAs, would be nice to have support for that as well.

zone/exp.cpp
fryguy503: Not sure that would work as outside the if/elseif they may be within the correct range.  I could completely flip the logic so its checking if you are over the AA Cap prior to checking if you have aaexp enabled, but I figured that would create needless checks. This way if a player is reimbursed AA's by a GM and had their AAEXP also set to 0 it wouldn't trigger the wipe. Its messy for sure.

zone/exp.cpp
```diff
@@ -541,6 +542,26 @@ void Client::AddEXP(uint32 in_add_exp, uint8 conlevel, bool resexp) {
...
+	if (aa_cap > 0 && aaexp > 0) {
+		char val1[20] = {0};
+		char val2[20] = {0};
+		
+		if (m_pp.aapoints == aa_cap) {
+			MessageString(Chat::Red, AA_CAP);
+			aaexp = 0;
+			m_epp.perAA = 0;
+		} else if (m_pp.aapoints > aa_cap) {
+			MessageString(Chat::Red, OVER_AA_CAP, ConvertArray(aa_cap, val1), ConvertArray(aa_cap, val2));
```
hgtw, line 558: `MessageString(Chat::Red, OVER_AA_CAP, fmt::format_int(aa_cap).c_str(), fmt::format_int(aa_cap).c_str());`
Would probably be better so you don't need the stack buffers

fryguy503: Thank you! I was chatting with Mackal about it, because I hated the way the ConvertArray looked.

Akkadius:
Another one almost a month now, @hgtw @Kinglykrab what is the consensus here

zone/exp.cpp
```diff
@@ -541,6 +542,23 @@ void Client::AddEXP(uint32 in_add_exp, uint8 conlevel, bool resexp) {
 		aaexp = had_aaexp;	//watch for wrap
 	}
 
+	// Check for Unused AA Cap.  If at or above cap, set AAs to cap, set aaexp to 0 and set aa percentage to 0.
+	// Doing this here means potentially one kill wasted worth of experience, but easiest to put it here than to rewrite this function.
+	int aa_cap = RuleI(AA, UnusedAAPointCap);
+
+	if (aa_cap > 0 && aaexp > 0) {
```
Kinglykrab, line 549: >= 0 to indicate -1 being the disabled value and 0 being the first valid value.

zone/exp.cpp
```diff
@@ -505,8 +505,9 @@ void Client::AddEXP(uint32 in_add_exp, uint8 conlevel, bool resexp) {
 	uint32 exp = 0;
 	uint32 aaexp = 0;
 
-	if (m_epp.perAA<0 || m_epp.perAA>100)
-		m_epp.perAA=0;	// stop exploit with sanity check
+	if (m_epp.perAA<0 || m_epp.perAA>100) {
```
Kinglykrab, line 508: Formatting.

Kinglykrab (approved):
Small comments. 👍

PR #2424, opened 2022-09-05
[Zoning] Possible zoning under world fix

Akkadius:
Force resending of position after client actually connects since the issue appears to be related to a race condition of packets and the client puts itself under the world, lets bring it back to where it should be (hopefully)

This might need additional investigation but is worth testing on PEQ test server

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -918,6 +918,10 @@ void Client::CompleteConnect()
 		GoToSafeCoords(ZoneID("arena"), 0);
 		return;
 	}
+
+	// force resending of position when finally complete
+	// in hopes to alleviating race condition under-world issues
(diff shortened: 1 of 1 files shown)


Akkadius:
Issue conversation here https://discord.com/channels/212663220849213441/212684816104030209/1012424374814507038

Akkadius:
Need someone to at least test and approve that you don't go underworld when zoning to places

Kinglykrab:
Will take a look this afternoon. Need to create a good box crew on PEQ or something.

Akkadius:
Not sure that we necessarily need a box crew - just that we don't legitimately break zoning for a single box. I can test with PEQ's crazy crews

Kinglykrab:
> Not sure that we necessarily need a box crew - just that we don't legitimately break zoning for a single box. I can test with PEQ's crazy crews

Sounds good.

Kinglykrab (approved):
Zoned around to a bunch of random zones, seemed fine. :+1:

PR #2428, opened 2022-09-07
[Feature] Update HateMod used by SPA 114 to Int32.

Aeadoin:
HateMod was previously Int8, which would cause overflow issues when using newer spells from Live.


mackal (approved):

Kinglykrab (approved):
👍

PR #2430, opened 2022-09-11
[Pathing] Fix pathing z-correctness for certain models

Akkadius:
Issue originally reported in #support-pathing (Discord) by Symetrix.

**Reproduction**

Pull Xygoz in Veeshan's Peek up/down steep hill from spawn and he will clip in and out of the floor

12b2a5043f8869ded889b54c552a5d79

**Problem(s)**

1. Xygoz is a fixed race model (5 size is default) but the server data has him set to 1
2. Dragon race 49 is missing proper Z offsets for its model

**Solution(s)**

* Updated GetZOffset to include remaining models from `Resources/SPOffsets.ini`
* Set static race dragon (49) to always be size 5 regardless of what is set in the database

Both of these solutions combined result in a dragon that does not clip through the floor under steep terrain and calculates Z positioning correctly with the server

### DIFF
--- a/zone/npc.cpp
+++ b/zone/npc.cpp
@@ -146,7 +146,9 @@ NPC::NPC(const NPCType *npc_type_data, Spawn2 *in_respawn, const glm::vec4 &posi
 
 	swarm_timer.Disable();
 
-	if (size <= 0.0f) {
+	// lava dragon is a fixed size model and should always use its default
+	// otherwise pathing issues
+	if (size <= 0.0f || race == RACE_LAVA_DRAGON_49) {
 		size = GetRaceGenderDefaultHeight(race, gender);
 	}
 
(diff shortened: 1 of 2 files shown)


zone/npc.cpp
```diff
@@ -150,6 +150,12 @@ NPC::NPC(const NPCType *npc_type_data, Spawn2 *in_respawn, const glm::vec4 &posi
 		size = GetRaceGenderDefaultHeight(race, gender);
 	}
 
+	// lava dragon is a fixed size model and should always use its default
+	// otherwise pathing issues
+	if (race == RACE_LAVA_DRAGON_49) {
+		size = 6;
```
Kinglykrab, line 156: Should this be 5 since you said size is 5?

Kinglykrab (approved):
👍

mackal (approved):

PR #2437, opened 2022-09-20
[Bug Fix] Fix Spellinfo Command to work with SpellIDs above int16

### LINKED ISSUE #2320: command_spellinfo broken for spells with an ID above a signed 16 bit int
Using spellinfo with an ID above 32767 (0b111111111111111) fails. I'm assuming it's because you're using atoi instead of strtoul
 here
https://github.com/EQEmu/Server/blob/0550fcfd3f4757fd7f73d56b9743525f97e5c2da/zone/gm_commands/spellinfo.cpp#L9

Aeadoin:
Fixes #2320


mackal:
Should be signed int instead. `int32_t`

mackal (approved):

Kinglykrab (approved):

PR #2440, opened 2022-09-22
[File Paths] Implement Path Manager

Akkadius:
This PR implements a unified file path manager within the server.

Before, we had paths relatively referenced all of the place, hard coded and inconsistent. Lots of if/else logic all over the source building paths when it should have been unified somewhere else. 

We were building paths in edges all over the code. We also cannot directly run binaries when inside the `bin` directory, while not a huge issue - is made possible by having a proper and centralized path management to reference.

Having a centralized server path management utility unifies how we deal with paths. If we need to change behavior down albeit rarely the line its done in one place. If we have to add new paths, its done in one place, etc.

Almost every reference to reading and writing files within the source also was swapped out to anchor to the new helpers which includes but not limited to.

* EQEmu Config (eqemu_config.json)
* Export client files
* Import client files
* Loginserver config (login.json)
* Logs
* Lua Mods
* Lua Modules
* Map files
* Nav files
* Opcodes
* Plugins
* Quests
* Water files
* eqemu_server.pl

**Helper example(s)**

```cpp
	[[nodiscard]] const std::string &GetLogPath() const;
	[[nodiscard]] const std::string &GetLuaModsPath() const;
	[[nodiscard]] const std::string &GetLuaModulesPath() const;
	[[nodiscard]] const std::string &GetMapsPath() const;
	[[nodiscard]] const std::string &GetPatchPath() const;
	[[nodiscard]] const std::string &GetPluginsPath() const;
	[[nodiscard]] const std::string &GetQuestsPath() const;
	[[nodiscard]] const std::string &GetServerPath() const;
	[[nodiscard]] const std::string &GetSharedMemoryPath() const;
```

**Usage Example(s)**

```cpp
// first look for /quests/zone/npcid.ext (precedence)
filename = fmt::format("{}/{}/{}", path.GetQuestsPath(), zone->GetShortName(), npcid);
```

```cpp
std::string file = fmt::format(
	"{}/{}",
	(file_path.empty() ? path.GetServerPath() : file_path),
	EQEmuConfig::ConfigFile
);
```

**Testing**

All testing done on Linux

* Logged into zone, quests worked, no #questerrors
* Booted `zone` in context of base server directory in and in `bin` folder
* Booted `world` in context of base server directory in and in `bin` folder
* Booted `ucs` in context of base server directory in and in `bin` folder
* Booted `loginserver` in context of base server directory in and in `bin` folder
* Booted `queryserv` in context of base server directory in and in `bin` folder
* Ran `shared_memory` in context of base server directory in and in `bin` folder
* Ran `import_client_files` in context of base server directory in and in `bin` folder
* Ran `export_client_files` in context of base server directory in and in `bin` folder

Need testing done in Windows

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -15,7 +15,7 @@ volumes:
 steps:
   - name: server-build
     # Source build script https://github.com/Akkadius/akk-stack/blob/master/containers/eqemu-server/Dockerfile#L20
-    image: akkadius/eqemu-server:latest
+    image: akkadius/eqemu-server:v11
     commands:
       - sudo chown eqemu:eqemu /drone/src/ * -R
       - sudo chown eqemu:eqemu /home/eqemu/.ccache/ * -R
--- a/.drone.yml
+++ b/.drone.yml
@@ -19,9 +19,9 @@ steps:
     commands:
       - sudo chown eqemu:eqemu /drone/src/ * -R
       - sudo chown eqemu:eqemu /home/eqemu/.ccache/ * -R
-      - git submodule init && git submodule update && mkdir -p build && cd build && cmake -DEQEMU_BUILD_LOGIN=ON -DEQEMU_ENABLE_BOTS=ON -DEQEMU_BUILD_LUA=ON -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_FLAGS_RELWITHDEBINFO:STRING="-O0 -g -DNDEBUG" -G 'Unix Makefiles' .. && make -j$((`nproc`-4))
+      - git submodule init && git submodule update && mkdir -p build && cd build && cmake -DEQEMU_BUILD_TESTS=ON -DEQEMU_BUILD_LOGIN=ON -DEQEMU_ENABLE_BOTS=ON -DEQEMU_BUILD_LUA=ON -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_FLAGS_RELWITHDEBINFO:STRING="-O0 -g -DNDEBUG" -G 'Unix Makefiles' .. && make -j$((`nproc`-4))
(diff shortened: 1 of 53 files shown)


mackal:
Why not use std::filesystem for more things?

Akkadius:
> Why not use std::filesystem for more things?

Actually wanted to originally. Opted to see if I could avoid it since a lot of people don't quite have GCC up to the level that it doesn't require either the experimental include or the flag being passed to the compiler. 

After HG tested on Windows I decided to go that route anyways to keep it all cross platform clean

Akkadius:
Windows working now as well. Does need some in game testing

image

Akkadius:
Build failing because GCC on Debian 10 doesn't have support by default. Image will need to be upgraded for Drone

mackal:
> Build failing because GCC on Debian 10 doesn't have support by default. Image will need to be upgraded for Drone

butts, I just double checked what PEQ was using and hoped we'd be all good for that suggestion :P

Kinglykrab (approved):
**Windows 10 Testing**
- Logged into zone, quests worked, no #questerrors
- Booted `zone` in context of base server directory in and in `bin` folder
- Booted `world` in context of base server directory in and in `bin` folder
- Booted `ucs` in context of base server directory in and in `bin` folder
- Booted `loginserver` in context of base server directory in and in `bin` folder
- Booted `queryserv` in context of base server directory in and in `bin` folder
- Ran `shared_memory` in context of base server directory in and in `bin` folder
- Ran `import_client_files` in context of base server directory in and in `bin` folder
- Ran `export_client_files` in context of base server directory in and in `bin` folder

Akkadius:
Thanks for the help testing. Not in a rush to merge this as this will break a lot of people. Trying to gives folks a heads up to get their things updated to support C++17

This includes EZ / PEQ

PR #2444, opened 2022-09-23
[Bug Fix] Fix Swarm Pet Flurry/Rampages Messages

Aeadoin:
Flurry & Rampage messages will now filter to proper chat window (Pet Flurry/Rampage filter) instead of using NPC Flurry/Rampage chat window filtering.

### DIFF
--- a/zone/mob_ai.cpp
+++ b/zone/mob_ai.cpp
@@ -2113,27 +2113,27 @@ bool Mob::Flurry(ExtraAttackOptions *opts)
 {
 	// this is wrong, flurry is extra attacks on the current target
 	Mob *target = GetTarget();
-	if (target) {
-		if (!IsPet()) {
-			entity_list.MessageCloseString(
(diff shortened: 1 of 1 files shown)


zone/mob_ai.cpp
```diff
@@ -2113,27 +2113,27 @@ bool Mob::Flurry(ExtraAttackOptions *opts)
...
-				this,
-				true,
-				200,
-				Chat::PetFlurry,
-				NPC_FLURRY,
-				GetCleanName(),
-				target->GetCleanName());
-		}
-
+    if (target) {
```
mackal, line 2116: These should be tabs, not spaces.

Aeadoin: Should hopefully be fixed now

Kinglykrab:
👍

mackal (approved):

PR #2453, opened 2022-09-28
[Logging] Add stack trace in code paths that shouldn't occur

Akkadius:
**Issue**

Folks are seeing 

```
[World] [Info] [GetZoneName] Failed to get zone name by zone_id [0] error_unknown [false]
[World] [Info] [GetZoneName] Failed to get zone name by zone_id [0] error_unknown [false]
[World] [Info] [GetZoneName] Failed to get zone name by zone_id [0] error_unknown [false]
[World] [Info] [GetZoneName] Failed to get zone name by zone_id [0] error_unknown [false]
```

Which are logs added recently to trap logic that shouldn't be occuring in the first place. We should not see this function get called with a `zone_id` of 0

Since this is called everywhere, we need to be able to see the stack, caller. Added https://github.com/bombela/backward-cpp cross-platform header library to print the stack when this log is produced to give more meaningful information.

image

Also fixed an issue caused by #2129 where the function parameter shadows a local variable creating issues

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -650,6 +650,7 @@ SET(common_headers
     patches/uf_limits.h
     patches/uf_ops.h
     patches/uf_structs.h
+    stacktrace/backward.hpp
     StackWalker/StackWalker.h
     util/memory_stream.h
     util/directory.h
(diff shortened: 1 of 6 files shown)


Kinglykrab (approved):

Akkadius:
Windows testing, had to add a workaround per https://github.com/bombela/backward-cpp/issues/206

image

PR #2463, opened 2022-10-01
[Feature] Add Support for "Show Mine Only" Filters

Aeadoin:
This adds support for the "Show Only Mine" Filter to function properly with the following Filter Types:

Spell Damage/Non-Melee Hits
Critical Melee Hits
Spell Crits.

I also noticed that DoT Damage was now working for "Show Group Only" where as I believe it was previously failing. 
DoT Damage could already be filtered with "Show Only Mine" I believe.

Tested commits by setting up two characters, and going through all possible filters for the above fields.

I'll raise a new PR if I find any issues with the other filter mode "Show Group Only" (that wasn't supported)

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -4050,12 +4050,14 @@ void Mob::CommonDamage(Mob* attacker, int64 &damage, const uint16 spell_id, cons
 							attacker->MessageString(Chat::DamageShield, OTHER_HIT_NONMELEE, GetCleanName(), ConvertArray(damage, val1));
 					}
 					else {
-						entity_list.MessageCloseString(
-							this, /* Sender */
-							true, /* Skip Sender */
+						entity_list.FilteredMessageCloseString(
+							attacker, /* Sender */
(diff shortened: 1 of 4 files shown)


Kinglykrab (approved):

zone/attack.cpp
```diff
@@ -4231,7 +4233,7 @@ void Mob::HealDamage(uint64 amount, Mob *caster, uint16 spell_id)
 						YOU_HEAL, GetCleanName(), itoa(acthealed));
 			}
 		}
-		else {
+		else if (this->CastToClient()->GetFilter(FilterHealOverTime) != (FilterShowSelfOnly || FilterHide)) {
```
Akkadius, line 4236: Remove `this->`

Akkadius (approved):

zone/attack.cpp
Aeadoin: I'm going to have to close this PR, I believe I had deleted the original branch. I've cloned it now and will submit a new PR with the change.

PR #2466, opened 2022-10-02
[Hate] Cleaned up Hate Override Logic

Aeadoin:
This cleans up logic for the field named "HateAdded" this field is an override for the amount of hate generated by casting a spell. Before we were doing this check after a ton of other calculations which were just being thrown away.


Akkadius:
Hard to grok what is actually changed with all of the formatting / indent changes. Would recommend splitting up formatting changes from logic changes

Aeadoin:
Will review, and resubmit if needed.

PR #2467, opened 2022-10-02
[Feature] Soft Delete Bots on Character Soft Delete

Aeadoin:
Only added support for Soft Deletes if the Character Soft Deletes Rule is used. Would have been too much effort currently to add support for Hard Deletes.

### DIFF
--- a/common/database.cpp
+++ b/common/database.cpp
@@ -412,23 +412,40 @@ bool Database::DeleteCharacter(char *character_name) {
 
 		QueryDatabase(query);
 
-		return true;
+	} else {
+
+		for (const auto& iter : DatabaseSchema::GetCharacterTables()) {
+			std::string table_name               = iter.first;
+			std::string character_id_column_name = iter.second;
+
+			QueryDatabase(fmt::format("DELETE FROM {} WHERE {} = {}", table_name, character_id_column_name, character_id));
+		}
+		LogInfo("DeleteCharacter | Character [{}] ({}) is being [{}]", character_name, character_id, delete_type);
 	}
 
 	LogInfo("DeleteCharacter | Character [{}] ({}) is being [{}]", character_name, character_id, delete_type);
 
-	for (const auto& iter : DatabaseSchema::GetCharacterTables()) {
-		std::string table_name               = iter.first;
-		std::string character_id_column_name = iter.second;
-
-		QueryDatabase(fmt::format("DELETE FROM {} WHERE {} = {}", table_name, character_id_column_name, character_id));
-	}
-
 #ifdef BOTS
 	query = StringFormat("DELETE FROM `guild_members` WHERE `char_id` = '%d' AND GetMobTypeById(%i) = 'C'", character_id); // note: only use of GetMobTypeById()
 	QueryDatabase(query);
-#endif
 
+	if (RuleB(Character, SoftDeletes)) {
+		std::string query = fmt::format(
+			SQL(
+				UPDATE
+				bot_data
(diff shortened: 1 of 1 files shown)


common/database.cpp
```diff
@@ -412,23 +412,40 @@ bool Database::DeleteCharacter(char *character_name) {
 
 		QueryDatabase(query);
 
-		return true;
```
Akkadius, line 415: Instead of removing the return here, just inline the bots preprocessor in the existing `if (RuleB(Character, SoftDeletes)) {`

common/database.cpp
```diff
@@ -412,23 +412,40 @@ bool Database::DeleteCharacter(char *character_name) {
...
+				bot_data
+				SET
+				name = SUBSTRING(CONCAT(name, '-deleted-', UNIX_TIMESTAMP()), 1, 64)
+				WHERE
+				owner_id = '{}'
+			),
+			character_id
+		);
+		QueryDatabase(query);
+		LogInfo("DeleteCharacterBOTS | CharacterBOTS [{}] ({}) is being [{}]", character_name, character_id, delete_type);
```
Akkadius, line 445: Suggest something like

`[DeleteCharacter] Bots soft deletion name [{}] id [{}] delete_type [{}]`

common/database.cpp
```diff
@@ -412,23 +412,40 @@ bool Database::DeleteCharacter(char *character_name) {
...
 	}
 
+	LogInfo("DeleteCharacter | Character [{}] ({}) is being [{}]", character_name, character_id, delete_type);
+
 #ifdef BOTS
 	query = StringFormat("DELETE FROM `guild_members` WHERE `char_id` = '%d' AND GetMobTypeById(%i) = 'C'", character_id); // note: only use of GetMobTypeById()
 	QueryDatabase(query);
-#endif
 
+	if (RuleB(Character, SoftDeletes)) {
```
Akkadius, line 432: Per previous comment, inline with the other check instead of doing the if/else

Akkadius:
Not sure I see the spot where bots actually get hard deleted

common/database.cpp
Aeadoin: Changed to what you suggested.

Aeadoin:
> Not sure I see the spot where bots actually get hard deleted

Correct, I only added logic for Soft Deletes with this PR.

Hard deletes would need us to deal with a multitude of foreign key constraints 
   
TABLE_NAME | COLUMN_NAME | CONSTRAINT_NAME | REFERENCED_TABLE_NAME | REFERENCED_COLUMN_NAME
-- | -- | -- | -- | --
bot_buffs | bot_id | FK_bot_buffs_1 | bot_data | bot_id
bot_group_members | bot_id | FK_bot_group_members_2 | bot_data | bot_id
bot_groups | group_leader_id | FK_bot_groups_1 | bot_data | bot_id
bot_heal_rotation_members | bot_id | FK_bot_heal_rotation_members_2 | bot_data | bot_id
bot_heal_rotations | bot_id | FK_bot_heal_rotations | bot_data | bot_id
bot_inventories | bot_id | FK_bot_inventories_1 | bot_data | bot_id
bot_pets | bot_id | FK_bot_pets_1 | bot_data | bot_id
bot_stances | bot_id | FK_bot_stances_1 | bot_data | bot_id
bot_timers | bot_id | FK_bot_timers_1 | bot_data | bot_id

Akkadius (approved):

Akkadius (approved):

PR #2468, opened 2022-10-03
Change SpellTypes Enum to be Bitmask values.

Aeadoin:
Changed to use bitmask vs bitshift to help with readability, and no point in the additional math being done.


mackal:
I don't think this really helps. It also makes adding new ones more difficult.

Aeadoin:
> I don't think this really helps. It also makes adding new ones more difficult.

That's fair!

PR #2471, opened 2022-10-04
Change all executables to use the default event loop run

KimLS:
Akkadius wanted me to change the executables that still were using event loop process to instead use event loop run because event loop process causes issues on windows under high load.

I ran through getting into a zone from login->world->zone and everything appears to work but could probably use some outside verification.


Akkadius:
Sweet, I’ll take a look later this week

Akkadius (approved):
Looking good on Linux

PR #2472, opened 2022-10-05
[Bug Fix] Client::QuestReward overload SummonItem call

mackal:
The overload that took the packet struct wasn't calling SummonItem with -1 charges, resulting it chargeless items if they had any


Kinglykrab (approved):
👍

PR #2476, opened 2022-10-09
[Feature] Add Data Bucket support to Bot Spells Entries

Aeadoin:
Initial Work on Bot Spell Data Buckets

Will work similar to Merchant Data Buckets, where a comparison will be made against the Characters Data Bucket, if the comparison returns Valid, the Bot will be eligible to use the spell in question.


Aeadoin:
I'll be cleaning this up under a new PR, as I've had to make some changes to how Bot Spells are loaded versus NPC spells.

PR #2479, opened 2022-10-11
[Zoning] Fix zone race condition

Akkadius:
**Problem**

Players end up in seemingly random locations in the destination zone

This is a race condition that can occur on all servers, but shows itself more prevalent on servers like PEQ where folks are running 30-60 box armies.

When a player crosses a zone line, a save is issued in `character_data` representing the destination `zone_id` `zone_instance` and `x` `y` `z` etc. The receiving zone then loads this to place the character properly on the other side.

The race condition is that when players cross a zone line, while x/y/z gets saved, there is a small chance that the players current X,Y,Z in the zone prior to the actual zoning gets updated when the client sends its client position update one last time before the client gets deconstructed (Where save is called again, saving the wrong x/y/z from the source zone)

**Solution**

When we perform a successful zone, we save the destination x/y/z and we lock further location over writing in the database in the very small time before the client object is desconstructed. 

**Testing**

Tested with a player on PEQ Sandbox with 35 characters. Zoned successfully with no character displacements in all attempts (20+ zones) whereas it would happen fairly consistenly between Nexus and Netherbian

### DIFF
--- a/zone/zoning.cpp
+++ b/zone/zoning.cpp
@@ -467,6 +467,8 @@ void Client::DoZoneSuccess(ZoneChange_Struct *zc, uint16 zone_id, uint32 instanc
 	//Force a save so its waiting for them when they zone
 	Save(2);
 
+	lock_save_position = true;
+
 	if (zone_id == zone->GetZoneID() && instance_id == zone->GetInstanceID()) {
 		// No need to ask worldserver if we're zoning to ourselves (most
 		// likely to a bind point), also fixes a bug since the default response was failure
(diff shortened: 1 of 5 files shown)


Kinglykrab (approved):

zone/zoning.cpp
```diff
@@ -467,6 +467,8 @@ void Client::DoZoneSuccess(ZoneChange_Struct *zc, uint16 zone_id, uint32 instanc
 	//Force a save so its waiting for them when they zone
 	Save(2);
 
+	lock_save_position = true;
```
hgtw, line 470: This would need to be reset if world responds with an error (zone failed to bootup for example, where zoning gets cancelled)

Akkadius: Updated

hgtw (approved):

PR #2501, opened 2022-10-27
[Feature] Change #scribespells to be aware of spellgroups & ranks

### LINKED ISSUE #727: #scribespells isn't aware of ranked spells
This command doesn't handle this correctly.

Aeadoin:
#scribespells previously wasn't aware of ranked spells.

This change will check if there's a higher rank spell available based on the max/min levels specified, and only learn the highest rank spell.

Fixes #727

### DIFF
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -10507,6 +10507,8 @@ std::vector<int> Client::GetMemmedSpells() {
 
 std::vector<int> Client::GetScribeableSpells(uint8 min_level, uint8 max_level) {
 	std::vector<int> scribeable_spells;
+	std::vector<int> spell_group_cache = BuildSpellGroupCache(max_level);
+
 	for (uint16 spell_id = 0; spell_id < SPDAT_RECORDS; ++spell_id) {
 		bool scribeable = true;
 		if (!IsValidSpell(spell_id)) {
@@ -10548,17 +10550,19 @@ std::vector<int> Client::GetScribeableSpells(uint8 min_level, uint8 max_level) {
 		}
 
 		if (spells[spell_id].spell_group) { 
-			uint32 highest_spell_id = GetHighestSpellinSpellGroup(spells[spell_id].spell_group);
-			if (
-				spells[highest_spell_id].classes[m_pp.class_ - 1] >= min_level &&
-				spells[highest_spell_id].classes[m_pp.class_ - 1] <= max_level &&
-				spell_id != highest_spell_id
-			) {
-				continue;
+			for (auto it = spell_group_cache.begin(); it != spell_group_cache.end(); it++) {
+				if (
+					spells[*it].classes[m_pp.class_ - 1] >= min_level &&
+					spells[*it].classes[m_pp.class_ - 1] <= max_level &&
+					spell_id == *it
+				) {
+					if (scribeable) {
+						scribeable_spells.push_back(spell_id);
+					}
+					continue;
+				}
 			}
-		}
-		
-		if (scribeable) {
+		} else if (scribeable) {
 			scribeable_spells.push_back(spell_id);
 		}
 	}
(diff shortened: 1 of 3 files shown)


zone/client.cpp
```diff
@@ -10547,6 +10547,17 @@ std::vector<int> Client::GetScribeableSpells(uint8 min_level, uint8 max_level) {
 			scribeable = false;
 		}
 
+		if(spells[spell_id].spell_group) { 
+			int highest_spell_id = GetHighestSpellinSpellGroup(spells[spell_id].spell_group);
```
Kinglykrab, line 10551: uint32 or auto since method returns uint32.

zone/client.cpp
```diff
@@ -10547,6 +10547,17 @@ std::vector<int> Client::GetScribeableSpells(uint8 min_level, uint8 max_level) {
...
+			if(spells[highest_spell_id].classes[m_pp.class_ - 1] <= max_level) {
```
Kinglykrab, line 10552: Formatting here and below. `if (` versus `if(`.

zone/client.cpp
Aeadoin: Fixed.

zone/client.cpp
```diff
@@ -10547,6 +10547,17 @@ std::vector<int> Client::GetScribeableSpells(uint8 min_level, uint8 max_level) {
...
+			uint32 highest_spell_id = GetHighestSpellinSpellGroup(spells[spell_id].spell_group);
+			if (spells[highest_spell_id].classes[m_pp.class_ - 1] <= max_level) {
+				if (spells[highest_spell_id].classes[m_pp.class_ - 1] >= min_level) {
+					if (spell_id != highest_spell_id) {
+						continue;
+					}
+				}
+			}
+		}
+
```
Kinglykrab, line 10560: Can consolidate the conditions being they all have to occur.
```cpp
if (spells[spell_id].spell_group) { 
	uint32 highest_spell_id = GetHighestSpellinSpellGroup(spells[spell_id].spell_group);
	if (
		spells[highest_spell_id].classes[m_pp.class_ - 1] >= min_level &&
		spells[highest_spell_id].classes[m_pp.class_ - 1] <= max_level &&
		spell_id != highest_spell_id
	) {
		continue;
	}
}
```

zone/spells.cpp
```diff
@@ -5520,6 +5520,26 @@ uint32 Client::GetHighestScribedSpellinSpellGroup(uint32 spell_group)
...
+				if (highest_rank < spells[i].rank) {
+					highest_rank = spells[i].rank;
+					highest_spell_id = i;
+				}
+			}
+		}
+	}
+	return highest_spell_id;
+}
+
```
Kinglykrab, line 5542: Consolidation similar to above.
```cpp
uint32 Client::GetHighestSpellinSpellGroup(uint32 spell_group)
{
	//Typical live spells follow 1/5/10 rank value for actual ranks 1/2/3, but this can technically be set as anything.

	int highest_rank = 0; //highest ranked found in spellgroup
	uint32 highest_spell_id = 0;  //spell_id of the highest ranked spell

	for (uint16 i = 0; i < SPDAT_RECORDS; i++) {
		if (
			IsValidSpell(i) &&
			spells[i].spell_group == spell_group &&
			highest_rank < spells[i].rank
		) {
			highest_rank = spells[i].rank;
			highest_spell_id = i;
		}
	}

	return highest_spell_id;
}
```

zone/client.cpp
Aeadoin: consolidated

zone/spells.cpp
Aeadoin: consolidated

Aeadoin:
> Maybe we could use a SQL query based on spells that are going to be scribed and get the highest ranked spell in each of the groups. Currently we could possibly iterate 10,000 times before we find the highest spell rank in a spell group, and the check is for any spell that shows up in the GetScribeableSpells() code and has a spell group. Has this been tested with huge amounts of spells scribing? Like level 1 to 70 all at once?

Agreed, I've implemented a query that'll reduce the iterations required.

Kinglykrab (approved):
Seems to work to me. If anyone else could take a look at it and see if this is the best way to do this, that'd be great. 👍

PR #2504, opened 2022-10-29
[Quest API] Add EVENT_AA_BUY and EVENT_AA_GAIN to Perl/Lua.

Kinglykrab:
# Perl
- Add EVENT_AA_BUY to Perl.
  - Exports `$aa_cost`, `$aa_id`, `$aa_previous_id`, and `$aa_next_id`
- Add EVENT_AA_GAIN to Perl.
  - Exports `$aa_gained`
- Add quest::getaaname(aa_id) to Perl.

# Lua
- Add event_aa_buy to Lua.
  - Exports `e.aa_cost`, `e.aa_id`, `e.aa_previous_id`, and `e.aa_next_id`
- Add event_aa_gain to Lua.
  - Exports `e.aa_gained`
- Add eq.get_aa_name(aa_id) to Lua.


zone/zone.cpp
```diff
@@ -2949,3 +2949,27 @@ void Zone::LoadDynamicZoneTemplates()
...
+std::string Zone::GetAAName(int aa_id)
+{
+	if (!aa_id) {
+		return std::string();
+	}
+
+	int current_aa_id = 0;
+
+	for (const auto& r : aa_ranks) {
+		if (r.second.get()->id == aa_id && r.second.get()->base_ability) {
```
hgtw, line 2962: Should be able to just `.find(aa_id)` instead of looping to find it since it looks like the id is the hash key for `Zone::aa_ranks` unless I'm missing something

Kinglykrab: Changed, good call.

hgtw (approved):

PR #2506, opened 2022-10-29
[Feature] Add special ability to block /open

hgtw:
This adds the IMMUNE_OPEN (53) special ability to prevent /open on LDON_TREASURE classes.


Kinglykrab (approved):
:+1:

PR #2508, opened 2022-10-29
[Feature] Add player /inspect quest event

hgtw:
Returning non-zero from EVENT_INSPECT will prevent default message


Kinglykrab (approved):
Beautiful.

PR #2514, opened 2022-10-30
[Quest API] Add Lua handlers for zone controller events

hgtw:
This cleans up some of the NPC::Death event dispatch code.

Adds handlers for EVENT_SPAWN_ZONE and EVENT_DEATH_ZONE used by zone controller and fixes the death handler exports which were incorrect.


Kinglykrab (approved):

Kinglykrab:
How does this work if the NPC is gone when event fires? Is it memory just long enough to fire this event?

hgtw:
Yea, the dead NPC pointer is still valid for the quest event since the zone controller event is called within the dead NPC's `NPC::Death`. It doesn't exit that function until the quest script's event function returns.

PR #2515, opened 2022-10-31
[Bots] Add support for Bot scripting.

Kinglykrab:
# Perl
- Add support for `zone/bot.pl` and `zone/bot_v#.pl`.
- Add support for `global/global_bot.pl`.
- Add `$bot->SignalBot(signal_id)` to Perl.
- Add `$bot->OwnerMessage(message)` to Perl.
- Add `$entity_list->SignalAllBotsByOwnerCharacterID(character_id, signal_id)` to Perl.
- Add `$entity_list->SignalBotByBotID(bot_id, signal_id)` to Perl.
- Add `$entity_list->SignalBotByBotName(bot_name, signal_id)` to Perl.
- Add `EVENT_SPELL_EFFECT_BOT` to Perl.
- Add `EVENT_SPELL_EFFECT_BUFF_TIC_BOT` to Perl.

# Lua
- Add support for `zone/bot.lua` and `zone/bot_v#.lua`.
- Add support for `global/global_bot.lua`.
- Add `bot:SignalBot(signal_id)` to Lua.
- Add `bot:OwnerMessage(message)` to Lua.
- Add `entity_list:SignalAllBotsByOwnerCharacterID(character_id, signal_id)` to Lua.
- Add `entity_list:SignalBotByBotID(bot_id, signal_id)` to Lua.
- Add `entity_list:SignalBotByBotName(bot_name, signal_id)` to Lua.
- Add `EVENT_SPELL_EFFECT_BOT` to Lua.
- Add `EVENT_SPELL_EFFECT_BUFF_TIC_BOT` to Lua.

# Supported Bot Events

1. `EVENT_CAST`
2. `EVENT_CAST_BEGIN`
3. `EVENT_CAST_ON`
4. `EVENT_COMBAT`
5. `EVENT_DEATH`
6. `EVENT_DEATH_COMPLETE`
7. `EVENT_NPC_SLAY`
8. `EVENT_POPUP_RESPONSE`
9. `EVENT_SAY`
10. `EVENT_SIGNAL`
11. `EVENT_SLAY`
12. `EVENT_SPAWN`
13. `EVENT_TARGET_CHANGE`
14. `EVENT_TIMER`
15. `EVENT_USE_SKILL`

# Common
- Convert NPC pointers in common events to Mob pointers so bots are supported.
- Convert signal IDs to `int` where it wasn't already, allowing negative signals to be sent properly.


hgtw:
I'm not familiar with bots, but it looks like they derive from NPC. Do they have unique NPCTypeIDs? I'm wondering if bots can just be bound to the same handlers as EventNPC and something lower level in event dispatch tapped in to handle "bot" and "global_bot" if IsBot() returns true. That would prevent the need for separate EventBot and handlers

Kinglykrab:
> I'm not familiar with bots, but it looks like they derive from NPC. Do they have unique NPCTypeIDs? I'm wondering if bots can just be bound to the same handlers as EventNPC and something lower level in event dispatch tapped in to handle "bot" and "global_bot" if IsBot() returns true. That would prevent the need for separate EventBot and handlers

They are derived from NPC, but a lot of their functionality is separate. Whatever works best. This is just the only way I knew to get it to work. Event exports have to be separate as well because they can’t do everything a normal NPC can do. I figured this would give us the most flexibility with it going forward if everything was separate and #ifdef’d out.

Kinglykrab:
Karana has been testing this for a few days and it seems to be working well for them.

Akkadius:
> I'm not familiar with bots, but it looks like they derive from NPC. Do they have unique NPCTypeIDs? I'm wondering if bots can just be bound to the same handlers as EventNPC and something lower level in event dispatch tapped in to handle "bot" and "global_bot" if IsBot() returns true. That would prevent the need for separate EventBot and handlers

This was my first thought as I read through this PR. There's a lot of code duplication that could pipe through similar logic. Lower level functions can `IsBot` to parse different bot scripts versus creating separate event handlers everywhere

Most other changes I don't really have an issue with, but duplicating handlers is definitely a code smell

Akkadius:
For example

```
parse->EventNPC(EVENT_COMBAT, CastToNPC(), attacker, "1", 0);
```
is duplicated with

```
parse->EventBot(EVENT_COMBAT, CastToBot(), attacker, "1", 0);
```

In the lower level functions such as `DispatchEventNPC` you can check for `IsBot` to load the proper bot scripts instead of the default npc ones. Utilize the same event hooks, same parameters and avoid duplication

Kinglykrab:
> For example
> 
> ```
> parse->EventNPC(EVENT_COMBAT, CastToNPC(), attacker, "1", 0);
> ```
> 
> is duplicated with
> 
> ```
> parse->EventBot(EVENT_COMBAT, CastToBot(), attacker, "1", 0);
> ```
> 
> In the lower level functions such as `DispatchEventNPC` you can check for `IsBot` to load the proper bot scripts instead of the default npc ones. Utilize the same event hooks, same parameters and avoid duplication

If we eventually separate Bots from NPCs I’ll have to rewrite all of this anyway. Bots also don’t use every single event an NPC can, so I had them separate so I could only parse where necessary. Whatever works though. Been sitting for long enough.

zone/event_codes.h
```diff
@@ -106,6 +106,10 @@ typedef enum {
 	EVENT_TASK_BEFORE_UPDATE,
 	EVENT_AA_BUY,
 	EVENT_AA_GAIN,
+ #ifdef BOTS
```
Akkadius, line 109: Small: there's an extra space before the `#ifdef`

Akkadius (approved):

PR #2543, opened 2022-11-14
[Quest API] Add GetRandomBot() to Perl/Lua

Kinglykrab:
# Perl
- Add `$entity_list->GetRandomBot()` to Perl.
- Add `$entity_list->GetRandomBot(x, y, z, distance)` to Perl.
- Add `$entity_list->GetRandomBot(x, y, z, distance, exclude_bot)` to Perl.

# Lua
- Add `eq.get_entity_list():GetRandomBot()` to Lua.
- Add `eq.get_entity_list():GetRandomBot(x, y, z, distance)` to Lua.
- Add `eq.get_entity_list():GetRandomBot(x, y, z, distance, exclude_bot)` to Lua.

# Notes
- Allows operators to grab a random Bot from entity list similar to Client, Mob, and NPC.
- Perl distance was sending as already squared, Lua was not.
- Send as non-squared and square in the method so that both work the same.


zone/entity.cpp
```diff
@@ -1863,21 +1863,56 @@ Client *EntityList::GetClientByLSID(uint32 iLSID)
...
+		) {
+			bots_in_range.push_back(b);
+		}
+	}
+
+	if (bots_in_range.empty()) {
+		return nullptr;
+	}
+
+	return bots_in_range[zone->random.Int(0, bots_in_range		.size() - 1)];
```
Akkadius, line 1894: Formatting with `bots_in_range` and `.size`

zone/entity.cpp
```diff
@@ -1863,21 +1863,56 @@ Client *EntityList::GetClientByLSID(uint32 iLSID)
 	return nullptr;
 }
 
+#ifdef BOTS
+Bot* EntityList::GetRandomBot(const glm::vec3& location, float distance, Bot* exclude_bot)
+{
+	auto is_whole_zone = false;
+	if (location.x == 0.0f && location.y == 0.0f) {
```
Akkadius, line 1870: This is kind of strange logic, why include whole zone when coordinates are 0 ?

Akkadius (approved):

zone/entity.cpp
Kinglykrab: If you don’t supply a location, it assumes you want the entire zone. glm::vec3 is set to 0.0 for XYZ in header if not sent.

PR #2546, opened 2022-11-15
[Commands] Remove #profiledump and #profilereset Commands.

Kinglykrab:
- These commands don't seem to be used anymore.


Akkadius (approved):
These were for KLS's built in profiler work that was experimental, it never got actual usage.

There are tools (Visual Studio and Linux perf) that can perform these same responsibilities

PR #2552, opened 2022-11-18
[Bug Fix] Fix issue with Bot Loading with 0 Health

Aeadoin:
Query was off by 1, causing HP to load incorrectly.


Kinglykrab (approved):
:+1:

PR #2570, opened 2022-11-23
[Merchant] LDoNSellBackRate support for Merchant:EnableAltCurrencySell Rule

Aeadoin:
Adds support for Merchants to buyback Alt Currency items at a percentage value using the item LDoNSellBackRate attribute.

Currently this was only being supported for Adventure Merchants.

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -2695,7 +2695,8 @@ void Client::Handle_OP_AltCurrencySell(const EQApplicationPacket *app)
 				continue;
 
 			if (item->ID == inst->GetItem()->ID) {
-				cost = ml.alt_currency_cost;
+				// 06/11/2016 This formula matches RoF2 client side calculation.
+				cost = (ml.alt_currency_cost + 1) * item->LDoNSellBackRate / 100;
 				found = true;
 				break;
 			}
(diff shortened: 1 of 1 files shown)


Kinglykrab:
Formula being used here is used here as well, can anyone confirm this is how Live does it?: https://github.com/EQEmu/Server/blob/645251992dae8c572558a3b60432ee2f029c6aa3/zone/client_packet.cpp#L2194

Kinglykrab:
Just curious if the `+ 1` is to offset some weird math they do, or what, if `LDoNSellBackRate` is set to `100`, you'll get more than 100% of the value.

Assume LDoNPrice and LDoNSellBackRate are `100`, this returns `101`:
```cpp
int32 price = (100 + 1) * 100 / 100;```

Aeadoin:
Correct, I took the formula from our existing Adventure Merchant formula... which if you set LDoNSellBackRate to 100 it would sell back for 1 more then you buy it for.

LDoNSellBackRate is default to 0 however on items.

``// 06/11/2016 This formula matches RoF2 client side calculation.
 int32 price = (item->LDoNPrice + 1) * item->LDoNSellBackRate / 100;``

Kinglykrab (approved):
LGTM. :+1:

PR #2578, opened 2022-11-26
[Bots] Add Rule AllowBotEquipAnyRaceGear

Aeadoin:
Adds optional Rule for Bots "AllowBotEquipAnyRaceGear" which Allows Bots to wear Equipment even if their race is not valid.
Set to False by Default. 

Below is an example of a usable item with the Rule set to True. Class & Level restrictions are still applied.

image


zone/bot.cpp
```diff
@@ -4728,8 +4728,9 @@ void Bot::PerformTradeWithClient(int16 begin_slot_id, int16 end_slot_id, Client*
 				return;
 			}
 		}
-
-		if (!trade_instance->IsEquipable(GetBaseRace(), GetClass()) || (GetLevel() < trade_instance->GetItem()->ReqLevel)) { // deity checks will be handled within IsEquipable()
+		
+		bool bots_ignore_race = RuleB(Bot, AllowBotEquipAnyRaceGear);
+		if (!trade_instance->IsClassEquipable(GetClass()) || (GetLevel() < trade_instance->GetItem()->ReqLevel) || (!trade_instance->IsRaceEquipable(GetRace()) && !bots_ignore_race)) { // deity checks will be handled within IsEquipable()
```
Kinglykrab, line 4733: `GetLevel() < trade_instance->GetItem()->ReqLevel` versus `(GetLevel() < trade_instance->GetItem()->ReqLevel)`

Kinglykrab (approved):

zone/bot.cpp
Aeadoin: Fixed.

PR #2586, opened 2022-11-27
[Bot] Add Support for Bots to receive Auras, and other AoE Spells

Aeadoin:
Also provides support for Mercs.

Player Auras will land on Bots & Mercs now.
PBAE Buffs willl land on Bots & Mercs now.
MGB Buffs will land on Bots & Mercs now.
NPC detrimental spells should also function if they didn't before (PBAE spells)

### DIFF
--- a/zone/entity.cpp
+++ b/zone/entity.cpp
@@ -2922,7 +2922,7 @@ void EntityList::ScanCloseMobs(
 	for (auto &e : mob_list) {
 		auto mob = e.second;
 
-		if (!mob->IsNPC() && !mob->IsClient()) {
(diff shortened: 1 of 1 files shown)


Kinglykrab (approved):
Looks good to me, not sure why this was the way it was, if anyone can give insight that'd be great. Don't want to merge if there's a valid reason not to include bots/mercenaries.

Akkadius (approved):

Akkadius:
> Looks good to me, not sure why this was the way it was, if anyone can give insight that'd be great. Don't want to merge if there's a valid reason not to include bots/mercenaries.

Reason is I never added them to the scanning for this. Mercs are almost non-existent and bots are usually an afterthought until they’ve gotten a lot more love in the past year or so

PR #2591, opened 2022-11-28
[Commands] Cleanup #scale Command.

Kinglykrab:
- Cleanup messages and logic.


Akkadius:
I'll review this later in the week

Akkadius:
I'm not sure why changing the messages is necessary in this command

Originally when I wrote this, it would show what it was doing in each phase (Searching) or (Applying) and also just show one entity per line, now we have two lines per entity and it's harder to read when trying to scan quickly

https://docs.eqemu.io/server/npc/npc-scaling/#name-search

**Before**

image

**After**

image

Otherwise, I don't really have issues with the changes as long as it functionally doesn't break anything and output is back to one line per entity

Functionally it at least appears to look the same

image

Kinglykrab:
I'll change the message back to one line.

Kinglykrab:
Made it back to one line.
image

Akkadius (approved):

PR #2593, opened 2022-11-29
[Commands] Cleanup #rules Command. 

Kinglykrab:
- Cleanup messages and logic.
- Rewrite all rules logic to use `std::string` and repositories.

### DIFF
--- a/common/repositories/rule_values_repository.h
+++ b/common/repositories/rule_values_repository.h
@@ -44,6 +44,87 @@ public:
      */
 
 	// Custom extended repository methods here
+	static std::vector<std::string> GetRuleNames(Database &db, int rule_set_id)
+	{
+		std::vector<std::string> v;
+
+		auto results = db.QueryDatabase(
+			fmt::format(
+				"SELECT rule_name FROM {} WHERE ruleset_id = {}",
+				TableName(),
+				rule_set_id
+			)
+		);
+		if (!results.Success() || !results.RowCount()) {
+			return v;
+		}
+
+		for (auto row : results) {
+			v.push_back(row[0]);
+		}
+
+		return v;
+	}
+
+	static std::vector<std::string> GetGroupedRules(Database &db)
+	{
+		std::vector <std::string> v;
+
+		auto results = db.QueryDatabase(
+			fmt::format(
+				"SELECT rule_name FROM {} GROUP BY rule_name",
+				TableName()
+			)
+		);
+		if (!results.Success() || !results.RowCount()) {
+			return v;
+		}
+
+		for (auto row : results) {
+			v.push_back(row[0]);
+		}
+
+		return v;
+	}
+
+	static bool DeleteOrphanedRules(Database& db, std::vector<std::string>& v)
+	{
+		const auto query = fmt::format(
+			"DELETE FROM {} WHERE rule_name IN ({})",
+			TableName(),
+			Strings::ImplodePair(",", std::pair<char, char>('\'', '\''), v)
+		);
+
+		return db.QueryDatabase(query).Success();
+	}
+
+	static bool InjectRules(Database& db, std::vector<std::tuple<int, std::string, std::string, std::string>>& v)
+	{
+		const auto query = fmt::format(
+			"REPLACE INTO {} (`ruleset_id`, `rule_name`, `rule_value`, `notes`) VALUES {}",
+			TableName(),
+			Strings::ImplodePair(
+				",",
+				std::pair<char, char>('(', ')'),
+				join_tuple(",", std::pair<char, char>('\'', '\''), v)
+			)
+		);
+
+		return db.QueryDatabase(query).Success();
+	}
+
+	static bool UpdateRuleNote(Database& db, int rule_set_id, std::string rule_name, std::string notes)
+	{
+		const auto query = fmt::format(
+			"UPDATE {} SET notes = '{}' WHERE ruleset_id = {} AND rule_name = '{}'",
+			TableName(),
+			notes,
+			rule_set_id,
+			rule_name
+		);
+
+		return db.QueryDatabase(query).Success();
+	}
 
 };
 
--- a/common/repositories/rule_values_repository.h
+++ b/common/repositories/rule_values_repository.h
@@ -118,7 +118,7 @@ public:
 		const auto query = fmt::format(
 			"UPDATE {} SET notes = '{}' WHERE ruleset_id = {} AND rule_name = '{}'",
 			TableName(),
-			notes,
+			Strings::Escape(notes),
 			rule_set_id,
 			rule_name
 		);
--- a/common/rulesys.cpp
+++ b/common/rulesys.cpp
@@ -145,7 +145,7 @@ bool RuleManager::SetRule(const std::string &rule_name, const std::string &rule_
 			LogRules("Set rule [{}] to value [{}]", rule_name, m_RuleIntValues[index]);
 			break;
 		case RealRule:
-			m_RuleRealValues[index] = atof(rule_value);
+			m_RuleRealValues[index] = atof(rule_value.c_str());
 			LogRules("Set rule [{}] to value [{:.2f}]", rule_name, m_RuleRealValues[index]);
 			break;
 		case BoolRule:
--- a/common/rulesys.cpp
+++ b/common/rulesys.cpp
@@ -27,7 +27,7 @@
 #include "../common/repositories/rule_sets_repository.h"
 #include "../common/repositories/rule_values_repository.h"
 
-std::string RuleManager::s_categoryNames[_CatCount + 1] = {
+const char *RuleManager::s_categoryNames[_CatCount + 1] = {
 	#define RULE_CATEGORY(category_name) \
 		#category_name ,
 	#include "ruletypes.h"
@@ -55,7 +55,7 @@ RuleManager::RuleManager()
 	ResetRules(false);
 }
 
-RuleManager::CategoryType RuleManager::FindCategory(std::string category_name)
+RuleManager::CategoryType RuleManager::FindCategory(const std::string &category_name)
 {
 	for (int i = 0; i < _CatCount; i++) {
 		if (Strings::Contains(category_name, s_categoryNames[i])) {
@@ -66,7 +66,7 @@ RuleManager::CategoryType RuleManager::FindCategory(std::string category_name)
 	return InvalidCategory;
 }
 
-bool RuleManager::ListRules(std::string category_name, std::vector<std::string> &l)
+bool RuleManager::ListRules(const std::string &category_name, std::vector<std::string> &l)
 {
 	CategoryType category_type = InvalidCategory;
 	if (!category_name.empty()) {
@@ -96,7 +96,7 @@ bool RuleManager::ListCategories(std::vector <std::string> &l)
 	return true;
 }
 
-bool RuleManager::GetRule(std::string rule_name, std::string &rule_value)
+bool RuleManager::GetRule(const std::string &rule_name, std::string &rule_value)
 {
 	RuleType type;
 	uint16 index;
@@ -112,14 +112,14 @@ bool RuleManager::GetRule(std::string rule_name, std::string &rule_value)
 			rule_value = fmt::format("{}", m_RuleRealValues[index]);
 			break;
 		case BoolRule:
-			rule_value = fmt::format("{}", m_RuleBoolValues[index] ? "true" : "false");
+			rule_value = m_RuleBoolValues[index] ? "true" : "false";
 			break;
 	}
 
 	return true;
 }
 
-bool RuleManager::SetRule(std::string rule_name, std::string rule_value, Database *db, bool db_save, bool reload)
+bool RuleManager::SetRule(const std::string &rule_name, const std::string &rule_value, Database *db, bool db_save, bool reload)
 {
 	if (rule_name.empty() || rule_value.empty()) {
 		return false;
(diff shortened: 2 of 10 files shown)


Akkadius:
I'll review this later in the week

common/rulesys.cpp
```diff
@@ -24,462 +24,451 @@
...
-	int rule_count = CountRules();
-	for (i = 0; i < rule_count; i++) {
-		const RuleInfo &rule = s_RuleInfo[i];
-		if (strcmp(rule_name, rule.name) == 0) {
-			type_into = rule.type;
-			index_into = rule.rule_index;
-			return(true);
+	for (int i = 0; i < CountRules(); i++) {
+		const auto& r = s_RuleInfo[i];
+		if (Strings::Contains(rule_name, r.name)) {
```
Akkadius, line 198: This changes behavior

We should have a `Strings::Equals` since `strcmp` is checking for exact equality of strings. We'll see side affects with partials here

common/rulesys.cpp
```diff
@@ -24,462 +24,451 @@
...
-	char value_string[100];
+void RuleManager::_SaveRule(Database *db, RuleType type, uint16 index) {
+	const auto rule_name = _GetRuleName(type, index);
 
-	if (type == IntRule && strcasecmp(_GetRuleName(type, index), "World:ExpansionSettings") == 0)
-		return;
-	if (type == BoolRule && strcasecmp(_GetRuleName(type, index), "World:UseClientBasedExpansionSettings") == 0)
+	if (
+		(type == BoolRule && Strings::Contains(rule_name, "World:UseClientBasedExpansionSettings")) ||
+		(type == IntRule && Strings::Contains(rule_name, "World:ExpansionSettings"))
```
Akkadius, line 346: This also changes behavior `strcasecmp` is a comparison if strings are equal (case insensitive) - a contains is probably fine at least in this case

common/rulesys.cpp
```diff
@@ -24,462 +24,451 @@
...
+	std::string rule_value;
 
 	switch (type) {
 		case IntRule:
-			sprintf(value_string, "%d", m_RuleIntValues[index]);
+			rule_value = fmt::format("{}", m_RuleIntValues[index]);
 			break;
 		case RealRule:
-			sprintf(value_string, "%.13f", m_RuleRealValues[index]);
+			rule_value = fmt::format("{:.2f}", m_RuleRealValues[index]);
```
Akkadius, line 358: Seems like we're losing precision here? Looks like we allowed 13 places above

Might matter with some rules and create unintended bugs

common/rulesys.cpp
```diff
@@ -24,462 +24,451 @@
...
-			break;
-		default:
-			break;
+		switch (r.type) {
+			case IntRule:
+				rule_data[r.name].first = fmt::format("{}", m_RuleIntValues[r.rule_index]);
+				rule_data[r.name].second = &r.notes;
+				break;
+			case RealRule:
+				rule_data[r.name].first = fmt::format("{:.2f}", m_RuleRealValues[r.rule_index]);
```
Akkadius, line 426: Losing precision here - should probably use `.13f`

common/rulesys.cpp
```diff
@@ -491,50 +480,36 @@ bool RuleManager::UpdateOrphanedRules(Database *db, bool quiet_update)
...
+	const auto& l = RuleValuesRepository::GetGroupedRules(*db);
+	if (l.empty()) {
 		return false;
 	}
 
 	// build rule data entries
-	for (const auto &ri_iter : s_RuleInfo) {
-		if (strcasecmp(ri_iter.name, "Invalid Rule") == 0) {
+	for (const auto &r : s_RuleInfo) {
+		if (Strings::Contains(r.name, "Invalid Rule")) {
```
Akkadius, line 490: This changes behavior as well

`strcasecmp` compares if strings are equal case insensitive whereas `Contains` is a partial

common/rulesys.cpp
```diff
@@ -546,139 +521,83 @@ bool RuleManager::UpdateOrphanedRules(Database *db, bool quiet_update)
...
 
 	int update_count = 0;
-	for (auto row = results.begin(); row != results.end(); ++row) {
-
-		auto rule = [](const char *rule_name) {
+	for (const auto& e : l) {
 
+		auto rule = [](std::string rule_name) {
 			for (auto rule_iter : s_RuleInfo) {
-				if (strcasecmp(rule_iter.name, rule_name) == 0) {
```
Akkadius, line 562: This changes behavior as well

`strcasecmp `compares if strings are equal case insensitive whereas `Contains` is a partial

common/strings.cpp
```diff
@@ -716,3 +716,20 @@ uint32 Strings::TimeToSeconds(std::string time_string)
 
 	return duration;
 }
+
+bool Strings::ToBool(std::string bool_string)
```
Akkadius, line 720: I love it

common/rulesys.cpp
Kinglykrab: I'll add `Strings::Equals`.

common/rulesys.cpp
Kinglykrab: Will change to `Strings::Equals`.

common/rulesys.cpp
Kinglykrab: Will use `Strings::Equals`.

common/rulesys.cpp
Kinglykrab: Will use `Strings::Equals`.

common/strings.cpp
Kinglykrab: :heart:

common/rulesys.cpp
```diff
@@ -24,462 +24,451 @@
...
-	set (cat) (rule) (value)
-	values [catname] -> show the values of all rules in the specified category/
-*/
+#include "../common/repositories/rule_sets_repository.h"
+#include "../common/repositories/rule_values_repository.h"
 
-const char *RuleManager::s_categoryNames[_CatCount+1] = {
-	#define RULE_CATEGORY(catname) \
-		#catname ,
+std::string RuleManager::s_categoryNames[_CatCount + 1] = {
```
hgtw, line 30: Changing this to std::string could cause static initialization problems since RuleManager is a singleton. Its constructor might try to access `s_categoryNames` before the strings are initialized. I'd recommend reverting this change and leaving it as is or investigating changing this in a separate patch

common/rulesys.cpp
```diff
@@ -24,462 +24,451 @@
...
+	for (int i = 0; i < _CatCount; i++) {
+		l.push_back(s_categoryNames[i]);
 	}
-	return(true);
+
+	return true;
 }
 
-bool RuleManager::GetRule(const char *rule_name, std::string &return_value) {
+bool RuleManager::GetRule(std::string rule_name, std::string &rule_value)
```
hgtw, line 99: Probably want to pass strings by const ref (here and other functions) to avoid a lot of unnecessary copy construction

common/rulesys.cpp
```diff
@@ -24,462 +24,451 @@
...
+			rule_value = fmt::format("{}", m_RuleIntValues[index]);
 			break;
 		case RealRule:
-			sprintf(tmp, "%f", m_RuleRealValues[index]);
+			rule_value = fmt::format("{}", m_RuleRealValues[index]);
 			break;
 		case BoolRule:
-			std::string tmp_val = m_RuleBoolValues[index] ? "true" : "false";
-			sprintf(tmp, "%s", tmp_val.c_str());
+			rule_value = fmt::format("{}", m_RuleBoolValues[index] ? "true" : "false");
```
hgtw, line 115: No need to use fmt here, can just assign the strings directly

common/rulesys.cpp
```diff
@@ -24,462 +24,451 @@
...
+		return (false);
+	}
 
 	if (reload) {
-		if (strcasecmp(rule_name, "World:ExpansionSettings") == 0)
-			return(false);
-		if (strcasecmp(rule_name, "World:UseClientBasedExpansionSettings") == 0)
-			return(false);
+		bool is_client = Strings::Contains(rule_name, "World:UseClientBasedExpansionSettings");
+		bool is_world = Strings::Contains(rule_name, "World:ExpansionSettings");
```
hgtw, line 136: These were doing exact matches before, is the change for partial matches intentional?

common/rulesys.cpp
```diff
@@ -24,462 +24,451 @@
...
+		if (is_client || is_world) {
+			return false;
+		}
 	}
 
-	switch(type) {
+	switch (type) {
 		case IntRule:
-			m_RuleIntValues[index] = atoi(rule_value);
+			m_RuleIntValues[index] = std::stoi(rule_value);
```
hgtw, line 144: I'd recommend using std::strtol if changing these since it doesn't throw. `std::stoi` and variants are just wrappers that throw exceptions instead of returning 0 for bad input. Either that or catch the exceptions and log bad inputs since changing this is a behavior change and can throw now

common/rulesys.cpp
```diff
@@ -24,462 +24,451 @@
...
-	int rule_count = CountRules();
-	for (i = 0; i < rule_count; i++) {
-		const RuleInfo &rule = s_RuleInfo[i];
-		if (strcmp(rule_name, rule.name) == 0) {
-			type_into = rule.type;
-			index_into = rule.rule_index;
-			return(true);
+	for (int i = 0; i < CountRules(); i++) {
+		const auto& r = s_RuleInfo[i];
+		if (Strings::Equal(rule_name, r.name)) {
```
hgtw, line 198: Can just do `if (rule_name == r.name)` here

common/rulesys.cpp
```diff
@@ -24,462 +24,451 @@
...
-			sprintf(value_string, "%d", m_RuleIntValues[index]);
+			rule_value = fmt::format("{}", m_RuleIntValues[index]);
 			break;
 		case RealRule:
-			sprintf(value_string, "%.13f", m_RuleRealValues[index]);
+			rule_value = fmt::format("{:.13f}", m_RuleRealValues[index]);
 			break;
 		case BoolRule:
-			sprintf(value_string, "%s", m_RuleBoolValues[index] ? "true" : "false");
+			rule_value = fmt::format("{}", m_RuleBoolValues[index] ? "true" : "false");
```
hgtw, line 361: Can just assign the strings here instead of using fmt

common/rulesys.cpp
```diff
@@ -24,462 +24,451 @@
...
+			case IntRule:
+				rule_data[r.name].first = fmt::format("{}", m_RuleIntValues[r.rule_index]);
+				rule_data[r.name].second = &r.notes;
+				break;
+			case RealRule:
+				rule_data[r.name].first = fmt::format("{:.13f}", m_RuleRealValues[r.rule_index]);
+				rule_data[r.name].second = &r.notes;
+				break;
+			case BoolRule:
+				rule_data[r.name].first = fmt::format("{}", m_RuleBoolValues[r.rule_index] ? "true" : "false");
```
hgtw, line 430: Same as the other one, can just assign these strings without fmt

common/rulesys.cpp
```diff
@@ -24,462 +24,451 @@
...
 			injected_rule_entries.push_back(
 				std::tuple<int, std::string, std::string, std::string>(
-					ruleset_id,								// `ruleset_id`
-					rd_iter.first,							// `rule_name`
-					rd_iter.second.first,					// `rule_value`
-					Strings::Escape(*rd_iter.second.second)	// `notes`
+					rule_set_id,
+					d.first,
+					d.second.first,
+					*d.second.second
```
hgtw, line 446: Lost string escapes on `notes` field

common/rulesys.cpp
```diff
@@ -546,139 +521,83 @@ bool RuleManager::UpdateOrphanedRules(Database *db, bool quiet_update)
...
 				}
 			}
 
-			return s_RuleInfo[_IntRuleCount+_RealRuleCount+_BoolRuleCount];
-		}(row[1]);
+			return s_RuleInfo[_IntRuleCount + _RealRuleCount + _BoolRuleCount];
+		}(e.rule_name);
 
-		if (strcasecmp(rule.name, row[1]) != 0) {
+		if (Strings::Contains(rule.name, e.rule_name)) {
```
hgtw, line 542: This was previously an exact match but is partial now, guessing this is unintentional

common/repositories/rule_values_repository.h
```diff
@@ -44,6 +44,87 @@ class RuleValuesRepository: public BaseRuleValuesRepository {
...
+
+		return db.QueryDatabase(query).Success();
+	}
+
+	static bool UpdateRuleNote(Database& db, int rule_set_id, std::string rule_name, std::string notes)
+	{
+		const auto query = fmt::format(
+			"UPDATE {} SET notes = '{}' WHERE ruleset_id = {} AND rule_name = '{}'",
+			TableName(),
+			notes,
```
hgtw, line 121: `notes` should be string escaped here I think

common/rulesys.cpp
```diff
@@ -24,462 +24,451 @@
...
+	switch (type) {
 		case IntRule:
-			m_RuleIntValues[index] = atoi(rule_value);
+			m_RuleIntValues[index] = atoi(rule_value.c_str());
 			LogRules("Set rule [{}] to value [{}]", rule_name, m_RuleIntValues[index]);
 			break;
 		case RealRule:
-			m_RuleRealValues[index] = atof(rule_value);
-			LogRules("Set rule [{}] to value [{}]", rule_name, m_RuleRealValues[index]);
+			m_RuleRealValues[index] = std::stof(rule_value);
```
hgtw, line 148: This one can throw on bad input now too

hgtw (approved):

PR #2594, opened 2022-11-30
[Quest API] Adjustment to depop_all function.

fryguy503:
Adjustment to depop_all function to no longer require an owner under all conditions (allows use inside encounters)

### DIFF
--- a/zone/questmgr.cpp
+++ b/zone/questmgr.cpp
@@ -870,11 +870,15 @@ void QuestManager::depop_withtimer(int npc_type) {
 
 void QuestManager::depopall(int npc_type) {
 	QuestManagerCurrentQuestVars();
-	if(owner && owner->IsNPC() && (npc_type > 0)) {
+	if (npc_type > 0) {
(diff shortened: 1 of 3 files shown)


zone/questmgr.cpp
```diff
@@ -870,11 +870,15 @@ void QuestManager::depop_withtimer(int npc_type) {
...
 void QuestManager::depopall(int npc_type) {
 	QuestManagerCurrentQuestVars();
-	if(owner && owner->IsNPC() && (npc_type > 0)) {
+	if (npc_type > 0) {
 		entity_list.DepopAll(npc_type);
-	}
-	else {
-		LogQuests("QuestManager::depopall called with nullptr owner, non-NPC owner, or invalid NPC Type ID. Probably syntax error in quest file");
+	} else {
+		if (owner && owner->IsNPC() && (npc_type > 0)) {
```
Kinglykrab, line 876: We’ll never reach this check if `npc_type` is > 0. Need to check for owner stuff first. Then fall back to this.

PR #2595, opened 2022-11-30
[Quest API] Adjustment to depop_all function.

fryguy503:
Adjustment to depop_all function to no longer require an owner under all conditions (allows use inside encounters)

### DIFF
--- a/zone/questmgr.cpp
+++ b/zone/questmgr.cpp
@@ -870,11 +870,10 @@ void QuestManager::depop_withtimer(int npc_type) {
 
 void QuestManager::depopall(int npc_type) {
 	QuestManagerCurrentQuestVars();
-	if(owner && owner->IsNPC() && (npc_type > 0)) {
+	if ((owner && owner->IsNPC() && (npc_type > 0)) || npc_type > 0) {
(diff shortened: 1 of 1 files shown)


zone/questmgr.cpp
```diff
@@ -870,11 +870,10 @@ void QuestManager::depop_withtimer(int npc_type) {
 
 void QuestManager::depopall(int npc_type) {
 	QuestManagerCurrentQuestVars();
-	if(owner && owner->IsNPC() && (npc_type > 0)) {
+	if ((owner && owner->IsNPC() && (npc_type > 0)) || npc_type > 0) {
```
Kinglykrab, line 873: Can probably just remove the owner checks and QuestManagerCurrentQuestVars() altogether since they’re not used and don’t matter.

Kinglykrab: Only check for NPC type > 0 basically.

fryguy503: Fair enough, for some reason my mind was set on continuing to check for ownership and IsNPC but that is not relevant with this.

Kinglykrab (approved):

PR #2612, opened 2022-12-04
[Bots] Save Bot Toggle Archer Setting between Loads.

Aeadoin:
^bottogglearcher settings weren't previously being saved, so would have to be reset when anytime Bots were being Loaded.


Kinglykrab (approved):
Gorgeous.

PR #2613, opened 2022-12-04
[Bug Fix] Fix case-sensitivity in #suspend Command.

Kinglykrab:
# Notes
- This command required you to properly send the name as it appears in the database otherwise it wouldn't kick the player.
- Solution is to `Strings::ToLower` then `Strings::UcFirst`.

### DIFF
--- a/zone/gm_commands/suspend.cpp
+++ b/zone/gm_commands/suspend.cpp
@@ -13,12 +13,12 @@ void command_suspend(Client *c, const Seperator *sep)
 		return;
 	}
 
-	const std::string character_name = sep->arg[1];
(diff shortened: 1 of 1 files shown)


Akkadius:
It's an edge case, but character names can end up being all lower case

It would be more resilient to cast the column WHERE in the MySQL to `LOWER(charname)` and ToLower on the input side

Kinglykrab:
> It's an edge case, but character names can end up being all lower case
> 
> It would be more resilient to cast the column WHERE in the MySQL to `LOWER(charname)` and ToLower on the input side

Makes sense to me.

Akkadius (approved):

PR #2614, opened 2022-12-04
[Bots] Convert Load, Save, SaveNew, and Delete to Repositories.

Kinglykrab:
# Notes
- General code cleanup, as manually adding to these queries doesn't scale very well.

### DIFF
--- a/zone/bot_database.cpp
+++ b/zone/bot_database.cpp
@@ -24,6 +24,7 @@
 #include "../common/strings.h"
 #include "../common/eqemu_logsys.h"
 
+#include "../common/repositories/bot_data_repository.h"
 #include "../common/repositories/bot_inventories_repository.h"
 
 #include "zonedb.h"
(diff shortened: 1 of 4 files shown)


zone/bot_database.cpp
```diff
@@ -379,86 +380,47 @@ bool BotDatabase::LoadBot(const uint32 bot_id, Bot*& loaded_bot)
...
-		std::stoul(row[14]), // Eye Color 1
-		std::stoul(row[15]), // Eye Color 2
-		std::stoul(row[16]), // Beard
-		std::stoul(row[17]), // Beard Color
-		std::stoul(row[18]), // Drakkin Heritage
-		std::stoul(row[19]), // Drakkin Tattoo
-		std::stoul(row[20]), // Drakkin Details
-		std::stoi(row[21]), // Health
-		std::stoi(row[22]), // Mana
+		l[0].spells_id,
```
Akkadius, line 404: Would really prefer we use the `FindOne` method when we intend on working with one object

It looks like the repository template will need to be updated to use the primary key in the `fmt::format`

```cpp
	static BotData FindOne(
		Database& db,
		int bot_data_id
	)
	{
		auto results = db.QueryDatabase(
			fmt::format(
				"{} WHERE id = {} LIMIT 1",
				BaseSelect(),
				bot_data_id
			)
		);
```

Akkadius: Then you are accessing just the one object via `l->` versus `l[0]->` which is a bit strange to read

zone/bot_database.cpp
```diff
@@ -480,304 +442,181 @@ bool BotDatabase::LoadBot(const uint32 bot_id, Bot*& loaded_bot)
...
+			bot_inst->GetBotID()
+		)
 	);
-	auto results = database.QueryDatabase(query);
-	if (!results.Success())
+	if (l.empty()) {
 		return false;
+	}
+
+	l[0].owner_id               = bot_inst->GetBotOwnerCharacterID();
```
Akkadius, line 554: Same here, similar to feedback prior

Akkadius (approved):
Minor comments, love to see the repository usage

zone/bot_database.cpp
Kinglykrab: Ah yeah, forgot about FindOne while I was working on this. No worries.

PR #2620, opened 2022-12-05
[Quest API] Add EVENT_LEVEL_DOWN to Perl/Lua.

Kinglykrab:
# Perl
- Add `EVENT_LEVEL_DOWN`, exports `$levels_lost`.
- Add `$levels_gained` export to `EVENT_LEVEL_UP`.

# Lua
- Add `event_level_down`, exports `e.levels_lost`.
- Add `e.levels_gained` export to `event_level_up`.

# Notes
- Allows operators to perform actions on level down.
- Allows operators to tell how many levels were lost or gained in case people are gaining/losing multiple levels and they want to keep track or use this as a mechanic in their code somewhere.


zone/embparser.cpp
```diff
@@ -163,7 +163,8 @@ const char *QuestEventSubroutines[_LargestEventID] = {
 	"EVENT_TASK_BEFORE_UPDATE",
 	"EVENT_AA_BUY",
 	"EVENT_AA_GAIN",
-	"EVENT_PAYLOAD"
+	"EVENT_PAYLOAD",
+	"EVENT_LEVEL_DOWN"
   #ifdef BOTS
 	,
```
hgtw, line 169: nit: I'd remove this floating comma and just put it at the end of `EVENT_LEVEL_DOWN` (and probably one after `EVENT_SPELL_EFFECT_BUFF_TIC_BOT`).

Even if BOTS is undefined the comma at the end will make it cleaner when adding new events since only one new line will need added.

Also remove the indent from `#ifdef BOTS`

Kinglykrab: Will fix this when at PC.

hgtw (approved):

PR #2621, opened 2022-12-05
[Quest API] Add MaxSkills() to Perl/Lua.

Kinglykrab:
# Perl
- Add `$client->MaxSkills()`.

# Lua
- Add `client:MaxSkills()`.

# Notes
- Allows operators an easy short hand for maxing a player's skills for their level without needing to do all the looping and stuff on their own.

### DIFF
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -12000,14 +12000,13 @@ std::string Client::GetGuildPublicNote()
 
 void Client::MaxSkills()
 {
-	for (const auto &skills_iter : EQ::skills::GetSkillTypeMap()) {
-		auto skill_id            = skills_iter.first;
+	for (const auto &s : EQ::skills::GetSkillTypeMap()) {
 		auto current_skill_value = (
-			EQ::skills::IsSpecializedSkill(skill_id) ?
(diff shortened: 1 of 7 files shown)


zone/client.cpp
```diff
@@ -11997,3 +11997,18 @@ std::string Client::GetGuildPublicNote()
 
 	return gci.public_note;
 }
+
+void Client::MaxSkills()
+{
+	for (const auto &s : EQ::skills::GetSkillTypeMap()) {
+		auto current_skill_value = (
+			EQ::skills::IsSpecializedSkill(s.first) ?
+			50 :
```
joligario, line 12006: Is 50 hard-coded or defined somewhere?

joligario (approved):
Recommend name change to something more descriptive like `SetMaxSkills`

zone/client.cpp
Kinglykrab: It’s the max a specialized skill can go before you set a specific specialization.

joligario: Ah, yeah. May be worth adding to constants - no big deal.

PR #2623, opened 2022-12-06
[Quest API] Add CopyHateList() to Perl/Lua.

Kinglykrab:
# Perl
- Add `$mob->CopyHateList(to_mob)`.

# Lua
- Add `mob:CopyHateList(to_mob)`.

# Notes
- Allows operators to easily copy and entire hatelist from one mob to another.


joligario (approved):
Should we have a protection of `to` being a client?

PR #2625, opened 2022-12-06
[Rules] rule gate pet zoning

fryguy503:
Allows preventing pet zoning from happening for more classic servers.


Kinglykrab (approved):
:+1:

PR #2636, opened 2022-12-12
[Commands] Nested Command Aliases

Akkadius:
Certain commands like `#rq` and `#size` are hard wired in our lizard brains after years of muscle memory and usage. These commands can be conveniently aliased using the commands aliases system but only for top level commands. 

Some commands have been consolidated to a single place like `#reload` which makes far more sense, but when typing short-hands for example I'm used to `#rq` reloading quests and that is now `#reload quest`.

This PR adds some hard-coded command aliasing for some nested commands.

* `#rq` equates to `#reload quests`
* `#size` equates to `#feature size`
* `#rl` equates to `#reload logs`

**Help Menu Improvement**

`#help <search>` now searches descriptions instead of just the commands themselves.

image

**Misc**

* Add `Strings::IsFloat` to check if a string is a float similar to `Strings::IsNumber`

**#size**

image

**#rq**

image

### DIFF
--- a/common/strings.cpp
+++ b/common/strings.cpp
@@ -200,6 +200,16 @@ bool Strings::IsNumber(const std::string &s)
 	}
 }
 
+bool Strings::IsFloat(const std::string &s)
+{
+	try {
+		auto r = stof(s);
+		return true;
+	}
+	catch (std::exception &) {
+		return false;
+	}
+}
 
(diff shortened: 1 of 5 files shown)


Kinglykrab (approved):

Kinglykrab:
Fantastic.

PR #2648, opened 2022-12-15
Add ground spawns to discovered items feature.

noudess:
Added code to allow ground spawns to be discovered items.

### DIFF
--- a/zone/object.cpp
+++ b/zone/object.cpp
@@ -535,6 +535,13 @@ bool Object::HandleClick(Client* sender, const ClickObject_Struct* click_object)
 			sender->PutItemInInventory(EQ::invslot::slotCursor, *m_inst, false);
 			sender->SendItemPacket(EQ::invslot::slotCursor, m_inst, ItemPacketTrade);
 
+			// Could be an undiscovered ground_spawn
+			if (m_ground_spawn && (RuleB(Character, EnableDiscoveredItems)))
+			{
+				if (!sender->GetGM() && !sender->IsDiscovered(item->ID))
+					sender->DiscoverItem(item->ID);
+			}
+
 			if(cursordelete)	// delete the item if it's a duplicate lore. We have to do this because the client expects the item packet
 				sender->DeleteItemInInventory(EQ::invslot::slotCursor);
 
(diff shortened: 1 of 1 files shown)


Kinglykrab (approved):

zone/object.cpp
```diff
@@ -535,6 +535,13 @@ bool Object::HandleClick(Client* sender, const ClickObject_Struct* click_object)
 			sender->PutItemInInventory(EQ::invslot::slotCursor, *m_inst, false);
 			sender->SendItemPacket(EQ::invslot::slotCursor, m_inst, ItemPacketTrade);
 
+			// Could be an undiscovered ground_spawn
+			if (m_ground_spawn && (RuleB(Character, EnableDiscoveredItems)))
+			{
+				if (!sender->GetGM() && !sender->IsDiscovered(item->ID))
```
Kinglykrab, line 541: Brackets.

noudess: Want a set around the one liner, or you want different indenting?

noudess: okies

nytmyr: I'm curious about this as well, I have noticed the recent commits have been bracketing one liners whereas most original code seems to not bracket.

I personally like no brackets for legibility but not opposed either way.

However, merge conflicts for custom code can become problematic if brackets are added to areas where there previously weren't. I had a ton of sections that threw errors where auto merges weren't adding either the first or second brackets if there were surrounding custom code. Easy fixes and it's only a custom code problem not related to master but just something I noticed.

fryguy503: Oneliners should always be bracketed due to prior history of issues related to implied brackets in complex code not being noticed. All PR's in the last roughly 2 years have been much more strict on overall formatting to keep a clean codebase.

Kinglykrab: > Oneliners should always be bracketed due to prior history of issues related to implied brackets in complex code not being noticed. All PR's in the last roughly 2 years have been much more strict on overall formatting to keep a clean codebase.

This. We haven’t always had the standards and leads to some issues over time where scope isn’t explicit. We’re trying to be cleaner about it now.

Aeadoin (approved):

PR #2655, opened 2022-12-20
[FMT] Bump to 7.1.3 from 5.3.0

Akkadius:
This bumps the [FMT](https://github.com/fmtlib/fmt) library to 7.1.3 - breaking our usage of `fmt::join` and swapped for our strings function.

More work will need to be done to get us to 9.1.0

This does at least shave us down some compile time roughly 10-20 seconds

```
# cpp17 with fmt 5.3.0 (3m7.506s)
# cpp17 with fmt 7.1.3 (2m56.933s)
```

### DIFF
--- a/common/repositories/character_expedition_lockouts_repository.h
+++ b/common/repositories/character_expedition_lockouts_repository.h
@@ -75,7 +75,7 @@ public:
 		Database& db, const std::vector<uint32_t>& character_ids,
 		const std::string& expedition_name, const std::string& ordered_event_name)
 	{
-		auto joined_character_ids = fmt::join(character_ids, ",");
+		auto joined_character_ids = Strings::Join(character_ids, ",");
 
 		auto results = db.QueryDatabase(fmt::format(SQL(
 			SELECT
--- a/common/repositories/dynamic_zone_members_repository.h
+++ b/common/repositories/dynamic_zone_members_repository.h
@@ -172,7 +172,7 @@ public:
 				DELETE FROM {}
 				WHERE dynamic_zone_id IN ({});
 			),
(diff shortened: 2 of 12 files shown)


KimLS:
Can we go over what exactly has broken about join that needs this change?

I see in 6.0:
`Disallowed passing the result of join as an lvalue to prevent misuse.`

If we're misusing join then perhaps we can just fix the places we're misusing it instead?

Akkadius:
> Can we go over what exactly has broken about join that needs this change?
> 
> I see in 6.0: `Disallowed passing the result of join as an lvalue to prevent misuse.`
> 
> If we're misusing join then perhaps we can just fix the places we're misusing it instead?

I'm not sure that we were actually "misusing" the result as we were using it as it was documented, but the compiler wasn't having it. Using it as an lvalue meant that we were using the memory address - which is strange. All uses of it were trying to simply use the string result.

Regardless - we have a common `Strings` utility class that covers all of our string needs, I was able to get `fmt::join` to play nice as being what actually processes the join logic for both of the types and also use templating to bind to multiple range types.

My guess is something was playing weird with compilation logic or static assertion with FMT and less of an issue with how we were using it. In the end the functionality should be no different now.

**Code**

```cpp
std::vector<uint32_t> id = {1, 2, 3, 4};
LogInfo("Here are some ids [{}]", Strings::Join(id, ","));
std::vector<std::string> ids = {"1", "2", "3", "4"};
LogInfo("Here are some ids [{}]", Strings::Join(ids, ","));
```

**Output**

```
./bin/world test:test
[World] [Info] Here are some ids [1,2,3,4]
[World] [Info] Here are some ids [1,2,3,4]
```

Akkadius:
Closing in favor of another branch

PR #2663, opened 2022-12-20
[Client] Remove unimplemented Client Insight Method.

Aeadoin:
This functions was never hooked up to anything/used anywhere,

If someone still wants it, it should probably be rewritten.

Also cleaned up more uses of Magic Numbers.


zone/mob.cpp
```diff
@@ -4057,37 +4057,6 @@ void Mob::SetNextIncHPEvent( int inchpevent )
 	nextinchpevent = inchpevent;
 }
```
Kinglykrab, line 4059: Any reason this was removed? Was it just not used anywhere?

Aeadoin: Believe you meant ```int16 Mob::GetResist(uint8 type) const``` but, yes removed as it's not used, was replaced by ```int Mob::GetResist(uint8 resist_type)``` that contains all of the newer resist types.

Kinglykrab (approved):

PR #2668, opened 2022-12-22
[Commands] Fix flymode command

Aeadoin:
Fixes Flymode command to properly show help if invalid arguments are supplied, also cleaned up some code.


Aeadoin:
Will reopen, and focus changes.

PR #2676, opened 2022-12-26
[Feature] Add Experience Gain Toggle.

Kinglykrab:
# Perl
- Add `$client->IsEXPEnabled()`.
- Add `$client->SetEXPEnabled(is_exp_enabled)`.

# Lua
- Add `client:IsEXPEnabled()`.
- Add `client:SetEXPEnabled(is_exp_enabled)`.

# Commands
- Add `#exptoggle [Toggle] - Toggle your or your target's experience gain.`.

# Notes
- Allows operators to turn on/off a player's experience gain individually without changing their rule values.
- The command allows operators to give players access to the command to disable their own experience gain.


zone/command.cpp
```diff
@@ -359,6 +359,7 @@ int command_init(void)
 		command_add("wpinfo", "Show waypoint info about your NPC target", AccountStatus::GMAreas, command_wpinfo) ||
 		command_add("worldwide", "Performs world-wide GM functions such as cast (can be extended for other commands). Use caution", AccountStatus::GMImpossible, command_worldwide) ||
 		command_add("xtargets", "[New Max XTargets] - Show your or your target's XTargets and optionally set max XTargets.", AccountStatus::GMImpossible, command_xtargets) ||
+		command_add("xptoggle", "[Toggle] - Toggle your or your target's experience gain.", AccountStatus::QuestTroupe, command_xptoggle) ||
```
Akkadius, line 362: Use full word “exp” vs “xp”

Kinglykrab: Sounds good.

Akkadius:
I’ll look closer tomorrow, about to 😴

Aeadoin (approved):
Not seeing any issues, only thing I was thinking was that exp_enabled would need to be initialized in the client constructor... but that's taken care of via: 

`memset(&m_epp, 0, sizeof(m_epp));` 

Will let Akk take another look before merging.

common/extprofile.h
```diff
@@ -56,6 +56,7 @@ struct ExtendedProfile_Struct {
 	int32				mercState; /* Not Used */
 	uint32				last_invsnapshot_time; /* Used */
 	uint32				next_invsnapshot_time; /* Used */
+	bool 				exp_enabled;
```
Akkadius, line 59: There's really no good reason why this should live here. This can simply be a private member on `Client` and the setters/getters can mutate it 

eg

```cpp
private:
bool m_exp_enabled = {};
```

zone/client.cpp
```diff
@@ -9136,6 +9136,25 @@ void Client::SetDevToolsEnabled(bool in_dev_tools_enabled)
...
+
+	c.exp_enabled = is_exp_enabled;
+
+	auto updated = CharacterDataRepository::UpdateOne(database, c);
+
+	if (!updated) {
+		return;
+	}
+
+	m_epp.exp_enabled = is_exp_enabled;
```
Akkadius, line 9155: Update to feedback above

zone/gm_commands/exptoggle.cpp
```diff
@@ -0,0 +1,28 @@
...
+	t->SetEXPEnabled(is_exp_enabled);
+
+	c->Message(
+		Chat::White,
+		fmt::format(
+			"Experience gain for {} is now {}abled.",
+			c->GetTargetDescription(t, TargetDescriptionType::LCSelf),
+			is_exp_enabled ? "en" : "dis"
+		).c_str()
+	);
```
Akkadius, line 27: Clean

zone/zonedb.cpp
```diff
@@ -1378,6 +1380,7 @@ bool ZoneDatabase::SaveCharacterData(uint32 character_id, uint32 account_id, Pla
 		Strings::Escape(pp->title).c_str(),						  // " title,                     "
 		Strings::Escape(pp->suffix).c_str(),						  // " suffix,                    "
 		pp->exp,						  // " exp,                       "
+		m_epp->exp_enabled,				  // " exp_enabled,               "
```
Akkadius, line 1383: We won't have access to the client object in this method so lets just pass the client pointer for now and pull things out of the client object within this save routine.

I'd avoid converting this to repositories for now

All of this client save stuff needs to get cleaned up

Kinglykrab: Yeah the reason I put it in extended profile is that we didn’t pass anything but player profile and extended profile to this method.

common/extprofile.h
Kinglykrab: Reason was we don’t pass client object to save method.

Akkadius (approved):
LGTM; assuming the character save changes were thoroughly tested

PR #2688, opened 2023-01-01
[Quest API] Add DoesAugmentFit() to Perl/Lua.

Kinglykrab:
# Perl
- Add `quest::does_augment_fit(item, augment_id)`.

# Lua
- Add `eq.does_augment_fit(item, augment_id)`.

# Notes
- Returns the augment slot index where the augment fits within the item instance provided, returns `-1` if it doesn't fit or has no open slot to fit in.


Aeadoin (approved):
LGTM

Issue #2500, opened 2022-10-26
Spell formula 122 (Splurt) should be able to be used beneficially as well as detrimentally.


MidnightMusicClub:
The click effect "Soul Well" on the Shissar Deathspeaker Staff should function as a "Splurt" style lifetap over time. As damage is done to an NPC, that same amount should be added to the caster's HP. The way it is now the recourse does damage to the caster instead of healing them.

Aeadoin:
I was able to get to "Soul Well Recourse" work by Setting it to the following:

[image](https://user-images.githubusercontent.com/109764533/210415259-c8f74fc3-ad27-4a55-a7fd-b341c3018b70.png)

Can see the Regen ticking up using #mystats (the client inventory though shows it broken)

[image](https://user-images.githubusercontent.com/109764533/210415080-ef2c6b6c-a46a-4d66-af4c-1939421f3fbc.png)

My actual recommendation is to ditch the Recourse on "Soul Well" and implement it like Live has, it looks like they ditched the recourse aspect, and made it a resource tap.

[image](https://user-images.githubusercontent.com/109764533/210415699-245485af-736f-4910-ac25-1b4347c518ee.png)

[image](https://user-images.githubusercontent.com/109764533/210415668-9fa579c9-8bea-434f-a262-d01034b0a4c8.png)

It looks like Live has a number of different Splurt related Formulas (Soul Well Recourse before Soul Well was changed to SPA 457, was using Formula 1012, and Soul Well itself uses 4012)

Isaaru:
@Aeadoin I like that. Do you know when live changed it?

Aeadoin:
Looks like they changed that a couple years ago. 

![image](https://user-images.githubusercontent.com/109764533/210636907-d7d16b8d-9f5e-4aa6-9790-450be50a0159.png)

and yes, that would make Soul Well regen stacks with those other abilities.

PR #2695, opened 2023-01-01
[Bug Fix] Fix Splurt Formula Causing Damage

Kinglykrab:
# Notes
- According to https://github.com/EQEmu/Server/issues/2500 the recourse causes damage instead of healing, this fixes that.

### DIFF
--- a/zone/spell_effects.cpp
+++ b/zone/spell_effects.cpp
@@ -3565,7 +3565,7 @@ snare has both of them negative, yet their range should work the same:
 			if(ticdif < 0)
 				ticdif = 0;
 
-			result = updownsign * (ubase - (12 * ticdif));
+			result = ubase - (12 * ticdif);
 			degenerating_effects = true;
 			break;
 		}


zone/spell_effects.cpp
```diff
@@ -3565,7 +3565,7 @@ snare has both of them negative, yet their range should work the same:
 			if(ticdif < 0)
 				ticdif = 0;
 
-			result = updownsign * (ubase - (12 * ticdif));
+			result = ubase - (12 * ticdif);
```
Aeadoin, line 3568: Removing this doesn't even seem to fix "Soul Well Recourse" it actually prevents the 1 method where it does work from working.
Without making this change I was able to get to "Soul Well Recourse" work by Setting it to the following:

image

Can see the Regen ticking up using #mystats (the client inventory though shows it broken)

image

My actual recommendation is to ditch the Recourse on  "Soul Well" and implement it like Live has, it looks like they ditched the recourse aspect, and made it a resource tap.

image

image

PR #2699, opened 2023-01-02
[Bug Fix] Fix #door Save

Kinglykrab:
# Notes
- Using `#door create` then `#door save` was overwriting doors instead of using the next highest ID.
- Remove the following unused methods.
```cpp
uint32 GetGuildEQID(uint32 guilddbid);
void UpdateDoorGuildID(int doorid, int guild_id);
int32 GetDoorsCount(uint32* oMaxID, const char *zone_name, int16 version);```

### DIFF
--- a/zone/doors.cpp
+++ b/zone/doors.cpp
@@ -697,7 +697,7 @@ uint32 ZoneDatabase::GetDoorsCountPlusOne()
 int ZoneDatabase::GetDoorsDBCountPlusOne(std::string zone_short_name, int16 version)
 {
 	const auto query = fmt::format(
-		"SELECT MAX(doorid) FROM doors "
+		"SELECT COALESCE(MAX(doorid), 0) FROM doors "
 		"WHERE zone = '{}' AND (version = {} OR version = -1)",
 		zone_short_name,
 		version
--- a/zone/doors.cpp
+++ b/zone/doors.cpp
@@ -704,7 +704,7 @@ int ZoneDatabase::GetDoorsDBCountPlusOne(std::string zone_short_name, int16 vers
 	);
 	auto results = QueryDatabase(query);
 	if (!results.Success() || !results.RowCount()) {
-		return 255;
+		return -1;
 	}
 
 	auto row = results.begin();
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: zone/doors.cpp [hunk 1]
```diff
@@ -685,82 +687,24 @@ void Doors::ToggleState(Mob *sender)
...
-}
-
-int32 ZoneDatabase::GetDoorsDBCountPlusOne(const char *zone_name, int16 version)
-{
-
-	uint32 oMaxID = 0;
-
-	std::string query   = StringFormat(
+	const auto query = fmt::format(
 		"SELECT MAX(doorid) FROM doors "
```
Akkadius, line 700: Max can be null, shouldn't assume `std::stoi` will convert safely on its own in this edge case

zone/doors.cpp
```diff
@@ -685,82 +687,24 @@ void Doors::ToggleState(Mob *sender)
...
-	auto        results = QueryDatabase(query);
-	if (!results.Success()) {
-		return -1;
-	}
-
-	if (results.RowCount() != 1) {
-		return -1;
+	auto results = QueryDatabase(query);
+	if (!results.Success() || !results.RowCount()) {
+		return 255;
```
Akkadius: 255 ?

Kinglykrab: Changed to `-1` like it was previously.

zone/doors.cpp
Kinglykrab: I'll use COALESCE here similar to the other methods I've seen.

zone/doors.cpp
```diff
@@ -685,81 +687,23 @@ void Doors::ToggleState(Mob *sender)
...
 }
 
-int32 ZoneDatabase::GetDoorsCountPlusOne(const char *zone_name, int16 version)
+int ZoneDatabase::GetDoorsDBCountPlusOne(std::string zone_short_name, int16 version)
 {
-	std::string query   = StringFormat(
-		"SELECT MAX(id) FROM doors WHERE zone = '%s' AND version = %u",
-		zone_name,
+	const auto query = fmt::format(
+		"SELECT COALESCE(MAX(doorid), 0) FROM doors "
```
Akkadius, line 700: Wouldn't you want this value to be at least `1` instead of `0`

Akkadius (approved):
Minor comment, lgtm

zone/doors.cpp
Kinglykrab: True enough.

PR #2705, opened 2023-01-05
[Rules] Add Rule to restrict too powerful NPC buffs.

noudess:
Added a new rule (false by default to retain existing behavior) to allow server ops to restrict NPC casted buffs using the same logic as PC restricted buffs.

Restructured the code a bit to centralize duplicate code and clarify.

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -3218,19 +3218,19 @@ bool Mob::CheckSpellLevelRestriction(Mob *caster, uint16 spell_id)
 	}
 
 	if (check_for_restrictions) {
-		int SpellLevel = GetMinLevel(spell_id);
+		int spell_level = GetMinLevel(spell_id);
 
 		// Only check for beneficial buffs
 		if (IsBuffSpell(spell_id) && IsBeneficialSpell(spell_id)) {
-			if (SpellLevel > 65) {
+			if (spell_level > 65) {
 				if (IsGroupSpell(spell_id) && GetLevel() < 62) {
 					can_cast = false;
 				}
 				else if (GetLevel() < 61) {
 					can_cast = false;
 				}
-			} else if (SpellLevel > 50) { // 51-65
-				if (GetLevel() < (SpellLevel / 2 + 15)) {
+			} else if (spell_level > 50) { // 51-65
+				if (GetLevel() < (spell_level / 2 + 15)) {
 					can_cast = false;
 				}
 			}
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -1153,7 +1153,7 @@ void Mob::InterruptSpell(uint16 message, uint16 color, uint16 spellid)
 	}
 
 	if(IsNPC()) {
-		CastToNPC()->AI_Event_SpellCastFinished(false, CastingSlot::Gem1);
+		CastToNPC()->AI_Event_SpellCastFinished(false, 1);
 	}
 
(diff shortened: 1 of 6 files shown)


mackal:
CheckSpellLevelRestriction needs to be virtual still.

noudess:
> CheckSpellLevelRestriction needs to be virtual still.

All npcs, mobs and clients all use the inherrited function now.   Is this a C++ thing I'm not getting?

noudess:
I'm also fixing a bug atm, so hold on any approvals.  Centralizaton of the code means I have to pass caster in.

noudess:
> CheckSpellLevelRestriction needs to be virtual still.

Fixed.  Also fixed pets version.

mackal:
> > CheckSpellLevelRestriction needs to be virtual still.
> 
> All npcs, mobs and clients all use the inherrited function now. Is this a C++ thing I'm not getting?

It should be possible to rewrite everything so it's not a virtual function, but I didn't see you touching the pets version, so I assumed you didn't move that into the single function :P

noudess:
> > > CheckSpellLevelRestriction needs to be virtual still.
> > 
> > 
> > All npcs, mobs and clients all use the inherrited function now. Is this a C++ thing I'm not getting?
> 
> It should be possible to rewrite everything so it's not a virtual function, but I didn't see you touching the pets version, so I assumed you didn't move that into the single function :P

Yeah, because I'm an idiot and didn't realize at first that I could not move the IsClient() checks into the function without adding an argument.  Then, once I did, I realized I couldn't let the pet function alone.

I think I got it right now.  Looking at the pet version, it seemed broken before I touched it.  The function is supposed to check the level of (this) against the spells.  Instead, the pet version calls the owner version and uses the owners level.  I left that alone and just fixed it to pass in the caster for the GM check.

zone/spells.cpp
```diff
@@ -3211,29 +3201,50 @@ int Mob::CheckStackConflict(uint16 spellid1, int caster_level1, uint16 spellid2,
...
-		if (SpellLevel > 65) {
-			if (IsGroupSpell(spell_id) && GetLevel() < 62)
-				return false;
-			else if (GetLevel() < 61)
-				return false;
-		} else if (SpellLevel > 50) { // 51-65
-			if (GetLevel() < (SpellLevel / 2 + 15))
-				return false;
+	if (check_for_restrictions) {
+		int SpellLevel = GetMinLevel(spell_id);
```
Kinglykrab, line 3221: Camel case, `spell_level`.

noudess: Fixed

noudess:
I have tested the various cases and I feel this is now working as intended, after fixing the messaging.

noudess:
I found that with the rule off - it all works - so no harm to existing code.  With the rule on, it works, but perhaps stops the casting NPC for awhile - which is clearly a bug.  I need to work out what impact this is having on AI so I changed it to draft status.

noudess:
Ok, I found that if an NPC spell is deemed non-castable for any reason, prior to casting_spell_id being set, it kills off the AI session for that NPC.  Perhaps the best solution is to better organize the entire spell casting stuff so that this variable, as well as the slot are set earlier.  However, it seems fragile to me.  Instead we simply reengage the AI engine if InterruptSpell is called for any NPC.

noudess:
I want to either fix this correctly or not at all.  Closing and will resubmit if I have a cleaner solution.

PR #2708, opened 2023-01-07
[Rule] Add rule for NPC Level Based Buff Restrictions.

noudess:
Added a rule that allows server ops to apply "SPELL_TOO_POWERFUL" rule to NPCs just like PCs.

NPCBuffLevelRestrictions (default off to retain existing functionality)

While debuging this I found that NPC spells from the AI that fail in the CastSpell() portion of the spell casting chain do not recover ever - as the code relies on the casting_spell_id  and the spell slot to be valid in every function that calls the AI with AI_Event_SpellCastFinished.  As these variables are not set until later in the casting chain, the  AI was never getting reset.

This bug was much more obvious when setting my new rule, as NPC spells fail in the checking stage alot more.

This is a DRAFT PR until someone more adept with spells than I ok's the change.

I did test all the variant cases and it works, and I put a log locally if an NPC ever cast a spell not using slot 1 and I didnt see any.  Not sure that matters, but the AI only resets on slot 1, and since slot was not set, I pass in 1.

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -3215,9 +3215,14 @@ bool Mob::CheckSpellLevelRestriction(Mob *caster, uint16 spell_id)
 		return false;
 	}
 
+	if (caster->IsClient() && caster->CastToClient()->GetGM()) {
+		LogSpells("[CheckSpellLevelRestriction] GM casting - No restrictions");
+		return true;
+	}
+
 	// NON GM clients might be restricted by rule setting
 	if (caster->IsClient()) {
-		if (RuleB(Spells, BuffLevelRestrictions) && !caster->CastToClient()->GetGM()) {
+		if (RuleB(Spells, BuffLevelRestrictions)) {
 			check_for_restrictions = true;
 		}
 	}
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -1149,7 +1149,7 @@ void Mob::InterruptSpell(uint16 message, uint16 color, uint16 spellid)
 			bard_song_mode = true;
 		} else {
 			spellid = casting_spell_id;
-		} 
+		}
 	}
 
 	LogSpells("Interrupt: casting_spell_id [{}] casting_spell_slot [{}]",
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -734,10 +734,6 @@ bool Mob::DoCastingChecksOnTarget(bool check_on_casting, int32 spell_id, Mob *sp
 	*/
 
 	bool ignore_on_casting = false;
-	bool ignore_if_npc_or_gm = false;
-	if (!IsClient() || (IsClient() && CastToClient()->GetGM())) {
-		ignore_if_npc_or_gm = true;
-	}
 
 	if (check_on_casting){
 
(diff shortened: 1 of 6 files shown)


noudess:
Tested all iterations and I am happy with results.

Akkadius:
Will look at this one later today

zone/spells.cpp
```diff
@@ -3211,29 +3205,55 @@ int Mob::CheckStackConflict(uint16 spellid1, int caster_level1, uint16 spellid2,
...
+		check_for_restrictions = true;
+	}
 
-	return true;
+	if (check_for_restrictions) {
+		int spell_level = GetMinLevel(spell_id);
+
+		// Only check for beneficial buffs
+		if (IsBuffSpell(spell_id) && IsBeneficialSpell(spell_id)) {
+			if (spell_level > 65) {
```
Akkadius, line 3234: Can we use 65 as a constant, magic number

Akkadius: Same with the others below

Akkadius: At the very least somewhere in `eq_constants.h`

zone/spells.cpp
```diff
@@ -3211,29 +3205,55 @@ int Mob::CheckStackConflict(uint16 spellid1, int caster_level1, uint16 spellid2,
...
-			if (IsGroupSpell(spell_id) && GetLevel() < 62)
-				return false;
-			else if (GetLevel() < 61)
-				return false;
-		} else if (SpellLevel > 50) { // 51-65
-			if (GetLevel() < (SpellLevel / 2 + 15))
-				return false;
+	// NON GM clients might be restricted by rule setting
+	if (caster->IsClient()) {
+		if (RuleB(Spells, BuffLevelRestrictions) && !caster->CastToClient()->GetGM()) {
```
Akkadius, line 3220: We should emit a log when this logic checks GM

So many times people waste hours troubleshooting a non-issue when they realize an effect wasn't working because they just had the GM flag on.

Something in `LogSpells` at the very least

zone/spells.cpp
```diff
@@ -3211,29 +3205,55 @@ int Mob::CheckStackConflict(uint16 spellid1, int caster_level1, uint16 spellid2,
...
-	return true;
-}
+	bool check_for_restrictions = false;
+	bool can_cast = true;
 
-bool Client::CheckSpellLevelRestriction(uint16 spell_id)
-{
-	int SpellLevel = GetMinLevel(spell_id);
+	if (!caster) {
+		LogSpells("CheckSpellLevelRestriction: No caster");
```
Akkadius, line 3214: Prefix log with `[CheckSpellLevelRestriction]` please vs `CheckSpellLevelRestriction:`

Akkadius:
So if I'm understanding your PR comments correctly, this both fixes a bug with AI logic pausing entirely and separately adds a rule for additional behavior?

Code looks fairly straightforward - I realize that my comments around constants are not your code originally, but we should try to make improvements while we're in the context of the changes

Would like someone versed in this area to also take an additional look, otherwise LGTM

noudess:
> So if I'm understanding your PR comments correctly, this both fixes a bug with AI logic pausing entirely and separately adds a rule for additional behavior?
> 
> Code looks fairly straightforward - I realize that my comments around constants are not your code originally, but we should try to make improvements while we're in the context of the changes
> 
> Would like someone versed in this area to also take an additional look, otherwise LGTM

Yes, I found the bug and had to fix it for my new feature to work.  I did ping Kayen to take a peek.

zone/spells.cpp
noudess: If I add this, it will be a LogMessage that comes out every time a GM casts a spell when RuleB(Spells, BuffLevelRestrictions is set, because the code does not restrict GMS regardless of the rule.  Do you want that?

Akkadius: Doesn't hurt if its in a categorized log (Spells)

zone/spells.cpp
noudess: I'm happy to do this, but any names I come up with make the code less easy to understand IMO.  Like 65 is what, UPPER_BENEFICIAL_SPELL_LIMIT_THRESHOLD maybe?  And then 61 and 62 and 50?  I could use similar names but it feels like they make it less readable?

zone/spells.cpp
noudess: Done.

zone/spells.cpp
noudess: Done.

Akkadius (approved):

PR #2714, opened 2023-01-09
[Zones] Add Max Level Check to Zones.

Kinglykrab:
# Perl
- Add `$client->CanEnterZone(zone_short_name)`.
- Add `$client->CanEnterZone(zone_short_name, instance_version)`.

# Lua
- Add `client:CanEnterZone(zone_short_name)`.
- Add `client:CanEnterZone(zone_short_name, instance_version)`.

# Rules
- Add `World:FailedRequirementBootZoneID`.

# Notes
- Allows operators to set a boot zone for when players fail requirements and somehow have entered the zone anyway.
- Allows operators to limit zones to a maximum level.
- Allows operators to see if a player can enter a zone before sending them.
- Keeps players from entering via `#peqzone` and `#zone` as well if they do not meet the requirements or are not high enough status to bypass the requirements.

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -904,9 +904,9 @@ void Client::CompleteConnect()
 	heroforge_wearchange_timer.Start(250);
 
 	// enforce some rules..
-	if (!CanBeInZone()) {
+	if (!CanEnterZone()) {
 		LogInfo("Kicking character [{}] from zone, not allowed here (missing requirements)", GetCleanName());
-		GoToSafeCoords(ZoneID("arena"), 0);
+		GoToSafeCoords(RuleI(World, FailedRequirementBootZoneID), 0);
 		return;
 	}
 }
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -906,7 +906,7 @@ void Client::CompleteConnect()
 	// enforce some rules..
 	if (!CanEnterZone()) {
 		LogInfo("Kicking character [{}] from zone, not allowed here (missing requirements)", GetCleanName());
-		GoToSafeCoords(RuleI(World, FailedRequirementBootZoneID), 0);
+		GoToBind();
 		return;
 	}
 }
--- a/zone/zoning.cpp
+++ b/zone/zoning.cpp
@@ -1215,7 +1215,7 @@ void Client::SetPEQZoneFlag(uint32 zone_id) {
 	}
 }
 
-bool Client::CanBeInZone() {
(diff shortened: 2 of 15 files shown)


zone/client_packet.cpp
```diff
@@ -904,9 +904,9 @@ void Client::CompleteConnect()
 	heroforge_wearchange_timer.Start(250);
 
 	// enforce some rules..
-	if (!CanBeInZone()) {
+	if (!CanEnterZone()) {
 		LogInfo("Kicking character [{}] from zone, not allowed here (missing requirements)", GetCleanName());
-		GoToSafeCoords(ZoneID("arena"), 0);
+		GoToSafeCoords(RuleI(World, FailedRequirementBootZoneID), 0);
```
Akkadius, line 909: It's possible that the client still can't even enter what's configured for a default here.

It might make sense to have it reroute them to bind

zone/zoning.cpp
```diff
@@ -1224,38 +1224,55 @@ bool Client::CanBeInZone() {
...
 	}
 
-	safe_x       = z->safe_x;
-	safe_y       = z->safe_y;
-	safe_z       = z->safe_z;
-	safe_heading = z->safe_heading;
-	min_status   = z->min_status;
-	min_level    = z->min_level;
+	if (GetLevel() < z->min_level) {
+		LogDebug(
```
Akkadius, line 1237: Let's change all of these `LogDebug`'s to `LogInfo`

`LogDebug` is nice during development but it should probably live somewhere, either a category oriented log sink or just the default "Info"

This information would be useful to at the very minimum see in the info category

zone/zoning.cpp
```diff
@@ -1215,7 +1215,7 @@ void Client::SetPEQZoneFlag(uint32 zone_id) {
 	}
 }
 
-bool Client::CanBeInZone() {
```
Akkadius, line 1218: It's great we're aggregating validation in one function for this, cleans things up.

I'm thinking this should be used in `Handle_OP_ZoneChange` since that is where we send zone cancels if they do not meet requirements.

We already duplicate some of this logic there and it would make sense to have it call it there as well.

I would say that would eliminate the need to have the checks in the commands, but when zone cancel is called - it actually sends the client back to safe coordinates and sometimes drops them from the ceiling which is another issue that we need to trace down and fix.

Regardless, I would add a check of `CanBeInZone` in `Handle_OP_ZoneChange`

Akkadius (approved):
Clean. Minor comments, LGTM

Kinglykrab:
Will consolidate code this afternoon.

PR #2736, opened 2023-01-15
[Lua] Resolve stoi Exception

Akkadius:
### What

It's been reported by numerous operators since https://github.com/EQEmu/Server/pull/2635 that Lua is throwing exceptions for `stoi` calls within the source. While this may not necessarily be breaking anything, it can cause Lua to throw exceptions and become noise to operators.

image

### Solution

This may not fix all places that this occurs, but a common place where `std::stoi` is called and many assumptions are made about inputs are in all of the event bindings where we pass in inputs from event calls. To cleanly mitigate this issue for now, I've added a strings wrapper that will detect for a validity of a string before providing a fallback `default -1` and a fallback can be provided in general for other use cases.

### DIFF
--- a/zone/lua_parser_events.cpp
+++ b/zone/lua_parser_events.cpp
@@ -120,12 +120,12 @@ void handle_npc_event_hp(
 	if(extra_data == 1) {
 		lua_pushinteger(L, -1);
 		lua_setfield(L, -2, "hp_event");
-		lua_pushinteger(L, std::stoi(data));
+		lua_pushinteger(L, Strings::Stoi(data));
 		lua_setfield(L, -2, "inc_hp_event");
 	}
 	else
 	{
-		lua_pushinteger(L, std::stoi(data));
+		lua_pushinteger(L, Strings::Stoi(data));
 		lua_setfield(L, -2, "hp_event");
 		lua_pushinteger(L, -1);
 		lua_setfield(L, -2, "inc_hp_event");
@@ -191,7 +191,7 @@ void handle_npc_task_accepted(
 	l_client_o.push(L);
 	lua_setfield(L, -2, "other");
 
-	lua_pushinteger(L, std::stoi(data));
+	lua_pushinteger(L, Strings::Stoi(data));
 	lua_setfield(L, -2, "task_id");
 }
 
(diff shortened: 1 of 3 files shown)


Kinglykrab:
Should we name it `Strings::ToInt` like `Strings::ToBool`? May need one for `std::stoul` as well.

Akkadius:
Going to kill this for better error handling

Kinglykrab:
👍

Akkadius:
Re-opened this as this is largely the cleanest solution and best developer experience server side and quest side

The only risks this could pose is that if any event is potentially not passing values, almost all events pass values and don't have optional inputs to exports. 

I went through all exports and events and there's nothing that I'm concerned about

xackery:
Just some observations:
https://github.com/EQEmu/Server/blob/df0d5f92a419cf6c2e753c2852e19a6a9482778a/zone/lua_parser_events.cpp#L1231
appears to still use std:stoi

Also any thoughts on std::stoll and std:stoul wrappers too? Kingly noted it actually earlier, but figured i'd bring it up while on topic.

xackery (approved):
Looks good!

Kinglykrab (approved):
Still curious if we need this for `uint32`, `int64`, and `uint64` support.

zone/lua_parser_events.cpp
```diff
@@ -227,7 +227,7 @@ void handle_npc_waypoint(
 	l_mob_o.push(L);
 	lua_setfield(L, -2, "other");
 
-	lua_pushinteger(L, std::stoi(data));
+	lua_pushinteger(L, Strings::Stoi(data, -1));
```
Kinglykrab, line 230: Do we need the `-1` since this is the default value anyway?

Akkadius:
> Still curious if we need this for `uint32`, `int64`, and `uint64` support.

Probably wouldn't hurt

zone/lua_parser_events.cpp
Akkadius: 0 appears to be the default value of a waypoint unless I'm misreading

Kinglykrab: I meant that the default second parameter of `Strings::Stoi` is `-1`.

Akkadius: Ahh, good call. I remember making this 0 before. It's already been weeks and forgot what I had done. The fallback should be 0 by default

Kinglykrab: Awesome, sounds good.

Issue #338, opened 2015-01-15
Replace abs, abs64, ABS, fabs, etc with std::abs


mackal:
We should now be using compilers that have all the correct C++11 overloads for these functions. Most of the 'cheap abs' macros will probably not actually matter with optimizations enabled.

Benefit? No silent narrowing or silly macros.

Kinglykrab:
Resolved by https://github.com/EQEmu/Server/pull/2739 unless there's more I don't see.

PR #2739, opened 2023-01-15
[Cleanup] Make use of std::abs where possible.

Kinglykrab:
# Notes
- Resolves https://github.com/EQEmu/Server/issues/338


Akkadius (approved):
<3

PR #2744, opened 2023-01-15
[Bug Fix] HasPet() Zone Crashes

Aeadoin:
Fixes numerous Zone Crashes that I've seen on KMRA recently where null instance is passed into methods.

here's some where HasPet() returns true, but GetPet() returns null... causing a crash.

[crash_lopingplains_version_0_inst_id_0_port_7061_684812.log](https://github.com/EQEmu/Server/files/10421306/crash_lopingplains_version_0_inst_id_0_port_7061_684812.log)
[crash_oldfieldofbone_version_0_inst_id_0_port_7033_3364593.log](https://github.com/EQEmu/Server/files/10421307/crash_oldfieldofbone_version_0_inst_id_0_port_7033_3364593.log)
[crash_lopingplains_version_0_inst_id_0_port_7043_1953522.log](https://github.com/EQEmu/Server/files/10421308/crash_lopingplains_version_0_inst_id_0_port_7043_1953522.log)

### DIFF
--- a/zone/npc.cpp
+++ b/zone/npc.cpp
@@ -3573,8 +3573,8 @@ void NPC::AIYellForHelp(Mob *sender, Mob *attacker)
 
 	LogAIYellForHelp(
 		"NPC [{}] ID [{}] is starting to scan",
-		(!this ? "NULL MOB" : GetCleanName()),
-		(!this ? 0 : GetID())
+		GetCleanName(),
+		GetID()
 	);
 
 	for (auto &close_mob : entity_list.GetCloseMobList(sender)) {
--- a/zone/npc.cpp
+++ b/zone/npc.cpp
@@ -3573,8 +3573,8 @@ void NPC::AIYellForHelp(Mob *sender, Mob *attacker)
 
 	LogAIYellForHelp(
 		"NPC [{}] ID [{}] is starting to scan",
-		GetCleanName(),
-		GetID()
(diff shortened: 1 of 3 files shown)


zone/npc.cpp
```diff
@@ -3573,8 +3573,8 @@ void NPC::AIYellForHelp(Mob *sender, Mob *attacker)
 
 	LogAIYellForHelp(
 		"NPC [{}] ID [{}] is starting to scan",
-		GetCleanName(),
-		GetID()
+		(!this ? "NULL MOB" : GetCleanName()),
```
mackal, line 3576: This change shouldn't be necessary. `this` can never be null in well formed code, so something else has to be the cause of the crash if it's crashing here (or maybe the line numbers were wrong?)


(checking `this` against null I guess was a thing back in the 90s, but that's because the compilers/implementations were special ...)

Aeadoin: Okay, that makes sense. not sure what else it would have been there. like you said, maybe the line numbers are different now, since this crash was last week.

zone/pets.cpp
```diff
@@ -529,6 +529,24 @@ Mob* Mob::GetPet() {
 	return(tmp);
 }
 
+bool Mob::HasPet() const {
+
```
Kinglykrab, line 533: Extra new line here.

mackal (approved):

Kinglykrab (approved):

PR #2746, opened 2023-01-15
[Bug Fix] Fix ST_TargetsTarget Spells with Restrictions

Aeadoin:
This fixes ST_TargetsTarget Spells that are triggered from spell effects such as SE_ApplyEffect or SE_TriggerCast.

This was causing spells like Remote Manaflux which triggers a healing spell "Manaflux Enervation" on the TargetsTarget to fail as it has a restriction of "PCs and Mercs". It was doing the Target Restrictions against the original target of "Remote Manaflux" not the TargetsTarget (Client, Bot, Merc)


Kinglykrab (approved):
:1st_place_medal:

PR #2749, opened 2023-01-17
[Bug Fix] Move EVENT_SPAWN for adding NPCs back to original spot, also add NPCs…

Natedog2012:
… to npc/mob list before parsing

Allows you to change NPCs before spawning rather than after


Kinglykrab (approved):

mackal:
Yeah, PEQ takes advantage of this happening where it used ... (mageSwarm pets)

Kinglykrab:
> Yeah, PEQ takes advantage of this happening where it used ... (mageSwarm pets)

Was an issue with `$npc` being undefined in spots and the previous code the NPC wasn't always a part of `$entity_list`.

PR #2751, opened 2023-01-17
Update Gitignore, and untrack Visual Studio Code settings.json file.

Aeadoin:
@xackery Adding VSCode to Gitignore, and untracking the settings.json file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,8 +55,12 @@ bin/
 /x64
 /client_files/**/CMakeFiles/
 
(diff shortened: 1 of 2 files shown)


xackery:
@Aeadoin .vscode can be versioned, just your settings.json isn't good.
Like the tasks I added work with the .devcontainer/ folder.
But they aren't actively maintained, so we may want to consider removing them too..

xackery:
This can be closed as it is replaced with #2753 2753

PR #2752, opened 2023-01-17
[C++20] Arithmetic on different enums is deprecated on C++20

mackal:
Just do some casts and add some constants to make life easier

### DIFF
--- a/common/rulesys.cpp
+++ b/common/rulesys.cpp
@@ -34,7 +34,7 @@ const char *RuleManager::s_categoryNames[_CatCount + 1] = {
 	"InvalidCategory"
 };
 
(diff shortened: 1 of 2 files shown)


mackal:
```
common/rulesys.cpp:539:57: warning: arithmetic between different enumeration types ‘RuleManager::IntType’ and ‘RuleManager::RealType’ is deprecated [-Wdeprecated-enum-enum-conversion]
  539 |                         return s_RuleInfo[_IntRuleCount + _RealRuleCount + _BoolRuleCount];
```

example of warnings.

Akkadius (approved):
LGTM

Tested locally, doesn't seem to blow anything up

PR #2753, opened 2023-01-17
[Code] Removed vscode settings

xackery:
This removes the .vscode and .devcontainer folders


Aeadoin (approved):

Akkadius (approved):
Makes sense to me.

If there are docs that folks can create as to show what works best for them in their development flows - that would help empower others

Kinglykrab:
Should we add `.devcontainer` to ignore as well?

PR #2754, opened 2023-01-17
Add HasItemOnCorpse to quest APIs

mackal:
Main motivation is PEQ's client_ext HasItem. The current API has SIGNIFICANT overhead and lacks the ability to search augs, which the rest of the HasItem searches do.

### DIFF
--- a/zone/zonedb.cpp
+++ b/zone/zonedb.cpp
@@ -4071,6 +4071,19 @@ uint32 ZoneDatabase::GetCharacterCorpseItemAt(uint32 corpse_id, uint16 slotid) {
 	return itemid;
 }
 
+bool ZoneDatabase::CharacterHasItemOnCorpse(uint32 corpse_id, uint32 item_id) {
+	// just returns true/false, we don't check count or anything
+	auto query = fmt::format(
+	    "SELECT ci.item_id FROM character_corpse_items ci INNER JOIN character_corpses cc on "
+	    "cc.id = ci.corpse_id WHERE cc.charid = {0} AND (ci.item_id = {1} OR ci.aug_1 = {1} OR ci.aug_2 "
+	    "= {1} OR ci.aug_3 = {1} OR ci.aug_4 = {1} OR ci.aug_5 = {1} OR ci.aug_6 = {1}) LIMIT 1",
+	    corpse_id, item_id);
+	auto results = QueryDatabase(query);
+	if (results.Success() && results.RowCount() > 0)
+		return true;
+	return false;
+}
+
 bool ZoneDatabase::LoadCharacterCorpseData(uint32 corpse_id, PlayerCorpse_Struct* pcs){
 	std::string query = StringFormat(
 		"SELECT           \n"
--- a/zone/zonedb.cpp
+++ b/zone/zonedb.cpp
@@ -4071,13 +4071,13 @@ uint32 ZoneDatabase::GetCharacterCorpseItemAt(uint32 corpse_id, uint16 slotid) {
 	return itemid;
 }
 
-bool ZoneDatabase::CharacterHasItemOnCorpse(uint32 corpse_id, uint32 item_id) {
+bool ZoneDatabase::CharacterHasItemOnCorpse(uint32 charid, uint32 item_id) {
 	// just returns true/false, we don't check count or anything
 	auto query = fmt::format(
 	    "SELECT ci.item_id FROM character_corpse_items ci INNER JOIN character_corpses cc on "
 	    "cc.id = ci.corpse_id WHERE cc.charid = {0} AND (ci.item_id = {1} OR ci.aug_1 = {1} OR ci.aug_2 "
 	    "= {1} OR ci.aug_3 = {1} OR ci.aug_4 = {1} OR ci.aug_5 = {1} OR ci.aug_6 = {1}) LIMIT 1",
-	    corpse_id, item_id);
+	    charid, item_id);
 	auto results = QueryDatabase(query);
 	if (results.Success() && results.RowCount() > 0)
 		return true;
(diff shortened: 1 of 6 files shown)


zone/zonedb.cpp
```diff
@@ -4071,6 +4071,19 @@ uint32 ZoneDatabase::GetCharacterCorpseItemAt(uint32 corpse_id, uint16 slotid) {
...
+bool ZoneDatabase::CharacterHasItemOnCorpse(uint32 corpse_id, uint32 item_id) {
+	// just returns true/false, we don't check count or anything
+	auto query = fmt::format(
+	    "SELECT ci.item_id FROM character_corpse_items ci INNER JOIN character_corpses cc on "
+	    "cc.id = ci.corpse_id WHERE cc.charid = {0} AND (ci.item_id = {1} OR ci.aug_1 = {1} OR ci.aug_2 "
+	    "= {1} OR ci.aug_3 = {1} OR ci.aug_4 = {1} OR ci.aug_5 = {1} OR ci.aug_6 = {1}) LIMIT 1",
+	    corpse_id, item_id);
+	auto results = QueryDatabase(query);
+	if (results.Success() && results.RowCount() > 0)
+		return true;
```
Kinglykrab, line 4083: Brackets for explicit scoping.

zone/zonedb.h
```diff
@@ -454,6 +454,7 @@ class ZoneDatabase : public SharedDatabase {
 	uint32		GetCharacterCorpseCount(uint32 char_id);
 	uint32		GetCharacterCorpseID(uint32 char_id, uint8 corpse);
 	uint32		GetCharacterCorpseItemAt(uint32 corpse_id, uint16 slotid);
+	bool		CharacterHasItemOnCorpse(uint32 charid, uint32 item_id);
```
Kinglykrab, line 457: Have it as charid here but corpseid in the method itself.

zone/zonedb.cpp
```diff
@@ -4071,6 +4071,19 @@ uint32 ZoneDatabase::GetCharacterCorpseItemAt(uint32 corpse_id, uint16 slotid) {
 	return itemid;
 }
 
+bool ZoneDatabase::CharacterHasItemOnCorpse(uint32 corpse_id, uint32 item_id) {
```
Kinglykrab, line 4074: Have it as corpse_id here but charid in header.

Kinglykrab (commented):
Should we have a quest::/eq. method that takes character ID so we can check other characters than current client reference? The method itself already allows for this.

mackal:
> Should we have a quest::/eq. method that takes character ID so we can check other characters than current client reference? The method itself already allows for this.

It's designed as an alternative to GetCorpseItemAt, which isn't exported like that.

Akkadius:
I appreciate what this PR is trying to do - but checking corpses really needs to occur in memory to make a meaningful dent in the problems we've had with checking for existence of items on corpses. 

Myself and others have wasted dozens of hours of time troubleshooting lag bomb problems that ended up being corpse item checks in nested loops that are querying the database causing massive lag spikes.

**Examples**

- Example in a player event script, checking for the existence of an item when a player dies can hang a zone for a few seconds because that one player generates 100's of queries. While the server hardware might be able to absorb some of that for that one player, what about a raid wipe? 10 clients just die, we have 1,000's of queries and a zone that hangs for a minute.
- Checking for an existence of an item while zoning into Wall of Slaughter on PEQ - same issue
- Back in the day we had a `HasItem` check in Plane of Knowledge that would bring the zone to a crawl

People will unknowingly continue to shoot themselves in the foot by continuing to use `HasItem` checks until we come up with a more meaningful solution that doing database hits here.

`HasItem` checks memory for every other inventory type outside of corpses. Corpses are just an extension of a players inventory

mackal:
Queries aren't the problem here. https://github.com/EQEmu/Server/blob/bd29f1c5bb1655b089667d86f2926c2ce96d20e6/zone/zonedb.cpp#L4063 ends up creating a corpse, loading it from the DB, getting the item ID of the item in the requested slot, deallocating the corpse, returning the item ID. Then doing that for EVERY slot. It's thrashing the shit out of memory, something just querying the data won't do.

This code simply does a query, with a limit of 1 since we just want a yes/no. It should be significantly faster, maybe someone has well over 1000 corpses it could be an issue, but the current code runs into problems far sooner.

Akkadius:
> Queries aren't the problem here.
> 
> https://github.com/EQEmu/Server/blob/bd29f1c5bb1655b089667d86f2926c2ce96d20e6/zone/zonedb.cpp#L4063
> 
> ends up creating a corpse, loading it from the DB, getting the item ID of the item in the requested slot, deallocating the corpse, returning the item ID. Then doing that for EVERY slot. It's thrashing the shit out of memory, something just querying the data won't do.
> This code simply does a query, with a limit of 1 since we just want a yes/no. It should be significantly faster, maybe someone has well over 1000 corpses it could be an issue, but the current code runs into problems far sooner.

I'm sorry but memory trashing even in this instance doesn't show up on a performance radar, it's the queries that do.

I agree this is less expensive than what is being done currently, but this should all still be occurring in memory

mackal:
At the end of the day, "just checking in memory" is a SIGNIFICANT task at the moment. Since the corpse items aren't currently in your inventory server side. So either we can get this solved now with this or just deal with the horrible problem until the inventory is rewritten in a few years that lets us "just check in memory"

Akkadius:
> At the end of the day, "just checking in memory" is a SIGNIFICANT task at the moment. Since the corpse items aren't currently in your inventory server side. So either we can get this solved now with this or just deal with the horrible problem until the inventory is rewritten in a few years that lets us "just check in memory"

It really isn't that difficult to do, doesn't require "an entire inventory rewrite". I already told you earlier this is quite an improvement from what is currently done. It is worth swapping out existing plugin checks with this but my comment still stands we should work towards removing the database hit for checking corpses even if that means a separate effort. 

All I'm looking for is acknowledgement and being a team player here, not excuses, "quoting things" in passive aggression.

Please drop the passive aggressiveness and be professional. 

What are your plans for incorporating this into the plugin logic that we have today?

PR #2757, opened 2023-01-19
 [Bots] Cleanup and remove preprocessors.

Kinglykrab:
# Notes
- Removes every `#ifdef BOTS` we have and locks bots behind `Bots:AllowBots` rule.
- Bot updates are now done by default similar to regular database updates.
- Modify `CMakeLists.txt`, `.drone.yml`, and `BUILD.md` to match the removal of `EQEMU_ENABLE_BOTS`.


zone/main.cpp
```diff
@@ -604,9 +600,9 @@ int main(int argc, char** argv) {
 	safe_delete(task_manager);
 	safe_delete(npc_scale_manager);
 	command_deinit();
-#ifdef BOTS
-	bot_command_deinit();
-#endif
+	if (RuleB(Bots, AllowBots)) {
```
Akkadius, line 603: Nit: `AllowBots` seems strange, how about `Bots:BotsEnabled`

common/guild_base.cpp
```diff
@@ -866,20 +866,11 @@ bool BaseGuildManager::QueryWithLogging(std::string query, const char *errmsg) {
 	return(true);
 }
 
-//factored out so I dont have to copy this crap.
-#ifdef BOTS
 #define GuildMemberBaseQuery \
```
Akkadius, line 869: This should still if/else considering this relies on views and different tables depending on whether bots are enabled or not

We need to eliminate these views in code and need to be substituted for queries

common/guild_base.cpp
```diff
@@ -969,12 +960,7 @@ bool BaseGuildManager::GetCharInfo(uint32 char_id, CharGuildInfo &into) {
 	}
 
 	//load up the rank info for each guild.
-	std::string query;
-#ifdef BOTS
```
Akkadius, line 973: This should still if/else with the rule

common/ruletypes.h
```diff
@@ -585,9 +585,8 @@ RULE_INT(Range, CriticalDamage, 80, "The packet range in which critical hit mess
 RULE_INT(Range, MobCloseScanDistance, 600, "Close scan distance")
 RULE_CATEGORY_END()
 
-
-#ifdef BOTS
 RULE_CATEGORY(Bots)
+RULE_BOOL(Bots, AllowBots, false, "Enable of disable bot functionality, default is false")
```
Akkadius, line 589: Nit: `AllowBots` seems strange, how about `Bots:Enabled`

zone/spell_effects.cpp
```diff
@@ -6819,12 +6811,9 @@ void Mob::CheckNumHitsRemaining(NumHit type, int32 buff_slot, uint16 spell_id)
 	} else if (type == NumHit::MatchingSpells) {
 		if (buff_slot >= 0) {
 			if (--buffs[buff_slot].hit_number == 0) {
-
-#ifdef BOTS
 				buff_name = spells[buffs[buff_slot].spellid].name;
```
Akkadius, line 6814: This seems to change how this logic fundamentally works when bots are enabled, at the very least rule gate this still or lets ask why this behaves differently when bots are enabled because it affects any mob

zone/spell_effects.cpp
```diff
@@ -6789,23 +6785,19 @@ void Mob::CheckNumHitsRemaining(NumHit type, int32 buff_slot, uint16 spell_id)
...
-#endif
 
 	//Spell specific procs [Type 7,10,11]
 	if (IsValidSpell(spell_id)) {
 		for (int d = 0; d < buff_max; d++) {
 			if (buffs[d].spellid == spell_id && buffs[d].hit_number > 0 &&
 			    spells[buffs[d].spellid].hit_number_type == static_cast<int>(type)) {
 
-#ifdef BOTS
 				buff_name = spells[buffs[d].spellid].name;
```
Akkadius, line 6798: This seems to change how this logic fundamentally works when bots are enabled, at the very least rule gate this still or lets ask why this behaves differently when bots are enabled because it affects any mob

zone/spell_effects.cpp
```diff
@@ -6839,11 +6828,9 @@ void Mob::CheckNumHitsRemaining(NumHit type, int32 buff_slot, uint16 spell_id)
 			if (IsValidSpell(buffs[d].spellid) && buffs[d].hit_number > 0 &&
 			    spells[buffs[d].spellid].hit_number_type == static_cast<int>(type)) {
 
-#ifdef BOTS
 				buff_name = spells[buffs[d].spellid].name;
```
Akkadius, line 6831: This seems to change how this logic fundamentally works when bots are enabled, at the very least rule gate this still or lets ask why this behaves differently when bots are enabled because it affects any mob

Akkadius (approved):
Few comments, LGTM, nice work.

Views really need to be removed entirely

Kinglykrab:
Will take a look this afternoon. Do we want to delete views in the scope of this pull request? What is our suggested alternative being bots/players use the same tables for groups and guilds? A `bot_id` column in these tables so we know it’s a bot versus a player?

zone/spell_effects.cpp
Aeadoin: Bot logic here feeds into a SendMarqueeMessage to the Client whenever the Bot is hit and have a buff with a hit counter... honestly this seems very annoying to me, and high spam.

I'd consider just removing it,.. not sure who would actually use this, worst case just lock it behind a rule.

`			bot_owner->CastToClient()->SendMarqueeMessage(
				Chat::Yellow,
				510,
				0,
				1000,
				3000,
				StringFormat(
					"%s has [%u] hit%s remaining on '%s'",
					GetCleanName(),
					buff_counter,
					(buff_counter == 1 ? "" : "s"),
					buff_name.c_str()
				)
			);`

zone/spell_effects.cpp
Aeadoin: Same as above.

zone/spell_effects.cpp
Aeadoin: Same as above.

Aeadoin (approved):
LGTM besides the minor things Akk pointed out.

Akkadius:
> Will take a look this afternoon. Do we want to delete views in the scope of this pull request? What is our suggested alternative being bots/players use the same tables for groups and guilds? A `bot_id` column in these tables so we know it’s a bot versus a player?

Thank you sir.

Normally I'd say yes, let's decouple these changes in separate PR's - but I am not going to support views in PEQ dumps just for bots.

We have 6 queries in the source with `vw_` in the prefix that we'd need to figure out how we alternatively solve for. I don't have an answer to all of them as I don't have context of the bot code but I can look through it and help give some suggestions if you would like.

Either way I think this PR is really close and this is a huge W for simplification

Akkadius:
We can use `is_bot` where we really need to, similar to in the `group_id` table we do an `ismerc`

Some of the cases we may not need anything special at all and could be opportunity to simplify

zone/spell_effects.cpp
nytmyr: This is toggled with owner options, right? I agree it is a pretty pointless and annoying option but is disabled by default.

Aeadoin: Ahh yes, correct. the Bot logic here would be fine then.

utils/sql/git/required/2023_01_18_bots_enable.sql
```diff
@@ -0,0 +1 @@
+UPDATE `rule_values` SET `rule_value` = 'true' WHERE `rule_name` = 'Bots:AllowBots';
```
Akkadius, line 1: This shouldn't be in required because not everyone wants to enable bots.

The rule itself will be auto injected into the database at whatever default value as well

PR #2772, opened 2023-01-21
[Bug Fix] Fix bots equipping augments.

Kinglykrab:
# Notes
- Bots didn't check if the item was an augment before equipping, this stops them from equipping them.


Aeadoin (approved):
LGTM

PR #2775, opened 2023-01-21
[Commands] Remove #guildapprove, #guildcreate, and #guildlist Commands

Kinglykrab:
# Notes
- Removes `#guildapprove`, `#guildcreate`, and `#guildlist`.
- Removes associated functions, classes, etc. that were only used in these commands.
- These commands are unused and/or covered by the general `#guild` command.
- Approvals don't really seem to be a thing anymore and the variable itself doesn't exist in stock PEQ database anyway.


joligario (approved):

Aeadoin (approved):
LGTM

PR #2779, opened 2023-01-21
[Commands] #scribespells triggered error on mysql keyword rank

noudess:
#scribespells was causing errors on my server (using mySQL) due to using the keywork rank.

image

escaping the keyword fixed the problem and #scribespells works for me.

Can someone vouch for MariaDB with this fix?


Aeadoin (approved):
Confirmed working on mariadb.

PR #2783, opened 2023-01-23
[Cleanup] Cleanup #door Command.

Kinglykrab:
# Notes
- Resolves an issue with `.1` versus `0.1`.
- Resolves an issue with updating a pre-existing door.
- Adds `heading` to the update/insert.

### DIFF
--- a/zone/doors.cpp
+++ b/zone/doors.cpp
@@ -837,37 +837,39 @@ void Doors::CreateDatabaseEntry()
 				GetDoorID()
 			);
 		}
-	} else {
-		auto e = DoorsRepository::NewEntity();
 
-		e.id           = GetDoorDBID();
-		e.doorid       = GetDoorID();
-		e.zone         = zone->GetShortName();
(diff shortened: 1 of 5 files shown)


zone/doors.cpp
```diff
@@ -802,20 +802,73 @@ void Doors::CreateDatabaseEntry()
...
+		if (!updated) {
+			LogError(
+				"Failed to update door in Zone [{}] Version [{}] Database ID [{}] ID [{}]",
+				zone->GetShortName(),
+				zone->GetInstanceVersion(),
+				GetDoorDBID(),
+				GetDoorID()
+			);
+		}
+	} else {
```
Akkadius, line 840: Inside the `l.empty()` just add a return so we can remove the `else`

Akkadius (approved):

zone/doors.cpp
Kinglykrab: Word.

PR #2784, opened 2023-01-23
[Bug Fix] Fix scenario where dereferenced objects could be null.

Aeadoin:


Kinglykrab (approved):

zone/bot_database.cpp
```diff
@@ -2358,7 +2358,7 @@ bool BotDatabase::LoadLeaderIDByBotGroupID(const uint32 group_id, uint32& leader
 bool BotDatabase::LoadBotGroupNameByBotGroupID(const uint32 group_id, std::string& botgroup_name)
 {
 	if (!group_id) {
-		false;
+		return false;
```
Kinglykrab, line 2361: What a weird one, good catch, been there for like 6 years.

PR #2793, opened 2023-01-25
[Feature] ResetItemCooldown added to lua/perl and fix item re-cast times to show properly

Natedog2012:
Properly apply the cooldown to the item header with SetRecastTimestamp before sending item packet to player. 
Recast timer SHOULD show up in corpse window now as well as when the player loots the item into inventory.
Recast timer SHOULD show up in inventory when using SummonItem
All timers now saved to database  `character_item_recast` table
-Shared timers are - charid | recast_type | timestamp
-Items that have their own timer (-1 recast type) saved as - charid | item_id | timestamp
-SQL to alter recast_type to INT for this change to work

Implemented the `pTimerNegativeItemReuse` which is just pTimerNegativeItemReuse * itemid so timers stay unique

### DIFF
--- a/common/eq_packet_structs.h
+++ b/common/eq_packet_structs.h
@@ -4545,7 +4545,7 @@ struct ItemVerifyReply_Struct {
 struct ItemRecastDelay_Struct {
 /*000*/	uint32	recast_delay;	// in seconds
 /*004*/	uint32	recast_type;
-/*008*/	uint32	unknown008;
+/*008*/	bool	ignore_casting_requirement; //Ignores recast times allows items to be reset?
 /*012*/
 };
 
--- a/common/shareddb.cpp
+++ b/common/shareddb.cpp
@@ -721,10 +721,14 @@ bool SharedDatabase::GetInventory(uint32 char_id, EQ::InventoryProfile *inv)
 			inst->SetCharges(charges);
 
 		if (item->RecastDelay) {
-			if (timestamps.count(item->RecastType))
+			if (item->RecastType != -1 && timestamps.count(item->RecastType)) {
 				inst->SetRecastTimestamp(timestamps.at(item->RecastType));
-			else
+			} else if (item->RecastType == -1 && timestamps.count(item->ID)) {
+				inst->SetRecastTimestamp(timestamps.at(item->ID));
+			}
+			else {
 				inst->SetRecastTimestamp(0);
+			}
 		}
 
 		if (item->IsClassCommon()) {
--- a/common/shareddb.cpp
+++ b/common/shareddb.cpp
@@ -721,9 +721,9 @@ bool SharedDatabase::GetInventory(uint32 char_id, EQ::InventoryProfile *inv)
 			inst->SetCharges(charges);
 
 		if (item->RecastDelay) {
-			if (item->RecastType != -1 && timestamps.count(item->RecastType)) {
+			if (item->RecastType != RECAST_TYPE_UNLINKED_ITEM && timestamps.count(item->RecastType)) {
 				inst->SetRecastTimestamp(timestamps.at(item->RecastType));
-			} else if (item->RecastType == -1 && timestamps.count(item->ID)) {
+			} else if (item->RecastType == RECAST_TYPE_UNLINKED_ITEM && timestamps.count(item->ID)) {
 				inst->SetRecastTimestamp(timestamps.at(item->ID));
 			}
 			else {
--- a/utils/sql/db_update_manifest.txt
+++ b/utils/sql/db_update_manifest.txt
@@ -471,7 +471,7 @@
 9215|2023_01_08_zone_max_level.sql|SHOW COLUMNS FROM `zone` LIKE 'max_level'|empty|
 9216|2023_01_15_merc_data.sql|SHOW TABLES LIKE 'mercs'|empty|
 9217|2023_01_15_chatchannel_reserved_names.sql|SHOW TABLES LIKE 'chatchannel_reserved_names'|empty|
-9218|2023_01_24_item_recast.sql|SELECT * FROM db_version WHERE version >= 9218|empty|
+9218|2023_01_24_item_recast.sql|show columns from character_item_recast like '%recast_type%'|contains|smallint
 
 # Upgrade conditions:
 # 	This won't be needed after this system is implemented, but it is used database that are not
--- a/utils/sql/db_update_manifest.txt
+++ b/utils/sql/db_update_manifest.txt
@@ -471,6 +471,7 @@
 9215|2023_01_08_zone_max_level.sql|SHOW COLUMNS FROM `zone` LIKE 'max_level'|empty|
 9216|2023_01_15_merc_data.sql|SHOW TABLES LIKE 'mercs'|empty|
 9217|2023_01_15_chatchannel_reserved_names.sql|SHOW TABLES LIKE 'chatchannel_reserved_names'|empty|
+9218|2023_01_24_item_recast.sql|SELECT * FROM db_version WHERE version >= 9218|empty|
 
 # Upgrade conditions:
 # 	This won't be needed after this system is implemented, but it is used database that are not
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -10483,7 +10483,7 @@ void Client::ResetItemCooldown(uint32 item_id)
 			item = GetInv().GetItem(slot_id);
 			if (item) {
 				item_d = item->GetItem();
-				if (item->GetID() == item_id || (item_d->RecastType != -1 && item_d->RecastType == recast_type)) {	
+				if (item_d && item->GetID() == item_id || (item_d->RecastType != -1 && item_d->RecastType == recast_type)) {	
 					item->SetRecastTimestamp(0);
 					DeleteItemRecastTimer(item_d->ID);
 					SendItemPacket(slot_id, item, ItemPacketCharmUpdate);
@@ -10506,8 +10506,8 @@ void Client::RemoveItem(uint32 item_id, uint32 quantity)
 		{ EQ::invbag::SHARED_BANK_BAGS_BEGIN, EQ::invbag::SHARED_BANK_BAGS_END },
 	};
 	int16 removed_count = 0;
-	const size_t size = sizeof(slots) / sizeof(slots[0]);
-	for (int slot_index = 0; slot_index < size; ++slot_index) {
+	const size_t slot_index_count = sizeof(slots) / sizeof(slots[0]);
+	for (int slot_index = 0; slot_index < slot_index_count; ++slot_index) {
 		for (int slot_id = slots[slot_index][0]; slot_id <= slots[slot_index][1]; ++slot_id) {
 			if (removed_count == quantity) {
 				break;
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -10462,6 +10462,12 @@ int Client::CountItem(uint32 item_id)
 void Client::ResetItemCooldown(uint32 item_id)
 {
 	EQ::ItemInstance *item = nullptr;
+	const EQ::ItemData* item_d = database.GetItem(item_id);
+	if (!item_d) {
+		return;
+	}
+	int recast_type = item_d->RecastType;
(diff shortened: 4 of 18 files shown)


zone/spells.cpp
```diff
@@ -6357,6 +6357,7 @@ void Client::SendItemRecastTimer(int32 recast_type, uint32 recast_delay)
 		ItemRecastDelay_Struct *ird = (ItemRecastDelay_Struct *)outapp->pBuffer;
 		ird->recast_delay = recast_delay;
 		ird->recast_type = static_cast<uint32>(recast_type);
+		ird->ignore_casting_requirement = ignore_casting_requirement; //True allows reset of item cast timers
```
Akkadius, line 6360: We need to rename one of these since they are the same, I suggest renaming the function argument. 

When we've named an input argument the same name as something else (local scoped or otherwise) it has caused unintended bugs

common/eq_packet_structs.h
```diff
@@ -4545,7 +4545,7 @@ struct ItemVerifyReply_Struct {
 struct ItemRecastDelay_Struct {
 /*000*/	uint32	recast_delay;	// in seconds
 /*004*/	uint32	recast_type;
-/*008*/	uint32	unknown008;
+/*008*/	bool	ignore_casting_requirement; //Ignores recast times allows items to be reset?
```
Akkadius, line 4548: Does making this a bool cause unintended consequences with the byte size difference? (1 byte vs 4 byte)

Did we test a case where we're not ignoring casting requirements just to make sure it works that way as well

Natedog2012: Size difference was tested and it never mattered

utils/sql/db_update_manifest.txt
```diff
@@ -471,6 +471,7 @@
 9215|2023_01_08_zone_max_level.sql|SHOW COLUMNS FROM `zone` LIKE 'max_level'|empty|
 9216|2023_01_15_merc_data.sql|SHOW TABLES LIKE 'mercs'|empty|
 9217|2023_01_15_chatchannel_reserved_names.sql|SHOW TABLES LIKE 'chatchannel_reserved_names'|empty|
+9218|2023_01_24_item_recast.sql|SELECT * FROM db_version WHERE version >= 9218|empty|
```
Akkadius, line 474: Can we change this to be more specific? We try not to check against arbitrary versions since it's not idempotent in this case.

Suggest:

Look for the condition of the previous state to trigger running the desired state. Character recast currently shows as a `smallint` column

```
9218|2023_01_24_item_recast.sql|show columns from character_item_recast like '%recast_type%'|contains|smallint
```

zone/client.cpp
```diff
@@ -10459,6 +10459,40 @@ int Client::CountItem(uint32 item_id)
...
+	static const int16 slots[][2] = {
+		{ EQ::invslot::POSSESSIONS_BEGIN, EQ::invslot::POSSESSIONS_END },
+		{ EQ::invbag::GENERAL_BAGS_BEGIN, EQ::invbag::GENERAL_BAGS_END },
+		{ EQ::invbag::CURSOR_BAG_BEGIN, EQ::invbag::CURSOR_BAG_END},
+		{ EQ::invslot::BANK_BEGIN, EQ::invslot::BANK_END },
+		{ EQ::invbag::BANK_BAGS_BEGIN, EQ::invbag::BANK_BAGS_END },
+		{ EQ::invslot::SHARED_BANK_BEGIN, EQ::invslot::SHARED_BANK_END },
+		{ EQ::invbag::SHARED_BANK_BAGS_BEGIN, EQ::invbag::SHARED_BANK_BAGS_END },
+	};
+	const size_t size = sizeof(slots) / sizeof(slots[0]);
```
Akkadius, line 10480: Can we `slot_index_count` vs `size` to be more descriptive

zone/client.cpp
```diff
@@ -10459,6 +10459,40 @@ int Client::CountItem(uint32 item_id)
...
+	for (int slot_index = 0; slot_index < size; ++slot_index) {
+		for (int slot_id = slots[slot_index][0]; slot_id <= slots[slot_index][1]; ++slot_id) {
+			item = GetInv().GetItem(slot_id);
+			if (item) {
+				item_d = item->GetItem();
+				if (item->GetID() == item_id || (item_d->RecastType != -1 && item_d->RecastType == recast_type)) {
```
Akkadius, line 10486: Add `item_d` before rest of checks to validate fetching the item at this level

Akkadius: eg 

```cpp
if (item_d && item->GetID() == item_id || (item_d->RecastType != -1 && item_d->RecastType == recast_type)) {	
```

zone/spells.cpp
Natedog2012: Fixed and renamed to avoid this

zone/client.h
```diff
@@ -1516,8 +1517,9 @@ class Client : public Mob
 
 	void SendReloadCommandMessages();
 
-	void SendItemRecastTimer(int32 recast_type, uint32 recast_delay = 0);
+	void SendItemRecastTimer(int32 recast_type, uint32 recast_delay = 0, bool ignore = false);
```
Akkadius, line 1520: `ignore` ignore what?

Natedog2012: ignore_casting_requirement but you asked me to change it :D

Akkadius: At least prefix with `in_` it may sound like nit picky feedback, but it has caused uninitialized values in the past having same name variables

`in_ignore_casting_requirement`

Natedog2012: Sounds good, was just wondering what you meant

zone/client_packet.cpp
```diff
@@ -8927,12 +8927,17 @@ void Client::Handle_OP_ItemVerifyRequest(const EQApplicationPacket *app)
...
 					if (item->RecastDelay > 0)
 					{
-						if (!GetPTimers().Expired(&database, (pTimerItemStart + item->RecastType), false)) {
+						if (item->RecastType != -1 && !GetPTimers().Expired(&database, (pTimerItemStart + item->RecastType), false)) {
 							SendItemRecastTimer(item->RecastType); //Problem: When you loot corpse, recast display is not present. This causes it to display again. Could not get to display when sending from looting.
 							MessageString(Chat::Red, SPELL_RECAST);
 							SendSpellBarEnable(item->Click.Effect);
 							LogSpells("Casting of [{}] canceled: item spell reuse timer not expired", spell_id);
 							return;
+						} else if (item->RecastType == -1  && !GetPTimers().Expired(&database, (pTimerNegativeItemReuse * item->ID), false)) {
```
Akkadius, line 8936: For all of these checks that do `recast_type == -1` can we swap them out for a constant for readability? We've been making sincere efforts to remove magic numbers from the source for the past few years.

This makes code more readable in its intent for new travelers

Somewhere in say `eq_constants.h`

```cpp
constexpr uint16 RECAST_TYPE_ITEM = -1;
```

Then in code

```cpp
} else if (item->RecastType == RECAST_TYPE_ITEM ...
```

Akkadius (changes requested):
Left a few comments, otherwise it looks good to me! 

Very nice work!

It does smell a little bit that we are not using a separate column to just declare `item_id` in `character_item_recast` but this makes sense too

zone/client_packet.cpp
Natedog2012: -1 maybe UNLINKED_ITEM_TIMER  ?  because anything other than -1 is a linked timer

Akkadius: Or `RECAST_TYPE_ITEM_UNLINKED` with a description in comments at the end of the constant

The constant is not a timer so might be confusing

zone/client.cpp
Akkadius: Still have `slot_index_count` here, you updated the other (thank you)

zone/spells.cpp
```diff
@@ -6409,14 +6413,21 @@ void Client::SetItemRecastTimer(int32 spell_id, uint32 inventory_slot)
...
+		
 		if (recast_type != -1) {
-			database.UpdateItemRecastTimestamps(
+			GetPTimers().Start((pTimerItemStart + recast_type), static_cast<uint32>(recast_delay));
+			database.UpdateItemRecast(
 				CharacterID(),
 				recast_type,
 				GetPTimers().Get(pTimerItemStart + recast_type)->GetReadyTimestamp()
 			);
+		} else if (recast_type == -1) {
```
Akkadius, line 6424: Another spot for constant

common/shareddb.cpp
```diff
@@ -721,10 +721,14 @@ bool SharedDatabase::GetInventory(uint32 char_id, EQ::InventoryProfile *inv)
 			inst->SetCharges(charges);
 
 		if (item->RecastDelay) {
-			if (timestamps.count(item->RecastType))
+			if (item->RecastType != -1 && timestamps.count(item->RecastType)) {
```
Akkadius, line 724: Another spot for constant

common/shareddb.cpp
```diff
@@ -721,10 +721,14 @@ bool SharedDatabase::GetInventory(uint32 char_id, EQ::InventoryProfile *inv)
...
 				inst->SetRecastTimestamp(timestamps.at(item->RecastType));
-			else
+			} else if (item->RecastType == -1 && timestamps.count(item->ID)) {
```
Akkadius, line 726: Another spot for constant

zone/spells.cpp
```diff
@@ -6473,7 +6506,9 @@ bool Client::HasItemRecastTimer(int32 spell_id, uint32 inventory_slot)
 		return false;
 	}
 	//if time is not expired, then it exists and therefore we have a recast on this item.
-	if (!CastToClient()->GetPTimers().Expired(&database, (pTimerItemStart + recast_type), false)) {
+	if (recast_type != -1 && !CastToClient()->GetPTimers().Expired(&database, (pTimerItemStart + recast_type), false)) {
```
Kinglykrab, line 6509: Constant.

zone/spells.cpp
```diff
@@ -6473,7 +6506,9 @@ bool Client::HasItemRecastTimer(int32 spell_id, uint32 inventory_slot)
...
+		return true;
+	} else if (recast_type == -1 && !CastToClient()->GetPTimers().Expired(&database, (pTimerNegativeItemReuse * item_id), false)) {
```
Kinglykrab, line 6511: Constant.

Kinglykrab (commented):
Do we need DeleteItemRecastTimer as its own export if it doesn’t fully delete the recast? Could see that being confusing. ResetItemCooldown may be enough.

Natedog2012:
> Do we need DeleteItemRecastTimer as its own export if it doesn’t fully delete the recast? Could see that being confusing. ResetItemCooldown may be enough.

I was thinking the same thing, function doesn't need export, I can remove later today

Kinglykrab:
> > Do we need DeleteItemRecastTimer as its own export if it doesn’t fully delete the recast? Could see that being confusing. ResetItemCooldown may be enough.
> 
> I was thinking the same thing, function doesn't need export, I can remove later today

Beautiful.

zone/client_packet.cpp
Natedog2012: Pretty sure I got all the spots where this is referenced

zone/spells.cpp
Natedog2012: Fixed ty missed those 🗡️

Natedog2012:
> Left a few comments, otherwise it looks good to me!
> 
> Very nice work!
> 
> It does smell a little bit that we are not using a separate column to just declare `item_id` in `character_item_recast` but this makes sense too

I was thinking this too but it works really well the way it is currently written in the source  <=20 is Linked Recast any number over that is an item ID which normally start at 1001 anyways.

Natedog2012:
Tested Dropping item on ground and picking up - Timer is correct
Tested looting from NPC - Timer shows up on corpse and correctly when looted
Tested trading between players - Needs to be fixed
SummonItem - Timer shows correct

Akkadius (approved):

PR #2795, opened 2023-01-26
[Bugfix] Add SetItemCooldown to tell client exact timer of item when clicked a…

Natedog2012:
…nd export to perl and lua

Items that are on cooldown but client doesn't show.. clicking item will fix the timer in client


Kinglykrab (commented):
Will look more this afternoon.

Kinglykrab (approved):

PR #2800, opened 2023-01-27
[Bug Fix] Resolve issue with max buff count being 25 in ROF2.

Kinglykrab:
# Notes
- This allows ROF2 to properly utilize their max buff count.
- May cause issues with older clients.

### DIFF
--- a/common/eq_packet_structs.h
+++ b/common/eq_packet_structs.h
@@ -29,7 +29,7 @@
 #include "textures.h"
 
 
-static const uint32 BUFF_COUNT = 25;
+static const uint32 BUFF_COUNT = 42;
 static const uint32 PET_BUFF_COUNT = 30;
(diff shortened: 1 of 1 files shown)


Akkadius (changes requested):
Let's make sure we've done testing with different clients on this change

Natedog2012:
> Let's make sure we've done testing with different clients on this change

The encodes should keep them safe from breaking older clients, but loading up Titanium to see

Natedog2012:
Casting every buff on titanium and got my character up to 26 BuffCount (server side) and no crash from the client

Natedog2012 (approved):
Seems to work for me on titanium

Akkadius:
> Seems to work for me on titanium

Thanks for testing @Natedog2012 <3

Akkadius (approved):

PR #2802, opened 2023-01-28
[Bug Fix] Fixing % based mob see invis

fryguy503:
# Notes

This bug was preventing anything other than 0 or 1 to be passed to see_invis, but per the current design, see_invis can be anything from 0 to 25499, which can give you level 254 invis. See design notes in Mob::GetSeeInvisibleLevelFromNPCStat.


Kinglykrab (approved):
:fire:

Akkadius (approved):

PR #2806, opened 2023-01-29
[Feature] Add Min/Max Status to Merchants

Kinglykrab:
# Notes
- Allows operators to set a minimum and maximum status that an item will show for players.
- Allows operators to have items on a merchant that only a GM can see.
- Some servers may use status for different things, so having a minimum and a maximum will allow for more functionality.
- Default of `min_status` is `0` (Player) and default of `max_status` is `255` (Max).

### DIFF
--- a/common/eq_packet_structs.h
+++ b/common/eq_packet_structs.h
@@ -3632,17 +3632,19 @@ struct LevelAppearance_Struct { //Sends a little graphic on level up
 };
 
 struct MerchantList {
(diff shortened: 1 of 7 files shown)


Aeadoin (approved):
lgtm

joligario:
@Kinglykrab: Should this include the temp merchant list as well? Not tested, but I assume that someone with status could sell a hidden item which could then be seen?

PR #2807, opened 2023-01-29
[Quest API] Cleanup string copies and push_backs.

Kinglykrab:
# Notes
- Several places use `push_back` instead of `emplace_back`.
- Several places use `std::string` instead of `const std::string&`.
- Several places were using ternaries for `true` and `false` instead of just checking `!= 0`.


Aeadoin (approved):
Looks good, thanks for cleaning this up!

PR #2810, opened 2023-01-30
[Quest API] Add EVENT_ITEM_CLICK_CLIENT and EVENT_ITEM_CLICK_CAST_CLIENT to Perl/Lua.

Kinglykrab:
# Perl
- Add `EVENT_ITEM_CLICK_CLIENT`.
- Add `EVENT_ITEM_CLICK_CAST_CLIENT`.
- Both events export `$item_id`, `$item_name`, `$slot_id`, and `$spell_id`.

# Lua
- Add `event_item_click_client`.
- Add `event_item_click_cast_client`.
- Both events export `e.item_id`, `e.item_name`, `e.slot_id`, `e.spell_id`, and `e.item`.

# Notes
- Allows operators to handle item clicks in player scripts instead of item-specific scripts.
- Events check scripts for the sub.
- Cleanup some other issues with stuff that was merged.


zone/attack.cpp
```diff
@@ -3827,28 +3827,28 @@ void Mob::CommonDamage(Mob* attacker, int64 &damage, const uint16 spell_id, cons
 		const auto has_npc_given_event = (
 			(
 				IsNPC() &&
-				parse->HasQuestSub(CastToNPC()->GetNPCTypeID(), EVENT_DAMAGE_GIVEN)
+				parse->HasQuestSub(CastToNPC()->GetNPCTypeID(), EVENT_DAMAGE_GIVEN, true)
```
Akkadius, line 3830: Seems like we should check encounters regardless, this is a very strange smell that maybe we should cut early

Akkadius: I'd remove the optional bool and just make it default behavior

Kinglykrab: Sounds good.

Akkadius (approved):

PR #2811, opened 2023-01-30
[Quest API] Add GetItemCooldown to return the time remaining on items…

Natedog2012:
… in seconds


Kinglykrab (approved):
Looks good. Should it be `uint32` since it’ll never be negative and `final_time` is a `uint32`?

Akkadius (approved):

PR #2812, opened 2023-01-30
[Cleanup] Removed hard coding for loops related to bags, Bazaar, and merchants

brennonreid:
moved some variables into rof2_limits to clean things up as well and have things all in once place.


common/emu_constants.h
```diff
@@ -62,7 +62,7 @@ namespace EQ
 		using RoF2::invtype::KRONO_SIZE;
 		using RoF2::invtype::OTHER_SIZE;
 
-		using Titanium::invtype::TRADE_NPC_SIZE;
```
Akkadius, line 65: The rest of the changes in this PR are fairly straightforward - I'm leery of just changing the default constants in this file without fully realizing what kind of nasty regressions that could come about.

What kind of testing has been done?

brennonreid: I'm currently working on getting something public with more testing, Most of the testing has just been on my personal server that I create a lot of characters and test cases on. My goal was to not change any value or do anything that would modify any current database locations or modify any major state checks.  I am concerned about moving bazaar_size & merchant_size as changes since those values are 200 vs the previous hardcoded values of 80.  I was thinking that the safest bet would be to change those constant values to 80 if I wanted it to match with exactly how it works now.  I am working on a bigger test environment currently.  I will try to post if I find a better platform to test on.

Akkadius: Updates?

Seppuku83: I would be happy to help test this as I know there's been chatter for bigger bags (myself included).. I'm assuming I'd need to fire up Titanium and RoF clients and just go ham doing things that involve slot changes? I don't want to be looking at/for the wrong things.

Kinglykrab: > I would be happy to help test this as I know there's been chatter for bigger bags (myself included).. I'm assuming I'd need to fire up Titanium and RoF clients and just go ham doing things that involve slot changes? I don't want to be looking at/for the wrong things.

This doesn't add support for bigger bags, it just changes them to use the ROF2 file and allows for operators to change their source to support bigger bags.

Kinglykrab: Would still need translation queries to move old slot IDs to new slot IDs and an actual change to the values to make the bags bigger.

brennonreid: No, i just don't have the resources to test on a large db. I'm just not confident I have enough data to test reliably on.  so i can remove the pull request for awhile. and the issue isn't running a server from scratch, it's the issue of folding this into existing servers without having to change data locations.  there are still range checks and hard coded range location checks that would also have to be scrubbed in order to safely change bag sizes. this was just suppose to be a first step.

Kinglykrab:
Any plans to finish this up @brennonreid, or want to close it for the time being until you can revisit it in the future?

brennonreid:
Maybe next time.

PR #2818, opened 2023-02-01
[Quest API] Default ScaleNPC to always scale.

Kinglykrab:
# Notes
- ScaleNPC will now always override stats, with the option to override special abilities.

### DIFF
--- a/zone/lua_npc.cpp
+++ b/zone/lua_npc.cpp
@@ -761,16 +761,10 @@ void Lua_NPC::ScaleNPC(uint8 npc_level)
 	self->ScaleNPC(npc_level);
 }
 
-void Lua_NPC::ScaleNPC(uint8 npc_level, bool always_scale_stats)
+void Lua_NPC::ScaleNPC(uint8 npc_level, bool override_special_abilities)
 {
 	Lua_Safe_Call_Void();
-	self->ScaleNPC(npc_level, always_scale_stats);
-}
-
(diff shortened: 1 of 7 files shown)


Akkadius (approved):
lgtm - ty

Kinglykrab:
> lgtm - ty

:fire:

xackery:
Curious why it's called ScaleNPC? Isn't that redundant?

Akkadius:
> Curious why it's called ScaleNPC? Isn't that redundant?

Probably copy / pasta from the scale manager method `ScaleNPC` I agree it could have been simplified back when it was initially added to `Scale`

xackery:
> > Curious why it's called ScaleNPC? Isn't that redundant?
> 
> Probably copy / pasta from the scale manager method `ScaleNPC` I agree it could have been simplified back when it was initially added to `Scale`

Yeah, or we can define it better on what it is scaling..
e.g.

It scales NPC stats and special abilities, so perhaps

ScaleStatsAndSpecialAbilities()
Or ScaleStats() or something.

Akkadius:
A function that has the word `And` in it is a bit much I'd opt for just keeping it `Scale` level is implied by the argument

Kinglykrab:
> A function that has the word `And` in it is a bit much I'd opt for just keeping it `Scale` level is implied by the argument

So we want to add new naming for it to be `Scale` and keep `ScaleNPC` so we don’t break existing stuff?

PR #2820, opened 2023-02-01
[Rules] Add rule to ignore name filter on chat channel creation.

Valorith:
**Rule** : Chat:ChannelsIgnoreNameFilter 
**Description** : Ignore name filtering when creating new chat channels
**Default** : false


ucs/chatchannel.cpp
```diff
@@ -78,7 +78,7 @@ ChatChannel *ChatChannelList::CreateChannel(
 {
 	uint8 max_perm_player_channels = RuleI(Chat, MaxPermanentPlayerChannels);
 
-	if (!database.CheckChannelNameFilter(name)) {
+	if (!database.CheckChannelNameFilter(name) && !RuleB(Chat, ChannelsIgnoreNameFilter)) {
```
Kinglykrab, line 81: Would put rule first so we’re not doing an unnecessary query here.

Valorith: Good call. Fixing now

Kinglykrab (approved):

PR #2822, opened 2023-02-03
[Quest API] Add EVENT_TASKACCEPTED to Player scope

Valorith:
# Issue
Previously EVENT_TASKACCEPTED was only accesible in the NPC scope. 

# Change
This adds EVENT_TASKACCEPTED to the player scope for both Perl and Lua.

Tested on up-to-date binaries and works fine.

### DIFF
--- a/zone/lua_parser.cpp
+++ b/zone/lua_parser.cpp
@@ -242,6 +242,7 @@ LuaParser::LuaParser() {
 	PlayerArgumentDispatch[EVENT_DUEL_LOSE]                  = handle_player_duel_loss;
 	PlayerArgumentDispatch[EVENT_LOOT]                       = handle_player_loot;
 	PlayerArgumentDispatch[EVENT_TASK_STAGE_COMPLETE]        = handle_player_task_stage_complete;
+	PlayerArgumentDispatch[EVENT_TASK_ACCEPTED]				 = handle_player_task_accepted;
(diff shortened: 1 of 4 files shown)


Natedog2012 (approved):
Tested in perl/lua seems to work just fine! @Akkadius think this should be added to zone_controller as well?

Akkadius:
> Tested in perl/lua seems to work just fine! @Akkadius think this should be added to zone_controller as well?

Not sure that it needs to be in the zone controller

zone/lua_parser.cpp
```diff
@@ -242,6 +242,7 @@ LuaParser::LuaParser() {
 	PlayerArgumentDispatch[EVENT_DUEL_LOSE]                  = handle_player_duel_loss;
 	PlayerArgumentDispatch[EVENT_LOOT]                       = handle_player_loot;
 	PlayerArgumentDispatch[EVENT_TASK_STAGE_COMPLETE]        = handle_player_task_stage_complete;
+	PlayerArgumentDispatch[EVENT_TASK_ACCEPTED]				 = handle_player_task_accepted;
```
Akkadius, line 245: Extra spaces

Akkadius: Use spaces instead of tabs

Valorith: Updated

Kinglykrab (approved):

PR #2823, opened 2023-02-04
[Fix] Fix NPC ghosting at safe coordinates

Akkadius:
### What

When zoning into some zones, NPCs can be "ghosted" at what appears to be zone ins or safe coordinates. This happens when NPCs are clipped into the geometry during the initial spawn of an NPC. The client auto corrects this by placing the actor in the safe coordinates. 

The client does other similar auto correcting behavior when the Z of a model is slightly in the ground, the client will correct it client side even if the server is wrong and the server is pushing the actor into the ground. 

The behavior occurs because the client treats the initial spawn struct differently than a position update. Years ago we used to brute force re-send position updates to the client even if they were on the other side of the zone and it "masked" the original problem because we sent the position always after the initial spawn packet (which is different).

Over the years with server side optimizations, we don't brute force send position updates regularly and we only send position updates for NPC's that are near and then re-send the entire zone when the client moves 600 units anywhere in the zone. This has had side affects that have been subtle and hard to pin down until now

### Solution

We send position update after the initial spawn struct, both for the initial spawn of an NPC and when a client enters the zone

**Before**

image

**After**

image

### DIFF
--- a/zone/client.h
+++ b/zone/client.h
@@ -2083,6 +2083,7 @@ private:
 	bool m_bot_precombat;
 
 	bool CanTradeFVNoDropItem();
+	void SendMobPositions();
 };
 
 #endif
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -778,6 +778,12 @@ void Client::CompleteConnect()
 
 	parse->EventPlayer(EVENT_ENTER_ZONE, this, "", 0);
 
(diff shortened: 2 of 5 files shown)


Natedog2012 (approved):
Seems to be working for me for Hole / Nagafen's lair where I was seeing a lot of issues

PR #2824, opened 2023-02-04
[Doors] Remove door dev tools spam on client controlled doors

Akkadius:
### What 

In certain zones, there are client controlled doors that are triggered or linked to other doors that the client will try to force into a desired state. It causes spam for GM's in various zones. This is even more so when many clients are in the zone all trying to force the state of the door simultaneously. 

image

It's fine to let the client control these doors, but the spam can be unruly and lead to believe that there is a deeper issue.

### Solution

The solution for now is to distance check these doors for clients before trying to display the dev tools prompt and triggering quest door click events.

If not within distance, we still process `HandleClick` internally to let the client trigger the door back to the state it should be per normal


Natedog2012 (approved):
Looks good to me

PR #2828, opened 2023-02-05
[Lua] Fix stoi exception in death events

Akkadius:
### What

Since implementing quest errors - while they've been very helpful - they've been bringing certain errors to light that may not necessarily be service affecting but do cause confusion and concern. 

This fixes the generic `stoi` exception that most commonly occurs in death events. Other events will need to be gone through but this stops the bleeding for the most common events currently.

```
Zone | QuestError | AddError Lua Exception: stoi -- [qeynos2] (North Qeynos) inst_id [0]
```

### Solution

The solution is merely validating that the arg is a real number before trying to run `stoi` on it, thus causing the exception.

### DIFF
--- a/zone/lua_parser_events.cpp
+++ b/zone/lua_parser_events.cpp
@@ -309,13 +309,18 @@ void handle_npc_death(
 	lua_setfield(L, -2, "other");
 
 	Seperator sep(data.c_str());
-	lua_pushinteger(L, std::stoi(sep.arg[0]));
-	lua_setfield(L, -2, "killer_id");
 
-	lua_pushinteger(L, std::stoi(sep.arg[1]));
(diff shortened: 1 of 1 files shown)


xackery (approved):

xackery:
Hmm, I'm trying to remember lua's handling of empty fields.. since..

```cpp
if (Strings::IsNumber(sep.arg[0])) {
		lua_pushinteger(L, std::stoi(sep.arg[0]));
		lua_setfield(L, -2, "killer_id");
	}
```

This means the field won't be set if the value is deemed invalid, could this cause a nil error if it's called? 

Wonder if we want to just default fallback the value or not.. probably will end up case by case. Sorry just thinking out loud on it.

PR #2831, opened 2023-02-05
[Quest API] Add rule AlternateAugmentationSealer for using a different bagtype

Natedog2012:
…t bagtype as an alternate augmentation sealer. Use EVENT_COMBINE with UseAugmentContainer

Lua Example
```lua
function event_combine(e)
	local bag_id = e.self:GetItemIDAt(e.container_slot)
	if (bag_id == 1267) then
		e.self:UseAugmentContainer(e.container_slot)
		return 1;
	end
end
```

Perl Example
```perl
sub EVENT_COMBINE {
	my $bag_id = $client->GetItemIDAt($container_slot);
	if ($bag_id == 1267) {
		$client->UseAugmentContainer($container_slot); 
		return 1; #We did a custom combine return 1
	}
}
```

### DIFF
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -744,6 +744,7 @@ RULE_BOOL(Inventory, DeleteTransformationMold, true, "False if you want mold to 
 RULE_BOOL(Inventory, AllowAnyWeaponTransformation, false, "Weapons can use any weapon transformation")
 RULE_BOOL(Inventory, TransformSummonedBags, false, "Transforms summoned bags into disenchanted ones instead of deleting")
 RULE_BOOL(Inventory, AllowMultipleOfSameAugment, false, "Allows multiple of the same augment to be placed in an item via #augmentitem or MQ2, set to true to allow")
+RULE_INT(Inventory, AlternateAugmentationSealer, 57, "Allows RoF+ clients to augment items from a special container type")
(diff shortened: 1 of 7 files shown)


Akkadius:
This isn't a live like feature right? It shouldn't be enabled by default.

Akkadius:
Looks dope though, thanks for taking the time to wrap it up for PRing

Natedog2012:
> This isn't a live like feature right? It shouldn't be enabled by default.

Yeah my bad set it to 53 so its defaulted to just being the BagTypeAugmentationSealer.. so server owners can change it to a different bagtype
https://github.com/EQEmu/Server/blob/master/common/item_data.h#L296

Kinglykrab:
> > This isn't a live like feature right? It shouldn't be enabled by default.
> 
> Yeah my bad set it to 53 so its defaulted to just being the BagTypeAugmentationSealer.. so server owners can change it to a different bagtype https://github.com/EQEmu/Server/blob/master/common/item_data.h#L296

I'd have `-1` be the default and check if value is anything except `-1` for enabled, notes field should say that `-1` means disabled and is default, similar to other stuff recently added.

Akkadius (approved):
Thanks for making the adjustments. LGTM

PR #2834, opened 2023-02-06
[Bug Fix] Fix for undefined MySQL library behavior.

KimLS:
The MYSQL structure in the MySQL C api is not copy aware. This isn't enforced because, well, C doesn't enforce anything really. 
See: [MySQL Structures](https://dev.mysql.com/doc/c-api/8.0/en/c-api-data-structures.html)

This changes the copy of an existing connection to instead create a second backup connection with the same data as what is being copied.  This does introduce another connection into the mix which are a limited resource but given the potential seriousness of the bug in question it seems like a fair tradeoff for now.  

Long term we can rewrite the dbcore class to use a underlying pointer instead of a pinned in place variable.

### DIFF
--- a/client_files/export/main.cpp
+++ b/client_files/export/main.cpp
@@ -86,7 +86,16 @@ int main(int argc, char **argv)
 			return 1;
 		}
 	} else {
-		content_db.SetMysql(database.getMySQL());
+		if (!content_db.Connect(
+			Config->DatabaseHost.c_str(),
+			Config->DatabaseUsername.c_str(),
+			Config->DatabasePassword.c_str(),
+			Config->DatabaseDB.c_str(),
+			Config->DatabasePort
+		)) {
+			LogError("Cannot continue without a content database connection");
+			return 1;
+		}
 	}
 
 	LogSys.SetDatabase(&database)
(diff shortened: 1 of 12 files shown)


KimLS:
Gonna change some things up to get rid of the double connection, have to rewrite a bit of db core.  Don't merge yet.

Akkadius (changes requested):
Awaiting pointer changes

KimLS:
A few changes, have to test make sure they work.  Also I noticed we're not sharing the mutex that protects the connection pointer so have to think up a solution for that. 

(Sorry it's not all at once I'm doing this as I have time at night)

Akkadius:
Synced up with KLS

We agreed that a mutex should be shared across the database instances because they share the same underlying connection. This protects from concurrency / thread-safety issues.

**Reproduced Problem**

image

**Fixed, Running for several minutes**

https://user-images.githubusercontent.com/3319450/221323677-db5cf08a-eab1-405d-b7ba-21fe5122ff74.mp4

Akkadius (approved):

Akkadius:
Tested booting up all servers, logging in game, everything works as expected

PR #2835, opened 2023-02-06
[Rule] Added rule to bypass level based haste caps.

jcr4990:
Added a rule for custom servers with low level caps to bypass hard coded level based haste caps and to default to "Character:HasteCap" and "Character:Hastev3Cap" for controlling haste cap behavior.


Kinglykrab (approved):
Looks good to me.

PR #2847, opened 2023-02-07
[Logging] Add raw opcode when emu translated opcode is not found (OP_Unknown) via (C->S)

Akkadius:
### What

Adds the raw EQ opcode instead of `0x0000` when the opcode is `OP_Unknown`

We run the internal opcode through the patch translator and get the patch-specific opcode and when there's nothing that matches we simply get `0x0000` this at least gives us the protocol level opcode if we didn't find one through translators

image


Kinglykrab (approved):
:fire: 🥇

PR #2853, opened 2023-02-08
[Bots] Add Pet Power Support for Temp Pets.

Aeadoin:


zone/aa.cpp
```diff
@@ -53,9 +53,11 @@ void Mob::TemporaryPets(uint16 spell_id, Mob *targ, const char *name_override, u
 	// yep, even these need pet power!
 	int act_power = 0;
 
-	if (IsClient()) {
-		act_power = CastToClient()->GetFocusEffect(focusPetPower, spell_id);
-		act_power = CastToClient()->mod_pet_power(act_power, spell_id);
+	if (IsOfClientBot()) {
+		act_power = GetFocusEffect(focusPetPower, spell_id);
+		if (IsClient()) {
+			act_power = CastToClient()->mod_pet_power(act_power, spell_id);
```
Kinglykrab, line 59: This doesn't seem to do anything.

Aeadoin: I can always remove the mod_pet_power, but are any of the mod_functions actually doing anything?
Might be best to remove those as part of another PR.

Kinglykrab: I don't see anything using it, making another PR to remove it seems needless.

Kinglykrab: Doesn't broaden scope of this much to remove a method only used in `6` places in `5` files.
image

Akkadius: Refactors or removals shouldn't really be tucked into another PR in general

If we have to back out these changes for whatever reason they're now coupled.

`mod_` back in the day were "custom stubs" so people could write their own hooks in isolation easily enough

This is before we had the option to do lua modding which hasn't picked up a ton of steam

Kinglykrab: Yeah, makes sense for something bigger than a 6 line change.

Aeadoin: I raised #2855 as a reminder that mod_functions can likely be cleaned up/removed. I will not be touching them as part of this PR, as it has nothing to do with the scope of this PR.

Kinglykrab: https://github.com/EQEmu/Server/pull/2856

Kinglykrab (approved):

PR #2857, opened 2023-02-09
[Database] Implement native database migrations in server

Akkadius:
### What

This PR implements native database (updates) migrations in server source

* Automatic updates can now be turned off with eqemu_config option `server.auto_database_updates`
* Database updates for server operators can now be ran offline, without requiring an internet connection to pull down `.sql` files
* Database updates for developers can now be tested before they are merged into `master` testing with locally built binaries
* `eqemu_server.pl` script is no longer ran for database updates from world, they are now ran and built natively in the source
* Zone booting is now blocked if there are pending database changes
* Rules now load a little bit earlier in zone boot process so database checks can determine if bots are enabled via `Bots:Enabled`
* Removed World `ignore_db` CLI option in favor of config option `server.auto_database_updates`
* ONLY `DDL` statements are meant to be used going forward. Meaning, only queries that change schema and alter data to conform with server changes and minimize operator impact. This is not meant to run content insertions or data insertions.

### Disclaimer

This merge might break installers - of which has been on my list to house keep lately anyways. If installers break (windows/linux/akk-stack) I will own those issues and fix them

This also might cause hiccups un-updated servers because the manifest is no longer available where it used to be. The new `eqemu_server.pl` does not do any database checks and should largely no-op

```
perl ~/code/utils/scripts/eqemu_server.pl ran_from_world
[Info] Skipping self check
[Info] For EQEmu Server management utilities - run eqemu_server.pl
eqemu@8b03e640f90f:~/server$
```

### World Bootup

Below is an example of world detecting changes, running a backup before running updates

image

### User Prompt Update Skipping

Sometimes updates error because you already have the migration/update for whatever reason, we now ask the operator if they would simply like to skip the update

image

### Zone Boot Blocking

Zone booting is now blocked if there are pending database changes

image

### Bots Initialization Example

Below is a test run of enabling bots

image

### Creating a Migration

Depending on your type of change there are two manifests

* `common/database/database_update_manifest.cpp`
* `common/database/database_update_manifest_bots.cpp`

The latest server binary version is still held in `./common/version.h`

You simply need a manifest entry that matches your database version, a condition, match and then the SQL that is to be ran. Manifest entry definitions are as follows

```cpp
struct ManifestEntry {
	int         version{};     // database version of the migration
	std::string description{}; // description of the migration ex: "add_new_table" or "add_index_to_table"
	std::string check{};       // query that checks against the condition
	std::string condition{};   // condition or "match_type" - Possible values [contains|match|missing|empty|not_empty]
	std::string match{};       // match field that is not always used, but works in conjunction with "condition" values [missing|match|contains]
	std::string sql{};         // the SQL DDL that gets ran when the condition is true
};
```

```diff
diff --git a/common/database/database_update_manifest.cpp b/common/database/database_update_manifest.cpp
index 8bb269b5e..c38be7c79 100644
--- a/common/database/database_update_manifest.cpp
+++ b/common/database/database_update_manifest.cpp
@@ -3579,6 +3579,19 @@ ALTER TABLE `character_item_recast`
 ALTER TABLE `merchantlist`
 ADD COLUMN `min_status` tinyint(3) UNSIGNED NOT NULL DEFAULT 0 AFTER `level_required`,
 ADD COLUMN `max_status` tinyint(3) UNSIGNED NOT NULL DEFAULT 255 AFTER `min_status`;
+)",
+       },
+       ManifestEntry{
+               .version = 9220,
+               .description = "add_new_table.sql",
+               .check = "SHOW TABLES LIKE 'new_table'",
+               .condition = "empty",
+               .match = "",
+               .sql = R"(
+CREATE TABLE `new_table`  (
+  `id` int NOT NULL AUTO_INCREMENT,
+  PRIMARY KEY (`id`)
+);
 )",
        }
 };
diff --git a/common/version.h b/common/version.h
index 6370b4418..765804563 100644
--- a/common/version.h
+++ b/common/version.h
@@ -42,7 +42,7 @@
  * Manifest: https://github.com/EQEmu/Server/blob/master/utils/sql/db_update_manifest.txt
  */
 
-#define CURRENT_BINARY_DATABASE_VERSION 9219
+#define CURRENT_BINARY_DATABASE_VERSION 9220
 #define CURRENT_BINARY_BOTS_DATABASE_VERSION 9037
 
 #endif

```


Akkadius:
This one is for the most part ready to go, just going to wait to merge this one when I can play fire fighter and get all installers ran through afterwards

Akkadius:
I'm currently working on a whole bunch of related things behind the scenes with the installer that will roll up to all of this. Will post up everything when I've got it ready

Akkadius:
Update: Looking to incorporate this with a bunch of other work in the next week or so. Installation runs on multiple OS's will be time consuming and I'd rather do it all in one fell swoop through iterations

Akkadius:
Closing this until it's ready for review / merge to keep it out of the inbox

Akkadius:
Bringing this back online as I'm getting closer to wrapping up related work

Kinglykrab (approved):

KimLS (approved):
Nothing jumps out at me as wrong here, but it is a large set of changes so I reserve the right to be wrong.

PR #2866, opened 2023-02-11
[Bots & Mercs] Add Support for TrySympatheticProc

Aeadoin:
Bots and Mercs had no support for Sympathetic Procs

### DIFF
--- a/zone/spell_effects.cpp
+++ b/zone/spell_effects.cpp
@@ -6254,7 +6254,6 @@ uint16 Mob::GetSympatheticFocusEffect(focusType type, uint16 spell_id) {
 			if (SympatheticProcList.size() > MAX_SYMPATHETIC_PROCS)
 				continue;
 
-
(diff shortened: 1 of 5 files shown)


zone/spell_effects.cpp
```diff
@@ -6275,9 +6275,9 @@ uint16 Client::GetSympatheticFocusEffect(focusType type, uint16 spell_id) {
 				if (SympatheticProcList.size() > MAX_SYMPATHETIC_PROCS)
 					continue;
 
-				EQ::ItemInstance *aug = nullptr;
+				EQ::ItemInstance const* aug = nullptr;
```
Kinglykrab, line 6278: Why is this a const here when value changes one line later? Seems like it’s unnecessary to not define the variable as the value the next line sets it to.

Aeadoin: Cleaned this up, thanks!

Kinglykrab (approved):

Akkadius (approved):

PR #2874, opened 2023-02-12
[Bots] Add Lore Check for Augments.

Aeadoin:
Bots were not checking if they had an Augment equipped that matches the loregroup of an augment being given to them in a piece of equipment.

### DIFF
--- a/zone/bot.cpp
+++ b/zone/bot.cpp
@@ -4776,7 +4776,8 @@ void Bot::PerformTradeWithClient(int16 begin_slot_id, int16 end_slot_id, Client*
 			client->Message(
 					Chat::Yellow,
 					fmt::format(
-							"This bot already has {}, the trade has been cancelled!",
+							"{} already has {}, the trade has been cancelled!",
+							GetCleanName(),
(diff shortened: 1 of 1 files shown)


zone/bot.cpp
```diff
@@ -4758,6 +4758,32 @@ void Bot::PerformTradeWithClient(int16 begin_slot_id, int16 end_slot_id, Client*
...
+
+			linker.SetLinkType(EQ::saylink::SayLinkItemInst);
+			linker.SetItemInst(augment);
+
+			item_link = linker.GenerateLink();
+
+			client->Message(
+					Chat::Yellow,
+					fmt::format(
+							"This bot already has {}, the trade has been cancelled!",
```
Kinglykrab, line 4779: Maybe send bot name here? Like `X already has {}, the trade has been cancelled!`.

Kinglykrab (approved):

zone/bot.cpp
Aeadoin: Good idea, added that.

PR #2886, opened 2023-02-12
[Quest API] (Performance) Check event exists before export and execute EVENT_BOT_CREATE

Kinglykrab:
# Notes

### DIFF
--- a/zone/bot_command.cpp
+++ b/zone/bot_command.cpp
@@ -9935,7 +9935,6 @@ uint32 helper_bot_create(Client *bot_owner, std::string bot_name, uint8 bot_clas
 	);
 
 	bot_id = my_bot->GetBotID();
-	
 	if (parse->PlayerHasQuestSub(EVENT_BOT_CREATE)) {
 		const auto& export_string = fmt::format(
 			"{} {} {} {} {}",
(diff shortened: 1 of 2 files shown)


zone/bot_command.cpp
```diff
@@ -10009,7 +10009,7 @@ uint32 helper_bot_create(Client *bot_owner, std::string bot_name, uint8 bot_clas
 
 	bot_id = my_bot->GetBotID();
 
-	const auto export_string = fmt::format(
+	const auto& export_string = fmt::format(
```
Akkadius, line 10012: export string is still outside of the HasQuest check

Akkadius: This is wasted if we don't have the event to execute

zone/questmgr.cpp
```diff
@@ -2421,7 +2421,7 @@ bool QuestManager::createBot(const char *name, const char *lastname, uint8 level
 					).c_str()
 				);
 
-				const auto export_string = fmt::format(
+				const auto& export_string = fmt::format(
```
Akkadius, line 2424: export should be defined inside HasQuestSub

Akkadius (approved):

PR #2914, opened 2023-02-12
[Quest API] (Performance) Check event EVENT_ITEM_TICK or EVENT_WEAPON_PROC exist before export and execute

Kinglykrab:
# Notes


zone/client.cpp
```diff
@@ -8107,8 +8110,11 @@ void Client::TryItemTick(int slot)
 		{
 			if( GetLevel() >= zone->tick_items[iid].level && zone->random.Int(0, 100) >= (100 - zone->tick_items[iid].chance) )
 			{
-				EQ::ItemInstance* e_inst = (EQ::ItemInstance*)a_inst;
-				parse->EventItem(EVENT_ITEM_TICK, this, e_inst, nullptr, "", slot);
+				EQ::ItemInstance* e_inst = (EQ::ItemInstance*) a_inst;
```
Akkadius, line 8113: Move `e_inst` inside HasQuestSub check

Akkadius: Cancel

Akkadius (approved):

PR #2916, opened 2023-02-12
[Quest API] (Performance) Check event EVENT_FEIGN_DEATH exists before export and execute

Kinglykrab:
# Notes

### DIFF
--- a/zone/entity.cpp
+++ b/zone/entity.cpp
@@ -3674,8 +3674,6 @@ void EntityList::ClearFeignAggro(Mob *targ)
 
 				if (it->second->IsNPC()) {
 					if (parse->HasQuestSub(it->second->GetNPCTypeID(), EVENT_FEIGN_DEATH)) {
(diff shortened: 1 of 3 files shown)


zone/entity.cpp
```diff
@@ -3645,21 +3645,27 @@ void EntityList::ClearFeignAggro(Mob *targ)
...
+					int i = parse->EventPlayer(EVENT_FEIGN_DEATH, targ->CastToClient(), "", 0, &args);
 					if (i != 0) {
 						++it;
 						continue;
 					}
 				}
+
+				if (it->second->IsNPC()) {
+					if (parse->HasQuestSub(it->second->GetNPCTypeID(), EVENT_FEIGN_DEATH)) {
+						std::vector<std::any> args = { it->second };
```
Akkadius, line 3660: args isn't being passed here, it wasn't before either though for NPC

Akkadius: Removed

PR #2926, opened 2023-02-14
[Windows] Fix MSVC compilation bug via workaround

Akkadius:
### What

Fixes a MSVC compilation bug via workaround. There are other references in the class that follow the alias properly but it doesn't in the constructor. Not worth trying to fix and statically inserted the value. This is a class I inserted for JSON minification for player events

```
16>------ Skipped Build: Project: INSTALL, Configuration: Debug x64 ------
16>Project not selected to build for this solution configuration 
========== Build: 11 succeeded, 0 failed, 13 up-to-date, 5 skipped ==========
```


Aeadoin (approved):
Tested locally, and now able to compile.

PR #2930, opened 2023-02-14
[Groundspawns] Fix issue where groundspawns appear floating high off the ground

Akkadius:
### What

Correct Z on spawn for groundspawns to keep them from floating high off the ground.

Titanium would correct this automatically client side, RoF2+ does not. We did this for tradeskills and it worked well

**Before**

image

**After**

image


Kinglykrab (approved):
🔥

Issue #2923, opened 2023-02-13
[Bug] Casting self buffs getting SPELL_TOO_POWERFUL failure


Natedog2012:
If I am targeting a low level character when casting my self buff I get the error "SPELL_TOO_POWERFUL" for my intended target even though its a self-cast buff.

Spell target should always be self for this target-type so somewhere its getting whatever I'm targeting

Spell: Cloak of Discord 
Target was level 50~ish

noudess:
I'll take a look and repair it tomorrow.  This is along the same vein as the targetted NPC issue that I closed out on my PR.

Sorry!

Kinglykrab:
Resolved by #2931.

PR #2931, opened 2023-02-14
[Bug Fix] Self Only Spells now work again (BuffSpellRestrictions)

noudess:
My changes in PR: https://github.com/EQEmu/Server/pull/2708 broke self only spell (if someone else is targetted),

This PR assures that self only spells always work.

Should resolve issue: https://github.com/EQEmu/Server/issues/2923

@Natedog2012

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -3195,6 +3195,11 @@ bool Mob::CheckSpellLevelRestriction(Mob *caster, uint16 spell_id)
 	bool check_for_restrictions = false;
 	bool can_cast = true;
 
+	if (spells[spell_id].target_type == ST_Self) {
+		LogSpells("[CheckSpellLevelRestriction] Self Only spell - no restrictions");
+		return true;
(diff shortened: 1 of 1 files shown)


zone/spells.cpp
```diff
@@ -3195,6 +3195,11 @@ bool Mob::CheckSpellLevelRestriction(Mob *caster, uint16 spell_id)
 	bool check_for_restrictions = false;
 	bool can_cast = true;
 
+	if (spells[spell_id].target_type == ST_Self) {
```
Kinglykrab, line 3198: Maybe put above variable definitions just so they’re not needlessly defined? Looks good otherwise.

Kinglykrab (approved):

zone/spells.cpp
noudess: Done.

Natedog2012 (approved):
Seems to work now thanks again!

PR #2939, opened 2023-02-16
[Crash] Fix crash issue with dropping items and order of operations

joligario:
Should wait to null out invalid_drop until after it is used in log message.


Akkadius:
Thanks @joligario <3 Please use `[Category] Change description` formatting for PR titles in the future

Helps us keep formatting easy for https://github.com/EQEmu/Server/releases && http://spire.akkadius.com/dev/changelog

PR #2949, opened 2023-02-17
[Crash] Fix Crash in Bot::AI_Process

Aeadoin:
In http://spire.akkadius.com/dev/release/22.4.0?id=570 We appear to be crashing on Line 2472 which is:

```Client* leash_owner = (bot_group->GetLeader() && bot_group->GetLeader()->IsClient() ? bot_group->GetLeader()->CastToClient() : bot_owner);```

This may not be the least verbose, but would rather be safe.

### DIFF
--- a/zone/bot.cpp
+++ b/zone/bot.cpp
@@ -2468,9 +2468,10 @@ void Bot::AI_Process()
 	}
 
 	// We also need a leash owner and follow mob (subset of primary AI criteria)
+	auto group_leader = bot_group->GetLeader();
 	Client* leash_owner = nullptr;
-	if (bot_group->GetLeader() && bot_group->GetLeader()->IsClient()) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: zone/bot.cpp [hunk 1]
```diff
@@ -2469,13 +2468,19 @@ void Bot::AI_Process()
...
+	auto group_leader = bot_group->GetLeader();
+	Client* leash_owner = nullptr;
+	if (group_leader && group_leader->IsClient()) {
+		leash_owner = group_leader->CastToClient();
+	}
+	else {
+		leash_owner = bot_owner;
+	}
+
 	if (!leash_owner) {
```
Akkadius, line 2480: How does this end up being any different than what's happening in the ternary?

Aeadoin: dumb moment on my part! need to look into this further to determine where a poisoned ptr is coming from here.

PR #2951, opened 2023-02-17
[Bot] Change SaveTimers to Replace instead of Insert.

Aeadoin:
Change to Replace instead of Insert to reduce scenarios where the timer is expired, and the old entry hasn't been cleaned up yet.

Client Timers use a "Replace" Statement to store Timers as well.

### DIFF
--- a/zone/bot_database.cpp
+++ b/zone/bot_database.cpp
@@ -962,7 +962,10 @@ bool BotDatabase::SaveTimers(Bot* bot_inst)
 		if (bot_timers[timer_index] <= Timer::GetCurrentTime())
 			continue;
 
-		query = StringFormat("REPLACE INTO `bot_timers` (`bot_id`, `timer_id`, `timer_value`) VALUES ('%u', '%u', '%u')", bot_inst->GetBotID(), (timer_index + 1), bot_timers[timer_index]);
+		query = fmt::format(
+				"REPLACE INTO `bot_timers` (`bot_id`, `timer_id`, `timer_value`) VALUES ('%u', '%u', '%u')",
+				bot_inst->GetBotID(), (timer_index + 1), bot_timers[timer_index]
+		);
 		auto results = database.QueryDatabase(query);
 		if (!results.Success()) {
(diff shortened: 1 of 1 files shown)


zone/bot_database.cpp
```diff
@@ -962,7 +962,7 @@ bool BotDatabase::SaveTimers(Bot* bot_inst)
 		if (bot_timers[timer_index] <= Timer::GetCurrentTime())
 			continue;
 
-		query = StringFormat("INSERT INTO `bot_timers` (`bot_id`, `timer_id`, `timer_value`) VALUES ('%u', '%u', '%u')", bot_inst->GetBotID(), (timer_index + 1), bot_timers[timer_index]);
+		query = StringFormat("REPLACE INTO `bot_timers` (`bot_id`, `timer_id`, `timer_value`) VALUES ('%u', '%u', '%u')", bot_inst->GetBotID(), (timer_index + 1), bot_timers[timer_index]);
```
Kinglykrab, line 965: While we’re in here, maybe make this use `fmt::format`? :)

Kinglykrab (approved):

zone/bot_database.cpp
Aeadoin: Good call, fixed.

zone/bot_database.cpp
```diff
@@ -962,7 +962,10 @@ bool BotDatabase::SaveTimers(Bot* bot_inst)
 		if (bot_timers[timer_index] <= Timer::GetCurrentTime())
 			continue;
 
-		query = StringFormat("REPLACE INTO `bot_timers` (`bot_id`, `timer_id`, `timer_value`) VALUES ('%u', '%u', '%u')", bot_inst->GetBotID(), (timer_index + 1), bot_timers[timer_index]);
+		query = fmt::format(
+				"REPLACE INTO `bot_timers` (`bot_id`, `timer_id`, `timer_value`) VALUES ('%u', '%u', '%u')",
```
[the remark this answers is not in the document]
Aeadoin: oh good catch, thanks!

PR #2964, opened 2023-02-19
[Pets] Client Pet summoned by NPC should not change guard location.

noudess:
On live, if an NPC summons a client pet that is on guard, the pet returns to its guard position.

On eqemu, the pet's guard position is changed to the summon location, which is incorrect.

This PR fixes this case.

### DIFF
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -2374,14 +2374,14 @@ void Mob::ShowBuffList(Client* client) {
 	}
 }
 
-void Mob::GMMove(float x, float y, float z, float heading) {
+void Mob::GMMove(float x, float y, float z, float heading, bool save_guard_spot) {
 	m_Position.x = x;
 	m_Position.y = y;
 	m_Position.z = z;
 	SetHeading(heading);
 	mMovementManager->SendCommandToClients(this, 0.0, 0.0, 0.0, 0.0, 0, ClientRangeAny);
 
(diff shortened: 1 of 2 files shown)


Kinglykrab:
Looks like some submodule stuff accidentally got included here?

zone/mob.cpp
```diff
@@ -3714,10 +3714,19 @@ bool Mob::HateSummon() {
...
 
-			if (target->IsClient())
+			if (target->IsClient()) {
 				target->CastToClient()->MovePC(zone->GetZoneID(), zone->GetInstanceID(), new_pos.x, new_pos.y, new_pos.z, new_pos.w, 0, SummonPC);
-			else
-				target->GMMove(new_pos.x, new_pos.y, new_pos.z, new_pos.w);
+			}
+			else {
+				bool set_new_guard_spot;
+				bool target_is_client_pet;
```
Kinglykrab, line 3722: Why not initialize these variables with their values from here if you’re doing so a line below?

noudess: Trying to make the lines wrap less.  Want it changed?

Kinglykrab: We can just format it in a way that they're separate lines. Similar to how I did this.
https://github.com/ProjectEQ/projecteqquests/commit/a0462dfb389c4cd0f3a1b3873fb8fbb955378241

PR #2968, opened 2023-02-19
[Scaling] Add support for zone ID and instance version to NPC Scaling

Kinglykrab:
# Notes
- Adds `zone_id` and `instance_version` to `npc_scale_global_base` table to allow for zone and version-specific scaling.
- Defaults back to `zone_id` of `0` and `instance_version` of `0` for global scaling.

# Loading Precedence
- `zone_id` of current zone and `instance_version` of current instance
- `zone_id` of current zone and `instance_version` of `0`
- `zone_id` of `0` and `instance_version` of `0`


Aeadoin:
Excellent, I'll test this out some.

Akkadius:
I'll take a look later

zone/npc_scale_manager.cpp
```diff
@@ -487,14 +536,21 @@ bool NpcScaleManager::IsAutoScaled(NPC* npc)
  */
 bool NpcScaleManager::ApplyGlobalBaseScalingToNPCStatically(NPC *&npc)
 {
-	int8 npc_type  = GetNPCScalingType(npc);
-	int  npc_level = npc->GetLevel();
-
-	global_npc_scale g = GetGlobalScaleDataForTypeLevel(npc_type, npc_level);
+	auto npc_type         = GetNPCScalingType(npc);
+	auto npc_level        = npc->GetLevel();
+	auto zone_id          = npc->GetZoneID();
```
Akkadius, line 541: This won't find anything from the base tables and breaks original functionality if we assume `zone_id` 0

Code will need to look for the specific zone first and then try zone_id 0 after if it can't find it

zone/npc_scale_manager.cpp
```diff
@@ -30,21 +30,27 @@ void NpcScaleManager::ScaleNPC(
...
 		return;
 	}
 
-	int8 npc_type       = GetNPCScalingType(npc);
-	int  npc_level      = npc->GetLevel();
-	bool is_auto_scaled = IsAutoScaled(npc);
+	auto npc_type         = GetNPCScalingType(npc);
+	auto npc_level        = npc->GetLevel();
+	auto is_auto_scaled   = IsAutoScaled(npc);
+	auto zone_id          = npc->GetZoneID();
```
Akkadius, line 41: Just pull the zone id off the zone

zone/npc_scale_manager.cpp
```diff
@@ -30,21 +30,27 @@ void NpcScaleManager::ScaleNPC(
...
+	auto instance_version = npc->GetInstanceVersion();
 
-	global_npc_scale scale_data = GetGlobalScaleDataForTypeLevel(npc_type, npc_level);
+	global_npc_scale scale_data = GetGlobalScaleDataForTypeLevel(
+		npc_type,
+		npc_level,
+		zone_id,
```
Akkadius, line 47: Same thing here, 0 lookups will fail if we pass in a specific zone. We'll need to look for specific first and then fallback to 0 if not found

Akkadius:
I like it, small comments, needs testing

Missing database migration

Kinglykrab:
I think this is the fallback you’re referring to, right?: https://github.com/EQEmu/Server/pull/2968/files#diff-ada8c9116ff7777395065b7ad2c458942fec949545fe0fcfed9f2e2ffdd2bd89R278

Akkadius:
> I think this is the fallback you’re referring to, right?: https://github.com/EQEmu/Server/pull/2968/files#diff-ada8c9116ff7777395065b7ad2c458942fec949545fe0fcfed9f2e2ffdd2bd89R278

Good call. Was a bit late and cross eyed reading. Looks good

zone/npc_scale_manager.cpp
Kinglykrab: Changed.

Akkadius (approved):

PR #2974, opened 2023-02-21
[Database] Address deadlock in player events

Akkadius:
### What

This addresses and issue that has been manifesting in world deadlocks lately influenced by player events.

This issue has been seen by multiple servers (Lazarus, PEQ, EZ, KMRA etc.)

This has been running on EZ for 2 days, confirmed no deadlocks, other folks have ran the same changes without issues.

Other notable things

* This uses native MySQL string escaping instead of the one we rolled on our own that is used all over the source. The MySQL client library native function is going to be more con-formative to character sets configured on the database instance and give the best escaping overall
* The mutex in player events was not locking before reading if the queue was empty, this can lead to racey memory access, deadlocks and memory corruption

### DIFF
--- a/common/dbcore.cpp
+++ b/common/dbcore.cpp
@@ -70,17 +70,20 @@ DBcore::~DBcore()
 // Sends the MySQL server a keepalive
 void DBcore::ping()
 {
-	if (!MDatabase.trylock()) {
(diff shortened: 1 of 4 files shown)


Kinglykrab (approved):
🥇

PR #2977, opened 2023-02-21
[Tradeskills] Fix for Lore Conflict

Aeadoin:
This fixes a regression issue where components of the combine could be lore, causing the combine to fail on the lore check if the components were returned on success, failure, or salvage.

### DIFF
--- a/zone/tradeskills.cpp
+++ b/zone/tradeskills.cpp
@@ -475,7 +475,7 @@ void Object::HandleCombine(Client* user, const NewCombine_Struct* in_combine, Ob
 	                                 spec.recipe_id);
 	auto results = content_db.QueryDatabase(query);
 
-	if(!results.Success() || results.RowCount() < 1 || results.RowCount() > 10) {
+	if (!results.Success() || results.RowCount() < 1 || results.RowCount() > 10) {
 		auto outapp = new EQApplicationPacket(OP_TradeSkillCombine, 0);
 		user->QueuePacket(outapp);
 		safe_delete(outapp);
--- a/zone/tradeskills.cpp
+++ b/zone/tradeskills.cpp
@@ -1879,7 +1879,7 @@ bool Client::CheckTradeskillLoreConflict(int32 recipe_id)
 	for (auto& e : recipe_entries) {
 		auto item_inst = database.GetItem(e.item_id);
 		if (item_inst) {
-			if (item_inst->LoreGroup == 0 || e.componentcount > 0) {
+			if (item_inst->LoreGroup == 0 || e.componentcount > 0 || e.iscontainer) {
 				continue;
 			}
 			if (CheckLoreConflict(item_inst)) {
--- a/zone/tradeskills.cpp
+++ b/zone/tradeskills.cpp
@@ -468,23 +468,42 @@ void Object::HandleCombine(Client* user, const NewCombine_Struct* in_combine, Ob
 		}
 	}
 
+	//pull the list of components so we can compare during lore check
+	std::string query = fmt::format("SELECT tre.item_id "
+	                                 "FROM tradeskill_recipe_entries AS tre "
+	                                 "WHERE tre.componentcount > 0 AND tre.recipe_id = {}",
+	                                 spec.recipe_id);
+	auto results = content_db.QueryDatabase(query);
+
+	if (!results.Success()) {
+		auto outapp = new EQApplicationPacket(OP_TradeSkillCombine, 0);
+		user->QueuePacket(outapp);
+		safe_delete(outapp);
+		return;
+	}
+
+	if(results.RowCount() < 1) {
+		LogTradeskills("Error in HandleCombine: no components returned");
+		auto outapp = new EQApplicationPacket(OP_TradeSkillCombine, 0);
+		user->QueuePacket(outapp);
+		safe_delete(outapp);
+		return;
+	}
+
+	if(results.RowCount() > 10) {
+		LogTradeskills("Error in HandleCombine: too many components returned ({})", results.RowCount());
+		auto outapp = new EQApplicationPacket(OP_TradeSkillCombine, 0);
+		user->QueuePacket(outapp);
+		safe_delete(outapp);
+		return;
+	}
+
 	// Check if Combine would result in Lore conflict
-	for (const auto& e : spec.onsuccess) {
-		auto success_item_inst = database.GetItem(e.first);
-		if (success_item_inst->LoreGroup > 0) {
-			continue;
-		}
-		if (user->CheckLoreConflict(success_item_inst)) {
-			EQ::SayLinkEngine linker;
-			linker.SetLinkType(EQ::saylink::SayLinkItemData);
-			linker.SetItemData(success_item_inst);
-			auto item_link = linker.GenerateLink();
-			user->MessageString(Chat::Red, TRADESKILL_COMBINE_LORE, item_link.c_str());
-			auto outapp = new EQApplicationPacket(OP_TradeSkillCombine, 0);
-			user->QueuePacket(outapp);
-			safe_delete(outapp);
-			return;
-		}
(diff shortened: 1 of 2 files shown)


zone/tradeskills.cpp
```diff
@@ -468,23 +468,42 @@ void Object::HandleCombine(Client* user, const NewCombine_Struct* in_combine, Ob
 		}
 	}
 
+	//pull the list of components so we can compare during lore check
+	std::string query = fmt::format("SELECT tre.item_id "
```
Kinglykrab, line 472: `const auto` maybe? Do we need the alias in the query if you're only referencing a single table and can use the column names without the need for an alias?

zone/tradeskills.cpp
```diff
@@ -468,23 +468,42 @@ void Object::HandleCombine(Client* user, const NewCombine_Struct* in_combine, Ob
...
+		return;
+	}
+
+	if(results.RowCount() < 1) {
+		LogTradeskills("Error in HandleCombine: no components returned");
+		auto outapp = new EQApplicationPacket(OP_TradeSkillCombine, 0);
+		user->QueuePacket(outapp);
+		safe_delete(outapp);
+		return;
+	}
```
Kinglykrab, line 491: Can these be consolidated since they'd both be an error?

zone/tradeskills.cpp
```diff
@@ -468,23 +468,42 @@ void Object::HandleCombine(Client* user, const NewCombine_Struct* in_combine, Ob
...
+
+	if(results.RowCount() > 10) {
```
Kinglykrab, line 493: Formatting.

zone/tradeskills.cpp
```diff
@@ -622,9 +641,9 @@ void Object::HandleAutoCombine(Client* user, const RecipeAutoCombine_Struct* rac
 	}
 
     //pull the list of components
-	std::string query = StringFormat("SELECT tre.item_id, tre.componentcount "
+	std::string query = fmt::format("SELECT tre.item_id, tre.componentcount "
```
Kinglykrab, line 644: `const auto` maybe? Do we need the alias in the query if you're only referencing a single table and can use the column names without the need for an alias?

zone/tradeskills.cpp
```diff
@@ -1882,3 +1894,36 @@ bool ZoneDatabase::DisableRecipe(uint32 recipe_id)
 
 	return false;
 }
+
+bool Client::CheckTradeskillLoreConflict(DBTradeskillRecipe_Struct& spec, MySQLRequestResult& results)
+{
+	std::vector< std::pair<uint32,uint8>> lore_conflict_list;
+	std::merge(spec.onsuccess.begin(), spec.onsuccess.end(), spec.onfail.begin(), spec.onfail.end(), std::back_inserter(lore_conflict_list));
+	std::merge(lore_conflict_list.begin(), lore_conflict_list.end(), spec.salvage.begin(), spec.salvage.end(), std::back_inserter(lore_conflict_list));
```
Kinglykrab, line 1902: Formatting.

zone/tradeskills.cpp
```diff
@@ -1882,3 +1894,36 @@ bool ZoneDatabase::DisableRecipe(uint32 recipe_id)
...
+
+	for (auto& e : lore_conflict_list) {
+		if (e.first) {
+			auto item_inst = database.GetItem(e.first);
+			if (!item_inst || item_inst->LoreGroup > 0) {
+				continue;
+			}
```
Kinglykrab, line 1909: Maybe a newline here?

zone/tradeskills.cpp
```diff
@@ -1882,3 +1894,36 @@ bool ZoneDatabase::DisableRecipe(uint32 recipe_id)
...
+			for (auto row = results.begin(); row != results.end(); ++row) {
```
Kinglykrab, line 1910: `for (auto row : results)`

zone/tradeskills.cpp
```diff
@@ -1882,3 +1894,36 @@ bool ZoneDatabase::DisableRecipe(uint32 recipe_id)
...
+				if (row[0]) {
+					auto component_item_inst = atoi(row[0]);
+					if (component_item_inst == item_inst->ID) {
+						continue;
+					}
+				}
+			}
```
Kinglykrab, line 1917: Newline here?

zone/tradeskills.cpp
```diff
@@ -1882,3 +1894,36 @@ bool ZoneDatabase::DisableRecipe(uint32 recipe_id)
...
+
+	for (auto& e : lore_conflict_list) {
+		if (e.first) {
+			auto item_inst = database.GetItem(e.first);
+			if (!item_inst || item_inst->LoreGroup > 0) {
+				continue;
+			}
+			for (auto row = results.begin(); row != results.end(); ++row) {
+				if (row[0]) {
+					auto component_item_inst = atoi(row[0]);
```
Kinglykrab, line 1912: `Strings::ToInt`

zone/tradeskills.cpp
Aeadoin: cleaned up.

zone/tradeskills.cpp
Aeadoin: Yeah, I just merged all these checks, it's just verbose for no real gain.

zone/tradeskills.cpp
Aeadoin: cleaned up.

zone/tradeskills.cpp
Aeadoin: Sure, might as well for readability purposes separate these blocks a bit

zone/tradeskills.cpp
Aeadoin: Good stuff

zone/tradeskills.cpp
```diff
@@ -469,29 +469,13 @@ void Object::HandleCombine(Client* user, const NewCombine_Struct* in_combine, Ob
...
-		LogTradeskills("Error in HandleCombine: no components returned");
-		auto outapp = new EQApplicationPacket(OP_TradeSkillCombine, 0);
-		user->QueuePacket(outapp);
-		safe_delete(outapp);
-		return;
-	}
-
-	if(results.RowCount() > 10) {
-		LogTradeskills("Error in HandleCombine: too many components returned ({})", results.RowCount());
+	if(!results.Success() || results.RowCount() < 1 || results.RowCount() > 10) {
```
Kinglykrab, line 478: Formatting. `if (` versus `if(`

Aeadoin: fixed, thanks!

zone/tradeskills.cpp
```diff
@@ -468,23 +468,26 @@ void Object::HandleCombine(Client* user, const NewCombine_Struct* in_combine, Ob
 		}
 	}
 
+	//pull the list of components so we can compare during lore check
+	const auto query = fmt::format("SELECT item_id "
```
Akkadius, line 472: Instead of using this and passing the MySQL result object into another function, this is a perfect use case for using repositories. Please refactor to use 

There's already examples of them being used in tradeskills

```cpp
auto re = TradeskillRecipeEntriesRepository::GetWhere(
	database,
	fmt::format("recipe_id = {} ORDER BY id ASC", recipe_id)
);
```

Then you have a vector of rows that you can pass to the other function if you'd like and makes things far more flexible

Aeadoin: this is great feedback! made these changes.

Akkadius (approved):

crotonrz:
This is still having the same issue when combining a lore item that results in an item in the same lore group.

Aeadoin:
@crotonrz that was fixed in https://github.com/EQEmu/Server/pull/3005 and just pushed https://github.com/EQEmu/Server/pull/3009 up to resolve a crash where an item_id is specified but doesn't actually have an item instance.

PR #2981, opened 2023-02-23
[Bots] Cleanup BotDatabase::LoadBuffs

Aeadoin:
Added Checks to initialize all buff slots (mimics Client LoadBuffs method)
Cleaned up all atoi conversions, change to Strings::ToInt and atoul where applicable.

### DIFF
--- a/zone/bot_database.cpp
+++ b/zone/bot_database.cpp
@@ -639,10 +639,11 @@ bool BotDatabase::DeleteBot(const uint32 bot_id)
 
 bool BotDatabase::LoadBuffs(Bot* bot_inst)
 {
-	if (!bot_inst)
+	if (!bot_inst) {
 		return false;
+	}
 
-	query = StringFormat(
+	query = fmt::format(
 		"SELECT"
 		" `spell_id`,"
 		" `caster_level`,"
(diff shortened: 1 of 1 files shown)


zone/bot_database.cpp
```diff
@@ -663,45 +664,60 @@ bool BotDatabase::LoadBuffs(Bot* bot_inst)
...
-		" WHERE `bot_id` = '%u'",
+		" WHERE `bot_id` = '{}'",
 		bot_inst->GetBotID()
 	);
 	auto results = database.QueryDatabase(query);
-	if (!results.Success())
+	if (!results.Success()) {
 		return false;
-	if (!results.RowCount())
+	}
```
Kinglykrab, line 673: Newline between here?

zone/bot_database.cpp
```diff
@@ -663,45 +664,60 @@ bool BotDatabase::LoadBuffs(Bot* bot_inst)
 		" `extra_di_chance`,"
 		" `instrument_mod`"
 		" FROM `bot_buffs`"
-		" WHERE `bot_id` = '%u'",
+		" WHERE `bot_id` = '{}'",
```
Kinglykrab, line 667: `'{}'` is only necessary for strings, can just do `{}`.

Kinglykrab (commented):
May be worth converting this to load the data from repositories. Not now, but in the future.

zone/bot_database.cpp
```diff
@@ -663,45 +664,60 @@ bool BotDatabase::LoadBuffs(Bot* bot_inst)
...
+			bot_buffs[buff_count].counters = atoul(row[5]);
+		}
+		else if (CalculateCurseCounters(bot_buffs[buff_count].spellid) > 0) {
+			bot_buffs[buff_count].counters = atoul(row[6]);
+		}
+		else if (CalculateCorruptionCounters(bot_buffs[buff_count].spellid) > 0) {
+			bot_buffs[buff_count].counters = atoul(row[7]);
+		} else {
+			bot_buffs[buff_count].counters = 0;
+		}
```
Kinglykrab, line 709: Not sure our standard, but I usually try to do this.
```cpp
if (a) {
} else if (b) {
} else if (c) {
}
```

Akkadius: Chances are else can just be the default and else is not usually necessary

Aeadoin: cleaned up, thanks!

Kinglykrab (approved):

PR #2990, opened 2023-02-25
[Rules] Add ResurrectionEffectsBlock

nytmyr:
Notes:

This adds the rule Spells:ResurrectionEffectsBlock that will prevent anything from overwriting Resurrection Sickness Effects. Currently they are able to be overwritten by certain spells.

Default state: Disabled/False.

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -3161,6 +3161,11 @@ int Mob::CheckStackConflict(uint16 spellid1, int caster_level1, uint16 spellid2,
 		}
 		if (sp2_value != sp1_value)
 			values_equal = false;
+
+		if (RuleB(Spells, ResurrectionEffectsBlock) && IsResurrectionEffects(spellid1)) {
+			LogSpells("ResurrectionEffectsBlock triggered -- [{}] is blocked by [{}]", sp2.name, sp1.name);
+			return -1;	// can't stack
+		}
 		//we dont return here... a better value on this one effect dosent mean they are
 		//all better...
 
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -434,7 +434,7 @@ RULE_BOOL(Spells, IllusionsAlwaysPersist, false, "Allows Illusions to persist be
 RULE_BOOL(Spells, UseItemCastMessage, false, "Enable to use the \"item begins to glow\" messages when casting from an item.")
 RULE_BOOL(Spells, TargetsTargetRequiresCombatRange, true, "Disable to remove combat range requirement from Target's Target Spell Target Type")
 RULE_BOOL(Spells, NPCBuffLevelRestrictions, false, "Impose BuffLevelRestrictions on NPCs if true")
-RULE_BOOL(Spells, ResurrectionEffectsBlock, false, "If enabled, resurrection effects cannot be overwritten.")
+RULE_BOOL(Spells, ResurrectionEffectsBlock, true, "If enabled, resurrection effects cannot be overwritten.")
 RULE_CATEGORY_END()
 
 RULE_CATEGORY(Combat)
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -434,6 +434,7 @@ RULE_BOOL(Spells, IllusionsAlwaysPersist, false, "Allows Illusions to persist be
 RULE_BOOL(Spells, UseItemCastMessage, false, "Enable to use the \"item begins to glow\" messages when casting from an item.")
 RULE_BOOL(Spells, TargetsTargetRequiresCombatRange, true, "Disable to remove combat range requirement from Target's Target Spell Target Type")
 RULE_BOOL(Spells, NPCBuffLevelRestrictions, false, "Impose BuffLevelRestrictions on NPCs if true")
+RULE_BOOL(Spells, ResurrectionEffectsBlock, false, "If enabled, resurrection effects cannot be overwritten.")
 RULE_CATEGORY_END()
 
 RULE_CATEGORY(Combat)


Kinglykrab:
This may need to be made more specific in its logic, as it seems to currently just block ANY spell not just ones that can overwrite the Resurrection Effects. You may also want to check that Resurrection Effects are even enabled in conjunction with this rule you added, as you can’t use this rule without them being enabled.

noudess:
I'm all for this rule and change (do we need  rule, or do res effects still block str spells, etc on live), but as Kingly said, we only want to block certain spells.  I think anything with STR in it.  Am I right about that?

You should still be able to heal, add other buffs, etc.  At least that's my recollection.

nytmyr:
I'll dig in to this more this weekend when I'm off work. I've personally never "noticed" anything blocked that shouldn't have been, the same full raid buffs took with/without Rez Sickness but that doesn't mean there aren't outliers. I can definitely see how this could be a possibility now that y'all pointed it out.

As for the Res rule suggestion: do we want to tie them together by default or leave them separate? If they're already enabled this rule comes in to play; if they're not enabled, it in theory never should unless Res Effects were scripted for some reason. Just curious.

nytmyr:
So I did some testing with this as is and I don't see any spells blocked other than spells that should be blocked with this rule (Strength spells).

Any examples/scenarios y'all have that this would cause issues with?

noudess:
> So I did some testing with this as is and I don't see any spells blocked other than spells that should be blocked with this rule (Strength spells).
> 
> Any examples/scenarios y'all have that this would cause issues with?

I guess I'm blind.  How does this code limit what is blocked to just str spells?  I don't see any checks at all for the type of spell being cast, just the spell being overwritten.  Is that earlier in the code or implied?

I love this PR, cause I want res effects to stay.  I just must be blind to how this limits it to str spells.

nytmyr:
> > So I did some testing with this as is and I don't see any spells blocked other than spells that should be blocked with this rule (Strength spells).
> > Any examples/scenarios y'all have that this would cause issues with?
> 
> I guess I'm blind. How does this code limit what is blocked to just str spells? I don't see any checks at all for the type of spell being cast, just the spell being overwritten. Is that earlier in the code or implied?
> 
> I love this PR, cause I want res effects to stay. I just must be blind to how this limits it to str spells.

Spell checks, especially overwrites are super confusing lol. 

I could still be wrong here, but my understanding is that this is basically the very end of the overwrite checks where a potential conflict was detected of the effectids and it's now comparing the values to see which is strong. In which case nothing should pass a stronger check against res effects.

zone/spells.cpp
```diff
@@ -3161,6 +3161,12 @@ int Mob::CheckStackConflict(uint16 spellid1, int caster_level1, uint16 spellid2,
 		}
 		if (sp2_value != sp1_value)
 			values_equal = false;
+
+		if (RuleB(Spells, ResurrectionEffectsBlock) && IsResurrectionEffects(spellid1)) {
+			LogSpells("Spell [{}] is blocked by [{}]",
+				sp2.name, sp1.name);
```
Akkadius, line 3167: Doesn't need to be on a newline

zone/spells.cpp
```diff
@@ -3161,6 +3161,12 @@ int Mob::CheckStackConflict(uint16 spellid1, int caster_level1, uint16 spellid2,
 		}
 		if (sp2_value != sp1_value)
 			values_equal = false;
+
+		if (RuleB(Spells, ResurrectionEffectsBlock) && IsResurrectionEffects(spellid1)) {
```
Akkadius, line 3165: Haven't gotten a clear answer from anyone. Is this live behavior? If so, we could remove the rule entirely and leave this as sane default

nytmyr: I cannot confirm whether or not this is true live behavior, however it does exist on p99 which is what I based the rule off of. (Not that it means it's correct or not, but it's the closest thing I can go off of)

noudess:
I just tested this on live.  

Cleric Reckless Strenth, and two shaman str and str/hp buffs.

They all ADD a new buff but do not take down Rez effects.  Can we make that happen?

zone/spells.cpp
noudess: https://github.com/EQEmu/Server/pull/2990#issuecomment-1454754358

Aeadoin:
> 

What effect is the "Resurrection Sickness" on live? are you able to inspect it, or see the spell ID?

raidloot.com shows there are 4 different versions of  "Resurrection Sickness", only SpellID 757 has SPA 4 (Strength debuff) which is what we use by default. I don't see anything to indicate SpellID 757 is somehow letting other Spells with a conflicting SPA bypass that restriction (I don't know if there is even a flag for that on with live spell data)

Aeadoin:
> > 
> 
> What effect is the "Resurrection Sickness" on live? are you able to inspect it, or see the spell ID?
> 
> raidloot.com shows there are 4 different versions of "Resurrection Sickness", only SpellID 757 has SPA 4 (Strength debuff) which is what we use by default. I don't see anything to indicate SpellID 757 is somehow letting other Spells with a conflicting SPA bypass that restriction (I don't know if there is even a flag for that on with live spell data)

Found some info, Live changed their "Resurrection Sickness" a couple years ago.

```- Changed the way Resurrection Sickness works. It no longer reduces stats. It will now increase the cast time of detrimental spells by 10% and decrease your attack speed by 10%. Both of these values will stack with other spells rather than preventing haste or spell haste effects.
- Changed the way the PvP version of Resurrection Sickness works. The cast time increase and attack speed decrease have the same values, but these effects will stack with other spells rather than preventing haste or spell haste effects.
- Changed the way Revival Sickness works. It no longer caps mana and endurance at 4000. Instead it caps at 25% and slowly decreases the cap by 1% every 5 levels starting at 50. For example, the cap at level 85 is 18% and the cap at level 115 is 12%.```

nytmyr:
This was going to be my comment as well, they redid all the res sickness stuff so can't verify how it use to function. 

Anything is possible and we could definitely figure a workaround to move it to a different buff slot but keep Res Sickness if we were to choose to do so, but that's a decision above me to do.

If we were to do that we'd need to verify maxbuffslots again as well before pushing to the next available slot id

noudess:
Anything better than what we have.  Pure block in this patch with rule is good IMO.  Then if they want to add the new effect and stacking as default behavior without rule, so be it.

Akkadius:
Changed rule to be default true, I'm good with it in its current form until someone figures out otherwise

Akkadius (approved):

noudess:
@nytmyr On my server I can run around for like half a minute before the 0 str stops me.  Same there?

PR #3005, opened 2023-02-26
[Bug Fix] Fix for Lore Components where component is returned.

Aeadoin:
This resolves issues seen with BiC quest line combines, where the item is returned on success, fail, salvage. However the recipe entries for each different return type are split between multiple rows. 

My previous PR was meant to fix the above, but was incorrect logic. my testing failed to find this, as the checks worked correctly on the first component if it was lore, not later components.

The correct implementation is to loop through all components first before performing any lore checks, allowing us to exclude any components that are lore, but have a separate row for return on success, fail, salvage.

Verified all previous use cases work as expected as well, meaning:

- Combine uses a lore container, which is returned.
- Combines that have a Lore Group component (Epic Lore Group, or other above 0) and return the same lore-group on success.
- Components are lore, and returned on success, fail, salvage (but in same row entry)

### DIFF
--- a/zone/tradeskills.cpp
+++ b/zone/tradeskills.cpp
@@ -1876,35 +1876,40 @@ bool Client::CheckTradeskillLoreConflict(int32 recipe_id)
 		return false;
 	}
 
-	const EQ::ItemData* fi = nullptr;
-	const EQ::ItemData* ei = nullptr;
-	for (auto &f : recipe_entries) {
-		if (f.item_id) {
-			fi = database.GetItem(f.item_id);
-			for (auto &e : recipe_entries) {
-				bool fi_is_valid = e.item_id && fi && fi->LoreGroup != 0;
-				bool item_id_match = e.item_id == f.item_id;
+	// validate which items from the recipe we will call CheckLoreConflict on
+	for (const auto &tre : recipe_entries) {
+		if (tre.item_id) {
+			auto tre_inst = database.GetItem(tre.item_id);
+
+			// To compare items we iterate against each item in the recipe that have a loregroup.
+			for (auto &tre_update_item : recipe_entries) {
+				bool fi_is_valid = tre_update_item.item_id && tre_inst && tre_inst->LoreGroup != 0;
 
 				if (fi_is_valid) {
-					ei = database.GetItem(e.item_id);
-					bool ei_is_valid = ei && ei->LoreGroup != 0;
-					bool unique_lore_group_match = fi->LoreGroup > 0 && fi->LoreGroup == ei->LoreGroup;
-					bool component_count_is_valid = e.componentcount == 0 && f.componentcount > 0;
-					if (ei_is_valid && (item_id_match || unique_lore_group_match) && component_count_is_valid) {
-						e.item_id = 0;
+					auto tre_update_item_inst = database.GetItem(tre_update_item.item_id);
+					bool ei_is_valid = tre_update_item_inst && tre_update_item_inst->LoreGroup != 0;
+					bool unique_lore_group_match = tre_inst->LoreGroup > 0 && tre_inst->LoreGroup == tre_update_item_inst->LoreGroup;
+					bool component_count_is_valid = tre_update_item.componentcount == 0 && tre.componentcount > 0;
+
+					// If the recipe item is a component, and matches a unique lore group (> 0) or the item_id matches another entry in the recipe
+					// zero out the item_id, this will prevent us from doing a lore check inadvertently where
+					// the item is a component, and returned on success, fail, salvage.
+					// or uses an item that is part of a unique loregroup that returns an item of the same unique loregroup
+					if (ei_is_valid && (tre_update_item.item_id == tre.item_id || unique_lore_group_match) && component_count_is_valid) {
+						tre_update_item.item_id = 0;
 					}
 				}
 			}
 
-			if (fi) {
-				if (fi->LoreGroup == 0 || f.componentcount > 0 || f.iscontainer) {
+			if (tre_inst) {
+				if (tre_inst->LoreGroup == 0 || tre.componentcount > 0 || tre.iscontainer) {
 					continue;
 				}
 
-				if (CheckLoreConflict(fi)) {
+				if (CheckLoreConflict(tre_inst)) {
 					EQ::SayLinkEngine linker;
 					linker.SetLinkType(EQ::saylink::SayLinkItemData);
-					linker.SetItemData(fi);
+					linker.SetItemData(tre_inst);
 					auto item_link = linker.GenerateLink();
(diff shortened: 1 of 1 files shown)


zone/tradeskills.cpp
```diff
@@ -1866,37 +1866,59 @@ bool ZoneDatabase::DisableRecipe(uint32 recipe_id)
...
 	);
 	if (recipe_entries.empty()) {
 		return false;
 	}
 
-	for (auto& f : recipe_entries) {
-		for (auto &e: recipe_entries) {
-			if (f.componentcount > 0 && e.item_id == f.item_id && e.componentcount == 0) {
-				e.item_id = 0;
+	const EQ::ItemData* f_item_inst = nullptr;
```
Akkadius, line 1879: Please shorten to `fi`

Still a bit verbose given this is a receiver of a complex type and makes accessing it verbose. Making it verbose doesn't exactly give more to readability

Same with below

`ei`

zone/tradeskills.cpp
```diff
@@ -1866,37 +1866,59 @@ bool ZoneDatabase::DisableRecipe(uint32 recipe_id)
...
+	const EQ::ItemData* e_item_inst = nullptr;
+	for (auto &f : recipe_entries) {
+		if (f.item_id) {
+			f_item_inst = database.GetItem(f.item_id);
+			for (auto &e : recipe_entries) {
+				if (e.item_id && f_item_inst && f_item_inst->LoreGroup != 0) {
+					e_item_inst = database.GetItem(e.item_id);
+					if (
```
Akkadius, line 1887: This is insane to read, i'd like to see this broken up a bit more in bools that describe the intent of what's happening in this part to make it obvious to the next reader.

Example of local vars describing intent

```cpp
    for (int i = PlayerEvent::GM_COMMAND; i != PlayerEvent::MAX; i++) {
        bool is_in_database = std::find(db.begin(), db.end(), i) != db.end();
        bool is_deprecated  = Strings::Contains(PlayerEvent::EventName[i], "Deprecated");
        bool is_implemented = !Strings::Contains(PlayerEvent::EventName[i], "Unimplemented");
```

Imagine if that was all on one line or nested

zone/tradeskills.cpp
Aeadoin: done!

zone/tradeskills.cpp
Aeadoin: just retesting with these changes made, pretty sure I didn't mess anything up, but better to be safe.

zone/tradeskills.cpp
```diff
@@ -1866,37 +1866,53 @@ bool ZoneDatabase::DisableRecipe(uint32 recipe_id)
...
 		return false;
 	}
 
-	for (auto& f : recipe_entries) {
-		for (auto &e: recipe_entries) {
-			if (f.componentcount > 0 && e.item_id == f.item_id && e.componentcount == 0) {
-				e.item_id = 0;
+	const EQ::ItemData* fi = nullptr;
+	const EQ::ItemData* ei = nullptr;
+	for (auto &f : recipe_entries) {
```
Akkadius, line 1881: In this case, what is `f` and what is `e` ? 

It would be good to have some comments above this describing what is happening

zone/tradeskills.cpp
```diff
@@ -1866,37 +1866,53 @@ bool ZoneDatabase::DisableRecipe(uint32 recipe_id)
...
 	);
 	if (recipe_entries.empty()) {
 		return false;
 	}
 
-	for (auto& f : recipe_entries) {
-		for (auto &e: recipe_entries) {
-			if (f.componentcount > 0 && e.item_id == f.item_id && e.componentcount == 0) {
-				e.item_id = 0;
+	const EQ::ItemData* fi = nullptr;
```
Akkadius, line 1879: There a reason this isn't just inlined where we call `GetItem` ?

zone/tradeskills.cpp
```diff
@@ -1866,37 +1866,53 @@ bool ZoneDatabase::DisableRecipe(uint32 recipe_id)
...
+		if (f.item_id) {
+			fi = database.GetItem(f.item_id);
+			for (auto &e : recipe_entries) {
+				bool fi_is_valid = e.item_id && fi && fi->LoreGroup != 0;
+				bool item_id_match = e.item_id == f.item_id;
+
+				if (fi_is_valid) {
+					ei = database.GetItem(e.item_id);
+					bool ei_is_valid = ei && ei->LoreGroup != 0;
+					bool unique_lore_group_match = fi->LoreGroup > 0 && fi->LoreGroup == ei->LoreGroup;
```
Akkadius, line 1891: This is a great variable name, tells me exactly what's happening

zone/tradeskills.cpp
```diff
@@ -1866,37 +1866,53 @@ bool ZoneDatabase::DisableRecipe(uint32 recipe_id)
...
+					bool component_count_is_valid = e.componentcount == 0 && f.componentcount > 0;
```
Akkadius, line 1892: This is a great variable name, tells me exactly what's happening

zone/tradeskills.cpp
```diff
@@ -1866,37 +1866,53 @@ bool ZoneDatabase::DisableRecipe(uint32 recipe_id)
...
-			if (f.componentcount > 0 && e.item_id == f.item_id && e.componentcount == 0) {
-				e.item_id = 0;
+	const EQ::ItemData* fi = nullptr;
+	const EQ::ItemData* ei = nullptr;
+	for (auto &f : recipe_entries) {
+		if (f.item_id) {
+			fi = database.GetItem(f.item_id);
+			for (auto &e : recipe_entries) {
+				bool fi_is_valid = e.item_id && fi && fi->LoreGroup != 0;
+				bool item_id_match = e.item_id == f.item_id;
```
Akkadius, line 1886: This may not necessarily need a bool to describe intent, it's pretty clear to the next traveler that this check is looking for where item_id's are equal

zone/tradeskills.cpp
Aeadoin: good point!

zone/tradeskills.cpp
Aeadoin: makes sense to me, removed it.

zone/tradeskills.cpp
Aeadoin: thanks!

zone/tradeskills.cpp
Aeadoin: okay yeah good idea, changed the names to be more descriptive... and added commenting, hopefully it clears it up?

Aeadoin:
I've retested with the latest version, and all looks good.

Akkadius (approved):
LGTM; thanks for being receptive to feedback

crotonrz:
Thank Aead.

PR #3008, opened 2023-02-27
[Bug Fix] Fix for Discipline Loading from Database causing issues with slot_ids

Aeadoin:
If a Server Operator attempts to cleanup disciplines after they are learned from the database, they may encounter issues after the fact when players try to learn new disciplines due to how we were loading data into the Disciplines_Struct.

We were loading data in with no accounting for Slot_id, so when learning a new Discipline if an older row was deleted the Struct would not match up with the correct Slot_id numbers, causing players to overwrite disciplines they have already learned.

This allows a null entry to be added to the Disciplines_Struct.disciplines.value, so when we learn a new discipline it'll find the null and insert to the slot that is free, and not try to insert at the end.

### DIFF
--- a/zone/zonedb.cpp
+++ b/zone/zonedb.cpp
@@ -814,20 +814,24 @@ bool ZoneDatabase::LoadCharacterLeadershipAA(uint32 character_id, PlayerProfile_
 }
 
 bool ZoneDatabase::LoadCharacterDisciplines(uint32 character_id, PlayerProfile_Struct* pp){
-	std::string query = StringFormat(
-		"SELECT				  "
-		"disc_id			  "
-		"FROM				  "
+	auto query = fmt::format(
+		"SELECT "
+		"disc_id, slot_id "
+		"FROM "
 		"`character_disciplines`"
-		"WHERE `id` = %u ORDER BY `slot_id`", character_id);
+		"WHERE `id` = {} ORDER BY `slot_id`", character_id);
 	auto results = database.QueryDatabase(query);
-	int i = 0;
 
+	if (!results.Success() || !results.RowCount()) {
+		return false;
+	}
+
+	int i = 0;
 	/* Initialize Disciplines */
 	memset(pp->disciplines.values, 0, (sizeof(pp->disciplines.values[0]) * MAX_PP_DISCIPLINES));
 	for (auto& row = results.begin(); row != results.end(); ++row) {
 		if (i < MAX_PP_DISCIPLINES)
-			pp->disciplines.values[i] = atoi(row[0]);
+			pp->disciplines.values[Strings::ToInt(row[1])] = atoi(row[0]);
         ++i;
     }
 	return true;
--- a/zone/zonedb.cpp
+++ b/zone/zonedb.cpp
@@ -13,6 +13,7 @@
 #include "zonedb.h"
 #include "aura.h"
 #include "../common/repositories/criteria/content_filter_criteria.h"
+#include "../common/repositories/character_disciplines_repository.h"
 #include "../common/repositories/npc_types_repository.h"
 
 #include <ctime>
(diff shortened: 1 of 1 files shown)


zone/zonedb.cpp
```diff
@@ -814,20 +814,24 @@ bool ZoneDatabase::LoadCharacterLeadershipAA(uint32 character_id, PlayerProfile_
 }
 
 bool ZoneDatabase::LoadCharacterDisciplines(uint32 character_id, PlayerProfile_Struct* pp){
-	std::string query = StringFormat(
-		"SELECT				  "
-		"disc_id			  "
-		"FROM				  "
+	auto query = fmt::format(
```
Akkadius, line 817: Not suggesting it needs to be in this PR, but seems like a great opportunity to just use `CharacterDisciplinesRepository` here since its already made and available to clean this up and make it more readable

zone/zonedb.cpp
```diff
@@ -814,20 +814,24 @@ bool ZoneDatabase::LoadCharacterLeadershipAA(uint32 character_id, PlayerProfile_
...
+		return false;
+	}
+
+	int i = 0;
 	/* Initialize Disciplines */
 	memset(pp->disciplines.values, 0, (sizeof(pp->disciplines.values[0]) * MAX_PP_DISCIPLINES));
 	for (auto& row = results.begin(); row != results.end(); ++row) {
 		if (i < MAX_PP_DISCIPLINES)
-			pp->disciplines.values[i] = atoi(row[0]);
+			pp->disciplines.values[Strings::ToInt(row[1])] = atoi(row[0]);
```
Akkadius, line 834: We should add a IsValidSpell check here while we're at it

zone/zonedb.cpp
```diff
@@ -814,20 +814,24 @@ bool ZoneDatabase::LoadCharacterLeadershipAA(uint32 character_id, PlayerProfile_
...
 
+	if (!results.Success() || !results.RowCount()) {
+		return false;
+	}
+
+	int i = 0;
 	/* Initialize Disciplines */
 	memset(pp->disciplines.values, 0, (sizeof(pp->disciplines.values[0]) * MAX_PP_DISCIPLINES));
 	for (auto& row = results.begin(); row != results.end(); ++row) {
 		if (i < MAX_PP_DISCIPLINES)
```
Akkadius, line 833: Close off braces

zone/zonedb.cpp
Aeadoin: Might as well do it now :) won't get done if not!

fryguy503:
Oh man I got bit by this when I removed an Out of Era disc from the DB those classes have been having issues since with consistency of discs.

Aeadoin:
> Oh man I got bit by this when I removed an Out of Era disc from the DB those classes have been having issues since with consistency of discs.

Yeah, another issue I've noticed is Discs don't cleanup Discs that are a part of spellgroups, just inserts the new entry but only displays the highest rank. so it'll become an issue in later expansions where we have multiple ranks per disc. Going to try to clean that up soon.

zone/zonedb.cpp
```diff
@@ -814,22 +815,27 @@ bool ZoneDatabase::LoadCharacterLeadershipAA(uint32 character_id, PlayerProfile_
...
+	int i = 0;
 	/* Initialize Disciplines */
 	memset(pp->disciplines.values, 0, (sizeof(pp->disciplines.values[0]) * MAX_PP_DISCIPLINES));
-	for (auto& row = results.begin(); row != results.end(); ++row) {
-		if (i < MAX_PP_DISCIPLINES)
-			pp->disciplines.values[i] = atoi(row[0]);
-        ++i;
-    }
+	for (auto& row : character_disciplines) {
+		if (i < MAX_PP_DISCIPLINES && IsValidSpell(row.disc_id)) {
```
Akkadius, line 834: Shouldn't `row.slot_id` be the thing we're checking here and get rid of `i` altogether ?

Aeadoin: yes, that should be fine. let me do that.

PR #3011, opened 2023-02-28
[Bots] Prevent interrupt spam when OOM

nytmyr:
Notes:

https://user-images.githubusercontent.com/53322305/223123434-2bcba221-d13f-4320-9995-e6a3619ccb84.mp4

Bots currently do not do a mana check until the spell cast has already been started which results in an interrupt immediately after. If it is the last spell for a bot to cast, it tends to result in interrupt spam until the bot has enough mana. This will block the interrupt spam if the spell is invalid.

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -1158,10 +1158,16 @@ void Mob::InterruptSpell(uint16 message, uint16 color, uint16 spellid)
 	uint16 message_other;
 	bool bard_song_mode = false; //has the bard song gone to auto repeat mode
 	if (!IsValidSpell(spellid)) {
-		if(bardsong) {
+		if (bardsong) {
 			spellid = bardsong;
 			bard_song_mode = true;
-		} else {
+		}
+		else {
+			if (IsBot() && !message && !color && !spellid) { // this is to prevent bots from spamming interrupts messages when trying to cast while OOM
+				ZeroCastingVars();	// resets all the state keeping stuff
+				LogSpells("Spell [{}] has been interrupted - Bot [{}] doesn't have enough mana", spellid, GetCleanName());
+				return;
+			}
 			spellid = casting_spell_id;
 		}
 	}


Kinglykrab:
I'm not seeing this spam, what are you doing to see it?

nytmyr:
> I'm not seeing this spam, what are you doing to see it?

I'll rebuild a stock master and see if it happens.

Take a naked caster bot, get a training dummy and nuke it down to where the bot starts casting (95% by default IIRC), hold aggro on it so the caster doesn't die or get interrupted, let it nuke until it's OOM. Once it runs OOM it'll spam interrupt until it has enough mana to nuke again.

I saw someone report this and I had fixed it on my server, however there is a possibility that it was something I did when I redid my bot spell AI

Kinglykrab:
I'm not seeing anything message wise that leads me to believe it's spam interrupting.

nytmyr:
> I'm not seeing anything message wise that leads me to believe it's spam interrupting.

When I get a chance I'll check Master and see how it's done if different. 

How this is done in my case, the bot tries to cast the nuke which doesn't have a mana check before the cash starts (this should be addressed) so it proceeds through the castspell methods and then it fails during a mana check which triggers the interrupt method with all 0 values (suppose to be an invalid spell and cancel in the method) however because it is a valid spell, during the !SpellID check it validates there is a casted spell id so it defaults the spell id which then passes the final spellid check before outputting an interrupt message, so it then outputs the default interrupt message. This repeats each AI cycle till it passes the mana check.

nytmyr:
I went through the master code on mobile and that appears to be true on master as well, but when I get to a PC I'll actually build and test.

nytmyr:
https://github.com/EQEmu/Server/blob/d2c3c14ae0c391afa5389bbde8eecfbe6824995c/zone/botspellsai.cpp#L539

Don't use a cleric or enchanter to test btw or this will do a mana check.

nytmyr:
> I'm not seeing anything message wise that leads me to believe it's spam interrupting.

Attached a video. Confirmed on a stock master compile and stock PEQ DB that this does occur.

trentdm:
I've see this recently as well. I'll grab a screenshot next time I do

Aeadoin (approved):
I was able to reproduce this as well.

Kinglykrab (approved):

PR #3014, opened 2023-03-01
[Rules] Add Multiplier for Heroic Stats.

Aeadoin:
Allows Server Operators to increase/Reduce effect of Heroic Stats by setting a multiplier for each stat.

Have found most locations where heroic stats are leveraged, I've chosen not to allow the multipliers to work on Evasion based skills (as it could easily cause an operator to shoot their own foot)

Still need to ensure all fields from #mystats command are updating correctly. Max mana wasn't when I increased multiplier on Heroic Wisdom, but it was showing the increase correctly in the client inventory.


Aeadoin:
#mystats is now showing correctly for Mana and regen values provided by heroic stats.

Kinglykrab (approved):

Akkadius (approved):

PR #3015, opened 2023-03-01
[Scaling] Add support for pipe-separated zone IDs and versions

Kinglykrab:
# Notes
- Allows `|` separated zone IDs and instance versions within the scaling data.
- Loads scaling data on zone bootup as well, without needing to repop for it to initialize.


utils/sql/git/required/2023_02_28_npc_scaling_zone_list_version_list.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE `npc_scale_global_base`
+    CHANGE COLUMN `zone_id` `zone_id_list` text CHARACTER SET latin1 COLLATE latin1_swedish_ci NOT NULL DEFAULT 0 AFTER `level`,
+    CHANGE COLUMN `instance_version` `instance_version_list` text CHARACTER SET latin1 COLLATE latin1_swedish_ci NOT NULL DEFAULT 0 AFTER `zone_id_list`,
+    DROP PRIMARY KEY,
+    ADD PRIMARY KEY (`type`, `level`, `zone_id_list`(255), `instance_version_list`(255)) USING BTREE;
```
Akkadius, line 5: Since we're overloading `zone_id_list` and `instance_version_list` here - it probably doesn't make sense to make this part of the primary key

Akkadius (approved):

PR #3018, opened 2023-03-01
[Bug Fix] Fix an issue where EVENT_TIMER timers would not be cleaned up after zone

noudess:
Entries in the QTimerList for clients used mob as the key when deciding if a timer's client was in the zone and was the correct client.

If you log in character a, start a timer, log out character a and log in character b  (in the same zone from the same instance of client) character b inherits and runs character a's timer.

This PR adds charid as a secondary key for client timers.

Please note that the existing code cleans up timers when a client is no longer in zone.  My patch does the exact same thing if a clients character ID has changed.  I noticed that a previous patch (that I am not running yet) also checks to make sure that the client has an EVENT_TIMER subroutine.  That change is good, but it leaves the client timer in the QTimerList forever, until the Client leaves the zone.  So this fix also removed the timer in that case as well.

### DIFF
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -518,7 +518,9 @@ Mob::Mob(
 }
 
 Mob::~Mob()
-{
+{ 
+	quest_manager.stopalltimers(this);
+
 	mMovementManager->RemoveMob(this);
 
 	AI_Stop();
--- a/zone/questmgr.cpp
+++ b/zone/questmgr.cpp
@@ -82,6 +82,19 @@ QuestManager::QuestManager() {
 QuestManager::~QuestManager() {
 }
 
+QuestManager::QuestTimer::QuestTimer(int duration, Mob *_mob, std::string _name)
+	: mob(_mob), name(_name), Timer_(duration) {
+	Timer_.Start(duration, false);
+	// mob the same if a client logs out one char and logs in a new
+	// one in same zone.  Use charid as a secondary key.
+	if (_mob && _mob->IsClient()) {
+		charid = _mob->CastToClient()->CharacterID();
+		}
+	else {
+		charid = 0;
+	}
+}
+
 void QuestManager::Process() {
 	std::list<QuestTimer>::iterator cur = QTimerList.begin(), end;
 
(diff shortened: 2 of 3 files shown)


Aeadoin (changes requested):
Would recommend just stopping all timers in Mob::~Mob()

```quest_manager.stopalltimers(this);```

noudess:
> Would recommend just stopping all timers in Mob::~Mob()
> 
> `quest_manager.stopalltimers(this);`

Makes sense.  We still need this to handle disconnects and/or crashes right?  Or are we guaranteed to run destructor in all cases?  The existing code previously handled zoning clients in the same place.  Would that also not be needed?

Aeadoin:
> 

Any scenario where the client leaves the zone would trigger the mob destructor, so that would cover disconnects/crashes/camping/zoning.

noudess:
> > 
> 
> Any scenario where the client leaves the zone would trigger the mob destructor, so that would cover disconnects/crashes/camping/zoning.

Ok, so one question still stands.  If you look at Process in QuestManager, they check if the timer is for a mob in the zone, and if not, they erase the timer.  That's existing - pre PR code - that should go away as well correct, since it won't be possible?

Torn between wasted code and CPU time, and safe rails.

noudess:
> > 
> 
> I'd think it could be left.

Only I'll add the changes and remove my new stuff.  Thanks!

Will retest tomorrow.

noudess:
I'm gonna take a shot at removing the check:

if (cur->mob && entity_list.IsMobInZone(cur->mob)) {

out of the code.  This will save this processing timer every single timer expiration on the whole server.  It really seems to be just catching timers leftover from dead mobs or zoned clients.  If we're cleaning up in ~Mob() as you suggested, seems like we ought to see what happens now, while the focus is here.

Aeadoin (approved):
Looks good to me, I'm assuming you've tested?

noudess:
> Looks good to me, I'm assuming you've tested?



> Looks good to me, I'm assuming you've tested?

Yes, and it fixes the 2nd char for the same client inheriting the timer.  But I'd like my server to run this for the weekend to make sure no crashes crop up from taking out the IsInZone() workaround.

Thanks.

Akkadius:
> Would recommend just stopping all timers in Mob::~Mob()
> 
> `quest_manager.stopalltimers(this);`

This is the feedback I gave in Discord. This should be good

PR #3020, opened 2023-03-03
[Bug Fix] Cursor Coin Upon Death

cybernine186:
Upon death if a character has coin on cursor it will not transfer to the corpse, rather stay on cursor after death. This fix will remove all coin on cursor and save it to the corpse.


Kinglykrab (commented):
Is this Live-like? If so, looks good to me.

Aeadoin:
> Is this Live-like? If so, looks good to me.

Agreed, this i great as we don't want to be leaving things dangling on cursors if we can help it!
if we can get someone to confirm on Live this would be great... or even if they just transfer the coin to inventory that would be good too.

Kinglykrab:
Tested on Live, money stays on cursor on death. Would make more sense to make this rule-based than have it be the default functionality, as it's inconsistent with Live. Default `false` for the rule to be disabled by default.

Kinglykrab:
Added a rule toggle for this.

Aeadoin (approved):
lgtm

Kinglykrab (approved):

PR #3025, opened 2023-03-04
[Bug Fix] Ensure synchronization of pet taunt state with UI

catapultam-habeo:
This fixes a longstanding bug where the UI state of the pet taunt button has little to no connection with the command state of the pet.

### DIFF
--- a/zone/npc.cpp
+++ b/zone/npc.cpp
@@ -3964,10 +3964,10 @@ void NPC::HandleRoambox()
 	return;
 }
 
-void NPC::SetTaunting(bool tog) {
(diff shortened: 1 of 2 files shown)


zone/npc.cpp
```diff
@@ -3963,3 +3963,11 @@ void NPC::HandleRoambox()
 
 	return;
 }
+
+void NPC::SetTaunting(bool tog) {
+	this->taunting = tog;
```
Kinglykrab, line 3968: `this` is unnecessary as it’s implied.

Kinglykrab (approved):

PR #3027, opened 2023-03-05
[Bots] Place BOT_COMMAND_CHAR inside messages

trentdm:
Fixing an issue with bot portal/circle lists not sending the correct command when the player interacts. The bot command char, `^`, was not being properly formatted to a string.

image


Kinglykrab:
Fixed this by just putting the `^` character in the string itself, no reason to build it on to it when we don't do this in most other places.

Aeadoin (approved):

PR #3060, opened 2023-03-14
[Commands] Allow clients with access level to #undye to undye themselves

noudess:
Some servers (at least mine) used #undyeme at access level 0 for their players.  The removal of the #2776 removed that access.

Since the control to #undye lies in the database access level, this change allows anyone with the correct access level in the db to #undye themselves, but only GMs to #undye others.

This restores the functionality without the need for the extra command.

### DIFF
--- a/zone/gm_commands/undye.cpp
+++ b/zone/gm_commands/undye.cpp
@@ -3,16 +3,22 @@
 void command_undye(Client *c, const Seperator *sep)
 {
 	auto target = c;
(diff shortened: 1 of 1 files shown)


noudess:
> This shouldn’t be necessary. By default it only undyes yourself unless you’re tagged as a GM. https://github.com/EQEmu/Server/blob/master/zone/gm_commands/undye.cpp

Not sure how I missed this in initial testing.  I must have just not been paying enough attention.

PR #3061, opened 2023-03-16
[Illusions] Fix bug where spells like Ignite Bones left npc size inco…

noudess:
The default for non-PC races left many npc mobs at the wrong size when illusion wears off.

The other constants in this function are not needed as the previous statement sets race to 0, which already fixes size so that a simple GetSize() works.


fryguy503 (approved):
lgtm! Thank you!

Kinglykrab (approved):

PR #3123, opened 2023-03-18
[Fix] Fixes for corpses not properly saving some item instance data correctly.

KimLS:
This is preliminary work for adding guids to item instances and tracking a chain of custody for every item.  

The API corpses used before didn't allow non-pod types to be easily passed between the various components; hence why std::string custom_data was probably never implemented. Obviously chain of custody terminating when someone dies is not ideal so I went about fixing this first.

Changes:
- Corpses properly save custom_data and item ornamentation data now.
- Cleaned up the API a bit between the database and corpse loading; it also allocates less on the heap but it probably doesn't matter much on that.
- The database.CreateItem functions are more explicit in their usage, this adds a bit more boilerplate code in a few spots but it helps with seeing what's actually going on and where we could be losing data.
- Cleaned up some of the ItemInstance custom data api.


common/shareddb.h
```diff
@@ -119,25 +119,33 @@ class SharedDatabase : public Database {
...
-		bool attuned = 0
+		int16 charges,
+		uint32 aug1,
+		uint32 aug2,
+		uint32 aug3,
+		uint32 aug4,
+		uint32 aug5,
+		uint32 aug6,
+		bool attuned,
+		const std::string& custom_data,
```
Akkadius, line 130: Why should these be required for every single item instance creation? This is very verbose to call `CreateItem`

Similarly, why did all the previous defaults get removed? It unnecessarily makes our existing 57 calls to `CreateItem` needlessly verbose.

I would think we could do 

```cpp
inst->SetItemCustomData();
inst->SetOrnamentIcon();
```

For optionally filled fields after instance creation

This feels like maybe you did this on accident

world/worlddb.cpp
```diff
@@ -946,28 +946,7 @@ bool WorldDatabase::GetCharSelInventory(uint32 account_id, char *name, EQ::Inven
...
-					use_id = !use_id;
-					continue;
-				}
-
-				char v = data_str[i];
-				if (use_id)
-					idAsString.push_back(v);
-				else
-					value.push_back(v);
-			}
```
Akkadius, line 970: Nice cleanup of this redundancy in several places

Akkadius (changes requested):
Changes look great and straightforward to me outside of the `CreateItem` default args change. Seems like that was done on accident

common/shareddb.h
KimLS: I explained it in the post.

The database.CreateItem functions are more explicit in their usage, this adds a bit more boilerplate code in a few spots but it helps with seeing what's actually going on and where we could be losing data.

KimLS: I could undo it if it's an issue, but taking it out did lead me to discover a similar issue with guild banks which I will fix after this patch goes in either way.

Akkadius: I read the PR description but doesn't necessarily register with this kind of change in my mind as I read it - but you did call it out.

```
adds a bit more boilerplate code in a few spots
```

If was maybe 2,3 spots sure, maybe we can just shrug it off - but this is a very commonly used function in the code called 53 times in this branch. It also adds 12 optional args as now required args in most of those 53 places. Typing that out in normal code writing would feel cumbersome in any circumstance.

image

```
helps with seeing what's actually going on and where we could be losing data
```

How does this help though? Outside of seeing everything in a quick search as args in this one instance of auditing this problem in the code. 

I agree with explicitness 100%, but arguably the optional parameters are a code smell and I'm amazed we've loaded this constructor function as much as we have as it is. https://blog.devgenius.io/code-smell-19-optional-arguments-c0714855dbbb

At the cost of extreme code verbosity it doesn't help readability in anyway. I'm having a hard time seeing these fields as required in most cases to be part of `CreateItem` when they can be set on the instance after the item is created in the places in the code where we need to mutate it (and any other item related thing we need to set)

We can trace all the places we call `CreateItem` and see if we do anything in that part of the code to mutate item related data after the instance has been instantiated

Outside of being able to see all calls to CreateItem and see where we're not passing some data in I don't really see the benefits outweighing the verbosity, readability, maintainability smells. It's just as easy to click through those parts of the code and see if we're calling the setters/getters to load/fetch those fields

I really do think these changes are great though, I don't want my strong thoughts on this signature change to invalidate that. I'm excited to see how this helps the rest of the GUID and instance work

KimLS: I've definitely come to appreciate being explicit with your intentions especially with code that is prime for misuse like that.  That said, I'm not married to the idea and will go through and change it back tomorrow.

KimLS:
Updated to largely remove the changes to the Database.CreateItem api.

Kinglykrab:
Will look over this once I’m off of work. Glad to see GUIDs moving forward.

common/item_instance.cpp
```diff
@@ -699,7 +699,7 @@ void EQ::ItemInstance::PutAugment(SharedDatabase *db, uint8 slot, uint32 item_id
 	if (item_id == 0) { return; }
 	if (db == nullptr) { return; /* TODO: add log message for nullptr */ }
 
-	const ItemInstance* aug = db->CreateItem(item_id);
+	const ItemInstance* aug = db->CreateItem(item_id, 0, 0, 0, 0, 0, 0, 0, false, "", 0, 0, 0);
```
Kinglykrab, line 702: Is this still necessary?

zone/corpse.cpp
```diff
@@ -784,25 +809,43 @@ uint32 Corpse::CountItems() {
...
+void Corpse::AddItem(uint32 itemnum, 
+	uint16 charges, 
+	int16 slot, 
+	uint32 aug1, 
+	uint32 aug2, 
+	uint32 aug3, 
+	uint32 aug4, 
+	uint32 aug5, 
+	uint32 aug6, 
+	uint8 attuned,
```
Kinglykrab, line 821: Any reason attuned is still a `uint8` here?

zone/questmgr.cpp
```diff
@@ -3609,7 +3609,16 @@ void QuestManager::UpdateZoneHeader(std::string type, std::string value) {
 
 EQ::ItemInstance *QuestManager::CreateItem(uint32 item_id, int16 charges, uint32 augment_one, uint32 augment_two, uint32 augment_three, uint32 augment_four, uint32 augment_five, uint32 augment_six, bool attuned) const {
 	if (database.GetItem(item_id)) {
-		return database.CreateItem(item_id, charges, augment_one, augment_two, augment_three, augment_four, augment_five, augment_six, attuned);
+		return database.CreateItem(item_id,
```
Kinglykrab, line 3612: Any reason not to put `item_id,` on a newline, too?

zone/zonedb.h
```diff
@@ -285,6 +285,55 @@ struct ClientMercEntry {
...
+	int16 equip_slot;
+	uint16 charges;
+	uint16 lootslot;
+	uint32 aug_1;
+	uint32 aug_2;
+	uint32 aug_3;
+	uint32 aug_4;
+	uint32 aug_5;
+	uint32 aug_6;
+	int8 attuned;
```
Kinglykrab, line 300: Any reason attuned is an `int8` here and a `bool` or `uint8` elsewhere?

Kinglykrab (approved):
Looks good, small comments.

KimLS:
I agree with the attuned comments, I hadn't really noticed it since I was just sort of passing what was already there forward.  We should use a bool.  Let me change that before merge.

KimLS:
Added changes requested.

common/shareddb.h
Akkadius: Thank you <3

Akkadius (approved):

PR #3126, opened 2023-03-19
[Crash] Fixes Crash when Zoning with XTarget when Bots are in group.

Aeadoin:
This fixes a crash where Group members were not being cleaned out prior to Entity bot_list was cleared out, causing conflicts with dangling pointers in Group Members.

As noticed here http://spire.akkadius.com/dev/release/22.4.5-dev?id=1697

### DIFF
--- a/zone/groups.cpp
+++ b/zone/groups.cpp
@@ -544,7 +544,6 @@ bool Group::UpdatePlayer(Mob* update) {
 	return updateSuccess;
 }
 
-
 void Group::MemberZoned(Mob* removemob) {
 	uint32 i;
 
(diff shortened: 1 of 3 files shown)


Kinglykrab (approved):

zone/groups.cpp
```diff
@@ -554,11 +553,10 @@ void Group::MemberZoned(Mob* removemob) {
...
-	for (i = 0; i < MAX_GROUP_MEMBERS; i++) {
-		if (members[i] == removemob) {
-			members[i] = nullptr;
-			//should NOT clear the name, it is used for world communication.
-			break;
+	//should NOT clear the name, it is used for world communication.
+	for (auto & m : members) {
+		if (m && (m == removemob || m->IsBot() && m->CastToBot()->GetBotOwner() == removemob)) {
+			m = nullptr;
 		}
```
Kinglykrab, line 560: Add a `break;` here?

Aeadoin: Can't break cause we need to set any Bots that removemob is the owner of to nullptr here, or we crash later on.

PR #3127, opened 2023-03-19
[Feature] Add Item Extra Skill Damage Percent Modifier

Kinglykrab:
# Notes
- Allows `Character:ItemExtraDmgCap` to be disabled if set to `-1` or lower.
- Allows operators to set `Character:ItemExtraSkillDamageCalcAsPercent` to `true` to allow skill damage for Frenzy, Backstab, Bash, Slam, Kick, and all Monk attacks to scale with a percentage based on `extradmgamt` values from items and spells.

# Images
image
image


Aeadoin (approved):
good stuff

PR #3138, opened 2023-03-22
[Fix] Change SPA 193 Weapon Damage to allow values over 65,535

Aeadoin:
Values for Weapon Damage were capped at uint16 for SPA 193, changed to signed int32 as we use signed Int for Weapon Damage elsewhere, and don't want to allow overflowing of weapon dmg if done by mistake in a spell entry.


Kinglykrab (approved):
🔥

PR #3142, opened 2023-03-23
[Cleanup] Cleanup Strings::ToInt uses.

Aeadoin:
Corrected a number of instances where Strings::ToInt was being used incorrectly, where ToBool or ToUnsignedInt would be preferred. Will continue cleanup of these as I have time.

### DIFF
--- a/common/shareddb.cpp
+++ b/common/shareddb.cpp
@@ -964,7 +964,7 @@ void SharedDatabase::LoadItems(void *data, uint32 size, int32 items, uint32 max_
 
 		// Flags
 		item.ArtifactFlag = Strings::ToBool(row[ItemField::artifactflag]);
-		item.Attuneable = !disable_attuneable && Strings::ToBool(row[ItemField::attuneable]) != 0;
+		item.Attuneable = !disable_attuneable && Strings::ToBool(row[ItemField::attuneable]);
 		item.BenefitFlag = Strings::ToBool(row[ItemField::benefitflag]) != 0;
 		item.FVNoDrop = Strings::ToInt(row[ItemField::fvnodrop]) != 0;
 		item.Magic = Strings::ToBool(row[ItemField::magic]) != 0;
--- a/common/shareddb.cpp
+++ b/common/shareddb.cpp
@@ -965,9 +965,9 @@ void SharedDatabase::LoadItems(void *data, uint32 size, int32 items, uint32 max_
 		// Flags
 		item.ArtifactFlag = Strings::ToBool(row[ItemField::artifactflag]);
 		item.Attuneable = !disable_attuneable && Strings::ToBool(row[ItemField::attuneable]);
-		item.BenefitFlag = Strings::ToBool(row[ItemField::benefitflag]) != 0;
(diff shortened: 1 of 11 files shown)


common/shareddb.cpp
```diff
@@ -964,7 +964,7 @@ void SharedDatabase::LoadItems(void *data, uint32 size, int32 items, uint32 max_
 
 		// Flags
 		item.ArtifactFlag = Strings::ToBool(row[ItemField::artifactflag]);
-		item.Attuneable = !disable_attuneable && Strings::ToInt(row[ItemField::attuneable]) != 0;
+		item.Attuneable = !disable_attuneable && Strings::ToBool(row[ItemField::attuneable]) != 0;
```
Kinglykrab, line 967: I wonder if these can have wonky behavior since we’re comparing a Boolean to an integer. Should we just check if it’s true instead here and elsewhere I see it used here?

Akkadius: The != 0 should just be removed if you’re going to use tobool

common/shareddb.cpp
```diff
@@ -964,7 +964,7 @@ void SharedDatabase::LoadItems(void *data, uint32 size, int32 items, uint32 max_
 
 		// Flags
 		item.ArtifactFlag = Strings::ToBool(row[ItemField::artifactflag]);
-		item.Attuneable = !disable_attuneable && Strings::ToInt(row[ItemField::attuneable]) != 0;
+		item.Attuneable = !disable_attuneable && Strings::ToBool(row[ItemField::attuneable]);
 		item.BenefitFlag = Strings::ToBool(row[ItemField::benefitflag]) != 0;
```
Kinglykrab, line 968: Can remove `!= 0`.

common/shareddb.cpp
```diff
@@ -964,7 +964,7 @@ void SharedDatabase::LoadItems(void *data, uint32 size, int32 items, uint32 max_
...
 		item.FVNoDrop = Strings::ToInt(row[ItemField::fvnodrop]) != 0;
 		item.Magic = Strings::ToBool(row[ItemField::magic]) != 0;
```
Kinglykrab, line 970: Can remove `!= 0`.

common/shareddb.cpp
```diff
@@ -964,7 +964,7 @@ void SharedDatabase::LoadItems(void *data, uint32 size, int32 items, uint32 max_
 
 		// Flags
 		item.ArtifactFlag = Strings::ToBool(row[ItemField::artifactflag]);
-		item.Attuneable = !disable_attuneable && Strings::ToInt(row[ItemField::attuneable]) != 0;
+		item.Attuneable = !disable_attuneable && Strings::ToBool(row[ItemField::attuneable]);
 		item.BenefitFlag = Strings::ToBool(row[ItemField::benefitflag]) != 0;
 		item.FVNoDrop = Strings::ToInt(row[ItemField::fvnodrop]) != 0;
```
Kinglykrab, line 969: Maybe `Strings::ToBool` without `!= 0` here, too?

Kinglykrab (approved):

PR #3151, opened 2023-03-27
[Crash] Add Checks for out of bounds & dereferencing nullptrs

Aeadoin:
Added checks to prevent crashes as seen in the below instances:

http://spire.akkadius.com/dev/release/22.8.0?id=2826
http://spire.akkadius.com/dev/release/22.8.0?id=2830

Mob::SpellEffect needs refactored eventually, but these checks should help ensure we don't call an out of bounds memory address. 

buffslot can be either -2, -1 or a positive value (that should be valid) so anywhere we use buffslot to access buffs[buffslot] we need to ensure there's a check to ensure we're not out of bounds.

### DIFF
--- a/zone/bot.cpp
+++ b/zone/bot.cpp
@@ -2740,8 +2740,8 @@ bool Bot::IsValidTarget(Client* bot_owner, Client* leash_owner, float lo_distanc
 	}
 
 	bool valid_target_state = HOLDING || !tar->IsNPC() || tar->IsMezzed() || lo_distance > leash_distance || tar_distance > leash_distance;
-	bool valid_target = !GetAttackingFlag() && !CheckLosFN(tar) && !leash_owner->CheckLosFN(tar);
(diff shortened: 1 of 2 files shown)


zone/bot.cpp
```diff
@@ -2735,18 +2735,15 @@ void Bot::CalcMeleeDistances(const Mob* tar, const EQ::ItemInstance* const& p_it
...
-	)
-	{
+	if (!tar || !bot_owner || !leash_owner) {
+		return false;
+	}
+
+	bool valid_target_state = HOLDING || !tar->IsNPC() || tar->IsMezzed() || lo_distance > leash_distance || tar_distance > leash_distance;
+	bool valid_target = !GetAttackingFlag() && !CheckLosFN(tar) && !leash_owner->CheckLosFN(tar);
+	bool valid_bo_target= !GetAttackingFlag() && NOT_PULLING_BOT && !leash_owner->AutoAttackEnabled() && !tar->GetHateAmount(this) && !tar->GetHateAmount(leash_owner);
+
```
Kinglykrab, line 2745: Formatting is a bit weird here, maybe line up the `=`?

Kinglykrab (approved):

PR #3154, opened 2023-03-28
[Bug Fix] Check Rule "Bots Enabled" to prevent bot database calls on connect

Aeadoin:
Added check to verify Rule (Bots, Enabled) is true before we attempt to summon any Bots the owner previously had in their raid when they connect. Without this we were trying to run a database query against bot_data even if the server didn't have the bot schema.

https://cdn.discordapp.com/attachments/1090037315394613389/1090037447703937085/image.png

### DIFF
--- a/zone/bot_raid.cpp
+++ b/zone/bot_raid.cpp
@@ -300,7 +300,10 @@ void Bot::ProcessBotGroupAdd(Group* group, Raid* raid, Client* client, bool new_
 
 void Client::SpawnRaidBotsOnConnect(Raid* raid) {
 	std::list<BotsAvailableList> bots_list;
-	database.botdb.LoadBotsList(CharacterID(), bots_list);
+	if (!database.botdb.LoadBotsList(CharacterID(), bots_list) || bots_list.empty()) {
+		return;
+	}
+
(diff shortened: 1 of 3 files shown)


fryguy503 (approved):
lgtm - Thanks for addressing this

zone/bot_raid.cpp
```diff
@@ -296,4 +298,27 @@ void Bot::ProcessBotGroupAdd(Group* group, Raid* raid, Client* client, bool new_
...
 }
 
+void Client::SpawnRaidBotsOnConnect(Raid* raid) {
+	std::list<BotsAvailableList> bots_list;
+	database.botdb.LoadBotsList(CharacterID(), bots_list);
+	std::vector<RaidMember> r_members = raid->GetMembers();
+	for (const auto& m: r_members) {
+		if (strlen(m.member_name) != 0) {
+
+			for (const auto& b: bots_list) {
```
Kinglykrab, line 308: Should we check if bots_list is empty? Not that it’ll change anything. But we can just skip this sort if it’s empty.

Aeadoin: Might as well, added that in.

Kinglykrab (approved):

PR #3163, opened 2023-03-30
[Performance] Change to use Pass by reference where valid.

Aeadoin:
Clean uses where we were passing by value where the value isn't a nullptr.
Clean uses where we were using postfix vs prefix operators on a non-primitive type.
Clean uses of converting a std::string to a c_string where the expected type is a std::string.
Catches should be passed by reference.

### DIFF
--- a/ucs/chatchannel.h
+++ b/ucs/chatchannel.h
@@ -76,7 +76,7 @@ class ChatChannelList {
 
 public:
 	ChatChannel* CreateChannel(const std::string& name, const std::string& owner, const std::string& password, bool permanent, int minimum_status, bool save_to_database = false);
-	ChatChannel* FindChannel(const std::string& Name);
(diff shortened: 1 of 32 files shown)


ucs/chatchannel.h
```diff
@@ -76,7 +76,7 @@ class ChatChannelList {
 
 public:
 	ChatChannel* CreateChannel(const std::string& name, const std::string& owner, const std::string& password, bool permanent, int minimum_status, bool save_to_database = false);
-	ChatChannel* FindChannel(std::string name);
+	ChatChannel* FindChannel(const std::string& Name);
```
Kinglykrab, line 79: Why capitalize?

Aeadoin: just a typo, fixed.

Kinglykrab (approved):

Akkadius (approved):

PR #3177, opened 2023-04-02
[Rules] MeleePushChance applies to players

trentdm:
Currently, the `Combat:MeleePush` rule is being applied to both clients and NPCs, but the `Combat:MeleePushChance` rule is only being applied to NPCs.

I'm putting this change out there for consideration, since it seems like it could be a net benefit to have the MeleePush rules applied consistently to the same entity types, but I also don't know the entire history around this and could be missing important context.

Based on comments from the original implementation, https://github.com/EQEmu/Server/commit/06f4fd49efc1dde83c4f18b0f20a93b709798436, there appears to be some uncertainty around whether clients should always get pushed or not, but I haven't been able to find a follow-up.

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -4211,8 +4211,7 @@ void Mob::CommonDamage(Mob* attacker, int64 &damage, const uint16 spell_id, cons
 		else
 			a->special = 0;
 		a->hit_heading = attacker ? attacker->GetHeading() : 0.0f;
-		if (RuleB(Combat, MeleePush) && damage > 0 && !IsRooted() &&
-			(IsClient() || zone->random.Roll(RuleI(Combat, MeleePushChance)))) {
+		if (RuleB(Combat, MeleePush) && damage > 0 && !IsRooted() && zone->random.Roll(RuleI(Combat, MeleePushChance))) {
 			a->force = EQ::skills::GetSkillMeleePushForce(skill_used);
 			if (IsNPC()) {
 				if (attacker && attacker->IsNPC()) {


trentdm:
One other thought for consideration, if there is a desire to keep melee push chance rules of clients and NPCs distinct, I could add a separate rule just for players.

Something like `MeleePushChance` as the legacy rule that applies to NPCs and and `MeleePushChancePlayer`, that applies to the client.

noudess:
In reading the code, it looks to me like both clients and NPCs are pushed, clients always (if damage) and NPCs if the roll succeeds.  Am I reading that correctly?

Aeadoin:
> In reading the code, it looks to me like both clients and NPCs are pushed, clients always (if damage) and NPCs if the roll succeeds. Am I reading that correctly?

That is correct per my understanding. do we know if push is always applied on melee hit to clients on live? or is it meant to always be a chance? this would require research/confirmation 

@mackal are you aware of how live implements melee push against a client? is it guaranteed on each melee hit?

mackal:
> @mackal are you aware of how live implements melee push against a client? is it guaranteed on each melee hit?

They always send the data in the packet and the client reacts from what I can see. I think we might be sending that packet too many times to force fix things though, never got around to investigating it.

Akkadius:
At a glance these changes look fine, would like someone who knows push behavior better to validate

noudess:
From anecdotal evidence, it doesn't feel like push happens with melee anywhere near as much as live.  I remember constantly having to tell the guild to get on both sides of mobs on live to avoid pushing them off cliffs.

noudess:
@Akkadius I checked and do have the rule set.  My feeling on this is I don't see the kind of melee push that I did on live, but it's just gut feeling.

Kinglykrab:
Where are we at with this? Seems like a custom/optional rule, are we in favor or no?

nytmyr:
I like the idea of having it behind a rule until proven livelike. Preferably with rules for both clients/bots/mercs and NPCs to allow either or both.

trentdm:
Closing for now. Will make distinct rules for clients and bots.

PR #3178, opened 2023-04-02
[Cleanup] Fix always true/false/unnecessary statements in database conversion

Kinglykrab:
# Notes
- `runconvert == 1` was already checked on these and was unnecessary.
- `strlen(pp->bandoliers[i].Name) < 32` is unnecessary as `strlen(pp->bandoliers[i].Name)` will always be less than 32.
- `results` was redefined over and over unnecessarily.
- `pp->tribute_time_remaining < 0` can never happen since `pp->tribute_time_remaining` is unsigned.


Aeadoin (approved):

Kinglykrab:
Going to close this as we plan to remove this stuff anyway.

PR #3204, opened 2023-04-02
[Rules] Optional summoning when already in melee range

trentdm:
Add an optional rule to disable summon spells when the player is already within melee range of an NPC.

This kind of summoning is most apparent in North Temple of Veeshan, but might be in place elsewhere.


Aeadoin (approved):
looks good to me.

Kinglykrab (approved):

PR #3210, opened 2023-04-02
[Cleanup] Use default dtor instead of empty dtor for EQTime in eqtime.cpp/eqtime.h

Kinglykrab:
# Notes
- This is better than using an empty dtor.


Aeadoin (approved):
Agreed, being explicit is preferred in our code style

PR #3211, opened 2023-04-02
[Cleanup] Cleanup strlen() and cheap-to-copy references to use value instead in eqemu_logsys.cpp/eqemu_logsys.h

Kinglykrab:
# Notes
- Use `.length()` instead of calculating `strlen` each loop.
- Use pass by value instead of reference, as it's actually less performant on cheap-to-copy stuff like this.
- https://pvs-studio.com/en/docs/warnings/v835/

### DIFF
--- a/common/eqemu_logsys.cpp
+++ b/common/eqemu_logsys.cpp
@@ -292,7 +292,7 @@ void EQEmuLogSys::ProcessConsoleMessage(
 
 				bool is_upper = false;
 
-				for (int i = 0; i < strlen(e.c_str()); i++) {
+				for (int i = 0; i < e.length(); i++) {
 					if (isupper(e[i])) {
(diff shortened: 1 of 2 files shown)


Akkadius:
Have you actually measured this?

Kinglykrab:
> Have you actually measured this?

Which part? Just going based on comments by PVS-Studio that @Aeadoin sent me.

Akkadius:
Either. When we’re making presumed performance changes we should measure the before and after especially in such a hot path and run it through a benchmark at the very least using the test:test command

Akkadius:
Also what the compiler does to the final binary can be different between O0 and O1 and sometimes you find the change really made no difference or made things worse. It’s just good practice to show the data with the change if we can

PR #3215, opened 2023-04-02
[Cleanup] Remove extraneous loottable_id setting in WakeTheDead in aa.cpp

Kinglykrab:
# Notes
- We already set `made_npc->loottable_id` to `0` above.


Aeadoin (approved):
We should eventually look at switching anything like this to use in-class member initializers.

PR #3221, opened 2023-04-03
[Cleanup] Fix typo where itembonuses should have been used instead of spellbonuses

Kinglykrab:
# Notes
- `itembonuses` was the proper name here, not `spellbonuses`.


Aeadoin (approved):

Aeadoin:
This is a good catch, might explain some things I've noticed with Skill Attack Procs on items (not working that is lol)

PR #3237, opened 2023-04-03
[Cleanup] Remove always true/false conditions from bot.cpp

Kinglykrab:
# Notes
- Some of these conditions were always true or false based on previous conditions.

# Image
image

### DIFF
--- a/zone/bot.cpp
+++ b/zone/bot.cpp
@@ -6686,8 +6686,6 @@ void Bot::UpdateGroupCastingRoles(const Group* group, bool disband)
 
 	Mob* healer = nullptr;
 	Mob* slower = nullptr;
-	/*Mob* nuker = nullptr;
-	Mob* doter = nullptr;*/
 
 	for (auto iter : group->members) {
 		if (!iter)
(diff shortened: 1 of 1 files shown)


zone/bot.cpp
```diff
@@ -6686,8 +6686,8 @@ void Bot::UpdateGroupCastingRoles(const Group* group, bool disband)
 
 	Mob* healer = nullptr;
 	Mob* slower = nullptr;
-	Mob* nuker = nullptr;
-	Mob* doter = nullptr;
+	/*Mob* nuker = nullptr;
```
Aeadoin, line 6689: Why comment out, if not needed should just be deleted.

zone/bot.cpp
```diff
@@ -6803,10 +6803,10 @@ void Bot::UpdateGroupCastingRoles(const Group* group, bool disband)
 		healer->CastToBot()->SetGroupHealer();
 	if (slower && slower->IsBot())
 		slower->CastToBot()->SetGroupSlower();
-	if (nuker && nuker->IsBot())
+	/*if (nuker && nuker->IsBot())
```
Aeadoin, line 6806: Why comment out, if not needed should just be deleted.

Aeadoin (changes requested):

Aeadoin (approved):

PR #3264, opened 2023-04-03
[Cleanup] Fix ornamentation augment icons in inspect requests

Kinglykrab:
# Notes
- We were not setting `aug_item` to the ornamentation augment when we found one.


Aeadoin (approved):

Aeadoin:
lgtm assuming tested.

PR #3272, opened 2023-04-08
[Hotfix] Resolve loading of inventory

Aeadoin:
This resolves regression introduced in #3190, we need to return true from GetSharedBank() or the player inventory will fail to load.


Kinglykrab:
Could we just do if no results return true? So it doesn’t process the extra logic at all?

Aeadoin:
> 

Code wise, it should just hit the For Loop, find nothing to iterate and then jump straight to the return true;

Kinglykrab (approved):

PR #3281, opened 2023-04-09
[Feature] Make ornamentations work with any augment type

Kinglykrab:
# Notes
- On Live there are augments that are not type 20/21 and are ornamentations.
- We also only allow a singular augment type to be ornamentation augment types, this will allow you to use any augment type as an ornamentation if it has a proper Hero's Forge model or a non-IT63/non-IT64 idfile.
- Remove `Character:OrnamentationAugmentType` rule as it's no longer necessary.
- This will not break any servers using the previous rule functionality as it will check the augment itself for things that make it a valid ornamentation.

### DIFF
--- a/common/item_instance.cpp
+++ b/common/item_instance.cpp
@@ -659,11 +659,11 @@ bool EQ::ItemInstance::UpdateOrnamentationInfo()
 
 			return true;
 		}
-	} else {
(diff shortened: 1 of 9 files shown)


Aeadoin (approved):
lgtm assuming has been tested for regression.

PR #3288, opened 2023-04-10
[Rules] ResurrectionEffectBlock to prevent/allow/move buffs.

nytmyr:
Revamp of #2990 

This changes the rule ResurrectionEffectsBlock (Bool) to ResurrectionEffectBlock (Int)  

Default = 2

Setting to 0 = Functions as it did before any blocking changes, Focus of Spirit and Strength buffs can overwrite Resurrection Effects.

Setting to 1 = Blocks all buffs that could overwrite Resurrection Effects.

Setting to 2 = Allows all buffs that would overwrite Resurrection Effects to land, however they will be moved to a new buff slot if one is available to allow both the beneficial buff to land and detrimental effects of Resurrection Effects to stay in effect until the duration is expired.


nytmyr:
Would like to make a few more adjustments to this before it gets merged. There's some underlying blocks that happen with res sickness still. Form of the Great Bear for one.

nytmyr:
This should be good to go. Any overwrites/blocks related to res sickness should follow the rule setting.

zone/spells.cpp
```diff
@@ -3162,6 +3168,13 @@ int Mob::CheckStackConflict(uint16 spellid1, int caster_level1, uint16 spellid2,
 		if(sp2_value < 0)
 			sp2_value = 0 - sp2_value;
 
+		if (IsResurrectionEffects(spellid1)) {
+			int8 ResEffectCheck = DoResSicknessChecks(spellid1, spellid2);
```
Kinglykrab, line 3172: `res_effect_check`, we prefer snake case for variable names.

zone/spells.cpp
```diff
@@ -3037,6 +3037,12 @@ int Mob::CheckStackConflict(uint16 spellid1, int caster_level1, uint16 spellid2,
 
 					if(sp1_value < overwrite_below_value)
 					{
+						if (IsResurrectionEffects(spellid1)) {
+							int8 ResEffectCheck = DoResSicknessChecks(spellid1, spellid2);
```
Kinglykrab, line 3041: `res_effect_check`, we prefer snake case for variable names.

common/spdat.h
```diff
@@ -1505,6 +1505,7 @@ bool IsDisciplineBuff(uint16 spell_id);
 bool IsDiscipline(uint16 spell_id);
 bool IsCombatSkill(uint16 spell_id);
 bool IsResurrectionEffects(uint16 spell_id);
+int8 DoResSicknessChecks(uint16 spell_id1, uint16 spell_id2);
```
Kinglykrab, line 1508: This is a bit oddly named. Maybe `GetResurrectionSicknessCheck`? Verbosity doesn’t hurt when naming methods.

common/spdat.cpp
```diff
@@ -776,6 +776,21 @@ bool IsResurrectionEffects(uint16 spell_id)
 	return false;
 }
 
+int8 DoResSicknessChecks(uint16 spell_id1, uint16 spell_id2)
+{
+	if (RuleI(Spells, ResurrectionEffectBlock) == 1) {
```
Kinglykrab, line 781: Maybe turn these magic numbers into constants? Plenty of examples around.

common/ruletypes.h
```diff
@@ -445,7 +445,7 @@ RULE_BOOL(Spells, IllusionsAlwaysPersist, false, "Allows Illusions to persist be
 RULE_BOOL(Spells, UseItemCastMessage, false, "Enable to use the \"item begins to glow\" messages when casting from an item.")
 RULE_BOOL(Spells, TargetsTargetRequiresCombatRange, true, "Disable to remove combat range requirement from Target's Target Spell Target Type")
 RULE_BOOL(Spells, NPCBuffLevelRestrictions, false, "Impose BuffLevelRestrictions on NPCs if true")
-RULE_BOOL(Spells, ResurrectionEffectsBlock, true, "If enabled, resurrection effects cannot be overwritten.")
+RULE_INT(Spells, ResurrectionEffectBlock, 2, "0 = allow overwrites/rule disabled. If set to 1 = Block all buffs that would overwrite Resurrection Effects. If set to 2 = Will not overwrite Resurrection Effects, instead moves new buff to an empty slot if available. ")
```
Kinglykrab, line 448: Changing the type of rule like this will not fix peoples’ database values. May want to put a required SQL in that sets their database values based on if they had this rule set to `true`/`false` prior to this change.

common/spdat.cpp
```diff
@@ -776,6 +776,21 @@ bool IsResurrectionEffects(uint16 spell_id)
...
+		LogSpells("ResurrectionEffectBlock triggered -- [{}] is blocked by [{}]", spells[spell_id2].name, spells[spell_id1].name);
+		return -1;	// can't stack
+	}
+	else if (RuleI(Spells, ResurrectionEffectBlock) == 2) {
+		LogSpells("ResurrectionEffectBlock triggered -- [{}] is blocked by [{}], moving to empty slot if available", spells[spell_id2].name, spells[spell_id1].name);
+		return 2;	// move to empty slot if available
+	}
```
Kinglykrab, line 788: Don’t need an else, if it doesn’t hit the previous checks, can just return 0.

common/ruletypes.h
Kinglykrab: Issue is it will set the new rule to this value and it won't be based on the old rule, as it will auto delete the old one, meaning this could override the settings an operator already had in place.

nytmyr: Do you mind verifying if this is the correct way you'd like this done?

common/spdat.cpp
nytmyr: Curious as to how you'd like these done. I get where some things that are recycled are done as constants, just not sure how you'd like it done here.

nytmyr:
Bump.

Kinglykrab (approved):

utils/sql/db_update_manifest.txt
```diff
@@ -481,6 +481,7 @@
 9225|2023_01_21_bots_raid_members.sql|SHOW COLUMNS FROM `raid_members` LIKE 'bot_id'|empty|
 9226|2023_03_17_corpse_fields.sql|SHOW COLUMNS FROM `character_corpse_items` LIKE 'custom_data'|empty|
 9227|2023_03_24_npc_scale_global_base_verify.sql|SHOW COLUMNS FROM `npc_scale_global_base` LIKE 'heroic_strikethrough'|not_empty|
+9228|2023_04_15_update_resurrection_effects_block.sql|SELECT `rule_name` FROM `rule_values` WHERE `rule_name` LIKE 'Spells:ResurrectionEffectsBlock'|empty|
```
Akkadius, line 484: This migration is unnecessary because we inject rules automatically. My suggestion is to slightly rename the rule so the old one will be automatically deleted and the desired default and notes will be automatically injected. Remove this migration, version.h entry and the .sql

common/spdat.cpp
```diff
@@ -776,6 +776,19 @@ bool IsResurrectionEffects(uint16 spell_id)
...
 
+int8 GetResurrectionSicknessCheck(uint16 spell_id1, uint16 spell_id2)
+{
+	if (RuleI(Spells, ResurrectionEffectsBlock) == RES_EFFECTS_BLOCK) {
+		LogSpells("ResurrectionEffectsBlock triggered -- [{}] is blocked by [{}]", spells[spell_id2].name, spells[spell_id1].name);
+		return -1;	// can't stack
+	}
+	else if (RuleI(Spells, ResurrectionEffectsBlock) == RES_EFFECTS_BLOCK_WITH_BUFFS) {
+		LogSpells("ResurrectionEffectsBlock triggered -- [{}] is blocked by [{}], moving to empty slot if available", spells[spell_id2].name, spells[spell_id1].name);
+		return 2;	// move to empty slot if available
```
Akkadius, line 787: Why aren't we using the constants in the return here? We try to make a concerted effort to eliminate magic numbers for readability when we can

Akkadius (changes requested):
Minor comments, please address and looks good

utils/sql/db_update_manifest.txt
nytmyr: I originally had exactly that, rule renamed to "ResurrectionEffectBlock" but was informed to change it to a conversion to honor operators current setting.

Do we want to for sure rename to set the new default or leave the check so operators maintain the same setting currently and new builds default?

The query would have the same effect as the original PR just with catching edge cases that still allowed Resurrection Sickness to be overwritten. By changing and getting the new default it takes on the new functionality of this PR which is allowing both to be currently active and moving the buff to a new slot instead of being blocked completely by Resurrection Sickness.

I doubt anyone really has this original rule disabled to allow Resurrection Sickness overwrites but you never know.

noudess: I agree.

Akkadius (approved):

Kinglykrab (approved):

PR #3309, opened 2023-04-23
[Quest API] Add GetPet() to Perl

Kinglykrab:
# Perl
- Add `$mob->GetPet()`.

# Notes
- This exists in Lua, but not Perl.

### DIFF
--- a/zone/pets.cpp
+++ b/zone/pets.cpp
@@ -503,21 +503,22 @@ bool ZoneDatabase::GetPoweredPetEntry(const char *pet_type, int16 petpower, PetR
 }
 
 Mob* Mob::GetPet() {
-	if(GetPetID() == 0)
-		return(nullptr);
-
-	Mob* tmp = entity_list.GetMob(GetPetID());
-	if(tmp == nullptr) {
-		SetPetID(0);
-		return(nullptr);
+	if (!GetPetID()) {
+		return nullptr;
 	}
 
-	if(tmp->GetOwnerID() != GetID()) {
+	const auto m = entity_list.GetMob(GetPetID());
+	if (!m) {
 		SetPetID(0);
-		return(nullptr);
+		return nullptr;
 	}
 
-	return(tmp);
+	if (m->GetOwnerID() != GetID()) {
+		SetPetID(0);
+		return nullptr;
+	}
+
+	return m;
 }
 
 bool Mob::HasPet() const {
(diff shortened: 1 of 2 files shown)


MortimerGreenwald:
that return pet id? or just a true/false?

Kinglykrab:
> > that return pet id? or just a true/false?
> 
> A Mob*

This. @MortimerGreenwald

zone/pets.cpp
```diff
@@ -503,21 +503,22 @@ bool ZoneDatabase::GetPoweredPetEntry(const char *pet_type, int16 petpower, PetR
 }
 
 Mob* Mob::GetPet() {
-	if(GetPetID() == 0)
-		return(nullptr);
+	if (!GetPetID()) {
+		return nullptr;
+	}
 
-	Mob* tmp = entity_list.GetMob(GetPetID());
```
Akkadius, line 509: Curious why the change here? We're turning a pointer

Kinglykrab: `const auto` still shows as a pointer, would the `const` cause issues possibly?

Aeadoin: > Curious why the change here? We're returning a pointer

Why is this a concern? we return it as a Mob* from the actual method itself. this just is a promise that Mob* won't be modified within this method after it's assigned.

Kinglykrab:
Bump.

zone/pets.cpp
Akkadius: Sorry for not being clear enough.

At a glance it looked like the pointer was removed from the object; that is one thing I don't quite like about `auto` is that it is less declarative for readability in some cases. Since the entity return type is a pointer it should also return a pointer (At least we would assume)

I'm not concerned about `const` - we're not expecting to modify the object here. I think using `const` where we want to protect mutation is preferred - although in some cases can have nasty side affects if the existing code was written poorly. 

The main thing was no longer fetching by pointer and accidentally "copying" the object thus leading to possible unintended leaks

Akkadius (approved):

zone/pets.cpp
Akkadius: We confirmed 

image

```cpp
    const auto m = entity_list.GetMob(GetPetID());
    if(m == nullptr) {
        SetPetID(0);
        return(nullptr);
    }

    LogInfo("pointer address is [{}] (m)", fmt::ptr(m));

    Mob *m2 = entity_list.GetMob(GetPetID());
    LogInfo("pointer address is [{}] (m2)", fmt::ptr(m2));
```

PR #3330, opened 2023-05-03
[Feature] Intoxication setter/getter for source, getter for Perl/Lua

JasXSL:
Added SetIntoxication and GetIntoxication for client to match how things like hunger/thirst have setters and getters.
Added GetIntoxication functions for perl and Lua. 
No set function function, since it'd only take effect when zoning and might be confusing.

### DIFF
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -908,8 +908,8 @@ void Client::ChannelMessageReceived(uint8 chan_num, uint8 language, uint8 lang_s
 	}
 
 	// Garble the message based on drunkness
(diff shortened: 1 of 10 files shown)


zone/client.cpp
```diff
@@ -8340,6 +8340,11 @@ void Client::SetThirst(int32 in_thirst)
 	safe_delete(outapp);
 }
 
+void Client::SetIntoxication(int32 in_intoxication)
+{
+	m_pp.intoxication = std::min(200,std::max(0,in_intoxication));
```
Kinglykrab, line 8345: `EQ::Clamp(in_intoxication, 0, 200)` would work here. Just have to include `common/data_verification.h`.

JasXSL: Added

Kinglykrab (approved):

PR #3333, opened 2023-05-07
[Bug Fix] ReloadQuests() on Zone::Init() to avoid cached global quests/plugins

Kinglykrab:
# Notes
- Before this, quests/plugins would be cached in an old state, so you'd have to either enter the zone and `#reload quest` or `#reload world` to get them to update.

### DIFF
--- a/zone/zone.cpp
+++ b/zone/zone.cpp
@@ -1213,6 +1213,8 @@ bool Zone::Init(bool is_static) {
 
 	npc_scale_manager->LoadScaleData();
 
+	parse->ReloadQuests();
+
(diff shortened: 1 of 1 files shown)


Aeadoin (approved):

Akkadius:
Had to revert this one because encounters were not being loaded at all on Lazarus.

https://github.com/EQEmu/Server/commit/434f270f68bcd3c7b61c18d474e436e0b5ca1891

I understand the need for doing this change but we need to test a bit more to make sure there aren't side affects (such as encounters not loading)

PR #3340, opened 2023-05-10
[Performance] Character tribute is now bulk saved

Kinglykrab:
This pull request combines individual `character_tribute` queries during `Save()` into one.

This pull request also adds a primary key of `id` to `character_tribute` and renames the pre-existing `id` column to `character_id`, this allows us to use repositories for this table.

image

### DIFF
--- a/zone/zonedb.cpp
+++ b/zone/zonedb.cpp
@@ -970,7 +970,7 @@ void ZoneDatabase::LoadCharacterTribute(Client* c){
 	auto i = 0;
 
 	for (const auto& e : l) {
-		if (e.tribute != TRIBUTE_NONE){
+		if (e.tribute != TRIBUTE_NONE) {
 			c->GetPP().tributes[i].tier    = e.tier;
 			c->GetPP().tributes[i].tribute = e.tribute;
 			i++;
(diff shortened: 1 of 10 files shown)


zone/zonedb.cpp
```diff
@@ -1058,18 +1059,35 @@ bool ZoneDatabase::SaveCharacterDisc(uint32 character_id, uint32 slot_id, uint32
...
+			tribute.character_id = c->CharacterID();
+			tribute.tier         = t.tier;
+			tribute.tribute      = t.tribute;
+
+			tributes.emplace_back(tribute);
+		}
+	}
+
+	if (tribute_count > 0) {
+		CharacterTributeRepository::DeleteWhere(database, fmt::format("character_id = {}", c->CharacterID()));
```
Akkadius, line 1087: We should probably pull this before `if (tribute_count > 0) {` because we could end up leaving stale entries behind and re-loading them

Kinglykrab: I’ll move it this afternoon. Was following the recent bulk update stuff you did as an example.

Akkadius: Understood, situation-ally my code was wrong and you saw it was fixed in the recent PR

Akkadius (approved):

PR #3341, opened 2023-05-12
[Bug Fix] Fix Heroic INT/WIS Bonuses

RekkasGit:
So users on Retribution server were complaining about their manastones actually reducing mana from max mana, and basically capping  quite a bit under their real max mana. I traced it down to the patch done about 2 months ago on heroic stat changes. I made these modifications to correct it and seems to work fine. This seems to have only impacted Pal/SK/Ranger/Beastlord. 
Made new Is Checks to make sure I didn't impact logic that is located in other places.


Kinglykrab:
@Aeadoin

PR #3347, opened 2023-05-14
[Cleanup] Remove GetMaxRank() from aa_ability.cpp/aa_ability.h

Kinglykrab:
# Notes
- This is unused.


Aeadoin:
I could see this being useful, maybe it should be kept?

Akkadius (approved):

PR #3351, opened 2023-05-14
[Bug Fix] Mob scaling issue with min dmg set to zero while max dmg is not

RekkasGit:
A user on Ret submitted a bug where they were one shot by a low level mob. I was able to reproduce this and started to diagnose the issue. It seemed to be where a Min Dmg value of 0 and a max dmg value of 3 was set in the database for this NPC. 
image

The min dmg scaling kicks in, but you get into a situation where MIN DMG > MAX DMG (12 vs 3 in this case)
It leads to situations like thus. 
image

Looking at it deeper it seems to be in the calculation area for rounding. 

image

I worked with Kingly and setup some sanity code to prevent wild situations of being hit for millions of damage, but needed to go one step further. 

I also changed it so that if you set Min Dmg to 0, it means max damage is also auto calculated... this had the mob working as I would expect a lvl 28 mob to be. 

image

### DIFF
--- a/zone/npc.cpp
+++ b/zone/npc.cpp
@@ -2613,6 +2613,11 @@ void NPC::ModifyNPCStat(const std::string& stat, const std::string& value)
 	else if (stat_lower == "min_hit") {
 		min_dmg     = Strings::ToInt(value);
 		// TODO: fix DB
+		if (min_dmg > max_dmg) {
+			const auto temporary_damage = max_dmg;
+			max_dmg = min_dmg;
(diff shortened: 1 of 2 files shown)


Aeadoin (approved):

Kinglykrab (approved):

noudess:
This has side effects.  If both min and max are set to zero, the scale manager never properly sets min_damage, as when it calls to set min_dmg it actually gets stored in max_dmg and thrown away.

PR #3363, opened 2023-05-17
[Quest API] Add Memorize and Scribe Spell Events to Perl/Lua

Kinglykrab:
# Perl
- Add `EVENT_MEMORIZE_SPELL`.
- Add `EVENT_UNMEMORIZE_SPELL`.
- Add `EVENT_SCRIBE_SPELL`.
- Add `EVENT_UNSCRIBE_SPELL`.

# Lua
- Add `event_memorize_spell`.
- Add `event_unmemorize_spell`.
- Add `event_scribe_spell`.
- Add `event_unscribe_spell`.

# Notes
- Allows operators to perform events on memorization, unmemorization, scribe, or unscribe.
- Cleaned up target description messages for `#unscribespell`.


Akkadius (approved):

zone/client.cpp
```diff
@@ -2783,18 +2783,64 @@ void Client::GMKill() {
...
 }
 
-void Client::MemorizeSpell(uint32 slot,uint32 spellid,uint32 scribing, uint32 reduction){
-	if (slot < 0 || slot >= EQ::spells::DynamicLookup(ClientVersion(), GetGM())->SpellbookSize)
+void Client::MemorizeSpell(uint32 slot, uint32 spell_id, uint32 scribing, uint32 reduction){
+	if (
+		!EQ::ValueWithin(
+			slot,
+			0,
+			(EQ::spells::DynamicLookup(ClientVersion(), GetGM())->SpellbookSize - 1)
```
Akkadius, line 2791: Why is this newly `- 1` versus how it was before?

ValueWithin does slot >= SpellbookSize

```
bool ValueWithin(const T &value, const T &lower, const T &upper)
{
    return value >= lower && value <= upper;
}
```

zone/client.cpp
```diff
@@ -2783,18 +2783,64 @@ void Client::GMKill() {
...
+	) {
 		return;
-	if ((spellid < 3 || spellid > EQ::spells::DynamicLookup(ClientVersion(), GetGM())->SpellIdMax) && spellid != 0xFFFFFFFF)
+	}
+
+	if (
+		!EQ::ValueWithin(
+			spell_id,
+			3,
+			(EQ::spells::DynamicLookup(ClientVersion(), GetGM())->SpellIdMax - 1)
```
Akkadius, line 2801: Same here

zone/client.cpp
Kinglykrab: We were returning if >= this value so we check if it’s not between 0 and 1 less than this value then return if it isn’t.

zone/client.cpp
Kinglykrab: This one I’ll fix to not be `-1`, the other checked `>=`.

Akkadius (approved):

PR #3366, opened 2023-05-19
[Bug Fix] NPC Armor Upgrade to a slot not handled correctly

noudess:
If you trade a pet a new helm(replacing one currently in use), the helm does not replace as the item counted in Bonuses until you zone.  This was due to two items in itemlist having the same equip-slot.

### DIFF
--- a/zone/loottables.cpp
+++ b/zone/loottables.cpp
@@ -396,12 +396,14 @@ void NPC::AddLootDrop(
 							}
 							else {
 								// Unequip old item
-								ServerLootItem_Struct *olditem=GetItem(i);
+								auto* olditem = GetItem(i);
+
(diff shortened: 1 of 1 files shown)


zone/loottables.cpp
```diff
@@ -395,6 +395,10 @@ void NPC::AddLootDrop(
 								foundslot = i;
 							}
 							else {
+								// Unequip old item
+								ServerLootItem_Struct *olditem=GetItem(i);
```
Kinglykrab, line 399: Slap a couple spaces in here for formatting purposes.

Kinglykrab (approved):
Looks good. One small comment. 🔥

PR #3373, opened 2023-05-22
[Bug Fix] Fix issue with Group Pointers

Aeadoin:
When a member of a group would go linkdead/camp/leavegroup, depending on their index inside the Group Members Struct it could cause issues when we called Groups::LearnMembers to create a duplicate Mob*.

### DIFF
--- a/zone/groups.cpp
+++ b/zone/groups.cpp
@@ -1156,24 +1156,14 @@ bool Group::LearnMembers() {
 		);
 	}
 
-	int memberIndex = 0;
-	for (const auto& member : rows) {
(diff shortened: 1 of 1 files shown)


noudess:
Your fix loops to MAX_GROUP_MEMBERS, but does your call to rows[5].name.empty() work if GetWhere only returns 4 entries?  Is so, I don't understand exactly how GetWhere works then.

This is what I did after seeing your fix and I think understanding your intent to clear all rows after the valid ones found by GetWhere():

[groupbug.patch](https://github.com/EQEmu/Server/files/11544360/groupbug.patch)

PR #3374, opened 2023-05-23
[Bug Fix] Fix issue with Group Pointers/Member roles

Aeadoin:
We weren't properly nulling Group membership prior to loading Group membership from the database.
This could cause issues where duplicate entries could appear causing a client to be in the group multiple times. 

We also weren't correctly handling Member Roles, causing these to possibly break when the group changed in some fashion.


noudess (commented):
Setting MemberRoles[i] = 0 in LearnMembers..  Where do these get filled in again?

noudess (approved):
Looks good to me.  I wish there wasn't so much dependence in multiple functions on their sequencing being correct,. but that's a bigger issue.

Many thanks for finding this.  Lets hope it puts a big dent in grouping issues.

Akkadius (approved):

PR #3376, opened 2023-05-26
[Illusions] RandomizeFeatures and SetGender were killing db texture

noudess:
If you use RandomizeFeatures() or SetGender() from a quest, they were indirectly setting texture to FF for all PC races.

I say indirectly because for some reason SendIlusion does not use texture if the mob is a PC race.  

I don't think anyone using RandomizeFeatures() or SetGender() wants the side effect of trashing texture, so I modified both calls to pass in current mob texture, preserving it.

### DIFF
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -2576,7 +2576,7 @@ bool Mob::RandomizeFeatures(bool send_illusion, bool set_variables)
 {
 	if (IsPlayerRace(GetRace())) {
 		uint8 current_gender = GetGender();
-		uint8 new_texture = 0xFF;
+		uint8 new_texture = GetTexture();
 		uint8 new_helm_texture = 0xFF;
 		uint8 new_hair_color = 0xFF;
 		uint8 new_beard_color = 0xFF;
--- a/zone/perl_mob.cpp
+++ b/zone/perl_mob.cpp
@@ -1798,7 +1798,7 @@ void Perl_Mob_SetRace(Mob* self, int32 race) // @categories Stats and Attributes
 
 void Perl_Mob_SetGender(Mob* self, int32 gender) // @categories Stats and Attributes
 {
-	self->SendIllusionPacket(self->GetRace(), gender);
+	self->SendIllusionPacket(self->GetRace(), gender, self->GetTexture());
 }
 
 // todo: SendIllusion should be sent in a hash like lua


Kinglykrab:
Does SetRace() do this, too?

noudess:
> Does SetRace() do this, too?

I think so, yes.  What it boils down to is that they all call SendIllusion() and if you don't pass texture to SendIllusion, it hits this code:

    if (in_texture == 0xFF && !IsPlayerRace(race)) {
        new_texture = GetTexture();
    }

That only preserves texture for non-player races.  I don't know why - but it seemed long-standing so I didnt want to mess with it.  I feel pretty sure the two functions I changed don't want texture messed with.  

I'll leave this up for discussion.

Kinglykrab:
> > Does SetRace() do this, too?
> 
> I think so, yes. What it boils down to is that they all call SendIllusion() and if you don't pass texture to SendIllusion, it hits this code:
> 
> ```
> if (in_texture == 0xFF && !IsPlayerRace(race)) {
>     new_texture = GetTexture();
> }
> ```
> 
> That only preserves texture for non-player races. I don't know why - but it seemed long-standing so I didnt want to mess with it. I feel pretty sure the two functions I changed don't want texture messed with.
> 
> I'll leave this up for discussion.
> 
> I randomize gender and face etc all over the place in quests on spawn - and all my PC races lose the texture specified in the db.

I’d be fine with just removing the `!IsPlayerRace()` condition there. It wouldn’t really break anything, as I’d assume anyone using this sets the texture manually (if race is a player race) since it would eat texture anyway.

noudess:
> > > Does SetRace() do this, too?
> > 
> > 
> > I think so, yes. What it boils down to is that they all call SendIllusion() and if you don't pass texture to SendIllusion, it hits this code:
> > ```
> > if (in_texture == 0xFF && !IsPlayerRace(race)) {
> >     new_texture = GetTexture();
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That only preserves texture for non-player races. I don't know why - but it seemed long-standing so I didnt want to mess with it. I feel pretty sure the two functions I changed don't want texture messed with.
> > I'll leave this up for discussion.
> > I randomize gender and face etc all over the place in quests on spawn - and all my PC races lose the texture specified in the db.
> 
> I’d be fine with just removing the `!IsPlayerRace()` condition there. It wouldn’t really break anything, as I’d assume anyone using this sets the texture manually (if race is a player race) since it would eat texture anyway.

Me too.  But I sure wish I knew WHY it's in there.  It's in there for helmtexture as well.

I hate taking something out that I don't understand the reasoning behind, because it's usually a fix someone put there for a reason.

If everyone's behind it, Id take this !IsPlayerRace(race)) out in a couple of spots - but I was hoping someone would know why it's there.  Maybe it was used to "mask" the fact that an npc with a player race was wearing some piece of gear?

Kinglykrab:
> > > > Does SetRace() do this, too?
> > > 
> > > 
> > > I think so, yes. What it boils down to is that they all call SendIllusion() and if you don't pass texture to SendIllusion, it hits this code:
> > > ```
> > > if (in_texture == 0xFF && !IsPlayerRace(race)) {
> > >     new_texture = GetTexture();
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > That only preserves texture for non-player races. I don't know why - but it seemed long-standing so I didnt want to mess with it. I feel pretty sure the two functions I changed don't want texture messed with.
> > > I'll leave this up for discussion.
> > > I randomize gender and face etc all over the place in quests on spawn - and all my PC races lose the texture specified in the db.
> > 
> > 
> > I’d be fine with just removing the `!IsPlayerRace()` condition there. It wouldn’t really break anything, as I’d assume anyone using this sets the texture manually (if race is a player race) since it would eat texture anyway.
> 
> Me too. But I sure wish I knew WHY it's in there. It's in there for helmtexture as well.
> 
> I hate taking something out that I don't understand the reasoning behind, because it's usually a fix someone put there for a reason.
> 
> If everyone's behind it, Id take this !IsPlayerRace(race)) out in a couple of spots - but I was hoping someone would know why it's there. Maybe it was used to "mask" the fact that an npc with a player race was wearing some piece of gear?

Not sure, since most NPC races don't really have different faces/hair/beards and just have different textures and helmtexture combinations that make them look differently.

hgtw:
I haven't looked heavily into client side mob appearances and how texture/armor stuff is implemented on emu so this may be wrong.

Sending -1 for player races is supposed to tell the client to leave the actor's armor unchanged (keep using its visible armor slot data). If it's not -1 then it overrides the visible slots with the specified texture. For non-player races -1 sets the texture to 0 on the client I think, so that check was probably added with the intent to leave non-player races unchanged when the function was called without a specified texture (assuming server texture is in sync). 

Sending -1 to leave a player race unchanged does seem like a bug if that mob is supposed to use a texture instead of armor data, so I think removing the `!IsPlayerRace()` check would be ok as long as `GetTexture()` returns -1 when armor data exists.

There are also some unimplemented packet flags, one of which prevents updating armor for player races that could be investigated to leave them unchanged on illusion.

noudess:
> Sending -1 to leave a player race unchanged does seem like a bug if that mob is supposed to use a texture instead of armor data, so I think removing the `!IsPlayerRace()` check would be ok as long as `GetTexture()` returns -1 when armor data exists.

I can do that.

Kinglykrab (approved):

hgtw:
> I think maybe it's best to say we just should not use SetGender(), SetRace() or RandomizeFeatures() in EVENT_SPAWN.

Using the `ChangeGender()` variant functions and `RandomizeFeatures(send_illusion = false)` overload are probably best practice for changing appearances from EVENT_SPAWN even with this fix since they update the internal values without sending the excess illusion packet.

> It would be nice to check for this. Is there a place we can flag a runtime error if we try to do these in EVENT_SPAWN()?

It's possible but you would need to walk the Lua/perl stacks to find the calling event function name and add warnings in problematic apis, not sure if it's worth that effort.

I think this change is ok, I don't have a great handle on appearance code but I don't think anything relies on illusion setting the texture back to -1.

hgtw (approved):

PR #3381, opened 2023-06-06
[Logging] Fixed statements that logged incorrect data

noudess:
Due to the data types in the client packet, door_id was not being logged correctly.  Replaced the value from the packet with the simple m_door_id.

The old code displayed a seemingly random value for the same door, I assume because of casting.


Kinglykrab (approved):
🥇

PR #3383, opened 2023-06-09
[Targeting] /tar <bad target> should not untarget existing target

noudess:
On live if you have something targetted, and you type /tar Innothule, but Innothule is not around, you get an error message and your current target is unmodifed.

Prior to this patch, if you do this on eqemu, it "looks" like your current target is unchanged, but you no longer get updates.

This change fixes that.  But I have a couple of questions:

1. The code differentiates between AI controlled clients and non-AI controlled clients.  Is not the /tar command unable to be send if you are AI controlled, meaning I could also delete those lines?  If not, perhaps the code I moved down inside non-ai controlled needs to be down in AI controlled as well?  My take is that those lines can be deleted.
2. The ONLY time all of the code after the // HoTT comment executes if its a NON-AI-CONTROLLED successful target.  So I don't believe any of that needs to be changed, since that's the case I fixed.  Just looking for a second set of eyes on that.

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -14419,11 +14419,6 @@ void Client::Handle_OP_TargetCommand(const EQApplicationPacket *app)
 		return;
 	}
 
-	if (GetTarget())
-	{
-		GetTarget()->IsTargeted(-1);
-	}
-
 	// Locate and cache new target
 	ClientTarget_Struct* ct = (ClientTarget_Struct*)app->pBuffer;
 	pClientSideTarget = ct->new_target;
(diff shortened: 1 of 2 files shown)


zone/client_packet.cpp
```diff
@@ -14460,21 +14460,7 @@ void Client::Handle_OP_TargetCommand(const EQApplicationPacket *app)
...
-			if (g && g->HasRole(this, RoleAssist))
-				g->SetGroupAssistTarget(0);
-
-			if (g && g->HasRole(this, RoleTank))
-				g->SetGroupTankTarget(0);
-
-			if (g && g->HasRole(this, RolePuller))
-				g->SetGroupPullerTarget(0);
-
+			this->MessageString(Chat::Red, DONT_SEE_TARGET);
```
Kinglykrab, line 14463: `this` is implicit in this scope, can just do `MessageString`.

noudess: Fixed

Kinglykrab (approved):

zone/client_packet.cpp
```diff
@@ -14460,21 +14460,7 @@ void Client::Handle_OP_TargetCommand(const EQApplicationPacket *app)
 		}
 		else
 		{
-			SetTarget(nullptr);
```
Akkadius, line 14463: This seems innocent enough, do you think we'll have unintended side effects for removing all of these?

Akkadius: Sounds good. Let's send it

Akkadius (approved):

PR #3401, opened 2023-06-10
[Commands] Assign #opcode to a #reload alias

Kinglykrab:
# Notes
- Can use `#reload opcodes` or `#opcode` to reload opcodes now instead of it being its own command file.


zone/gm_commands/reload.cpp
```diff
@@ -1,16 +1,18 @@
...
-	int         arguments = sep->argnum;
+	const auto  arguments = sep->argnum;
 	if (!arguments && Strings::Contains(command, "#reload")) {
 		c->SendReloadCommandMessages();
 		return;
 	}
 
-	bool is_rq_alias             = sep->arg[0] && Strings::Contains(command, "#rq");
 	bool is_logs_reload_alias    = sep->arg[0] && Strings::Contains(command, "#rl");
+	bool is_opcodes_reload_alias = sep->arg[0] && Strings::Contains(command, "#opcode");
```
Akkadius, line 14: I like the alias touch

Akkadius (approved):
LGTM. Let's get this hooked into the reload API as well

`world/eqemu_api_world_data_service.cpp:137` add just add a `Reload` struct entry - should just work as long as we have world handlers for it

PR #3402, opened 2023-06-11
[Cleanup] Add GMFind_Struct to packet structures

Kinglykrab:
# Notes
- X and Y were swapped in GMSummon_Struct so it displayed XYZ incorrectly in /find, adding its own struct fixes this.

### DIFF
--- a/common/eq_packet_structs.h
+++ b/common/eq_packet_structs.h
@@ -1794,14 +1794,14 @@ struct GMSummon_Struct {
 };
 
 struct GMFind_Struct {
-char	charname[64];
-char	gmname[64];
-uint32	success;
-uint32	zoneID;
-float	x;
-float	y;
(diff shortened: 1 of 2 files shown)


common/eq_packet_structs.h
```diff
@@ -1793,6 +1793,17 @@ struct GMSummon_Struct {
 /*104*/	uint32	unknown2; // E0 E0 56 00
 };
 
+struct GMFind_Struct {
+/*  0*/	char	charname[64];
```
Akkadius, line 1797: The byte sizes are completely off in the comments (I know it was copy pasted). I would just remove them, the struct is not large enough to where someone is going to lose track counting the struct size

Akkadius (approved):
Looks good, minor comment

PR #3407, opened 2023-06-13
[Targeting] Fix bug when using /tar on invalid target

noudess:
Using /tar on an invalid target should not negate a current target. (Tested on live)

My previous attempt to fix this was invalid, as Handle_OP_Target was getting called for multiple packets, so my fix could not reside there alone.

In my digging I found that hitting escape (to remove target), mobs dying, clicking with the mouse, all issued direct OP_TargetMouse commands which should bahave as we have always done them.  However, our code also called the same code for OP_Target, which actually only seems to be meant to server as a validation of a /tar command.  If we accept the target, the client immediately sends the corresponding OP_TargetMouse.  If we reject it, the client sends nothing further.

This fix seems to solve the original issue while not breaking the rest, which is much better than the last shitty PR I submitted.

The diff is hard to read because of indentation changes caused by code removal.  Handle_TargetMouse is exactly whaty Handle_Target used to be, except it mo longer handes accept/reject.  HandleTarget has been refactored for clarity, and only handles accept/reject.

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -14423,8 +14423,6 @@ void Client::Handle_OP_TargetCommand(const EQApplicationPacket *app)
 	// Locate and cache new target
 	ClientTarget_Struct* ct = (ClientTarget_Struct*)app->pBuffer;
 
-	LogError("Handle_OP_TargetCommand newtar[{}]", ct->new_target);
-
 	bool can_target=false;
 	Mob *nt = entity_list.GetMob(ct->new_target);
 
@@ -14470,7 +14468,6 @@ void Client::Handle_OP_TargetMouse(const EQApplicationPacket *app)
 	// Locate and cache new target
 	ClientTarget_Struct* ct = (ClientTarget_Struct*)app->pBuffer;
 	pClientSideTarget = ct->new_target;
-	LogError("Handle_OP_TargetMouse newtar[{}]", ct->new_target);
 	if (!IsAIControlled())
 	{
 		Mob *nt = entity_list.GetMob(ct->new_target);
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -14416,6 +14416,48 @@ void Client::Handle_OP_SwapSpell(const EQApplicationPacket *app)
 void Client::Handle_OP_TargetCommand(const EQApplicationPacket *app)
 {
 	if (app->size != sizeof(ClientTarget_Struct)) {
+		LogError("OP size error: OP_TargetCommand expected:[{}] got:[{}]", sizeof(ClientTarget_Struct), app->size);
+		return;
+	}
+
+	// Locate and cache new target
+	ClientTarget_Struct* ct = (ClientTarget_Struct*)app->pBuffer;
+
+	LogError("Handle_OP_TargetCommand newtar[{}]", ct->new_target);
+
+	bool can_target=false;
+	Mob *nt = entity_list.GetMob(ct->new_target);
+
+	if (nt) {
+		if (GetGM() || (!nt->IsInvisible(this) && (DistanceSquared(m_Position, nt->GetPosition()) <= TARGETING_RANGE*TARGETING_RANGE))) {
+			if (nt->GetBodyType() == BT_NoTarget2 || 
+				nt->GetBodyType() == BT_Special || 
+				nt->GetBodyType() == BT_NoTarget) {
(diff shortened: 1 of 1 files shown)


Kinglykrab (approved):

Kinglykrab:
Looks good to me. Would probably be good to let some others chime in.

zone/client_packet.cpp
```diff
@@ -14510,157 +14550,108 @@ void Client::Handle_OP_TargetCommand(const EQApplicationPacket *app)
...
+		else if (GetTarget()->IsPetOwnerClient())
+		{
+			GetTarget()->IsTargeted(1);
+			return;
+		}
+		else if (GetBindSightTarget())
+		{
+			if (DistanceSquared(GetBindSightTarget()->GetPosition(), GetTarget()->GetPosition()) > (zone->newzone_data.maxclip*zone->newzone_data.maxclip))
 			{
-				if (DistanceSquared(GetBindSightTarget()->GetPosition(), GetTarget()->GetPosition()) > (zone->newzone_data.maxclip*zone->newzone_data.maxclip))
```
Akkadius, line 14613: Why the change here? This is underneath the GetBindSightTarget() logic chain, we should be using `GetBindSightTarget()->GetPosition()` not `m_Position`

zone/client_packet.cpp
```diff
@@ -14510,157 +14550,108 @@ void Client::Handle_OP_TargetCommand(const EQApplicationPacket *app)
...
-			}
-
-			QueuePacket(app);
-
+	if (GetTarget())
+	{
+		if (GetGM())
+		{
 			GetTarget()->IsTargeted(1);
-			SendHPUpdate();
```
Akkadius, line 14537: Why the removal of the HP update? This was here for a reason

zone/client_packet.cpp
```diff
@@ -14414,6 +14414,46 @@ void Client::Handle_OP_SwapSpell(const EQApplicationPacket *app)
...
+				can_target = false;
+			}
+			else {
+				can_target = true;
+			}
+		}
+	}
+
+	if (can_target) {
+		QueuePacket(app);
```
Akkadius, line 14443: A `SendHPUpdate` should be here

zone/client_packet.cpp
noudess: There really isn't  a change here, diff tool is misleading.

zone/client_packet.cpp
noudess: Not removed.  Again, diff is confused by indentation, etc.

zone/client_packet.cpp
noudess: I don't think so.  Handle_OP_Target should only accept/reject.  If it's a valid target and gets accepyted, Handle_OP_TargetMouse gets called immediately and does the update.

zone/client_packet.cpp
Akkadius: No it is definitely removed

image

zone/client_packet.cpp
Akkadius: That's fine

zone/client_packet.cpp
Akkadius: There is actually a change

image

zone/client_packet.cpp
Akkadius: Nevermind. It wasn't there in master. 

My suggestion in the future is to not do indentation / formatting changes with stuff like this as it makes it very difficult to understand what is actually changed. 

Formatting changes to this extent should be separate from logic changes.

Akkadius:
In general, as long as testing and function works as expected - I approve of this PR. For folks trying to pull this PR in to merge with their fork, this is going to be an absolute nightmare to understand what has changed for merge conflicts. 

This should be updated with only the changes and not the formatting changes.

zone/client_packet.cpp
noudess: Makes sense.

noudess:
> In general, as long as testing and function works as expected - I approve of this PR. For folks trying to pull this PR in to merge with their fork, this is going to be an absolute nightmare to understand what has changed for merge conflicts.
> 
> This should be updated with only the changes and not the formatting changes.

> 

My suggestions for a merge if auto merge doesnt work is to just copy these two functions verbatim over the top of existing ones.

PR #2268, opened 2022-06-15
[Feature] Add Water Line of Sight Checks

nytmyr:
RuleB(Combat, WaterMatchRequiredForAutoFireLoS) to True checks that both parties are in or out of the water for AutoFire to work.

RuleB(Spells, WaterMatchRequiredForLoS) to True checks that both parties are in or out of the water for spells to land.

PR #3408, opened 2023-06-14
[Feature] Add Water Line of Sight Checks

nytmyr:
Let's try this again :P --- This is a duplicate of #2268 which went stale.

This adds rules to enable or disable checks for spells and autofire to prevent casting or autofire from landing if the player or bot do not match their targets plane in regards to water.

Currently players and bots can cast or autofire if they are not in the water but their target is and vice versa, this should not be possible.

RuleB(Combat, WaterMatchRequiredForAutoFireLoS) set to True (default) checks that both parties are in or out of the water for AutoFire to work.

RuleB(Spells, WaterMatchRequiredForLoS) set to True (default) checks that both parties are in or out of the water for spells to land.

### DIFF
--- a/zone/client_process.cpp
+++ b/zone/client_process.cpp
@@ -2408,7 +2408,7 @@ bool Client::CheckWaterAutoFireLoS(Mob* m) // checks if both attacker and target
 	if (
 		!RuleB(Combat, WaterMatchRequiredForAutoFireLoS) ||
 		!zone->watermap
-	) { // if rule is set to false, bypass check
+	) {
 		return true;
 	}
 
--- a/zone/client_process.cpp
+++ b/zone/client_process.cpp
@@ -51,6 +51,7 @@
 #include "zone.h"
 #include "zonedb.h"
 #include "../common/events/player_event_logs.h"
+#include "water_map.h"
 
 extern QueryServ* QServ;
 extern Zone* zone;
@@ -336,7 +337,7 @@ bool Client::Process() {
 					if (ranged_timer.Check(false)) {
 						if (GetTarget() && (GetTarget()->IsNPC() || GetTarget()->IsClient()) && IsAttackAllowed(GetTarget())) {
 							if (GetTarget()->InFrontMob(this, GetTarget()->GetX(), GetTarget()->GetY())) {
-								if (CheckLosFN(GetTarget())) {
+								if (CheckLosFN(GetTarget()) && CheckWaterAutoFireLoS(this, GetTarget())) {
 									//client has built in los check, but auto fire does not.. done last.
 									RangedAttack(GetTarget());
 									if (CheckDoubleRangedAttack())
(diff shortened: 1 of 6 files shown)


zone/spells.cpp
```diff
@@ -7100,3 +7100,11 @@ const CombatRecord &Mob::GetCombatRecord() const
 {
 	return m_combat_record;
 }
+
+bool Mob::CheckWaterLoS(Mob* los_attacker, Mob* los_target) // checks if both attacker and target are both in or out of the water
```
Kinglykrab, line 7104: Should just have `los_target` as the only parameter and use `los_attacker` as `this` based on this being a member method.

```cpp
bool Mob::CheckWaterLos(Mob* los_target)
```

```cpp
return zone->watermap->InLiquid(GetPosition()) == zone->watermap->InLiquid(los_target->GetPosition());
```

zone/client_process.cpp
```diff
@@ -2401,3 +2402,11 @@ void Client::SendGuildLFGuildStatus()
 	worldserver.SendPacket(pack);
 	safe_delete(pack);
 }
+
+bool Client::CheckWaterAutoFireLoS(Mob* los_attacker, Mob* los_target) // checks if both attacker and target are both in or out of the water
+{
+	if (!RuleB(Combat, WaterMatchRequiredForAutoFireLoS) || zone->watermap == nullptr) { // if rule is set to false, bypass check
```
Kinglykrab, line 2408:
```cpp
|| !zone->watermap
```

zone/spells.cpp
```diff
@@ -7100,3 +7100,11 @@ const CombatRecord &Mob::GetCombatRecord() const
...
+{
+	if (!RuleB(Spells, WaterMatchRequiredForLoS) || zone->watermap == nullptr) { // if rule is set to false, bypass check
```
Kinglykrab, line 7106:
```cpp
|| !zone->watermap
```

Kinglykrab:
Few comments, when the functions are simplified to one parameter, the `this` can be removed in all the uses.

Kinglykrab:
Seems to work very well for me, would like @fryguy503 to chime in since he had crashes on the previous version.

Kinglykrab:
Me trying to cast on a fish in Surefall Glade while out of water, then casting in water.
image

fryguy503:
> Seems to work very well for me, would like @fryguy503 to chime in since he had crashes on the previous version.

My crashes were related to a zone not having a water map (It actually keyed me into a navmesh issue I was having with that zone which required a water map)

Since fixing that I have not seen any crashes.

Kinglykrab:
> > Seems to work very well for me, would like @fryguy503 to chime in since he had crashes on the previous version.
> 
> My crashes were related to a zone not having a water map (It actually keyed me into a navmesh issue I was having with that zone which required a water map)
> 
> Since fixing that I have not seen any crashes.

Awesome, so you already have this code in your source and can confirm it's good also?

Kinglykrab (approved):

nytmyr:
Thank you KK.

@fryguy503, did you happen to have those crash issues before the nullptr check was added? The initial commit was missing that check and I'm wondering if that was the initial crash issue for not having a watermap.

fryguy503:
> 

Yes, it was before the nullptr check was added.

Akkadius (approved):

PR #3409, opened 2023-06-15
[Telnet] Add cross zone/world wide cast and move functionality to Telnet

Kinglykrab:
# Notes
- Add `czcast`, `czmove`, `wwcast`, and `wwmove` to Telnet functionality.
- Allows operators to cast spells across zone/world-wide and move players across zone/world-wide from telnet.

### DIFF
--- a/world/console.cpp
+++ b/world/console.cpp
@@ -962,6 +962,268 @@ void ConsoleQuit(
 	connection->Close();
 }
 
+void ConsoleCrossZoneCastSpell(
+	EQ::Net::ConsoleServerConnection *connection,
+	const std::string &command,
+	const std::vector<std::string> &args
+) {
+	if (args.size() < 3) {
+		connection->SendLine("czcast character [character_id] [spell_id]");
+		connection->SendLine("czcast expedition [expedition_id] [spell_id]");
+		connection->SendLine("czcast group [group_id] [spell_id]");
+		connection->SendLine("czcast guild [guild_id] [spell_id]");
+		connection->SendLine("czcast name [character_name] [spell_id]");
+		connection->SendLine("czcast raid [raid_id] [spell_id]");
+		return;
+	}
+
+	const auto& type = Strings::ToLower(args[0]);
+
+	const auto is_character  = type == "character";
+	const auto is_expedition = type == "expedition";
+	const auto is_group      = type == "group";
+	const auto is_guild      = type == "guild";
+	const auto is_name       = type == "name";
+	const auto is_raid       = type == "raid";
+
+	if (
+		!is_character &&
+		!is_expedition &&
+		!is_group &&
+		!is_guild &&
+		!is_name &
+		!is_raid
+	) {
+		connection->SendLine("czcast character [character_id] [spell_id]");
+		connection->SendLine("czcast expedition [expedition_id] [spell_id]");
+		connection->SendLine("czcast group [group_id] [spell_id]");
+		connection->SendLine("czcast guild [guild_id] [spell_id]");
+		connection->SendLine("czcast name [character_name] [spell_id]");
+		connection->SendLine("czcast raid [raid_id] [spell_id]");
+		return;
+	}
+
+	const auto   update_identifier = !is_name ? Strings::ToInt(args[1]) : 0;
+	const auto&  name              = is_name ? Strings::UcFirst(Strings::ToLower(args[1])) : std::string();
+	const auto   spell_id          = Strings::ToUnsignedInt(args[2]);
+
+	uint8 update_type;
+
+	if (is_character) {
+		update_type = CZUpdateType_Character;
+	} else if (is_expedition) {
+		update_type = CZUpdateType_Expedition;
+	} else if (is_group) {
+		update_type = CZUpdateType_Group;
+	} else if (is_guild) {
+		update_type = CZUpdateType_Guild;
+	} else if (is_name) {
+		update_type = CZUpdateType_ClientName;
+	} else if (is_raid) {
+		update_type = CZUpdateType_Raid;
+	}
+
+	auto pack = new ServerPacket(ServerOP_CZSpell, sizeof(CZSpell_Struct));
+	auto* CZS = (CZSpell_Struct*) pack->pBuffer;
+
+	CZS->update_type       = update_type;
+	CZS->update_subtype    = CZSpellUpdateSubtype_Cast;
+	CZS->update_identifier = update_identifier;
+	CZS->spell_id          = spell_id;
+
+	strn0cpy(CZS->client_name, name.c_str(), sizeof(CZS->client_name));
+
+	zoneserver_list.SendPacket(pack);
+	safe_delete(pack);
+
+	connection->SendLine(
+		fmt::format(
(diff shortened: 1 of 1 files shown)


world/console.cpp
```diff
@@ -973,17 +1235,20 @@ void RegisterConsoleFunctions(std::unique_ptr<EQ::Net::ConsoleServer>& console)
 	console->RegisterCall("api", 200, "api", std::bind(ConsoleApi, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3));
 	console->RegisterCall("auction", 50, "auction [message]", std::bind(ConsoleAuction, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3));
 	console->RegisterCall("broadcast", 50, "broadcast [message]", std::bind(ConsoleBroadcast, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3));
+	console->RegisterCall("czcast", 50, "czcast [type] [identifier] [spell_id]", std::bind(ConsoleCrossZoneCastSpell, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3));
+	console->RegisterCall("czmove", 50, "czcast [type] [identifier] [instance_id|zone_short_name] - instance_id and zone_short_name are interchangeable", std::bind(ConsoleCrossZoneMove, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3));
```
Akkadius, line 1239: czmove description also says `czcast`

Akkadius:
```> ?
Commands:
  acceptmessages [on/off]
  api
  auction [message]
  broadcast [message]
  czcast [type] [identifier] [spell_id]
  czcast [type] [identifier] [instance_id|zone_short_name] - instance_id and zone_short_name are interchangeable
  echo [on/off]
```

world/console.cpp
```diff
@@ -962,6 +962,268 @@ void ConsoleQuit(
...
+		connection->SendLine("czmove guild [guild_id] [instance_id]");
+		connection->SendLine("czmove guild [guild_id] [zone_short_name]");
+		connection->SendLine("czmove name [character_name] [instance_id]");
+		connection->SendLine("czmove name [character_name] [zone_short_name]");
+		connection->SendLine("czmove raid [raid_id] [instance_id]");
+		connection->SendLine("czmove raid [raid_id] [zone_short_name]");
+		return;
+	}
+
+	const auto    update_identifier = !is_name ? Strings::ToInt(args[1]) : 0;
```
Akkadius, line 1139: Should validate earlier that these inputs are the types that we expect them to be (integer)

```
> czmove character something something
 World |    Info    | GetZoneID Failed to get zone_name [something] 
Moving player(s) to UNKNOWN by character with an identifier of 0.
```

Akkadius: Looks like some of these could be either, but we should still validate if they result to nothing before continuing on

```
> czmove character 1 nozone
 World |    Info    | GetZoneID Failed to get zone_name [nozone] 
Moving player(s) to UNKNOWN by character with an identifier of 1.
```

world/console.cpp
```diff
@@ -962,6 +962,268 @@ void ConsoleQuit(
...
+	strn0cpy(CZM->zone_short_name, zone_short_name.c_str(), sizeof(CZM->zone_short_name));
+	strn0cpy(CZM->client_name, name.c_str(), sizeof(CZM->client_name));
+
+	zoneserver_list.SendPacket(pack);
+	safe_delete(pack);
+
+	connection->SendLine(
+		fmt::format(
+			"Moving player(s) to {} by {} with an identifier of {}.",
+			!instance_id ? zone_store.GetZoneLongName(zone_store.GetZoneID(zone_short_name), true) : fmt::format("Instance ID {}", instance_id),
```
Akkadius, line 1177: This works, but fyi you can use the zone store to fetch the entire zone object by the short name

```cpp
ZoneRepository::Zone *GetZone(const std::string& in_zone_name);
```

```cpp
ZoneRepository::Zone z = zone_store.GetZone(zone_short_name);
// access via z->member
```

Akkadius: Also makes it easier to validate when the zone doesn't exist

```cpp
if (z->id == 0) {
    // return validation error
}
```

world/console.cpp
```diff
@@ -962,6 +962,268 @@ void ConsoleQuit(
...
+	const std::vector<std::string> &args
+) {
+	if (args.size() < 1) {
+		connection->SendLine("wwcast [spell_id]");
+		connection->SendLine("wwcast [spell_id] [min_status]");
+		connection->SendLine("wwcast [spell_id] [min_status] [max_status]");
+		return;
+	}
+
+	const auto spell_id = Strings::ToUnsignedInt(args[0]);
```
Akkadius, line 1061: We should cover some validation cases here

```
> wwcast 1 something 1
Casting spell ID 1 world wide for players with a status between 0 and 1.
> 
> 
> wwcast something 1 something
Casting spell ID 0 world wide for players with a status between 1 and 0.
> 
> 
> wwcast 1 something something
Casting spell ID 1 world wide for players with a status between 0 and 0.
```

Akkadius (approved):
Looks great. In general we could use more validation on inputs to be on the safe side especially with world crashes being higher impact in production.

world/console.cpp
Kinglykrab: Not sure how to validate here, I've tried using `spells[spell_id]` or whatever, but it doesn't like that, also tried `GetSpellName(spell_id)` and it didn't like that.

Kinglykrab: Tried adding a SharedDatabase method to get the spell's name from the database, but that didn't work either, just crashes world even on valid input.

Akkadius: I meant moreso `string` vs `int` input validation for the different args. I don't think we need to load data into world to validate spells specifically

Kinglykrab:
Added validation.

world/console.cpp
```diff
@@ -1195,11 +1263,27 @@ void ConsoleWorldWideMove(
...
+	const auto&  zone_short_name = !Strings::IsNumber(args[2]) ? args[2] : "";
+	const uint16 instance_id     = Strings::IsNumber(args[2]) ? static_cast<uint16>(Strings::ToUnsignedInt(args[2])) : 0;
 
-	const auto&  zone_short_name  = !Strings::IsNumber(args[0]) ? args[0] : "";
-	const uint16 instance_id      = Strings::IsNumber(args[0]) ? static_cast<uint16>(Strings::ToUnsignedInt(args[0])) : 0;
-	const uint8  min_status       = args.size() >= 2 ? static_cast<uint8>(Strings::ToUnsignedInt(args[1])) : 0;
-	const uint8  max_status       = args.size() >= 3 ? static_cast<uint8>(Strings::ToUnsignedInt(args[2])) : UINT8_MAX;
+	const auto& z = !zone_short_name.empty() ? zone_store.GetZone(zone_short_name) : nullptr;
+
+	if (!zone_short_name.empty() && !z->id) {
```
Akkadius, line 1272: Might want to just do `if (z && !z->id)` because `z` could be null and while `zone_short_name` might not be empty, it might not actually return a valid pointer and return `nullptr`

PR #3415, opened 2023-06-17
[CLI] Add `bots:enable` and `bots:disable` commands

Akkadius:
Adds the following commands

* `./bin/world bots:enable` Bootstraps bot tables if they don't exist and sets bots rule `Bots:Enabled` to `true`
* `./bin/world bots:disable` Deletes bot tables and sets bots rule `Bots:Enabled` to `false`


Kinglykrab (approved):

Akkadius:
Tested after merge (since it downloads the file from Github)

image

PR #3428, opened 2023-06-20
[Database] Fix database version checking edge case issue

Akkadius:
When a server operator does not have bots enabled, it is possible for `CheckVersions` to improperly return "not up to date" because bots version values can be incorrectly compared.

This fixes an edge case where if bots are **not** enabled - it will always return **up-to-date**


Kinglykrab (approved):

Akkadius:
Tossica validated this fixed his edge case - good to get this on a release

PR #3434, opened 2023-06-23
[Commands] Add #findlanguage Command

Kinglykrab:
# Notes
- Add `#findlanguage` command.


Akkadius (approved):
Might make sense to consolidate all find commands into `#find` and break out functionality by sub-cpp files

PR #3439, opened 2023-06-23
[Commands] Cleanup #showbuffs Command

Kinglykrab:
# Notes
- Cleaned up messages and logic.
- Removed unnecessary itembonuses/spellbonuses stuff.
- Removed `Mob::ShowBuffList` as it's just a copy of this method.
- Remove `#shownumhits` command since this command now serves its purpose.

# Image
image


Akkadius:
Probably an opportunity here to also consolidate all "show" commands into one command with sub-commands. Will improve compilation times and also consolidate where users expect to find "#show" type commands

Akkadius:
Also would be good to tie in all of the "show" commands into the DevTools targeting functionality

Akkadius (approved):

PR #3452, opened 2023-06-27
[Commands] Consolidate #findX commands to a singular #find Command

Kinglykrab:
# Notes
- All `#findX` commands will be subcommands of an all-encompassing `#find` command.
- Aliases such as `#finditem`, `#fz`, or `#findrace` will still function for those who have been using these for years.
- All command messages and logic have been standardized so they all print the same way.


Akkadius (approved):

crotonrz:
Not a huge deal, I think most people trying operate a server would understand to omit, but `#find` now returns every thousands comma separated numbers (ie 53,125 vs. 53125).  Again, not a huge issue for the vast majority, but using the numbers as displayed would cause issues in scripts, and doesn't work in game when entered manually (ie #npctypespawn 116,137).

image

Kinglykrab:
> Not a huge deal, I think most people trying to operate a server would understand to omit, but `#find` now returns every thousands comma separated numbers (ie 53,125 vs. 53125). Again, not a huge issue for the vast majority, but using the numbers as displayed would cause issues in scripts, and doesn't work in game when entered manually (ie #npctypespawn 116,137).
> 
> image

They’re commified for ease of reading. Not intended to be used as read.

PR #3453, opened 2023-06-27
[Bug Fix] Fix error in 2023_01_21_bots_raid_members.sql

mibastian:
world process did not start. An strace showed something wrong in 2023_01_21_bots_raid_members.sql.  Seemed like a snap&paste error.


mibastian:
To extend on the very short PR. When you attempt a new installation via akk-stack the world server will never come up because the ManifestEntry 2023_01_21_bots_raid_members.sql had a duplicate SQL from the ManifestEntry before it.
I revered to commit 1f25639dd35950148e1be4432a6823c813617be8, right before the revamp and identified and copied the proper SQL.

Akkadius (approved):

PR #3454, opened 2023-06-27
[Bug Fix] Merchant Open Flag set only for regular Merchants

Kinglykrab:
# Notes
- `merchant_open` was only set to `true` if class was `MERCHANT`.


fryguy503 (approved):
lgtm

Thank you for the quick bug fix

PR #3458, opened 2023-06-28
[Bug Fix] Add check for underscores in botcreate command

tuday2:
### NOTES
Check for underscores in bot names as the server will replace with a space and add to the database. The bot can't be deleted/summoned or changed except by editing the database.

Other symbols seem to be properly ignored (they are just deleted, like # or *). I added a simple check to prevent names with _'s. 

### Steps to Recreate
**Command:** ^botcreate Tuday_bot_healer 1 1 1

image

The server replaces _'s with spaces and adds the bot. Here is an example:

image

You can no longer delete/summon/alter the bot in anyway without editing SQL.

### Example after adding check

image

### DIFF
--- a/zone/bot_command.cpp
+++ b/zone/bot_command.cpp
@@ -5512,7 +5512,8 @@ void bot_subcommand_bot_create(Client *c, const Seperator *sep)
 
 	std::string bot_name = sep->arg[1];
 	bot_name = Strings::UcFirst(bot_name);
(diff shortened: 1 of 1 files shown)


Kinglykrab (approved):

Kinglykrab:
Looks great, `Strings::Contains` is a utility method we have to check if a string contains another string.

PR #3471, opened 2023-07-03
[Feature] Add Strings::BeginsWith() and Strings::EndsWith()

Kinglykrab:
# Notes
- These are useful so that we don't have to manually calculate size or perform a substring.


Akkadius (approved):

Akkadius:
Good with this once I know it was tested in game

PR #3481, opened 2023-07-04
[Rules] Add Skills:TrivialTradeskillCombinesNoFail Rule

JasXSL:
Adds a server rule that (when enabled) removes the fail chance for trivial tradeskill combinations.


zone/tradeskills.cpp
```diff
@@ -1050,6 +1050,9 @@ bool Client::TradeskillExecute(DBTradeskillRecipe_Struct *spec) {
 		// above critical still stands.
 		// Mastery modifier is: 10%/25%/50% for rank one/two/three
 		chance = 95.0f + (float(user_skill - spec->trivial) / 40.0f);
+		if(RuleB(Skills, TrivialNoFail)) {
```
Kinglykrab, line 1053: `if (` instead of `if(`

Kinglykrab: Some newlines around this would make it easier to read.

common/ruletypes.h
```diff
@@ -241,6 +241,7 @@ RULE_CATEGORY_END()
 RULE_CATEGORY(Skills)
 RULE_INT(Skills, MaxTrainTradeskills, 21, "Highest level for trading skills that can be learnt by the trainer")
 RULE_BOOL(Skills, UseLimitTradeskillSearchSkillDiff, true, "Enables the limit for the maximum difference between trivial and skill for recipe searches and favorites")
+RULE_BOOL(Skills, TrivialNoFail, false, "Makes all trivial recipes no-fail")
```
Kinglykrab, line 244: `TrivialRecipesNoFail` is a bit more verbose and explains what this does better.

utils/sql/svn/1392_recipe_learning.sql
```diff
@@ -9,4 +9,5 @@ alter table `tradeskill_recipe` add column `must_learn` tinyint not null default
 
 insert into rule_values (ruleset_id, rule_name, rule_value, notes) values
 (1, 'Skills:UseLimitTradeskillSearchSkillDiff', 'true', 'Enables the limit for the maximum difference between trivial and skill for recipe searches and favorites.'),
-(1, 'Skills:MaxTradeskillSearchSkillDiff', '50', 'The maximum difference in skill between the trivial of an item and the skill of the player if the trivial is higher than the skill. Recipes that have not been learnt or made at least once via the Experiment mode will be removed from searches based on this criteria.');
\ No newline at end of file
+(1, 'Skills:TrivialNoFail', 'false', 'Makes all trivial recipes no-fail.'),
```
Kinglykrab, line 12: Adding this here is unnecessary, rules are handled automatically by the source.

common/ruletypes.h
JasXSL: May as well go more verbose with "TrivialTradeskillCombinesNoFail", which better matches the otehr two Skills rules

Kinglykrab: Works for me. :+1:

utils/sql/svn/1392_recipe_learning.sql
JasXSL: So that SQL file is basically deprecated then?

Kinglykrab: Yes, 100%, it's from SVN (subversion) days which is over 10 years old at this point, long before this repository existed.

Kinglykrab: We automatically add new rules in `rulesys.cpp`.
```cpp
	for (const auto &d : rule_data) {
		if (std::find(v.begin(), v.end(), d.first) == v.end()) {
			injected_rule_entries.push_back(
				std::tuple<int, std::string, std::string, std::string>(
					rule_set_id,
					d.first,
					d.second.first,
					Strings::Escape(*d.second.second)
				)
			);

			if (!quiet_update) {
				LogInfo(
					"Adding new rule [{}] ruleset [{}] ({}) value [{}]",
					d.first,
					rule_set_name,
					rule_set_id,
					d.second.first
				);
			}
		}
	}
	```

Kinglykrab (approved):

PR #3486, opened 2023-07-07
[Commands] Consolidate #set-like commands into a singular #set command

Kinglykrab:
# Notes
- All `#set `commands like `#setaltcurrency` are now subcommands of `#set`.
- All aliases like `#setaltcurrency` still function.


fryguy503:
Consolidation is nice for sure, should some of the like commands be nested? 

Rather than

#set zone_color
#set zone_header
#set zone_locked

should it be

#set zone header foo bar
#set zone locked foo bar

That way you could do a `#set zone` and have it return "clipping, header, locked, safe coors, sky, underworld, etc

Kinglykrab:
> Consolidation is nice for sure, should some of the like commands be nested?
> 
> Rather than
> 
> #set zone_color #set zone_header #set zone_locked
> 
> should it be
> 
> #set zone header foo bar #set zone locked foo bar
> 
> That way you could do a `#set zone` and have it return "clipping, header, locked, safe coors, sky, underworld, etc

Could do that for sure.

zone/command.cpp
```diff
@@ -968,6 +824,131 @@ void command_bot(Client *c, const Seperator *sep)
 	}
 }
 
+std::vector<SubCommandRecord> GetSubCommandRecords()
```
Akkadius, line 827: BEAUTIFUL

Akkadius: 😍

zone/command.cpp
```diff
@@ -591,9 +484,52 @@ int command_realdispatch(Client *c, std::string message, bool ignore_status)
...
-		c->Message(Chat::White, "Your status is not high enough to use this command.");
-		return -1;
+
+	bool is_subcommand            = false;
+	bool can_use_subcommand       = false;
+	bool found_subcommand_setting = false;
+
+	const auto arguments = sep.argnum;
+
+	if (arguments >= 2) {
```
Akkadius, line 494: Nice doing this globally

zone/command.cpp
```diff
@@ -549,25 +416,51 @@ int command_add(std::string command_name, std::string description, uint8 admin,
...
-	auto command_status = commands_map[command_name];
-	return command_status;
+uint8 GetCommandStatus(std::string command_name)
+{
+	return commands_map[command_name];
 }
 
+std::vector<CommandSubsettingsRepository::CommandSubsettings> GetCommandSubsettingsByCommand(std::string command)
+{
+	std::vector<CommandSubsettingsRepository::CommandSubsettings> v;
```
Akkadius, line 441: We have `command_subsettings` in memory, we should use it instead of hitting the DB each command invocation

image

Akkadius:
image

Akkadius:
Small: usage does not match what the subcommand asks for in (#set exp) 

image

Akkadius:
Frozen output needs some adjustments

image

Akkadius:
server_locked appears to have a positional issue

image

Akkadius:
Min clip can have 0 values

image

Akkadius:
#ooc_mute

image

Akkadius:
Password has a positional arg issue, also not sure what password its setting so could benefit from description in the usage

image

Akkadius:
Skill has issues too, doesn't matter what skill ID I try (100+ was out of bounds checks)

image

Akkadius:
gender has flymode next to it
image

Akkadius:
gm_status has an arg positional issue
image

zone/command.cpp
```diff
@@ -267,146 +217,62 @@ int command_init(void)
...
+					cs.second.second.emplace_back(alias);
+				}
 			}
-		},
-	};
+		}
+	}
 
-	// inject static aliases
+	// inject database sub command aliases
```
Akkadius, line 266: I'm not actually seeing entries injected (in the database at least)

```
select * from command_subsettings;
Empty set
```

Akkadius:
One thing I think we should do as a quality of life thing is when operators type an old alias (for example #fi) we should shadow print the new command (Gray) so it informs them of where the command is now located and they can explore the other sub commands.

```
#fi is now located under #show
Using: #show item <x> (takes what's passed in) 
```

Akkadius:
image

Akkadius (approved):

PR #3488, opened 2023-07-08
[Bug Fix] Fix charmed pets to follow when charmed.

noudess:
Add on to a fix a few months back to make sure "stand" was correct on charm.  It turns out follow/guard and stop also need to be initialized.


Kinglykrab (approved):
:1st_place_medal:

PR #3490, opened 2023-07-10
[Bug Fix] Fix Tradeskill Combines with augmented items

Kinglykrab:
# Notes
- Keeps players from doing tradeskill combines with augmented items.
- Behavior is Live-like, we don't need a rule since `EVENT_COMBINE` is separate and processed prior to the recipe check.
- String is hard-coded as it's not part of ROF2 `eqstr_us.txt`.

# Live Screenshot
image

# EQEmulator Screenshot
image

### DIFF
--- a/zone/tradeskills.cpp
+++ b/zone/tradeskills.cpp
@@ -1556,7 +1556,7 @@ bool ZoneDatabase::GetTradeRecipe(
 
 	//Pull the on-success items...
 	query = fmt::format(
-		"SELECT item_id,successcount FROM tradeskill_recipe_entries WHERE successcount > 0 AND recipe_id = {}",
+		"SELECT item_id, successcount FROM tradeskill_recipe_entries WHERE successcount > 0 AND recipe_id = {}",
 		recipe_id
 	);
 	results = QueryDatabase(query);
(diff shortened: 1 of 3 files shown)


Akkadius:
There's a lot going on in this PR and does make me a bit nervous. 

What level of testing was done? @Aeadoin could you also give this a look since you did TS's recently?

zone/tradeskills.cpp
```diff
@@ -1476,85 +1505,91 @@ bool ZoneDatabase::GetTradeRecipe(
 				tradeskill_recipe
 				INNER JOIN tradeskill_recipe_entries ON tradeskill_recipe.id = tradeskill_recipe_entries.recipe_id
 			WHERE
-				tradeskill_recipe.id = %lu
-				AND tradeskill_recipe_entries.item_id %s
+				tradeskill_recipe.id = {}
+				AND tradeskill_recipe_entries.item_id {}
```
fryguy503, line 1509: Missing =

Kinglykrab: https://github.com/EQEmu/Server/pull/3490/files#diff-b26ee5d332638e203898010e5a5f8c7988e93a801a9b7acea85d4017d45d9c19R1485

Akkadius (approved):

Akkadius:
@fryguy503 says this has been sitting on his server and nothing core is broken so that's good enough for me

PR #3492, opened 2023-07-12
[Bug Fix] Fix rule check and add rule for pickpocket command

tuday2:
### NOTES
- Add rule for allowing pickpocket
- Fix method that is supposed to check rule, needed to check for false
- Changed Z axis to use Zoffset to account for differences in heights of models

### DIFF
--- a/zone/bot_command.cpp
+++ b/zone/bot_command.cpp
@@ -10041,7 +10041,9 @@ void bot_command_pickpocket(Client *c, const Seperator *sep)
 	glm::vec4 mob_distance    = (c->GetPosition() - target_mob->GetPosition());
 	float     mob_xy_distance = ((mob_distance.x * mob_distance.x) + (mob_distance.y * mob_distance.y));
 	float     mob_z_distance  = (mob_distance.z * mob_distance.z);
-	if (mob_z_distance >= 40 || mob_xy_distance > 250) {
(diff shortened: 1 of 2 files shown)


Kinglykrab (approved):

zone/bot_command.cpp
```diff
@@ -10037,7 +10041,7 @@ void bot_command_pickpocket(Client *c, const Seperator *sep)
 	glm::vec4 mob_distance    = (c->GetPosition() - target_mob->GetPosition());
 	float     mob_xy_distance = ((mob_distance.x * mob_distance.x) + (mob_distance.y * mob_distance.y));
 	float     mob_z_distance  = (mob_distance.z * mob_distance.z);
-	if (mob_z_distance >= 25 || mob_xy_distance > 250) {
+	if (mob_z_distance >= 40 || mob_xy_distance > 250) {
```
Akkadius, line 10044: This should use `GetZOffset` it takes in relation the model and the size of the mob - otherwise you will always find yourself tweaking this number

Akkadius: cc @tuday2

tuday2: Thanks did not know about this, I changed the code and tested. We should be good.

Akkadius (approved):

PR #3499, opened 2023-07-16
[Z Clipping] Don't issue zclip adjustments when NPC is not moving

Akkadius:
Don't issue zclip adjustments when NPC is not moving


noudess:
Thanks.  I knew your fix would be more comprehensive than my quick patch.

noudess (approved):
LGTM

PR #3501, opened 2023-07-18
[Rules] Rule to toggle enforcement of default zone port blocking

Valorith:
### **What**
This PR adds the following rule: World:DisableTeleportFromTutorialAndLoad (default = true)

If true, porting _from_ `TUTORIAL` and `LOAD` zones will be blocked, default is true. 

True is current (pre-PR) behavior, False ignores this hard coded port block.

### **Why**
Some servers use the tutorial zone as a hub and want to allow gating from the zone.

### DIFF
--- a/zone/data_bucket.cpp
+++ b/zone/data_bucket.cpp
@@ -21,20 +21,11 @@ void DataBucket::SetData(const std::string &bucket_key, const std::string &bucke
 
 void DataBucket::SetData(const DataBucketKey &k)
 {
-	auto r = DataBucketsRepository::GetWhere(
-		database,
-		fmt::format(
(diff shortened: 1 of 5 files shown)


Valorith:
Still testing this. I will update with results when testing is complete.

Akkadius:
The rule name is a bit broad, suggest something like `DisableTeleportFromTutorialAndLoad`

Valorith:
> The rule name is a bit broad, suggest something like `DisableTeleportFromTutorialAndLoad`

Makes sense. Updated

Kinglykrab:
Not sure why this was needed or added back when it was, can't be Live-like since Live doesn't use these zones. In my opinion we just remove the logic entirely.

Valorith:
Removing it sounds good to me. Anyone have any objections? @Akkadius ?

PR #3507, opened 2023-07-20
[Crash Fix] Guard against Spells:MaxTotalSlotsPET being set above client allowed maximum.

Valorith:
Currently, setting Spells:MaxTotalSlotsPET above the client allowed maximum of 30 will cause the server to crash and could also cause database corruption. This caps the rule at 30.

Still doing some testing on this.


Valorith:
Tests have been successful in stopping crashes and data corrupted that was occurring when exceeding client max for this rule value.

Kinglykrab (approved):

PR #3509, opened 2023-07-20
[Databuckets] Fix rarer same bucket name scoping overlap issue

Akkadius:
This fixes a rarer scenario where if you have two keys named the same but one exists globally and one exists on a player, it can result in one being fetched first over the other because global queries where not explicitly filtering out character,npc,bot entries


Akkadius:
Getting this one in as a hotfix

PR #3514, opened 2023-07-28
[Scaling/Bug Fix] Scaling where min and max damage was bugged

noudess:
This is a refinement of a fix by @Kinglykrab in the following PR:

https://github.com/EQEmu/Server/pull/3351

@Kinglykrab addressed an issue where min_dmg was 0 in db and max_dmg was not.  This caused the npc_scale_manager to only scale the min_dmg side, which can result in the min_dmg being bigger than the max_dmg.

This PR fixes a side effect of that PR wherein when min_dmg and max_dmg are both 0 in the db, setting min_dmg 1st results in min_dmg getting put into max_dmg (due to his attempts to keep max_dmg greater than min_dmg inside of ModifyNPCStat()) at all times.

My belief is that if _either min or max dmg is 0 in the db, both need to be processed.  His earlier PR assures that, so long as min_dmg is 0, but not when min_dmg is non-zero and max_dmg is 0.  Admittedly an unlikely case, but it simplifies the code tremendously to simply scale both if either is 0.

This was discovered because several special attacks like bash and kick (for scaled mobs) were saying the opponent was invulnerable.  This was due to math from the changes to npc.cpp which left min_damage (GetMinDamage()) negative and sometimes that value was -5 (DMG_INVULNERABLE) or -6 (DMG_RUNE).

eqlog_Ebola_Rolath.txt:[Tue Jun 13 15:14:40 2023] A wan ghoul knight tries to bash YOU, but  are INVULNERABLE!
eqlog_Ebola_Rolath.txt:[Tue Jun 13 15:15:12 2023] A wan ghoul knight tries to bash YOU, but  are INVULNERABLE!
eqlog_Ebola_Rolath.txt:[Tue Jun 13 15:49:58 2023] A wan ghoul knight tries to bash YOU, but  are INVULNERABLE!
eqlog_Ebola_Rolath.txt:[Tue Jun 13 16:06:20 2023] A zol ghoul knight tries to bash YOU, but  are INVULNERABLE!

### DIFF
--- a/zone/npc.cpp
+++ b/zone/npc.cpp
@@ -2583,12 +2583,16 @@ void NPC::ModifyNPCStat(const std::string& stat, const std::string& value)
 	}
 	else if (stat_lower == "min_hit") {
 		min_dmg = Strings::ToInt(value);
+		// Clamp max_dmg to be >= min_dmg
+		max_dmg = std::max(min_dmg, max_dmg);
 		base_damage = round((max_dmg - min_dmg) / 1.9);
 		min_damage  = min_dmg - round(base_damage / 10.0);
 		return;
 	}
 	else if (stat_lower == "max_hit") {
 		max_dmg = Strings::ToInt(value);
+		// Clamp min_dmg to be <= max_dmg
+		min_dmg = std::min(min_dmg, max_dmg);
 		base_damage = round((max_dmg - min_dmg) / 1.9);
 		min_damage  = min_dmg - round(base_damage / 10.0);
 		return;
--- a/zone/npc.cpp
+++ b/zone/npc.cpp
@@ -2583,30 +2583,12 @@ void NPC::ModifyNPCStat(const std::string& stat, const std::string& value)
 	}
 	else if (stat_lower == "min_hit") {
 		min_dmg = Strings::ToInt(value);
-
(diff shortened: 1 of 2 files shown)


noudess:
Please hold merge even if approved until Sunday as I'll be out of town tomorrow and want to be available if you have any issues.

Kinglykrab (approved):

Akkadius:
I'm for this change but we still need to be clamping values

noudess:
> I'm for this change but we still need to be clamping values

Can you explain?  If you mean in ModifyNPCStat() it would have to differ from what I removed.  The existing code there to make sure that min < max actually throws min away completely in this case:

min_dmg and max_dmg 0 in db
scale manager sets min, and then max

The setting of min actually puts min into max and min never gets touched again.

Leaving out the code in ModifyNPCStat works fine since npc_scale_manager always sets both.

But yeah, some quest could only set one and they could be hosed.  (always been the case).

Really we should not allow min or max to be changed individually  - one function to set both at once before the math...  But I don't want to break existing scripts.

Akkadius:
> I took a shot at what @Akkadius wants. This should make sure min is never > max and max is never < min.

Thank you sir

PR #3516, opened 2023-07-30
[Doors] Add door blacklist

Akkadius:
There was an issue on Discord that is incredibly obscure, apparently only on Windows, can't be reproduced in a way that makes it easy to trap. @hgtw @KimLS both tried to reproduce and failed to do so and they run Windows development setups as well.

The issue is stems from a door in the zone that is a switch that appears to also be client controlled. Never have I seen in any zone outside of this zone where the client tries to control a door entity in this way. Related PR and issue https://github.com/EQEmu/Server/pull/2824

When a NPC opens it and the door it is related to, the client will try to "force" the door into a zero state when the server hasn't done so itself fast enough and will spam "OP_CLICKDOOR" packets over and over.

Since we've dumped a ton of time into this with no real solution other than workarounds which appear to be not having the NPC mess with the switches, this PR adds a static blacklist for now which codifies the workaround.

https://discord.com/channels/212663220849213441/1134586539389755412

image

Original report 

"Valorith [Clumsy's World]
OP
 — 07/28/2023 3:41 PM
Looking for some validators/people to test an issue that we're seeing on several servers as it appears to be on master and seems to have started in/before May. There appears to be an issue where Temple of Sol Ro will crash after 5-10 minutes of being in zone. There won't be a crash log or any indicator other than you just getting disconnected. You will not be able to reconnect and zone into soltemple and you will not be able to shut the zone down in-game. The zone process must be nuked. If the zone is booted but in stasis (empty) this problem never appears. I have not been able to isolate it to any other zones (yet)

Edit: Plenty possible this is a windows server issue (as most seem to be).

This has been confirmed on multiple windows servers, to include with NO scripts or plugins active on binaries."

### DIFF
--- a/zone/doors.cpp
+++ b/zone/doors.cpp
@@ -83,6 +83,8 @@ Doors::Doors(const DoorsRepository::Doors &door) :
 	m_close_timer.Disable();
 
 	m_disable_timer = (door.disable_timer == 1 ? true : false);
+
+	m_is_blacklisted_to_open = GetIsDoorBlacklisted();
 }
 
 Doors::Doors(const char *model, const glm::vec4 &position, uint8 open_type, uint16 size) :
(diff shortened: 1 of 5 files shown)


Kinglykrab (approved):

Valorith:
Tested this on my local server and it 100% resolves the issue in `soltemple`. Thanks @Akkadius

PR #3520, opened 2023-08-01
[Bug Fix] Fix Appearance Issues

Kinglykrab:
# Notes
- Changing race, gender, or texture of a Mob could result in it changing sizes due to use not sending the size as part of the appearance packet.
- Also converts the parameterized method to a struct parameter so that we can optionally send things without back-filling multiple arguments.


zone/questmgr.cpp
```diff
@@ -2334,22 +2362,23 @@ void QuestManager::playerfeature(const char* feature, int setting)
...
-		DrakkinDetails,
-		Size
+		AppearanceStruct{
+			.beard = Beard,
+			.beard_color = BeardColor,
+			.drakkin_details = DrakkinDetails,
+			.drakkin_heritage = DrakkinHeritage,
+			.drakkin_tattoo = DrakkinTattoo,
+			.eye_color_one = EyeColor1,
+			.eye_color_two =EyeColor2,
```
Akkadius, line 2372: Space

zone/questmgr.cpp
```diff
@@ -2412,22 +2441,23 @@ void QuestManager::npcfeature(const char* feature, int setting)
...
```
Akkadius, line 2451: Space

PR #3524, opened 2023-08-02
[Feature] Update Raid Functions for Titanium and Underfoot

neckkola:
Good evening all.

I have updated various raid features for Titanium and Underfoot clients.

- Raid window now functional, including with BOTS
- Raid delegate assist/mark work
- Raid notes work
- Raid /rmark, /clearmarks work 

I also updated the structs/encode/decode routines for all clients for consistency.  I also updated the raid note/motd to std::string for the source struct and therefore updated the encode/decodes to convert as required.

Any suggestions or feedback, please let me know.


common/patches/rof.cpp
```diff
@@ -34,6 +34,7 @@
 #include "../rulesys.h"
 #include "../path_manager.h"
 #include "../races.h"
+#include "../../zone/raids.h"
```
Akkadius, line 37: common shouldn't be including anything from `world` or `zone`

Akkadius: Suggest moving the types you're referencing to `common` so that `zone` and `world` reference it instead of the other way around

Akkadius:
Awesome! Will give it a closer look in another day or so

titanium-forever:
I can confirm the raid management window works with Ti. I can invite bots to raids, assign them to groups and attach notes. I haven't tested adding/removing marks because my character doesn't have leadership points yet. I can set something up to test them if needed.

Akkadius:
Update: Still on my list to review, unfortunately have not had time recently

There are still files in common importing zone. `common` should be self contained and should not import project files zone/world/loginserver/ucs etc.

neckkola:
Thanks Akk and no worries on the timing.  I have not updated the common stuff as I figure there will be more comments and I will make them as a package.

common/patches/rof.cpp
```diff
@@ -2651,48 +2653,70 @@ namespace RoF
...
-		else
+		case raidSetNote:
+		{
+			auto in_note = (RaidNote_Struct*)__emu_buffer;
+			auto outapp = new EQApplicationPacket(OP_RaidUpdate, sizeof(structs::RaidNote_Struct));
+			auto note = (structs::RaidNote_Struct*)outapp->pBuffer;
+			note->general.action = raidSetNote;
+			strn0cpy(note->general.leader_name, in_note->general.leader_name, sizeof(note->general.leader_name));
+			strn0cpy(note->general.player_name, in_note->general.player_name, sizeof(note->general.leader_name));
+			strncpy(note->note, in_note->note.c_str(), in_note->note.length());
```
hgtw, line 2696: Same as motd above re:  using `strn0cpy` instead

common/patches/rof.cpp
```diff
@@ -2651,48 +2653,70 @@ namespace RoF
...
+			dest->QueuePacket(outapp);
+			safe_delete(outapp);
```
hgtw, line 2698: Can just use `FastQueuePacket` here and remove the `safe_delete` to be a little more efficient. `QueuePacket` will copy the buffer while `FastQueuePacket` will steal it and delete it

hgtw (commented):
Only got through some of it, I'll look at more later

common/eq_packet_structs.h
```diff
@@ -4165,9 +4164,14 @@ struct RaidAddMember_Struct {
...
 
+struct RaidNote_Struct {
+/*000*/ RaidGeneral_Struct general;
+/*140*/ std::string note;
+};
+
 struct RaidMOTD_Struct {
 /*000*/ RaidGeneral_Struct general; // leader_name and action only used
-/*136*/ char motd[0]; // max size is 1024, but reply is variable
+/*140*/ std::string motd; // max size is 1024, but reply is variable
```
hgtw, line 4174: After looking at this closer this should be safe even if the packet is queued as long as the struct containing the string is being dynamically allocated as the packet's buffer,

common/patches/rof.cpp
```diff
@@ -2651,48 +2653,70 @@ namespace RoF
...
-			strn0cpy(outmotd->general.player_name, inmotd->general.player_name, 64);
-			strn0cpy(outmotd->motd, inmotd->motd, strlen(inmotd->motd) + 1);
+			RaidMOTD_Struct* emu = (RaidMOTD_Struct*)__emu_buffer;
+			auto outapp = new EQApplicationPacket(OP_RaidUpdate, sizeof(structs::RaidMOTD_Struct));
+			structs::RaidMOTD_Struct* eq = (structs::RaidMOTD_Struct*)outapp->pBuffer;
+
+			OUT(general.action);
+			OUT_str(general.player_name);
+			OUT_str(general.leader_name);
+			strncpy(eq->motd, emu->motd.c_str(), emu->motd.length());
```
hgtw, line 2668: This applies to other client files as well (just to avoid repeating)

common/patches/titanium.cpp
```diff
@@ -1245,6 +1246,119 @@ namespace Titanium
...
+			ENCODE_LENGTH_EXACT(RaidAddMember_Struct);
+			SETUP_DIRECT_ENCODE(RaidAddMember_Struct, structs::RaidAdd_Struct);
+
+			eq->action = emu->raidGen.action;
+			eq->level = emu->level;
+			eq->_class = emu->_class;
+			eq->groupid = emu->raidGen.parameter;
+			eq->group_leader = emu->isGroupLeader;
+			strcpy(eq->leader_name, emu->raidGen.leader_name);
+			strcpy(eq->player_name, emu->raidGen.player_name);
```
hgtw, line 1302: `strn0cpys` here for some safety

common/patches/rof.cpp
```diff
@@ -2608,35 +2609,36 @@ namespace RoF
...
+		structs::RaidGeneral_Struct* general = (structs::RaidGeneral_Struct*)outapp_create->pBuffer;
 
 		general->action = 8;
 		general->parameter = 1;
 		strn0cpy(general->leader_name, raid_create->leader_name, 64);
 		strn0cpy(general->player_name, raid_create->leader_name, 64);
 
 		dest->FastQueuePacket(&outapp_create);
-		safe_delete(inapp);
+		safe_delete(outapp_create);
```
hgtw, line 2625: This applies to the other client files too (to avoid repeating)

zone/raids.cpp
```diff
@@ -1543,21 +1543,23 @@ void Raid::SendRaidGroupRemove(const char *who, uint32 gid)
...
+	temp.motd = motd;
 
-	size_t size = motd.size() + 1;
-	auto outapp = new EQApplicationPacket(OP_RaidUpdate, sizeof(RaidMOTD_Struct) + size);
-	auto rmotd = (RaidMOTD_Struct *)outapp->pBuffer;
-	rmotd->general.action = raidSetMotd;
-	strn0cpy(rmotd->general.player_name, c->GetName(), 64);
-	strn0cpy(rmotd->motd, motd.c_str(), size);
-	c->FastQueuePacket(&outapp);
+	auto outapp = new EQApplicationPacket(OP_RaidUpdate, sizeof(RaidNote_Struct));
```
hgtw, line 1558: Should this be `sizeof(RaidMOTD_Struct)`?

zone/raids.cpp
```diff
@@ -1543,21 +1543,23 @@ void Raid::SendRaidGroupRemove(const char *who, uint32 gid)
...
+	outapp->pBuffer = (unsigned char*)&temp;
+	QueuePacket(outapp);
+	outapp->pBuffer = nullptr;
+	safe_delete(outapp);
```
hgtw, line 1562: This will leak the existing `outapp->pBuffer` (which is allocated to `sizeof(RaidNote_Struct)`) since it's being re-assigned.

Also doing it like this would be unsafe if this packet is ever queued since the `pBuffer` copied by `QueuePacket` would be left dangling once `temp` is gone. 

Since outapp's buffer should already be a `RaidMOTD_Struct` it can just be casted and modifed directly instead of using `temp`:

```cpp
RaidMOTD_Struct* temp = reinterpret_cast<RaidMOTD_Struct*>(outapp->pBuffer);
temp->general.action = raidSetMotd;
...
```

zone/raids.cpp
```diff
@@ -2278,13 +2282,18 @@ void Raid::SendRaidNotes()
...
+		temp.general.parameter = 0;
+		temp.general.unknown1 = 0;
+		strn0cpy(temp.general.leader_name, c.member_name, 64);
+		strn0cpy(temp.general.player_name, GetLeaderName().c_str(), 64);
+		temp.note = c.note;
+
+		auto outapp = new EQApplicationPacket(OP_RaidUpdate, sizeof(RaidNote_Struct));
+		outapp->pBuffer = (unsigned char*)&temp;
 		QueuePacket(outapp);
+		outapp->pBuffer = nullptr;
```
hgtw, line 2296: Same as above with the `outapp` buffer leak, can just cast the buffer to `RaidNote_Struct` and modify directly

hgtw (commented):
The only serious issues I see are the `inapp` leak in all the client `ENCODE(OP_RaidJoin)` functions and the packet creation leaks in `Raid::SendRaidMOTD` and `Raid::SendRaidNotes`.

Otherwise some places strn0cpy should be used with dst buffer sizes instead of strncpy/strcpy using the src buffer for a little more safety and some unnecessary `safe_deletes` where `FastQueuePacket` is used.

neckkola:
Thanks for the review @hgtw .  I will take a look and make some updates.  I appreciate the explanations as well as it gives me some context for my own learning.

common/eq_packet_structs.h
hgtw: Looking at this again, I think these should be kept as `char[]` to avoid possible undefined behavior with memcpy and destruction

common/patches/rof2.cpp
```diff
@@ -5076,38 +5086,69 @@ namespace RoF2
...
+			//Map values and convert char to string to be used within source
+			IN(general.action);
+			IN(general.parameter);
+			IN_str(general.leader_name);
+			IN_str(general.player_name);
+			emu->motd = std::string(eq->motd);
+
+			//Point buffer to new structure
+			__packet->size = sizeof(RaidNote_Struct);
+			__packet->pBuffer = (unsigned char*)emu;
```
hgtw, line 5110: This will cause the `std::string` to leak in the packet struct because `RaidMOTD_Struct` never destructs.

The `std::string` in `RaidMOTD_Struct` makes it non-trivially destructible. When `pBuffer` is deleted later the `RaidMOTD_Struct` will not destruct (just its memory is deleted) which will result in the memory allocated by the std::string to leak. The `RaidMOTD_Struct` would have to be deleted here to destruct it properly, but the only way to do that is by memcpying to the buffer which would also be undefined behavior because of the `std::string`.

zone/raids.cpp
hgtw: This was kind of bad advice due to the struct having a `std::string` (ignoring that technically reinterpret_casting to the struct and then de-referencing it is already undefined behavior because it violates strict aliasing but we do that everywhere already).

Since the struct has the string it would have to actually be constructed to be used instead of just overlaying the memory like this. The struct would have to be created and then memcpy'd onto the buffer. But that's undefined behavior because it contains a std::string.

hgtw (commented):
After looking at this for a third time, I think the packet structs should be kept trivial + standard layout with a `char[]`. The std::string makes the packet structs non-trivially copyable which would result in undefined behavior if it's memcpy'd.

It also makes the struct non-trivially destructible so its string will leak when being assigned to the packet `pBuffer` in decoders because it's never properly deleted.

I think refactoring to just keep the server packet structs as `char[]` and copying into/out of them would be better

neckkola:
Thanks for the comments and apologies for my delay.  Been busy with rl and a few other projects.  Will take a look at this in the next few days and tiddy it up.

Akkadius:
> Thanks for the comments and apologies for my delay. Been busy with rl and a few other projects. Will take a look at this in the next few days and tidy it up.

Thanks @neckkola !

neckkola:
Good evening.  I have reviewed and updated.  I hope I hit all the comments though happy to adjust if I overlooked something.  I also made a few other updates to the Ti encoding and tested against Ti, UF and RoF2.

common/patches/ss_define.h
```diff
@@ -70,6 +70,9 @@
 #define OUT_str(x) \
 	strncpy(eq->x, emu->x, sizeof(eq->x)); \
         eq->x[sizeof(eq->x)-1] = '\0';
+#define OUT_str_x_y(x,y) \
```
hgtw, line 73: This looks unused, not sure if intended to use it somewhere or leftover from testing

hgtw (commented):
Only commented in one client translator but comments should apply to all of them

neckkola:
Thanks @hgtw .  My intent was to use the macros where possible as an attempt to make things consistent across all the translators and as had been previously done.  That lead me to having some mixed, using the macro and then not.  Do you have a preference?  I will also confirm the client packet for motd/note for Ti/UF/RoF2 and modify the variable length struct.  Thanks again for all your help on this.

Neckkola.

common/eq_packet_structs.h
```diff
@@ -4165,9 +4164,14 @@ struct RaidAddMember_Struct {
...
 
+struct RaidNote_Struct {
+/*000*/ RaidGeneral_Struct general;
+/*140*/ char note[];
+};
+
 struct RaidMOTD_Struct {
 /*000*/ RaidGeneral_Struct general; // leader_name and action only used
-/*136*/ char motd[0]; // max size is 1024, but reply is variable
+/*140*/ char motd[]; // max size is 1024, but reply is variable
```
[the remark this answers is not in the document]
neckkola: Good evening.  I ran some further testing with Ti/UF/RoF2.  
Motd is always 1024 from UF/RoF2.  Ti is variable to server.
Note is always 64 from all three clients.
Therefore, I set motd[1024] and note[64] for all, then added some decode translation for Ti to check for shorter length.

common/patches/rof.cpp
```diff
@@ -2608,40 +2609,41 @@ namespace RoF
...
+		strncpy_s(general->player_name, raid_create->leader_name, sizeof(raid_create->leader_name));
 
 		dest->FastQueuePacket(&outapp_create);
 		safe_delete(inapp);
 	}
 
 	ENCODE(OP_RaidUpdate)
 	{
-		EQApplicationPacket *inapp = *p;
-		*p = nullptr;
```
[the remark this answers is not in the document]
neckkola: Reworked the RoF2 and Ti encoders/decoders to following this pattern.  @hgtw, if you can take a look and is it is ok, I will push to the other client types.

common/patches/rof.cpp
```diff
@@ -2651,46 +2653,80 @@ namespace RoF
...
+			strn0cpy(note->general.leader_name, in_note->general.leader_name, sizeof(note->general.leader_name));
+			strn0cpy(note->general.player_name, in_note->general.player_name, sizeof(note->general.leader_name));
+			strn0cpy(note->note, in_note->note, size);
+
+			dest->FastQueuePacket(&outapp);
+			break;
+		}
+		case raidNoRaid:
+		{
+			dest->FastQueuePacket(&*p);
```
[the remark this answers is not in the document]
neckkola: Updated as noted above.

common/patches/rof.cpp
```diff
@@ -2651,46 +2653,80 @@ namespace RoF
...
+			strn0cpy(raid_general->leader_name, in_raid_general->leader_name, sizeof(in_raid_general->leader_name));
+			strn0cpy(raid_general->player_name, in_raid_general->player_name, sizeof(in_raid_general->player_name));
 			raid_general->action = in_raid_general->action;
 			raid_general->parameter = in_raid_general->parameter;
+
 			dest->FastQueuePacket(&outapp);
+			break;
 		}
-
 		safe_delete(inapp);
```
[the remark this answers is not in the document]
neckkola: Updated

common/patches/rof.cpp
```diff
@@ -4861,37 +4897,47 @@ namespace RoF
...
-				FINISH_DIRECT_DECODE();
-				break;
-			}
+		RaidGeneral_Struct* rgs = (RaidGeneral_Struct*)__packet->pBuffer;
+
+		switch (rgs->action)
+		{
+		case raidSetMotd:
+		{
+			SETUP_VAR_DECODE(RaidMOTD_Struct, structs::RaidMOTD_Struct, motd);
```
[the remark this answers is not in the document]
neckkola: With the struct changes, using IN_str/OUT_str which should resolve this one.

common/patches/rof.cpp
```diff
@@ -4861,37 +4897,47 @@ namespace RoF
...
+
+			IN(general.action);
+			IN(general.parameter);
+			IN_str(general.leader_name);
+			IN_str(general.player_name);
+			strn0cpy(emu->motd, eq->motd, strlen(eq->motd) + 1);
```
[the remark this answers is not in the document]
neckkola: With the struct changes, using IN_str/OUT_str which should resolve this one.

common/patches/rof.cpp
```diff
@@ -4861,37 +4897,47 @@ namespace RoF
...
+		}
+		case raidSetNote:
+		{
+			SETUP_VAR_DECODE(RaidNote_Struct, structs::RaidNote_Struct, note);
+
+			IN(general.action);
+			IN(general.parameter);
+			IN_str(general.leader_name);
+			IN_str(general.player_name);
+			strn0cpy(emu->note, eq->note, strlen(eq->note) + 1);
```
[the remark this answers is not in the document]
neckkola: With the struct changes, using IN_str/OUT_str which should resolve this one.

neckkola:
Good evening.  I have updated with further confirmation for the motd/note size and adjusted accordingly.  Updated RoF2/Ti translators accordingly.  If they are ok, I will update the others.  Further comments welcome.

Akkadius:
FYI - Linux build is failing - but not reporting properly http://drone.akkadius.com/EQEmu/Server/11076/1/2

I'll fix the reporting separately

common/patches/rof.cpp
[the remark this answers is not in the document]
hgtw: Looks good to me

hgtw:
> That lead me to having some mixed, using the macro and then not. Do you have a preference?

I don't have a preference here, your discretion

> Updated RoF2/Ti translators accordingly. If they are ok, I will update the others

Looked good to me, I didn't see any other issues except for the failing build from use of `strncpy_s`.

neckkola:
Updated the other translators and fixed the linux build issue.

zone/raids.cpp
```diff
@@ -1543,21 +1545,22 @@ void Raid::SendRaidGroupRemove(const char *who, uint32 gid)
...
-	rmotd->general.action = raidSetMotd;
-	strn0cpy(rmotd->general.player_name, c->GetName(), 64);
-	strn0cpy(rmotd->motd, motd.c_str(), size);
-	c->FastQueuePacket(&outapp);
+	data->general.action = raidSetMotd;
+	data->general.parameter = 0;
+	data->general.unknown1 = 0;
+	strncpy(data->general.leader_name, c->GetName(), sizeof(c->GetName()));
+	strncpy(data->general.player_name, GetLeaderName().c_str(), 64);
+	strncpy(data->motd, motd.c_str(), sizeof(data->motd));
```
hgtw, line 1560: All the regular strncpy calls should be changed to `strn0cpy` (ones in translators too that were renamed from strncpy_s)

neckkola: Good morning.  Updated raids.cpp (including another spot), translators, and the defines for OUT_str.

common/patches/ss_define.h
```diff
@@ -68,7 +68,7 @@
 #undef OUT
 #define OUT(x) eq->x = emu->x;
 #define OUT_str(x) \
-	strncpy(eq->x, emu->x, sizeof(eq->x)); \
+	strn0cpy(eq->x, emu->x, sizeof(eq->x)); \
```
hgtw, line 71: This one should be reverted since the macro already handled adding null in its own way and we probably don't want to change how this works in this patch

hgtw:
Didn't notice any other issues so I think this is good otherwise

common/patches/titanium.cpp
```diff
@@ -1421,7 +1535,7 @@ namespace Titanium
 			eq->Class[char_index] = 0;
 			eq->Face[char_index] = 0;
 
-			strncpy(eq->Name[char_index], "<none>", 6);
+			strn0cpy(eq->Name[char_index], "<none>", 6);
```
hgtw, line 1538: Sorry I missed this one, looks like an accidental change. This should be reverted too (it would replace the last '>' with a null)

neckkola: Aye, missed this one.  Updated.

hgtw (approved):

Aeadoin:
@Kinglykrab updated binaries today on KMRA, forming a raid instantly crashes the zone. 

I was able to reproduce this on my development box after updating binaries as well.

[crash_barren_version_0_inst_id_0_port_7033_1659789.log](https://github.com/EQEmu/Server/files/12908627/crash_barren_version_0_inst_id_0_port_7033_1659789.log)

This doesn't instill much confidence in me if just forming a raid after update is crashing zones. Going to revert the change for now.

PR #3529, opened 2023-08-03
[Fix] Fix issue with mob scanning when trying to use EVENT_SPAWN

Akkadius:
This wasn't an issue until recently when Quest API entries were exposed that add the ability to read the close mob list.

The close mob list is not scanned until after `EVENT_SPAWN` this moves scanning before EVENT_SPAWN so script callers can rely on the information being there.


noudess:
Wow that was a fast fix.

Kinglykrab (approved):

PR #3531, opened 2023-08-04
[Rules] Add rule to toggle pets accepting quest items

Valorith:
Adds the following rule: Pets:CanTakeQuestItems, default true.

As name suggests, rejects items flagged as "Quest" when traded to a pet.  The main purpose for this rule is to block players accidentally trading quest items to their pet instead of the desired quest NPC. 

Everything works so far in testing, but I want to torture test it some more before i take this out of draft.

### DIFF
--- a/common/item_data.cpp
+++ b/common/item_data.cpp
@@ -230,6 +230,10 @@ bool EQ::ItemData::IsTypeShield() const
 	return (ItemType == item::ItemTypeShield);
 }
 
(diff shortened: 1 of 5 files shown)


noudess:
Does the existing rule: Pets:CanTakeNoDrop do what you need?

Valorith:
Going to close and remake this to try to clean up the commit history.

PR #3539, opened 2023-08-07
[Spawns] Fixes a rarer issue where spawn2 is not being properly content filtered

Akkadius:
Fixes a rarer issue where spawn2 is not being properly content filtered

As observered when @regneq was working on PoK spawns and was splitting the same NPC onto two separate grids and the wrong pathgrid was being loaded 

**Before**

image

image

**After**

image

### DIFF
--- a/zone/spawn2.cpp
+++ b/zone/spawn2.cpp
@@ -26,6 +26,7 @@
 #include "worldserver.h"
 #include "zone.h"
 #include "zonedb.h"
+#include "../common/repositories/criteria/content_filter_criteria.h"
 
 extern EntityList entity_list;
 extern Zone* zone;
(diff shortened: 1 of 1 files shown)


zone/spawn2.cpp
```diff
@@ -485,7 +486,8 @@ bool ZoneDatabase::PopulateZoneSpawnList(uint32 zoneid, LinkedList<Spawn2*> &spa
 		"animation "
 		"FROM "
 		"spawn2 "
-		"WHERE zone = '%s' AND  (version = %u OR version = -1)",
+		"WHERE TRUE {} AND zone = '{}' AND (version = {} OR version = -1) ",
```
Kinglykrab, line 489: What does `TRUE` do here?

Akkadius: It’s for ANDing, you don’t know if content filter criteria will return anything so if it doesn’t the query will fail with a syntax error before AND. True is like a no-op placeholder so we can always AND regardless

Kinglykrab: 👍

PR #3548, opened 2023-08-15
[Feature] Add adjustability for AERampage Range.

fryguy503:
This functionality is needed for fights like Ture to be accurate, where their ramp range was 101% of their melee safe range.

Example in lua of utilizing this scripting

```
e.self:SetSpecialAbilityParam(SpecialAbility.area_rampage,8,101;
```

### DIFF
--- a/zone/aggro.cpp
+++ b/zone/aggro.cpp
@@ -1080,18 +1080,15 @@ bool Mob::CombatRange(Mob* other, float fixed_size_mod, bool aeRampage, ExtraAtt
 			}
 		}
 
-		if (aeramp_size <= 0 ) {
+		if (aeramp_size <= 0) {
 			aeramp_size = 0.90;
 		} else {
 			aeramp_size /= 100;
 		}
 
 		float ramp_range = size_mod * aeramp_size;
-		if (_DistNoRoot <= ramp_range) {
-			return true;
-		} else {
-			return false;
-		}
+		
+		return _DistNoRoot <= ramp_range;
 	}
 
 	if (_DistNoRoot <= size_mod) {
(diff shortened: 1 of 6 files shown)


zone/aggro.cpp
```diff
@@ -1071,18 +1070,31 @@ bool Mob::CombatRange(Mob* other, float fixed_size_mod, bool aeRampage)
...
-		float ramp_range = (size_mod * multiplyer);
+		float aeramp_size = RuleR(Combat, AERampageMaxDistance);
+
+		if (opts) {
+			if (opts->range_percent > 0) {
+				aeramp_size = opts->range_percent;
+			}
+		}
+
+		if (aeramp_size <= 0 ) {
```
Kinglykrab, line 1083: Extra space here.

zone/aggro.cpp
```diff
@@ -1071,18 +1070,31 @@ bool Mob::CombatRange(Mob* other, float fixed_size_mod, bool aeRampage)
...
+			aeramp_size = 0.90;
+		} else {
+			aeramp_size /= 100;
+		}
+
+		float ramp_range = size_mod * aeramp_size;
 		if (_DistNoRoot <= ramp_range) {
```
Kinglykrab, line 1090: Can use this instead.
```cpp
return _DistNoRoot <= ramp_range;```

Kinglykrab (approved):
Looks great, small comments. 🔥

zone/aggro.cpp
fryguy503: updated

zone/aggro.cpp
fryguy503: updated

PR #3549, opened 2023-08-15
[Charm] Fix to update target windows on charm on/off

noudess:
On classic servers that do not use the rule AlwaysSendTargetsBuffs, we can only see a charmed pet's buffs if we manually retarget after the charm.  These buffs also remain displayed (including the charm) after charm breaks unless we re-target.  This is a bug.

This PR fixes this issue by updating all client target windows when a mob they have targeted is charmed.  It also clears all the buffs from that clients sight when the charm drops (unless client is a GM, AlwaysSendTargetsBuffs is set, or the groupAAInspectBuffsis in play).

### DIFF
--- a/zone/entity.cpp
+++ b/zone/entity.cpp
@@ -1595,7 +1595,7 @@ void EntityList::RefreshClientXTargets(Client *c)
 }
 
 void EntityList::QueueClientsByTarget(Mob *sender, const EQApplicationPacket *app,
-		bool iSendToSender, Mob *SkipThisMob, bool ackreq, bool HoTT, uint32 ClientVersionBits, bool inspect_buffs, bool clear)
+		bool iSendToSender, Mob *SkipThisMob, bool ackreq, bool HoTT, uint32 ClientVersionBits, bool inspect_buffs, bool clear_target_window)
 {
 	auto it = client_list.begin();
 	while (it != client_list.end()) {
@@ -1611,7 +1611,7 @@ void EntityList::QueueClientsByTarget(Mob *sender, const EQApplicationPacket *ap
 
 		TargetsTarget = Target->GetTarget();
 
-		bool Send = clear;
+		bool Send = clear_target_window;
 
 		if (c == SkipThisMob)
 			continue;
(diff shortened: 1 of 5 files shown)


zone/entity.cpp
```diff
@@ -1611,7 +1611,7 @@ void EntityList::QueueClientsByTarget(Mob *sender, const EQApplicationPacket *ap
 
 		TargetsTarget = Target->GetTarget();
 
-		bool Send = false;
```
Kinglykrab, line 1614: Should we just use the parameter at this point?

zone/entity.cpp
```diff
@@ -1624,22 +1624,23 @@ void EntityList::QueueClientsByTarget(Mob *sender, const EQApplicationPacket *ap
...
-						if (!raid)
-							continue;
-						uint32 gid = raid->GetGroup(c);
-						if (gid > 11 || raid->GroupCount(gid) < 3)
-							continue;
-						if (raid->GetLeadershipAA(groupAAInspectBuffs, gid))
-							Send = true;
+						if (raid) {
+							uint32 gid = raid->GetGroup(c);
+							if (gid <= 11 && raid->GroupCount(gid) >= 3) {
```
Kinglykrab, line 1632: `< MAX_RAID_GROUPS `

zone/spells.cpp
```diff
@@ -6258,12 +6258,22 @@ void Mob::SendBuffsToClient(Client *c)
...
-	uint32 buff_count = for_target ? GetMaxBuffSlots() : GetMaxTotalSlots();
+	if (for_target)
+	{
+		buff_count = (clear) ? 0 : GetMaxBuffSlots();
+	}
+	else
+	{
+		buff_count = GetMaxTotalSlots();
+	}
+
```
Kinglykrab, line 6276: Formatting, one these are preferred.
```cpp
if {
}
else {
}
```
or
```cpp
if {
} else {
}```

Comment (Kinglykrab) [reply zone/spells.cpp:6276]: Could also use a double ternary if you wanted. Not sure if others would approve though.
```cpp
buff_count = for_target ? (clear ? 0 : GetMaxBuffSlots()) : GetMaxTotalSlots();
```

zone/entity.cpp
noudess: I don't think so.  It's kind of a crappy overload but it felt better than rewriting the whole thing.  When we send a packet to clear the window, we always send it.  Hence, if clear is set so is send..  But I think it's clearer as written.  Let me know.

zone/spells.cpp
noudess: Ok, updated this function.

zone/entity.cpp
```diff
@@ -1595,7 +1595,7 @@ void EntityList::RefreshClientXTargets(Client *c)
 }
 
 void EntityList::QueueClientsByTarget(Mob *sender, const EQApplicationPacket *app,
-		bool iSendToSender, Mob *SkipThisMob, bool ackreq, bool HoTT, uint32 ClientVersionBits, bool inspect_buffs)
+		bool iSendToSender, Mob *SkipThisMob, bool ackreq, bool HoTT, uint32 ClientVersionBits, bool inspect_buffs, bool clear)
```
Akkadius, line 1598: `clear` - Clear what? This should read it's intent

zone/spells.cpp
```diff
@@ -6258,29 +6258,33 @@ void Mob::SendBuffsToClient(Client *c)
 	}
 }
 
-EQApplicationPacket *Mob::MakeBuffsPacket(bool for_target)
+EQApplicationPacket *Mob::MakeBuffsPacket(bool for_target, bool clear)
```
Akkadius, line 6261: Same comment on clear as above

zone/entity.cpp
noudess: Yeah, I tend to agree.  This QueueClientsByTarget function kinda sucks and it isn't gonna read clearly no matter what because the internals and one of the arguments are non-generic but the function wants to think it is.  But I updated the variable a bit.

zone/spells.cpp
noudess: Done.

zone/entity.cpp
Akkadius: The code is certainly messy, but at least the one tweak should help

Akkadius (approved):
Good enough for now, the existing code is a bit of a mess

PR #3551, opened 2023-08-18
[Combat Messages] Fix issue where pet proc damage was not showing up

noudess:
@Kinglykrab reported a problem from his server where some pet damage from spells was being missed by observers.

While I think pet damage isn't seen by others on live, as there is no others-pet-filters, this was always a part of eqemu and many parsers used it.

The bug came from my PR back in 2022: https://github.com/EQEmu/Server/pull/2170/files

That PR removed many duplicate messages and added missing color messages, but omitted the ones this PR fixes.

New output as seen by owner of pet that is procing:

[Fri Aug 18 14:11:03 2023] Bristlebanes Warder begins to cast a spell. (Spirit of Scorpion Strike)
[Fri Aug 18 14:11:03 2023] Bristlebanes Warder strikes Kizdean Gix for 92 points of damage.
[Fri Aug 18 14:11:03 2023] Kizdean Gix spasms as the spirit of a scorpion rips through their body .

New output as seen by close observers:

[Fri Aug 18 14:11:09 2023] Bristlebanes Warder begins to cast a spell. (Spirit of Scorpion Strike)
[Fri Aug 18 14:11:09 2023] Bristlebanes Warder strikes Kizdean Gix for 92 points of damage.
[Fri Aug 18 14:11:09 2023] Kizdean Gix spasms as the spirit of a scorpion rips through their body .

This PR also puts the pet damage under the correct filter if you are the pet owner.  if you're not the pet owner, you're just gonna get the messages just like all the other pet messages.

I'd suggest long term maybe we need a rule to turn off others pet messages, but that's outside the scope of this fix.

@Kinglykrab Maybe you can have your guy do some testing before we merge this?  We tried to test it well, testing various situations, but only have like 2 hours of testing in.  Your call.

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -4225,12 +4225,11 @@ void Mob::CommonDamage(Mob* attacker, int64 &damage, const uint16 spell_id, cons
 		//Note: if players can become pets, they will not receive damage messages of their own
 		//this was done to simplify the code here (since we can only effectively skip one mob on queue)
 		eqFilterType filter;
-		Mob* skip = attacker;
 		if (attacker && attacker->GetOwnerID()) {
 			//attacker is a pet, let pet owners see their pet's damage
 			Mob* owner = attacker->GetOwner();
 			if (owner && owner->IsClient()) {
-				if ((IsValidSpell(spell_id) || (FromDamageShield)) && damage > 0) {
+				if (FromDamageShield && damage > 0) {
 					//special crap for spell damage, looks hackish to me
 					char val1[20] = { 0 };
 					owner->MessageString(Chat::NonMelee, OTHER_HIT_NONMELEE, GetCleanName(), ConvertArray(damage, val1));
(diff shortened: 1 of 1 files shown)


zone/attack.cpp
```diff
@@ -4225,12 +4225,11 @@ void Mob::CommonDamage(Mob* attacker, int64 &damage, const uint16 spell_id, cons
 		//Note: if players can become pets, they will not receive damage messages of their own
 		//this was done to simplify the code here (since we can only effectively skip one mob on queue)
 		eqFilterType filter;
-		Mob* skip = attacker;
 		if (attacker && attacker->GetOwnerID()) {
 			//attacker is a pet, let pet owners see their pet's damage
 			Mob* owner = attacker->GetOwner();
 			if (owner && owner->IsClient()) {
-				if ((IsValidSpell(spell_id) || (FromDamageShield)) && damage > 0) {
+				if (FromDamageShield && damage > 0) {
```
Akkadius, line 4232: Shouldn't we at least leave the `IsValidSpell(spell_id)` check here and && the rest of the logic ?

noudess: Yes!

Akkadius: That's fine, the spell is not used in this call chain regardless. The code that was here before needs a lot of help

noudess:
Refinement today.  I had removed some logic I thought no longer needed but we needed it for some melee to not be duplicated.

noudess:
Tested some on Rolath and Lazurus.  Seems good.

PR #3552, opened 2023-08-18
[Bug Fix] Fix Bot::CheckDataBucket to work with Owner Buckets.

Aeadoin:
Fixed regression from #3498, Bot::CheckDataBucket was not properly checking owner buckets as ```k.key``` was not being set after ```k = GetBotOwner()->GetScopedBucketKeys();```.


Kinglykrab (approved):
👍

PR #3554, opened 2023-08-19
[Rules] Add FinalRaidExpMultiplier Rule

Valorith:
# This PR adds a new rule:

`Character:FinalRaidExpMultiplier`
Default: 1.0
Description: Multiplies all raid experience by this value. Default: 1.0

***_This new Rule defaults to 1.0 and does not change any behavior without adjusting it._***

# What does this do that the other Raid Exp Modifier does not do?

Prior to this PR, the following rules existed to modify raid experience:

1) `Character:RaidExpMultiplier` : The experience gained in raids is multiplied by (1-RaidExpMultiplier).
2) `Character:EnableRaidMemberEXPModifier` : Enable or disable the raid experience modifier based on members in raid, default is true.

The issue (that this PR solves) is that the first rule doesnt do what its name would suggest to most people. Instead of being a straight multiplier against the unaltered xp value, it is actually an experience penalty, using the below formula:

```cpp
raid_experience = static_cast<uint64>(
static_cast<float>(raid_experience) *
(1.0f - RuleR(Character, RaidExpMultiplier))
);
```
This leaves server operators without an intuitive way to modify the default raid xp directly. This PR adds a rule that uses a more intuitive formula to directly modify the player's raid exp:

```cpp
raid_experience = static_cast<uint64>(
       static_cast<float>(raid_experience) *
       RuleR(Character, FinalRaidExpMultiplier)
}
```

Or more simply put:

Existing formula: `RaidXP = UnmodifiedXP x (1.0 - RaidExpMultiplier)`

New formula: `RaidXP = UnmodifiedXP x FinalRaidExpMultiplier`


Valorith:
Tested and working as expected

Akkadius:
Reading this PR it is not clear what this is looking to solve versus the other rules. We have a ton of EXP knobs already

Valorith:
> Reading this PR it is not clear what this is looking to solve versus the other rules. We have a ton of EXP knobs already keep in mind.
> 
> What is flat versus not flat? Is it intended as a final multiplier? If so we should continue using language that already exists, example
> 
> ```c++
> RULE_REAL(Character, FinalExpMultiplier, 1, "Added on top of everything else, easy for setting EXP events")
> ```
> 
> Above ^ 1 = 100% 2 = 200%
> 
> Not opposed to this change at all, just if I'm confused as to what it does, chances are others will be as well.

Thats fair...I really didnt clarify what this PR is solving. I updated the description above. Please let me know if there is still any lingering confusion. Thanks

Akkadius:
Thanks for adjusting the PR description, my point though is I should be able to understand what these rules do by their names and/or their notes fields.

If the original rules are faulty then why are they there?

Valorith:
> Thanks for adjusting the PR description, my point though is I should be able to understand what these rules do by their names and/or their notes fields.
> 
> If the original rules are faulty then why are they there?

I modified the rule to hopefully be more intuitively named/described, consistent with your original suggestion. I am leaving the already existing rule alone because servers may be using it and changing it would be somewhat intrusive with update queries, etc. 

This new rule is a final multiplier, after all other raid exp mods.

zone/exp.cpp
```diff
@@ -1213,6 +1213,12 @@ void Raid::SplitExp(const uint64 exp, Mob* other) {
 			}
 		}
 	}
+
```
Akkadius, line 1216: Extra newline

zone/exp.cpp
```diff
@@ -1182,8 +1182,8 @@ void Raid::SplitExp(const uint64 exp, Mob* other) {
 
 	if (RuleB(Character, EnableRaidEXPModifier)) {
 		raid_experience = static_cast<uint64>(
-			static_cast<float>(raid_experience) *
-			(1.0f - RuleR(Character, RaidExpMultiplier))
+				static_cast<float>(raid_experience) *
```
Akkadius, line 1185: This seems to be indented in one more level than needbe

zone/exp.cpp
```diff
@@ -1213,6 +1213,12 @@ void Raid::SplitExp(const uint64 exp, Mob* other) {
 			}
 		}
 	}
+
+
+	raid_experience = static_cast<uint64>(
```
Akkadius, line 1218: Can we put this on a single line

Valorith: Done

zone/exp.cpp
Valorith: Removed

zone/exp.cpp
Valorith: Converted to single line.

Akkadius (approved):

crotonrz:
I don't know if it's intended, and I am missing it, but this rule seems to cause raid experience to reduce to almost 0, unless you disable the Character:EnableRaidEXPModifier rule, which is default flagged as true. Keep in mind, this is EZ Server, everything is exaggerated, so the 24 AA gain is considered normal in this case. Also, this is with a 1000% experience buff, without the buff, you don't receive any experience message.

Rule flagged true (default) with no exp buff:
image

Rule flagged true (default) with 1000% exp buff:
image

Rule flagged false:
image

Valorith:
> I don't know if it's intended, and I am missing it, but this rule seems to cause raid experience to reduce to almost 0, unless you disable the Character:EnableRaidEXPModifier rule, which is default flagged as true. Keep in mind, this is EZ Server, everything is exaggerated, so the 24 AA gain is considered normal in this case. Also, this is with a 1000% experience buff, without the buff, you don't receive any experience message.
> 
> Rule flagged true (default) with no exp buff: image
> 
> Rule flagged true (default) with 1000% exp buff: image
> 
> Rule flagged false with 1000% exp buff: image

The issue here is the way that the legacy rule (`Character:EnableRaidEXPModifier`) is programmed. That rule for some unknown reason is subtracting the rule value from 1.0. So if you have a rule value of 1, it multiplies all xp by 0 (resulting in no xp). The legacy rule is not intuitive at all, which is why I created this PR in the first place. 

My recommendation is to set `Character:EnableRaidEXPModifier` to false and just use `Character:FinalRaidExpMultiplier` as a flat modifier.

PR #3556, opened 2023-08-20
[Expansions] Expansion settings tweaks

Akkadius:
* Adds rule `World:CharacterSelectExpansionSettings` to be able to override the server side bitmask for just character select
* Cleanup how ExpansionSettings gets loaded when client side expansion settings are not chosen
* Notify GM's that they are bypassing expansion checks when zoning
* Notify players when AA's are reloaded
* Reload player expansion settings when AA's are reloaded


zone/entity.cpp
```diff
@@ -5647,6 +5647,8 @@ void EntityList::StopMobAI()
 
 void EntityList::SendAlternateAdvancementStats() {
 	for (auto &c : client_list) {
+		c.second->Message(0, "Reloading AA");
```
Kinglykrab, line 5650: `Chat::White`.

Akkadius: Good catch, that was a Copilot suggestion :)

PR #3557, opened 2023-08-20
[Feature] Change #reload zone to reload zone headers globally.

Kinglykrab:
# Notes
- This allows operators to globally reload zone headers without going to every zone individually to do so.


zone/worldserver.cpp
```diff
@@ -191,6 +191,8 @@ void WorldServer::HandleMessage(uint16 opcode, const EQ::Net::Packet &p)
 	ServerPacket tpack(opcode, p);
 	ServerPacket *pack = &tpack;
 
+	bool reloaded_zones = false;
```
Akkadius, line 194: Why is this bool necessary?

Akkadius (approved):

PR #3574, opened 2023-08-31
[Messages] Remove duplicate message on tracking begin

noudess:
When you click on a target and hit "tracking", the you begin to track <target> message is coming out twice.

Removing this from the server results in one and only one message on both Titanium and RoF2.  

Any reason not to merge this?  Other test I should do?


Kinglykrab:
Looks good to me. I don’t have any client but ROF2 to test on.

fryguy503 (approved):
lgtm - running it since pr was pushed with no complaints.

Issue #2498, opened 2022-10-22
/split and /autosplit are not live like


mackal:
On live, the coin split feature splits each coin separately. Currently we evenly split the total amount and roll over the left overs into the next domination. This isn't how live works. All left over coins on live are just given back to the splitter.

Ex.

Splitting 4 plat, 1 gold in a group of 3. The other 2 members will each get 1 plat and the splitter will get 2 plat 1 gold. Currently we will make that 1 leftover plat 10 gold and continue the sharing, this isn't how live works.

mackal:
It appears /autosplit gives the left overs to someone random in the group

PR #3583, opened 2023-09-07
[Feature] Cleanup Group Split Money Logic

crdunwel:
Implements changes suggested in https://github.com/EQEmu/Server/issues/2498

### DIFF
--- a/zone/groups.cpp
+++ b/zone/groups.cpp
@@ -9,7 +9,7 @@
 	but WITHOUT ANY WARRANTY except by those people which sell it, which
 	are required to give you total support for your newly bought product;
 	without even the implied warranty of MERCHANTABILITY or FITNESS FOR
-	A PARTICULAR PURPOSE. See the GNU General Public License for more details.					
+	A PARTICULAR PURPOSE. See the GNU General Public License for more details.
 
 	You should have received a copy of the GNU General Public License
 	along with this program; if not, write to the Free Software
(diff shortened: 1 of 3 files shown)


zone/groups.cpp
```diff
@@ -114,91 +114,85 @@ Group::~Group()
...
+    uint8 splitter_index = MAX_GROUP_MEMBERS
+    // Count the number of actual players in the group.
+    for (uint32 i = 0; i < MAX_GROUP_MEMBERS; i++) {
+        if (members[i] && members[i]->IsClient()) {
+            member_count++;
+            if (splitter && members[i]->CastToClient()->CharacterID() == splitter->CharacterID()) {
+                splitter_index = i;
+            }
+        }
+    }
```
Kinglykrab, line 132: Can do this instead.
```cpp
for (const auto& m : members) {
	if (m && m->IsClient()) {
		member_count;
	}
}
```

Then just use the `splitter` Client that you're passing to the method instead of using `splitter_index`.

zone/groups.cpp
```diff
@@ -114,91 +114,85 @@ Group::~Group()
...
+    uint32 silver_remainder = silver % member_count;
+
+    uint32 gold_split = gold / member_count;
+    uint32 gold_remainder = gold % member_count;
+
+    uint32 platinum_split = platinum / member_count;
+    uint32 platinum_remainder = platinum % member_count;
+
+    // Loop through the group members to split the coins.
+    for (uint32 i = 0; i < MAX_GROUP_MEMBERS; i++) {
```
Kinglykrab, line 157: Same here.
```cpp
for (const auto& m : members) {
	if (m && m->IsClient()) {
```

zone/groups.cpp
```diff
@@ -114,91 +114,85 @@ Group::~Group()
...
+        if (members[i] && members[i]->IsClient()) {
+            Client* member_client = members[i]->CastToClient();
+
+            uint32 log_copper = copper_split;
+            uint32 log_silver = silver_split;
+            uint32 log_gold = gold_split;
+            uint32 log_platinum = platinum_split;
+
+            if (i == splitter_index) {
```
Kinglykrab, line 166: Using above logic just do this.
```cpp
if (m == splitter) {
```

Kinglykrab:
Couple comments. May way to validate `splitter` as well.

zone/groups.cpp
```diff
@@ -114,91 +114,85 @@ Group::~Group()
...
-			);
-		}
-	}
+    // Return early if no money to split.
+    if (!copper && !silver && !gold && !platinum) {
+        return;
+    }
+
+    uint8 member_count = 0;
+    uint8 splitter_index = MAX_GROUP_MEMBERS
```
Kinglykrab, line 123: Missing a `;` so compilation is failing, but will be fixed if we just don't use `splitter_index` anymore.

zone/groups.cpp
```diff
@@ -114,91 +114,85 @@ Group::~Group()
...
-				).c_str()
-			);
-		}
-	}
+    // Return early if no money to split.
+    if (!copper && !silver && !gold && !platinum) {
+        return;
+    }
+
+    uint8 member_count = 0;
```
Kinglykrab, line 122:
```cpp
uint8 member_count = GroupCount();```

Comment (crdunwel) [reply zone/groups.cpp:122]: Looking at `GroupCount()` it doesn't appear to count only clients? Don't think we want to split with non-clients here?

File: zone/groups.cpp (hunk 4 above)
Comment (crdunwel) [reply zone/groups.cpp:123]: woops

File: zone/groups.cpp (hunk 5 above)
Comment (Kinglykrab) [reply zone/groups.cpp:122]: We can change it to have a parameter to only count clients if we want.

Comment (crdunwel) [reply zone/groups.cpp:126]: Think we need to also check if the splitter is in the group too? Just pushed changes.

File: zone/groups.cpp [hunk 6]
```diff
@@ -114,91 +114,90 @@ Group::~Group()
...
+    uint32 gold_split = gold / member_count;
+    uint32 gold_remainder = gold % member_count;
+
+    uint32 platinum_split = platinum / member_count;
+    uint32 platinum_remainder = platinum % member_count;
+
+    // Loop through the group members to split the coins.
+    for (const auto& m : members) {
+        if (m && m->IsClient()) {
+            Client* member_client = members[i]->CastToClient();
```
Kinglykrab, line 164: `m`

zone/groups.cpp
Kinglykrab: Yeah, probably not a bad idea.

zone/groups.cpp
crdunwel: derp

crdunwel:
yay passing build!

Akkadius (approved):

Issue #2, opened 2013-02-17
[Underfoot]: Unable to interact with a trader/buyer if they started trader/buyer mode while you were in the zone


ghost:
http://www.peqtgc.com/phpBB3/viewtopic.php?f=17&t=13833

Currently one cannot interact with a trader/buyer in the bazaar if they started their trading while you were in the zone. /baz search window only shows active traders from before you zoned (trader tab), though a general search does show from all traders.
As is, if you wish to interact one needs to log or zone 1st.

PR #3585, opened 2023-09-08
[Feature] Trade Changes with NPCs

crdunwel:
https://github.com/EQEmu/Server/issues/932

This should address 1 and 3

### DIFF
--- a/zone/trading.cpp
+++ b/zone/trading.cpp
@@ -836,7 +836,7 @@ void Client::FinishTrade(Mob* tradingWith, bool finalizer, void* event_entry, st
 					);
 				}
                 // Can not complete trade with a mob when you are on their hate list
-                else if (tradingWith->hate_list.IsEntOnHateList(this)) {
+                else if (tradingWith->CheckAggro(this)) {
                     tradingWith->SayString(TRADE_BACK, GetCleanName());
                     PushItemOnCursor(*inst, true);
                 }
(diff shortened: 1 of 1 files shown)


zone/trading.cpp
```diff
@@ -839,6 +835,11 @@ void Client::FinishTrade(Mob* tradingWith, bool finalizer, void* event_entry, st
 						true
 					);
 				}
+                // Can not complete trade with a mob when you are on their hate list
```
Akkadius, line 838: Indenting is off with the rest of the blocks

Akkadius:
Can anyone validate live intended functionality here? @noudess seems like you'd know this one

noudess:
The not turning to face the trading PC looks good.

Two things about the not competing trade when NPC is aggro:

1) Do we want a GM to be able to do it?  I have no opinion either way.

2) I don't think this PR has that code in the correct place.  Right now I think the code you added is an else case to a large collection of trades that will just happen.

Right now you have:

if (GM | lots of othe cases)
  {
  do trade
  }
elseif (tradingWith->CheckAggro(this))
  {
  dont do trade
  }

They check all cases in the 1st if and explain them all in the elsifs.  You'd have to add your case to the 1st if and leave your code below, or maybe just check and reject up top.

Agree or an I reading it incorrectly?

noudess:
I like this change, and I might be able to fix  #2 in the issue mentioned as well.  Any response to my question above?  I could download a patch and test but I figured you'd either set me straight or make the corrections.

Akkadius:
Bump

Akkadius:
Closing until follow up from contributor

PR #3590, opened 2023-09-11
[Quest API] Add Caster ID Parameter to FindBuff in Perl/Lua

Kinglykrab:
# Perl
- Add `$mob->FindBuff(spell_id, caster_id)`.

# Lua
- Add `mob:FindBuff(spell_id, caster_id)`.

# Notes
- Allows operators to check if the spell ID is cast by a specific entity ID.
- We don't use `Mob*` reference here since the mob may have died, left zone, etc.

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -4402,14 +4402,20 @@ std::vector<uint16> Mob::GetBuffSpellIDs()
 	return l;
 }
 
-bool Mob::FindBuff(uint16 spell_id)
+bool Mob::FindBuff(uint16 spell_id, uint16 caster_id)
 {
-	uint32 buff_count = GetMaxTotalSlots();
+	const int buff_count = GetMaxTotalSlots();
 	for (int buff_slot = 0; buff_slot < buff_count; buff_slot++) {
-		auto current_spell_id = buffs[buff_slot].spellid;
+		const uint16 current_spell_id = buffs[buff_slot].spellid;
 		if (
-			IsValidSpell(current_spell_id) &&
-			current_spell_id == spell_id
+			(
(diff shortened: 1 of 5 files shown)


zone/spells.cpp
```diff
@@ -4402,14 +4402,18 @@ std::vector<uint16> Mob::GetBuffSpellIDs()
 	return l;
 }
 
-bool Mob::FindBuff(uint16 spell_id)
+bool Mob::FindBuff(uint16 spell_id, uint16 caster_id)
 {
-	uint32 buff_count = GetMaxTotalSlots();
+	const int buff_count = GetMaxTotalSlots();
 	for (int buff_slot = 0; buff_slot < buff_count; buff_slot++) {
-		auto current_spell_id = buffs[buff_slot].spellid;
```
Akkadius, line 4409: Thank you for translating these to simple types.

zone/spells.cpp
```diff
@@ -4402,14 +4402,18 @@ std::vector<uint16> Mob::GetBuffSpellIDs()
...
+		const uint16 current_spell_id = buffs[buff_slot].spellid;
 		if (
 			IsValidSpell(current_spell_id) &&
-			current_spell_id == spell_id
+			current_spell_id == spell_id &&
+			(
+				!caster_id ||
```
Akkadius, line 4414: Are we sure on this added check? This seems like it could unintentionally break things

Akkadius: Looks like !caster_id covers the default cause added - so it should be fine

zone/spells.cpp
Kinglykrab: No problem.

PR #3591, opened 2023-09-13
[Bug Fix] Fix Killed XYZH support in EVENT_DEATH in Perl.

Kinglykrab:
# Notes
- Fixes XYZH exports in Perl.


Akkadius:
The return of killer variables

Akkadius (approved):

PR #3595, opened 2023-09-16
[Bug Fix] NPC Faction War prevention.

fryguy503:
This should assist with prevention of NPC Faction waring. I have been using this code on my server for over 5 years with no coredumps or any noted bugs. This was brought up as users have reported a few events where they can trigger NPC's to fight each other.

### DIFF
--- a/zone/spell_effects.cpp
+++ b/zone/spell_effects.cpp
@@ -3889,11 +3889,12 @@ void Mob::DoBuffTic(const Buffs_Struct &buff, int slot, Mob *caster)
 
 				if (IsDetrimentalSpell(buff.spellid)) {
 					if (caster->IsClient()) {
-						if (!caster->CastToClient()->GetFeigned())
+						if (!caster->CastToClient()->GetFeigned()) {
 							AddToHateList(caster, -effect_value);
-					} else if (!IsClient()) // Allow NPC's to generate hate if casted on other
-								// NPC's.
+						}
+					} else if (!IsClient()) { // Allow NPC's to generate hate if casted on other NPC's
 						AddToHateList(caster, -effect_value);
+					}
 				}
 
 				effect_value = caster->GetActDoTDamage(buff.spellid, effect_value, this);
@@ -3994,6 +3995,12 @@ void Mob::DoBuffTic(const Buffs_Struct &buff, int slot, Mob *caster)
 		case SE_Charm: {
 			if (!caster || !PassCharismaCheck(caster, buff.spellid)) {
 				BuffFadeByEffect(SE_Charm);
+				
+				// Remove from hate list of any NPC's hate list and remove all NPCs this hate list
+				if (IsNPC()) {
+					entity_list.RemoveFromHateLists(this);
+					WipeHateList(true);
+				}
 			}
 
 			break;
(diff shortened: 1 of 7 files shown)


fryguy503:
> Have not yet tested. Looks fine at first view. Should be tested against wars that we wan't to happen. Thinking owlbears/sonic wolves/grimlings, bandits/cargo clockwork, etc.

I have not seen any issues with forced npc to npc combat (e.g. Hollowshade war) on my server. Just helps prevent player initiated but undesired npc wars.

zone/spells.cpp
```diff
@@ -4187,41 +4187,22 @@ bool Mob::SpellOnTarget(
...
-			spelltar->SetHateAmountOnEnt(this, std::max(newhate, static_cast<int64>(1)));
-		}
-	} else if (IsBeneficialSpell(spell_id) && !IsSummonPCSpell(spell_id)) {
-		if (this != spelltar && IsClient()){
-			if (spelltar->IsClient()) {
-				CastToClient()->UpdateRestTimer(spelltar->CastToClient()->GetRestTimer());
-			} else if (spelltar->IsPet()) {
-				auto* owner = spelltar->GetOwner();
-				if (owner && owner != this && owner->IsClient()) {
-					CastToClient()->UpdateRestTimer(owner->CastToClient()->GetRestTimer());
```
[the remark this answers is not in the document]
fryguy503: Good Catch, I pulled from my early (Pre Rest Timer) code. Let me add that in.

zone/spell_effects.cpp
```diff
@@ -3994,6 +3995,12 @@ void Mob::DoBuffTic(const Buffs_Struct &buff, int slot, Mob *caster)
 		case SE_Charm: {
 			if (!caster || !PassCharismaCheck(caster, buff.spellid)) {
 				BuffFadeByEffect(SE_Charm);
+				
+				// Remove from hate list of any NPC's hate list and remove all NPCs this hate list
+				if (IsNPC()) {
+					entity_list.RemoveFromHateLists(this);
+					WipeHateList(true);
+				}
```
hgtw, line 4003: I don't know the spell code at all but would it make more sense to handle this in the `SE_Charm` handler inside `BuffFadeBySlot`?

There's also already some hate list stuff there (https://github.com/EQEmu/Server/blame/1212ccefef7132221425201bc47ff53d83af1468/zone/spell_effects.cpp#L4422) added in https://github.com/EQEmu/Server/commit/af5cfb9bed7c4fdfbef3dcf42bb39b349cfb3888 so not sure if that is newer than the base for this code and something there might conflict

zone/spells.cpp
```diff
@@ -4187,41 +4187,39 @@ bool Mob::SpellOnTarget(
...
+	if (!(spelltar->IsNPC() && IsNPC() && !(IsPet() && GetOwner()->IsClient()))) {
```
hgtw, line 4190: If I'm understanding this right this is preventing hate being added if an npc (that's not a client charmed pet) casts on another npc? The downside for forced npc vs npc fights is healers or caster npcs won't accumulate hate but guessing this is the main cause of unintentional faction wars?

fryguy503: Correct. inherent npc to npc aggro (Conflicting Factions) should still continue to work but will not accumulate hate. In my limited tested years ago, the mobs would always target a real player on hate over a npc, so this didn't seem like an issue with this.

We just want to ensure that npc will not accidentally get onto the hatelist of another "friendly" or "indifferent" faction.

Prob not the most elegant way to prevent it, but it has worked well for me.

Akkadius:
We really need to get this to the finish line

fryguy503:
What is pending on this? The only question left was weather the DoBuffTic was the best location. I know it has been the most consistent location over the years.

hgtw (approved):

Issue #1523, opened 2021-09-08
A well timed NPC summon can pull a player across zones


Isaaru:
> so im looking at mob summoning since there seems to be an occasional race condition between gate/summon where the gate triggers a MovePC() but the client doesnt communicate the character has left the zone fast enough so the mob triggers a MovePC() via summon and summons the toon back across server lines:

https://github.com/EQEmu/Server/blob/master/zone/mob.cpp#L2786-L2804

Original Post: https://discord.com/channels/212663220849213441/213021985465303040/673588922298794002

Isaaru:
![image](https://github.com/EQEmu/Server/assets/19696544/bfe9c569-8964-4bd5-ac65-3e53253dcf09)

I know we're having issues replicating this, but it is a thing still.

nytmyr:
[...]

PR #3600, opened 2023-09-19
[Bug Fix] Add check to prevent npcs summoning across zones.

Valorith:
This is an attempt at fixing https://github.com/EQEmu/Server/issues/1523

Edit: Testing complete. Through exhaustive testing, I can confirm that adding this check does not break anything in this functionality. I cannot however 100% confirm that it resolves the underlying bug as I cannot seem to reproduce it.  My recommendation is to merge this and see if it does the trick for those that are experiencing the original bug. Worst case scenario, we are no worse off than we were prior to this check being added if it misses the mark on fixing the bug.

### DIFF
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -4786,6 +4786,11 @@ bool Mob::HateSummon() {
 			new_pos = target->TryMoveAlong(new_pos, 5.0f, angle);
 
 			if (target->IsClient()) {
+				auto target_zone_id = target->GetZoneID();
+				auto summoner_zone_id = zone->GetZoneID();
+				if (!(target_zone_id == summoner_zone_id)) { // Ensure the target is in the same zone as the summoner
+					return false;
+				}
 				target->CastToClient()->MovePC(
 					zone->GetZoneID(),
 					zone->GetInstanceID(),
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -4788,7 +4788,7 @@ bool Mob::HateSummon() {
 			if (target->IsClient()) {
 				auto target_zone_id = target->GetZoneID();
 				auto summoner_zone_id = zone->GetZoneID();
-				if ((target_zone_id != summoner_zone_id)) { // Ensure the target is in the same zone as the summoner
+				if (target_zone_id != summoner_zone_id) { // Ensure the target is in the same zone as the summoner
 					return false;
 				}
 				target->CastToClient()->MovePC(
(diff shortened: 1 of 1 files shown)


Valorith:
> I think
> 
> if (target_zone_id != summoner_zone_id)
> 
> is a little clearer.
> 
> What was happening? the AI was summoning as a clienrt zoned?

Agreed. Updated.

noudess:
This seems straight forward as a fix so I'd tend to improve it, but if I am reading chatr channels right, you're having troble reproducing it now, even without the fix?

Valorith:
> This seems straight forward as a fix so I'd tend to approve it, but if I am reading chatr channels right, you're having troble reproducing it now, even without the fix?

Correct. I am struggling to replicate the issue on current release binaries as to validate that it is fixed by this change. Multiple people have however confirmed that they have experienced this bug across multiple servers. 

I can, however, confirm that adding this check does not appear to cause any issues in my testing and it will likely resolve the reported bug.

nytmyr:
My only suggestion to possibly aide in reproducing is using a VPN to create a delay. It's a super rare bug but it has happened over the years.

Valorith:
> My only suggestion to possibly aide in reproducing is using a VPN to create a delay. It's a super rare bug but it has happened over the years.

This isnt something within my immediate capabilities. Others are welcome to give that a try if they like.

joligario (approved):
Extra parentheses left over.

No issues from my perspective. Just an added protection.

Valorith:
> Extra parentheses left over.
> 
> No issues from my perspective. Just an added protection.

Removed the extra parentheses, thanks.

Isaaru:
> This seems straight forward as a fix so I'd tend to approve it, but if I am reading chatr channels right, you're having troble reproducing it now, even without the fix?

It's not super common and as @nytmyr mentioned it may be delay/ping related. I was unable to reproduce after an hour of straight testing on our test box but have had it happen several times over the years on our live server (which I have less latency on). It's rare for sure and I wasn't even aware of it until someone mentioned it to me and then about a week ago it happened to me again so I brought it back up. Trust seems to be seeing the issue a bit more than we are though.

Akkadius:
> Since I have added this, I have been seeing random zone deadlocks. I rolled back each commit one by one and so far this is the only one it has been happening from.
> 
> I will try to get a zone perf of it to confirm.

Have you been able to confirm this is the case

Akkadius:
Closing this until someone actually tests this on a live server for a week and prove it’s not causing deadlocks

PR #3601, opened 2023-09-19
[Swarm Pets] Fix pet names to use '_' instead of ' '

noudess:
Swarm pets were created with a space in the name instead of an underscore.

Thereafter, all uses of GetCleanName() resulted in : PlayerName`spet in messages.


Kinglykrab (approved):
Awesome.

PR #3604, opened 2023-09-29
[Logs] Change pathing log messages from `Error` to `Pathing`.

joligario:
Call to `Pause` and `Resume` wandering happen with quest subs and shouldn't be considered errors.

### DIFF
--- a/zone/waypoints.cpp
+++ b/zone/waypoints.cpp
@@ -157,7 +157,7 @@ void NPC::ResumeWandering()
 		}
 		else
 		{
-			LogError("NPC not paused - can't resume wandering: [{}]", (unsigned long)GetNPCTypeID());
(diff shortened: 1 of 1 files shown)


fryguy503 (approved):
lgtm

noudess:
I tend to agree, but the "errors' on when those quest subs are used on mobs without grids - so techically at least warnings?

fryguy503:
I always considered Errors and Warnings related to standard infrastructure and functionality of the core server. Where as pathing would not fall into that bucket but rather a pathing specific log.

joligario:
The way I was looking at it was that if it should be able to pause/resume but can't - that's an error. If it shouldn't pause, but every call still goes forward, then it is more info.

PR #3613, opened 2023-10-07
[Feature] Add Extra Kick Classes

Kinglykrab:
# Notes
- Allows operators to add extra classes to the "Kick" skill.
- Without this only Warrior, Ranger, Monk, Beastlord, and Berserker could kick.

### DIFF
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -530,7 +530,7 @@ RULE_BOOL(Combat, BackstabIgnoresElemental, false, "Enable or disable Elemental 
 RULE_BOOL(Combat, BackstabIgnoresBane, false, "Enable or disable Bane weapon damage affecting backstab damage, false by default.")
 RULE_BOOL(Combat, SummonMeleeRange, true, "Enable or disable summoning of a player when already in melee range of the summoner.")
 RULE_BOOL(Combat, WaterMatchRequiredForAutoFireLoS, true, "Enable/Disable the requirement of both the attacker/victim being both in or out of water for AutoFire LoS to pass.")
-RULE_INT(Combat, ExtraAllowedKickClasses, 0, "Bitmask for allowing extra classes beyond Warrior, Ranger, Beastlord, and Berserker to kick, No Extra Classes (0) by default")
+RULE_INT(Combat, ExtraAllowedKickClassesBitmask, 0, "Bitmask for allowing extra classes beyond Warrior, Ranger, Beastlord, and Berserker to kick, No Extra Classes (0) by default")
 RULE_CATEGORY_END()
 
 RULE_CATEGORY(NPC)
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -530,6 +530,7 @@ RULE_BOOL(Combat, BackstabIgnoresElemental, false, "Enable or disable Elemental 
 RULE_BOOL(Combat, BackstabIgnoresBane, false, "Enable or disable Bane weapon damage affecting backstab damage, false by default.")
 RULE_BOOL(Combat, SummonMeleeRange, true, "Enable or disable summoning of a player when already in melee range of the summoner.")
 RULE_BOOL(Combat, WaterMatchRequiredForAutoFireLoS, true, "Enable/Disable the requirement of both the attacker/victim being both in or out of water for AutoFire LoS to pass.")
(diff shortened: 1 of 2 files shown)


zone/special_attacks.cpp
```diff
@@ -324,146 +337,193 @@ void Client::OPCombatAbility(const CombatAbility_Struct *ca_atk)
...
+		reuse_time = FrenzyReuseTime - 1 - skill_reduction;
+		reuse_time = (reuse_time * haste_modifier) / 100;
 
-		auto primary_in_use = GetInv().GetItem(EQ::invslot::slotPrimary);
+		const EQ::ItemInstance* primary_in_use = GetInv().GetItem(EQ::invslot::slotPrimary);
 		if (primary_in_use && GetWeaponDamage(GetTarget(), primary_in_use) <= 0) {
 			max_dmg = DMG_INVULNERABLE;
 		}
 
-		while (AtkRounds > 0) {
```
Akkadius, line 410: Negative values can technically return true

common/ruletypes.h
```diff
@@ -530,6 +530,7 @@ RULE_BOOL(Combat, BackstabIgnoresElemental, false, "Enable or disable Elemental
 RULE_BOOL(Combat, BackstabIgnoresBane, false, "Enable or disable Bane weapon damage affecting backstab damage, false by default.")
 RULE_BOOL(Combat, SummonMeleeRange, true, "Enable or disable summoning of a player when already in melee range of the summoner.")
 RULE_BOOL(Combat, WaterMatchRequiredForAutoFireLoS, true, "Enable/Disable the requirement of both the attacker/victim being both in or out of water for AutoFire LoS to pass.")
+RULE_INT(Combat, ExtraAllowedKickClasses, 0, "Bitmask for allowing extra classes beyond Warrior, Ranger, Beastlord, and Berserker to kick, No Extra Classes (0) by default")
```
Akkadius, line 533: Probably good to put Bitmask in the name `ExtraAllowedKickClassesBitmask`

zone/special_attacks.cpp
```diff
@@ -324,146 +337,193 @@ void Client::OPCombatAbility(const CombatAbility_Struct *ca_atk)
...
+			} else {
+				hate_override = damage = GetBaseSkillDamage(EQ::skills::SkillKick, GetTarget());
+			}
+
+			reuse_time = KickReuseTime - 1 - skill_reduction;
+			DoSpecialAttackDamage(GetTarget(), EQ::skills::SkillKick, damage, 0, hate_override, reuse_time);
 
-			ReuseTime = KickReuseTime - 1 - skill_reduction;
-			DoSpecialAttackDamage(GetTarget(), EQ::skills::SkillKick, dmg, 0, ht, ReuseTime);
+			goto send_reuse_time;
```
Akkadius, line 492: We really shouldn't be using the `goto` pattern in the project or even encourage it.

"NOTE − Use of goto statement is highly discouraged because it makes difficult to trace the control flow of a program, making the program hard to understand and hard to modify. Any program that uses a goto can be rewritten so that it doesn't need the goto."

**Reason to Avoid goto Statement**

The goto statement gives the power to jump to any part of a program but, makes the logic of the program complex and tangled.

In modern programming, the goto statement is considered a harmful construct and a bad programming practice.

zone/special_attacks.cpp
```diff
@@ -324,146 +337,193 @@ void Client::OPCombatAbility(const CombatAbility_Struct *ca_atk)
...
 				EQ::skills::SkillFlyingKick,
 				EQ::skills::SkillDragonPunch,
 				EQ::skills::SkillEagleStrike,
 				EQ::skills::SkillTigerClaw,
 				EQ::skills::SkillRoundKick
 			};
+
 			int extra = 0;
 			// always 1/4 of the double attack chance, 25% at rank 5 (100/4)
-			while (wuchance > 0) {
```
Akkadius, line 458: Negative values can be true

zone/special_attacks.cpp
```diff
@@ -474,37 +534,44 @@ void Client::OPCombatAbility(const CombatAbility_Struct *ca_atk)
...
+		goto send_reuse_time;
 	}
 
-	ReuseTime = (ReuseTime * HasteMod) / 100;
-	if (ReuseTime > 0) {
-		p_timers.Start(timer, ReuseTime);
+	reuse_time = 9 - skill_reduction;
+
+send_reuse_time:
+	reuse_time = (reuse_time * haste_modifier) / 100;
```
Akkadius, line 571: Should clamp this value and make sure it doesn't go negative

Akkadius (approved):

PR #3615, opened 2023-10-07
[Bots] Adjust Bot Movement Speed

Kinglykrab:
# Notes
- Bots were having a hard time keeping up with players.
- Part of this was due to using walk until a certain distance.
- Another part was their ctor only having `0.7f` run speed versus our Mob sanity check of `1.25f`.
- We check movement stuff now before idle checks so bots are more likely to start moving earlier.

### DIFF
--- a/zone/bot.cpp
+++ b/zone/bot.cpp
@@ -651,7 +651,7 @@ NPCType *Bot::FillNPCTypeStruct(
 	n->current_hp = hp;
 	n->max_hp = hp;
 	n->size = size;
-	n->runspeed = 0.7f;
+	n->runspeed = 1.25f;
 	n->gender = gender;
 	n->race = botRace;
 	n->class_ = botClass;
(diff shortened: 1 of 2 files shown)


zone/bot.cpp
```diff
@@ -2263,23 +2263,11 @@ bool Bot::TryNonCombatMovementChecks(Client* bot_owner, const Mob* follow_mob, g
...
-					running = false;
-				}
-
-				if (running) {
-					RunTo(Goal.x, Goal.y, Goal.z);
-				}
-				else {
-					WalkTo(Goal.x, Goal.y, Goal.z);
-				}
+				RunTo(Goal.x, Goal.y, Goal.z);
```
Akkadius, line 2270: Bots shouldn't always run, do you have a video of the new behavior change showing walk/run transitions?

Kinglykrab: I don’t have a video of the transition. Not having them run always means they have an even larger chance to lag behind because they take a second to start moving once owner moves then they have to wait for a certain distance to start running.

Psycho-EQ: 99.999999999% of the time a player has [ Run ] toggled on. When a player moves they switch from a standing position to run. There is no walk transition between standing still and running. Bots should also follow this same rule of movement.

Yes, while it does look more visually graceful for bots to transition from standing still to running and running to stopping by using walk it just creates more distance between the player and the bots due to the distance they need to walk instead of run.

The movement speed increase should help overall. Having walk transitions just adds distance between the player and the bots.

Akkadius: The transition part of the code is clearly broke which is why there’s a problem to begin with. It’d be nice to fix the root of the problem and have the graceful transition rather than force run the entire time. What this does is better than the current scenario.

Akkadius (approved):

PR #3620, opened 2023-10-12
[Quest API] Add target ID and spell exports to events

Kinglykrab:
# Notes
- Add `$spell` export to `EVENT_CAST`, `EVENT_CAST_BEGIN`, `EVENT_CAST_BEGIN`, `EVENT_ITEM_CLICK`, `EVENT_ITEM_CLICK_CAST`, `EVENT_ITEM_CLICK_CLIENT`, `EVENT_ITEM_CLICK_CAST_CLIENT`, `EVENT_SPELL_EFFECT_BUFF_TIC_BOT`, `EVENT_SPELL_EFFECT_BUFF_TIC_CLIENT`, `EVENT_SPELL_EFFECT_BUFF_TIC_NPC`, `EVENT_SPELL_EFFECT_BOT`, `EVENT_SPELL_EFFECT_CLIENT`, `EVENT_SPELL_EFFECT_NPC`, `EVENT_SPELL_FADE`, `EVENT_DEATH`, `EVENT_DEATH_COMPLETE`, `EVENT_DEATH_ZONE`, `EVENT_DAMAGE_GIVEN`, and `EVENT_DAMAGE_TAKEN` in Perl.
- Add `$target_id`and `$target` exports to `EVENT_CAST`, `EVENT_CAST_BEGIN`, and `EVENT_CAST_ON` in Perl.
- Add `e.target_id` and `e.target` exports to `EVENT_CAST`, `EVENT_CAST_BEGIN`, and `EVENT_CAST_ON` in Lua.

### DIFF
--- a/zone/embparser.cpp
+++ b/zone/embparser.cpp
@@ -1560,6 +1560,7 @@ void PerlembParser::ExportEventVariables(
 			if (extra_pointers && extra_pointers->size() == 1) {
 				ExportVar(package_name.c_str(), "target", "Mob", std::any_cast<Mob*>(extra_pointers->at(0)));
 			}
+
 			break;
 		}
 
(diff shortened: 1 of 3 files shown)


hgtw:
Think it would be better if target was exported as a Mob object instead of just the target id so quests could just access `e.target:Foo()`

Kinglykrab:
> Think it would be better if target was exported as a Mob object instead of just the target id so quests could just access `e.target:Foo()`

I'll do that. I'll just have to validate the target before export so it doesn't cause crashes.

hgtw (approved):
Could probably also just revert the target_id export since it's redundant now with the full target object

Kinglykrab:
> Could probably also just revert the target_id export since it's redundant now with the full target object

Figured I could leave it in case people want the target ID for some specific scripting functionality or incase `$target` is invalid once they cast the spell and it kills it or something.

PR #3621, opened 2023-10-12
[Scripts] Update 13th Floor script for legacy research tome bagtypes

joligario:
Legacy research tomes have the wrong bagtype (BACKPACK) blocking the clients from showing the combine button.


Kinglykrab (approved):

Akkadius (approved):

Akkadius:
We should check and diff data imported versus what’s in our table today to see if we’ve manually made other changes that we need to account for post import

joligario:
> We should check and diff data imported versus what’s in our table today to see if we’ve manually made other changes that we need to account for post import

Agreed.

PR #3644, opened 2023-10-21
[Bug Fix] Hotfix command without hotfix name

joligario:
If no hotfix name is provided, the hotfix command won't need the empty string.

Seen in http://spire.akkadius.com/dev/release/22.28.0?id=13585


joligario:
Although I guess it might be better that if they don't provide a hotfix name, we create a random generic hotfix name

Kinglykrab (approved):

PR #3648, opened 2023-10-22
[Perl] Reload perl quests on zone bootup

hgtw:
Perl wasn't implementing quest interface's Init which is called by Zone::Bootup. This should fix zone's that were in standby not using the latest version of perl plugin scripts.

This needs tested to make sure it doesn't break something. This doesn't touch lua but I'm not sure why https://github.com/EQEmu/Server/pull/3333/files would have broken encounters since lua parser already implements Init and calls ReloadQuests after zone init.

### DIFF
--- a/zone/embparser.cpp
+++ b/zone/embparser.cpp
@@ -205,6 +205,11 @@ PerlembParser::~PerlembParser()
 	safe_delete(perl);
 }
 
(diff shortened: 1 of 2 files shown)


Kinglykrab (approved):
Fantastic.

PR #3655, opened 2023-10-24
Return key in validate external login web api call

knervous:
In order to make a handshake with a world server we would need to provide the lsid and key--the login web api can provide this to bypass the need to connect with (web)sockets from a client's perspective.


knervous:
Abandoning since this api will not be public

PR #3664, opened 2023-10-30
[Spawn] Split spawn2 enabled into its own state table

Akkadius:
**Problem** 

Currently `spawn2` is classified as a content table. A content table should be read-only from a server mutability perspective. Meaning the only thing that should be changing content tables should be a developer. Server state should be separately maintained.

On PEQ the content tables are on a read only CDN so we get errors like 

```
[QueryErr] Zone [sharvahl] [1146] [Table 'peq.spawn2' doesn't exist]
[UPDATE spawn2 SET enabled = 1 WHERE id = 32803]
```

Because `spawn2` **enabled** field database calls are set to a non-content connection while `spawn2` is classified as a content table on the content connection

**Solution**

Split out the **enabled** field from `spawn2` onto its own table called `spawn2_disabled` which inversely reads from the prior implementation of **enabled** but it is so we are not storing all `spawn2` entries as **enabled** in another table. We only need to know about the ones that are disabled otherwise we assume all spawns are enabled.

**Additional Notes**

This also implements enabled/disabled to be instance aware and @fryguy503 (Trust) owes me a nice box of pie.

### DIFF
--- a/common/database/database_update_manifest.cpp
+++ b/common/database/database_update_manifest.cpp
@@ -5011,7 +5011,7 @@ CREATE TABLE `spawn2_disabled` (
   `instance_id` int(11) DEFAULT 0,
   `disabled` smallint(11) DEFAULT 0,
   PRIMARY KEY (`id`),
-  UNIQUE KEY `spawn2_id` (`spawn2_id`)
+  UNIQUE KEY `spawn2_id` (`spawn2_id`,`instance_id`) USING BTREE
 ) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=utf8mb4;
 INSERT INTO spawn2_disabled (spawn2_id, disabled) SELECT id, 1 FROM spawn2 WHERE enabled = 0;
 ALTER TABLE `spawn2` DROP COLUMN `enabled`;
--- a/common/database/database_update_manifest.cpp
+++ b/common/database/database_update_manifest.cpp
@@ -5008,6 +5008,7 @@ CREATE TABLE `spawn2_backup_2023_10_29` LIKE `spawn2`;
 CREATE TABLE `spawn2_disabled` (
   `id` bigint(11) NOT NULL AUTO_INCREMENT,
   `spawn2_id` int(11) DEFAULT NULL,
+  `instance_id` int(11) DEFAULT 0,
   `disabled` smallint(11) DEFAULT 0,
   PRIMARY KEY (`id`),
   UNIQUE KEY `spawn2_id` (`spawn2_id`)
--- a/common/database/database_update_manifest.cpp
+++ b/common/database/database_update_manifest.cpp
@@ -4997,6 +4997,25 @@ ADD PRIMARY KEY (`id`) USING BTREE,
 ADD UNIQUE INDEX(`varname`);
 )"
 	},
+	ManifestEntry{
+		.version = 9241,
+		.description = "2023_10_29_split_spawn2_enabled.sql",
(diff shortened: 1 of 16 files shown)


Kinglykrab (approved):

zone/questmgr.cpp
```diff
@@ -359,22 +359,25 @@ Mob *QuestManager::spawn_from_spawn2(uint32 spawn2_id)
...
-	database.UpdateSpawn2Status(spawn2_id, 1);
+	database.UpdateSpawn2Status(spawn2_id, 1, zone->GetInstanceID());
 	auto pack = new ServerPacket(ServerOP_SpawnStatusChange, sizeof(ServerSpawnStatusChange_Struct));
-	ServerSpawnStatusChange_Struct* ssc = (ServerSpawnStatusChange_Struct*) pack->pBuffer;
-	ssc->id = spawn2_id;
-	ssc->new_status = 1;
+	auto *ssc = (ServerSpawnStatusChange_Struct *) pack->pBuffer;
+	ssc->id          = spawn2_id;
+	ssc->new_status  = true;
+	ssc->instance_id = zone->GetInstanceID();
```
hgtw, line 367: Not sure if any current usage of these functions are called from outside the spawn2 id's zone but this could end up using the wrong instance id if so. Otherwise since this change ties these functions to the current zone's instance then there should be no need to send a world message to other zones and the spawns could just be toggled immediately.

This isn't super relevant for this pr but I think this needs to be looked at closer and the intent of the api decided in a followup to either add an instance id overload or remove a possible unnecessary world message.

hgtw (commented):
The new table will need to purged by Database::PurgeExpiredInstances() for any instance_ids that expire

common/database/database_update_manifest.cpp
```diff
@@ -4995,6 +4995,26 @@ ADD COLUMN `id` int(11) NOT NULL AUTO_INCREMENT FIRST,
...
+		.match = "",
+		.sql = R"(
+CREATE TABLE `spawn2_backup_2023_10_29` LIKE `spawn2`;
+CREATE TABLE `spawn2_disabled` (
+  `id` bigint(11) NOT NULL AUTO_INCREMENT,
+  `spawn2_id` int(11) DEFAULT NULL,
+  `instance_id` int(11) DEFAULT 0,
+  `disabled` smallint(11) DEFAULT 0,
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `spawn2_id` (`spawn2_id`)
```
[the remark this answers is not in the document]
Akkadius: Good eye. I had this on my local schema and not updated in the migration

zone/questmgr.cpp
Akkadius: Yeah I’ll look at it closer later

Akkadius: Looking into this code, its actually really strange as it is meant only for that zone - yet it doesn't filter by zone id so you could technically end up toggling another zone's spawns.

Going to localize this to a function

Akkadius: Updated, please review result

Akkadius: Actually I reverted it entirely. After thinking about it, there was a reason for the way it was done originally. 

If you wanted to disable spawns from another zone based on events that occurred in another, the code enables you to do so. Although, the current `quest::` `eq.` variants do assume current `instance_id` which in most of the use cases today being in normal zones, this won't be a problem.

If someone needs to add functionality for cross zone targetting. The API's could be added

hgtw (approved):

PR #3667, opened 2023-10-31
[Spells] Added IsNightTime() for Dance of the Fireflies

regneq:
Should only be cast at night from 7pm to 4am be cast at night from 7pm to 4am.

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -712,12 +712,18 @@ bool Mob::DoCastingChecksZoneRestrictions(bool check_on_casting, int32 spell_id)
 		return false;
 	}
 	/*
-		Zones where you can not cast a spell that is for nighttime only
+		Zones where you can not cast a spell that is for daytime or nighttime only
 	*/
-	if (spells[spell_id].time_of_day == 2 && !zone->zone_time.IsNightTime()) {
+	if (spells[spell_id].time_of_day == SpellTimeRestrictions::Day && !zone->zone_time.IsDayTime()) {
+		MessageString(Chat::Red, CAST_DAYTIME);
+		return false;
+	}
+
+	if (spells[spell_id].time_of_day == SpellTimeRestrictions::Night && !zone->zone_time.IsNightTime()) {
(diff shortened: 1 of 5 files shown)


Kinglykrab (approved):

Kinglykrab:
Do we not need to add a spot for using the day time only logic, even if there aren't any spells using it?

zone/spells.cpp
```diff
@@ -711,7 +711,13 @@ bool Mob::DoCastingChecksZoneRestrictions(bool check_on_casting, int32 spell_id)
 		Message(Chat::Red, "You cannot cast detrimental spells here.");
 		return false;
 	}
-
+	/*
+		Zones where you can not cast a spell that is for nighttime only
+	*/
+	if (spells[spell_id].time_of_day == 2 && !zone->zone_time.IsNightTime()) {
```
Kinglykrab, line 717: Could we make an enum for this so we eliminate the magic number?

regneq: I noticed there are some spells that uses 255.  I have no idea what that is for.      1 is day and 2 is night, although there is no spells that uses day.  I guess it's just there in case.

joligario: Moon doesn’t have day or night if I remember right

regneq: Luclin doesn't show time but it still uses it.  I can test this on live really quick.

regneq: image

Kinglykrab: > I noticed there are some spells that uses 255. I have no idea what that is for. 1 is day and 2 is night, although there is no spells that uses day. I guess it's just there in case.

Was just working on the assumption maybe a server operator may want that functionality for a spell eventually.

Kinglykrab: > Yeah I could. in eq_constants?

Yes.

regneq: updated.

joligario:
Or you could !`IsNightTime`

PR #3669, opened 2023-11-04
[Feature] Add Comment to Item Data/Quest API

Kinglykrab:
# Perl
- Add `quest::getitemcomment(item_id)`.
- Add `quest::getitemlore(item_id)`.
- Add `$questitemdata->GetComment()`.

# Lua
- Add `eq.get_item_comment(item_id)`.
- Add `eq.get_item_lore(item_id)`.
- Add `item:Comment()`.

# Notes
- Added the ability for operators to pull these fields from item data without a database hit.
- Fixed a bug in `embparser_api.cpp` where `GetZoneGravity` was connected to the wrong methods.

### DIFF
--- a/common/shareddb.cpp
+++ b/common/shareddb.cpp
@@ -51,7 +51,9 @@ namespace ItemField
 #define F(x) x,
 #include "item_fieldlist.h"
 #undef F
-		updated
+		updated,
+		minstatus,
+		comment,
 	};
 }
 
@@ -965,7 +967,7 @@ void SharedDatabase::LoadItems(void *data, uint32 size, int32 items, uint32 max_
 #define F(x) "`"#x"`,"
 #include "item_fieldlist.h"
 #undef F
-		"updated FROM items ORDER BY id";
+		"updated, minstatus, comment FROM items ORDER BY id";
 	auto results = QueryDatabase(query);
 	if (!results.Success()) {
 		return;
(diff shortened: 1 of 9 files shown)


common/shareddb.cpp
```diff
@@ -51,7 +51,9 @@ namespace ItemField
 #define F(x) x,
 #include "item_fieldlist.h"
 #undef F
-		updated
+		updated,
+		minstatus,
+		comment,
```
joligario, line 56: Extra comma

zone/embparser_api.cpp
```diff
@@ -5860,8 +5870,8 @@ void perl_register_quest()
 	package.add("GetZoneFogRed", (uint8(*)(uint32))&Perl__GetZoneFogRed);
 	package.add("GetZoneFogRed", (uint8(*)(uint32, uint8))&Perl__GetZoneFogRed);
 	package.add("GetZoneFogRed", (uint8(*)(uint32, uint8, int))&Perl__GetZoneFogRed);
-	package.add("GetZoneGravity", (float(*)(uint32))&Perl__GetZoneMaximumClip);
-	package.add("GetZoneGravity", (float(*)(uint32, int))&Perl__GetZoneMaximumClip);
+	package.add("GetZoneGravity", (float(*)(uint32))&Perl__GetZoneGravity);
```
joligario, line 5873: Accidental / ninja addition?

joligario (approved):
Some minor comments

common/shareddb.cpp
Kinglykrab: Can leave it since it doesn't change functionality.

zone/embparser_api.cpp
Kinglykrab: Purposeful change. Can put it in its own pull request though.

Akkadius: It really should be split, if we revert this for example this change also goes with it

PR #3676, opened 2023-11-04
[CRASH] Fix crash on CentOS when forming a raid with PCs or BOTs

neckkola:
A crash condition was being seen in CentOS after the conversion of raid notes from char[] to std::string.  It was not seen in other linux distros or on windows though it appears that it would have eventually been seen.

Issue was duplicated on a new CentOS 7 install and was traced to the memset usage in the raid constructors and LearnMembers as it was overwriting the std::string structure.  

Added default initializers to RaidMember struct as well.

Tested with the CentOS install and on a windows install by forming a raid with a Bot, moving/rearranging groups, attacking and zoning.  Would appreciate others testing as an added layer.  OP of crash is testing in the next few days.

Thanks,
Neckkola

### DIFF
--- a/zone/raids.cpp
+++ b/zone/raids.cpp
@@ -40,7 +40,6 @@ extern WorldServer worldserver;
 Raid::Raid(uint32 raidID)
 : GroupIDConsumer(raidID)
 {
-	memset(members ,0, (sizeof(RaidMember)*MAX_RAID_MEMBERS));
 	memset(&raid_aa, 0, sizeof(RaidLeadershipAA_Struct));
 	memset(group_aa, 0, sizeof(GroupLeadershipAA_Struct) * MAX_RAID_GROUPS);
 	for (auto& gm : group_mentor) {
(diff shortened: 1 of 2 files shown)


Kinglykrab (approved):

Kinglykrab:
Need to rebase.

zone/raids.cpp
```diff
@@ -2975,3 +2973,22 @@ void Raid::SendMarkTargets(Client* c)
...
+		members[i].level           = 0;
+		members[i].main_assister   = 0;
+		members[i].main_marker     = 0;
+		members[i]._class          = 0;
+		members[i].is_bot          = false;
+		members[i].is_looter       = false;
+		members[i].is_raid_leader  = false;
+		members[i].is_raid_main_assist_one = false;
+		members[i].member          = nullptr;
+		members[i].note            = '\0';
```
hgtw, line 2991: This should use `members[i].clear()` since it's a std::string. Otherwise this will leave the size unchanged which might break anything relying on it (e.g. note.empty() would still return false)

hgtw: Nevermind I misread this as `note[0] = '\0'` so it wouldn't preserve the old size. This should still use .clear though

neckkola: Thanks.  Updated with .clear()

Aeadoin (approved):

PR #3679, opened 2023-11-06
[Quest API] Add details to Lua event dispatch errors

hgtw:
This will add the event name to assist with finding source errors that break lua events

### DIFF
--- a/zone/lua_parser.cpp
+++ b/zone/lua_parser.cpp
@@ -438,9 +438,7 @@ int LuaParser::_EventNPC(std::string package_name, QuestEventID evt, NPC* npc, M
 
 		lua_pop(L, npop);
 	} catch(std::exception &ex) {
-		std::string error = "Lua Exception: ";
-		error += std::string(ex.what());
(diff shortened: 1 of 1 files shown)


Akkadius (approved):

Kinglykrab (approved):

zone/lua_parser.cpp
```diff
@@ -532,9 +530,7 @@ int LuaParser::_EventPlayer(std::string package_name, QuestEventID evt, Client *
 
 		lua_pop(L, npop);
 	} catch(std::exception &ex) {
-		std::string error = "Lua Exception: ";
-		error += std::string(ex.what());
-		AddError(error);
+		AddError(fmt::format("Lua Exception | [{}] for Player in [{}]: {}", sub_name, package_name, ex.what()));
```
Kinglykrab, line 533: No way to add character ID/name here similar to bots?

Kinglykrab: No problem. Understood.

PR #3690, opened 2023-11-18
[GM Commands] Add `#takeplatinum`

joligario:


common/database/database_update_manifest.cpp
```diff
@@ -5018,6 +5018,17 @@ INSERT INTO spawn2_disabled (spawn2_id, disabled) SELECT id, 1 FROM spawn2 WHERE
...
 )"
 	},
+	ManifestEntry{
+		.version = 9242,
+		.description = "2023_11_18_takeplatinum_command.sql",
+		.check = "SELECT * FROM `command_settings` WHERE `command` LIKE 'takeplatinum'",
+		.condition = "empty",
+		.match = "",
+		.sql = R"(
+INSERT INTO `command_settings` VALUES ('takeplatinum', '150', '');
```
Akkadius, line 5028: These do get injected automatically

https://github.com/EQEmu/Server/blob/master/zone/command.cpp#L342

zone/gm_commands/takeplatinum.cpp
```diff
@@ -0,0 +1,57 @@
+#include "../client.h"
+
+void command_takeplatinum(Client *c, const Seperator *sep)
+{
+	int arguments = sep->argnum;
+	if (!arguments || !sep->IsNumber(1)) { //must be a number
+		c->Message(Chat::Red, "Usage: #Usage: #takeplatinum [Platinum]");
```
Kinglykrab, line 7: Double usage.

zone/gm_commands/takeplatinum.cpp
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	Client *target = c;
+	if (c->GetTarget() && c->GetTarget()->IsClient()) {
+		target = c->GetTarget()->CastToClient();
+	}
+
+	uint32 platinum = Strings::ToUnsignedInt(sep->arg[1]);
+	if (!platinum) {
+		c->Message(Chat::Red, "Usage: #Usage: #takeplatinum [Platinum]");
```
Kinglykrab, line 18: Double usage.

zone/gm_commands/takeplatinum.cpp
```diff
@@ -0,0 +1,57 @@
...
+					platinum,
+					0,
+					0,
+					0
+				),
+				c->GetTargetDescription(target)
+			).c_str()
+		);
+	}
+	else {
```
Kinglykrab, line 42: Could do a return in `if (success)` instead.

Maybe don’t need to say it’s too much as that’s implied? “You do not have/X does not have Y Platinum.”

joligario: I think a return is pointless as it is at the end of the code block.

Yes, it was implied (a hint) as that is the main (only?) case I could think that it would fail. If you try to take too much money, it wouldn't tell you. I'll go with the other message you offered.

Kinglykrab (approved):

PR #3693, opened 2023-11-18
[Cleanup] Cleanup #show currencies Command

Kinglykrab:
# Notes
- Cleans up messages.
- Adds an enum for `MoneyTypes` and `MoneySubtypes` so we're not using magic numbers for `GetMoney` anymore.
- Converts money amounts to `uint64`.

### DIFF
--- a/zone/gm_commands/show/currencies.cpp
+++ b/zone/gm_commands/show/currencies.cpp
@@ -3,7 +3,7 @@
 
 void ShowCurrencies(Client *c, const Seperator *sep)
 {
-	Client* t = c;
+	Client *t = c;
 	if (c->GetTarget() && c->GetTarget()->IsClient()) {
 		t = c->GetTarget()->CastToClient();
 	}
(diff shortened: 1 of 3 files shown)


zone/gm_commands/show/currencies.cpp
```diff
@@ -56,6 +56,16 @@ void ShowCurrencies(Client *c, const Seperator *sep)
 		has_currency = true;
 	}
 
+	const int aa_points = t->GetAAPoints();
```
joligario, line 59: Personally, I don't see AAs as a currency. Just my two cents...

Kinglykrab: I can remove that, we just didn't have a way to view them as of right now.

joligario: Good point. I recommend breaking it out but let's add that ability somewhere.

Kinglykrab: I'll make a separate command for it.

joligario (approved):

PR #3702, opened 2023-11-21
[Scripts] Update 13th Floor Import Tool

joligario:


MortimerGreenwald:
How does this work? Is there a way to launch the import tool by an admin and import missing id's to a db? Or is this simply a data dump?

joligario:
> How does this work? Is there a way to launch the import tool by an admin and import missing id's to a db? Or is this simply a data dump?

This tool imports the items from the downloaded dump into its own table, compares your existing items, then updates your table with changes and additions.

Kinglykrab (approved):

PR #3724, opened 2023-11-28
[Bug Fix] Guild Message limits

neckkola:
Reports that guild mail would fail if sending to 50 members.

Analysis indicated that the buffer holding the inbound message was set with a 512 char limit.  This truncated the message and caused the issue. 

Refactored to adjust the size of the buffer based on the inbound message size.


Kinglykrab (approved):

noudess:
You know this fixed an issue that has been in issues for 7 years?  I added it in 2016!!!

Thanks @neckkola !!

PR #3726, opened 2023-12-01
[Database] Modify `updated` column in `items` table with proper default.

joligario:
From https://mariadb.com/kb/en/datetime/, `datetime` can be ZERO_DATE (`0000-00-00` with no time component) or between `1000-01-01 00:00:00.000000` and `9999-12-31 23:59:59.999999`. Currently, `updated` is the only datetime field that is `NOT NULL` and improperly defaulted to `0000-00-00 00:00:00`. This change matches existing structures of the other `datetime` columns.


Kinglykrab:
You'll need to update the repository as well.

Kinglykrab:
https://docs.eqemu.io/developer/repositories/#code-generation-repository-generator

Kinglykrab (approved):

PR #3728, opened 2023-12-01
[GM Commands] Fix formatting of wpinfo output.

noudess:
#wpinfo output was missing a space on lines that included a delay.

Capture


Kinglykrab (approved):
Good catch.

PR #3734, opened 2023-12-03
[Quest API] Add EVENT_ALT_CURRENCY_GAIN and EVENT_ALT_CURRENCY_LOSS to Perl/Lua

Kinglykrab:
# Perl
- Add `EVENT_ALT_CURRENCY_GAIN`.
- Add `EVENT_ALT_CURRENCY_LOSS`.
- Export `$currency_id`, `$amount`, and `$total`.

# Lua
- Add `event_alt_currency_gain`.
- Add `event_alt_currency_loss`.
- Export `e.currency_id`, `e.amount`, and `e.total`.

# Notes
- Convert `int8 method` to `bool is_scripted` in `Client::AddAlternateCurrencyValue`.
- Properly utilize `is_scripted` parameter in `perl_client.cpp`.
- Allows operators to perform events on alternate currency gains/losses.


joligario:
I think LOSS implies losing, where SPENT might be a better description? Just a recommendation, may be just pedantic, but perhaps EARNED/SPENT might be better?

Kinglykrab:
> I think LOSS implies losing, where SPENT might be a better description? Just a recommendation, may be just pedantic, but perhaps EARNED/SPENT might be better?
> 
> Never used ALT Currency before. Can you also trade them?

Just following the naming convention we used for AA gain/loss and Currency gain/loss.

You can't trade the alternate currencies directly, only if you reclaim them as an item.

Akkadius (approved):

Akkadius (approved):
LGTM. Rebased for you and compiled locally

PR #3739, opened 2023-12-05
[Quest API] Add EVENT_LOOT_ADDED to Perl/Lua

Kinglykrab:
# Perl
- Add `EVENT_LOOT_ADDED`.
- Exports `$item`, `$item_id`, `$item_name`, `$item_charges`, `$augment_one`, `$augment_two`, `$augment_three`, `$augment_four`, `$augment_five`, and `$augment_six`.

# Lua
- Add `event_loot_added`.
- Exports `e.item`, `e.item_id`, `e.item_name`, `e.item_charges`, `e.augment_one`, `e.augment_two`, `e.augment_three`, `e.augment_four`, `e.augment_five`, and `e.augment_six`.

# Notes
- Allows operators to perform events when loot is added to an NPC, such as removing the loot or keeping track of it.


zone/loottables.cpp
```diff
@@ -118,39 +133,37 @@ void ZoneDatabase::AddLootTableToNPC(NPC* npc, uint32 loottable_id, ItemList* it
...
-	}
-
-	if (loot_drop->NumEntries == 0) {
-		return;
-	}
-
-	if (!content_service.DoesPassContentFiltering(loot_drop->content_flags)) {
+	const auto *lds = GetLootDrop(lootdrop_id);
+	if (
+		!lds ||
```
Akkadius, line 138: I believe all of these need to be `&&` not `||`

zone/loottables.cpp
```diff
@@ -395,141 +405,136 @@ void NPC::AddLootDrop(
...
+			equipment[found_slot] = item2->ID;
+
 			found = true;
 		}
 
-		// @merth: IDFile size has been increased, this needs to change
-		uint16 emat;
-		if(item2->Material <= 0
-			|| (item2->Slots & ((1 << EQ::invslot::slotPrimary) | (1 << EQ::invslot::slotSecondary)))) {
-			memset(newid, 0, sizeof(newid));
```
Akkadius, line 446: Why was this removed ?

Akkadius:
Adding this event generally looks fine - but the biggest thing with this PR is it mixes a bunch of formatting, refactoring and minor logic changes with an event add. 

If we have to revert this we also revert all of the formatting which is not always merge/unmerge friendly.

I really appreciate the mindset behind cleaning code up while we're working with it though, I just think we should split these up

zone/loottables.cpp
Kinglykrab: Previous logic checked for these in order and returned if any were true. So `||` is correct as far as I can tell.

zone/loottables.cpp
Kinglykrab: I am using `equipment_material = Strings::ToUnsignedInt(&item2->IDFile[2])` instead so we don’t need the temporary variable.

PR #3742, opened 2023-12-05
[Quest API] Add EVENT_LDON_POINTS_GAIN and EVENT_LDON_POINTS_LOSS to Perl/Lua

Kinglykrab:
# Perl
- Add `EVENT_LDON_POINTS_GAIN`.
- Add `EVENT_LDON_POINTS_LOSS`.
- Exports `$theme_id` and `$points`.

# Lua
- Add `event_ldon_points_gain`.
- Add `event_ldon_points_loss`.
- Exports `e.theme_id` and `e.points`.

# Notes
- Allows operators to track gain/loss of LDoN Points of any theme.


zone/client.cpp
```diff
@@ -1410,126 +1410,173 @@ void Client::SetMaxHP() {
...
 			}
+
 			m_pp.ldon_points_tak += points;
 			break;
 		}
 	}
+
 	m_pp.ldon_points_available += points;
+
+	if (!is_loss) {
```
Akkadius, line 1546: Wouldn't this read slightly easier if we just did `if (is_loss)` and flipped

Akkadius: Even further - we could simplify with 

```cpp
	auto event_id = is_loss ? EVENT_LDON_POINTS_LOSS : EVENT_LDON_POINTS_GAIN;
	if (parse->PlayerHasQuestSub(event_id)) {
		const std::string &export_string = fmt::format("{} {}", theme_id, std::abs(points));
		parse->EventPlayer(event_id, this, export_string, 0);
	}
```

Kinglykrab: Sounds good to me.

Akkadius (approved):
LGTM. Rebased for you and checked compilation locally.

PR #3752, opened 2023-12-09
[Cleanup] Cleanup classes.cpp/classes.h

Kinglykrab:
# Notes
- Remove unused methods and cleanup logic.


joligario (commented):
While we're in the classes, we could probably go ahead and add the LVL 75 titles: https://everquest.allakhazam.com/wiki/EQ:Special_Titles - not sure if there are more above 75 and don't see Drakkin in there which probably have a title.

joligario (approved):

PR #3760, opened 2023-12-12
[Rules] Add DOT and HOT Rules

Kinglykrab:
# Notes
- Add `Spells:DOTDamageBonusSplitOverDuration` rule.
- Allows operators to disable the DOT bonus damage being split over duration and instead adds the full amount every tic.
- Add `Spells:HOTBonusHealingSplitOverDuration` rule.
- Allows operators to disable the HOT bonus healing being split over duration and instead adds the full amount every tic.


Akkadius:
Doesn't seem like these rules should be on by default unless I'm missing something ?

Kinglykrab:
> Doesn't seem like these rules should be on by default unless I'm missing something ?

The current functionality is to divide extra damage over duration. These being on by default leaves functionality as is except for those who’d like to opt out.

Akkadius (approved):

PR #3765, opened 2023-12-13
[UCS] Merge mailserver and chatserver configs into UCS.

### LINKED ISSUE #700: Combine Mailserver and Chatserver config into one UCS block

joligario:
Fixes #700.
Requires manual adjustment of existing config files as identified below.
Old config:
```
          "mailserver" : {
               "host" : "localhost",
               "port" : "7778"
          },
          "chatserver" : {
               "host" : "localhost",
               "port" : "7778"
          },
```
New config:
```
          "ucs" : {
               "host" : "localhost",
               "port" : "7778"
          },
```

Coordination:
- [ ] Occulus
- [ ] Spire
- [ ] Akk-stack
- [x] EQEmu Docs
- [x] TalkEQ
- [x] Installer Template

### DIFF
--- a/common/eqemu_config.cpp
+++ b/common/eqemu_config.cpp
@@ -114,10 +114,10 @@ void EQEmuConfig::parse_config()
 	/**
 	 * UCS
 	 */
-	ChatHost = _root["server"]["chatserver"].get("host", "eqchat.eqemulator.net").asString();
-	ChatPort = Strings::ToUnsignedInt(_root["server"]["chatserver"].get("port", "7778").asString());
-	MailHost = _root["server"]["mailserver"].get("host", "eqmail.eqemulator.net").asString();
-	MailPort = Strings::ToUnsignedInt(_root["server"]["mailserver"].get("port", "7778").asString());
+	ChatHost = _root["server"]["ucs"].get("host", "eqchat.eqemulator.net").asString();
+	ChatPort = Strings::ToUnsignedInt(_root["server"]["ucs"].get("port", "7778").asString());
+	MailHost = _root["server"]["ucs"].get("host", "eqmail.eqemulator.net").asString();
+	MailPort = Strings::ToUnsignedInt(_root["server"]["ucs"].get("port", "7778").asString());
 
 	/**
 	 * Database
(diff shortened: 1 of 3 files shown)


Kinglykrab:
Will this break anything for those who already have UCS going?

joligario:
> Will this break anything for those who already have UCS going?

This only changes the name of where it gets the config. No changes to the loading or functionality of UCS.

Akkadius:
> > Will this break anything for those who already have UCS going?
> 
> This only changes the name of where it gets the config. No changes to the loading or functionality of UCS.

Occulus / Spire / Akk-stack all need changes to coordinate with this

joligario:
And docs

Akkadius:
TalkEQ as well which has strict typing and uses the config cc @xackery 

Installer base configs will need to be updated

Akkadius:
We also need an automated way to convert peoples configuration files

joligario:
I disagree with “need”. Nice to have? Yes. Outside of the XML to JSON conversion, we haven’t been doing any automated config file edits. In theory it sounds ideal to reduce errors, but you add just as many variable configurations out there to introduce errors. This is a minor modification. We already have users modify their own config file as it is good to know what is in their as an operator…

xackery:
> TalkEQ as well which has strict typing and uses the config cc @xackery

DiscordEQ (the predecessor of TalkEQ) did inject lines into eqemu_config, but it has been deprecated for a long time now.

TalkEQ does not use eqemu_config in any way, so should be a non-issue for above.

So should be green on this perspective.

Akkadius:
> I disagree with “need”. Nice to have? Yes. Outside of the XML to JSON conversion, we haven’t been doing any automated config file edits. In theory it sounds ideal to reduce errors, but you add just as many variable configurations out there to introduce errors. This is a minor modification. We already have users modify their own config file as it is good to know what is in their as an operator…

I’m sorry but it’s not acceptable and lazy on our part to make a change and not make this seamless for operators. We have done tons of automatic updates for years and I’m the one who ends up cleaning up after all of the lazy manual intervention changes and it pisses operators off

PR #3772, opened 2023-12-16
[Bots] Fix ^oo autodefend from sending bots/pets to invalid haters

nytmyr:
Bots and especially their pets could get in a loop where they try to attack an invalid target. This would result in their pets spamming "That is not a legal target" messages.


nytmyr:
Been an extremely long time since I fixed this on my server but IIRC this was how to prevent their pets from spamming you with illegal targets. I believe the best way to replicate was charm a mob the bots were already attempting to attack.

Kinglykrab (approved):

PR #3763, opened 2023-12-13
[Cleanup] Cleanup race constants and references.

Kinglykrab:
# Notes
- All race constants have been converted to an enumerator.
- Race bitmask contants have been converted to an enumerator.
- Race methods have been cleaned up and consolidated as best as possible.
- All references to race use the use enumerator.

PR #3773, opened 2023-12-16
[Code Cleanup] Race constants refactor

Akkadius:
This is helping Kingly in PR review https://github.com/EQEmu/Server/pull/3763#issuecomment-1858703434

The previous PR does too many sweeping changes at once, has errors and makes it hard to merge / review / dissect / revert

**Suggestions** 

* 1:1 race constant changes from defines to scoped constexpr in another PR
* 1:1 gender constant changes from defines to scoped constexpr in another PR
* Refactors / logic changes to switches in a separate PR

This current branch only does race constants. It uses a small Go app located in `./utils/raceconstantrename/` - keep in mind the app is quick and dirty but does the job very accurately.

* Parses races.h for the old definitions
* Parses races.h for the new scoped `Race` namespace definitions (transplanted manually)
* Walks through all cpp/h files in the project, parses for occurrences of the old definitions and replaces them with new definitions using specific matching so we don't do blind matches and accidentally replace things like strings or partial matches
* All matches are done only if there is an exact match on the value of the old definition to to an equivalent new definition

This is to ensure 100% accuracy in the constant migration. It makes this change far easier to review, revert, merge etc.

Kingly if you want to take PR credit and recut this branch using the script by all means - I just care more that we minimize issues with this large of change.


Kinglykrab:
This is awesome.

Kinglykrab (approved):

common/races.cpp
```diff
@@ -1420,26 +1420,26 @@ uint32 GetPlayerRaceBit(uint16 race_id) {
...
+		case Race::DarkElf:
+		case Race::HalfElf:
+		case Race::Dwarf:
+		case Race::Troll:
+		case Race::Ogre:
+		case Race::Halfling:
+		case Race::Gnome:
 			return player_race_value;
-		case PLAYER_RACE_IKSAR:
+		case Race::Aviak:
```
hgtw, line 1436: These are wrong starting here, the `player_race_value` doesn't correspond to race id for this function

Akkadius: Good call - I need the script to ignore some of the other constant values

Akkadius:
Re-cutting this one with another script run https://github.com/EQEmu/Server/pull/3782

PR #3779, opened 2023-12-17
[Bug Fix] Send Entity ID in Death Events to resolve #3721

### LINKED ISSUE #3721: zone_controller.pl EVENT_DEATH_ZONE sometimes not providing killed_entity_id
In the code below, if you #kill an npc instantly,  killed_entity_id will have the proper value.  If you aggro the mob first, then damage it down and kill it, then killed_entity_id will be zero all the time.

```
sub EVENT_DEATH_ZONE {
        quest::debug("killed_entity_id:".$killed_entity_id);
        quest::debug("killed_npc_id:".$killed_npc_id);

```
image

image

Kinglykrab:
# Notes
- Due to some pointers becoming invalid you may get an invalid entity ID on the killed mob if we don't just directly send the entity ID in the export string.
- Resolves #3721.

# Screenshot
image

### DIFF
--- a/zone/lua_parser_events.cpp
+++ b/zone/lua_parser_events.cpp
@@ -330,9 +330,6 @@ void handle_npc_death(
 		l_npc_o.push(L);
 		lua_setfield(L, -2, "killed");
 	}
-
-	lua_pushinteger(L, Strings::ToUnsignedInt(sep.arg[4]));
-	lua_setfield(L, -2, "killed_entity_id");
 }
 
 void handle_npc_cast(
--- a/zone/lua_parser_events.cpp
+++ b/zone/lua_parser_events.cpp
@@ -330,6 +330,9 @@ void handle_npc_death(
 		l_npc_o.push(L);
 		lua_setfield(L, -2, "killed");
 	}
+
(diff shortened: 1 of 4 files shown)


Akkadius:
Did you test this

Kinglykrab:
> Did you test this

Yes. I can log back ingame to get some screenshot really quick.

Kinglykrab:
`#kill` works fine. Using `#killallnpcs` seems to still result it in exporting as `0` for some reason, but that's not a standard use-case for this event anyway.

zone/lua_parser_events.cpp
```diff
@@ -330,6 +330,9 @@ void handle_npc_death(
 		l_npc_o.push(L);
 		lua_setfield(L, -2, "killed");
 	}
+
```
Akkadius, line 333: Does there become a weird ordering behavior with extra parameters and the extra pointers ?

Kinglykrab: Shouldn't be any issue since we're exporting them named and the order shouldn't matter much.

I may be able to just remove the `GetID()` export altogether since we now have it functioning properly and don't need this in Lua since we'll have access to the pointer.

Akkadius (approved):

PR #3780, opened 2023-12-17
[Bug Fix] Fix NPCs routing to 0.0, 0.0 on #summon

### LINKED ISSUE #2474: Using #summon on an NPC produces weird pathing
I've noticed this a few times when trying to #summon a mob to me to then use #spawnfix it almost instantly runs off. Not sure to where! 

It happens on both static mobs and mobs with pathing grids set.

Presumably static mobs should definitely stand still after summoning - mobs with pathing grids may be different?

Kinglykrab:
# Notes
- Resolves #2474.

### DIFF
--- a/zone/gm_commands/summon.cpp
+++ b/zone/gm_commands/summon.cpp
@@ -3,19 +3,16 @@
 
 extern WorldServer worldserver;
 
-#include "../corpse.h"
-
 void command_summon(Client *c, const Seperator *sep)
 {
-	int arguments = sep->argnum;
+	const int arguments = sep->argnum;
 	if (!arguments && !c->GetTarget()) {
 		c->Message(Chat::White, "Usage: #summon - Summon your target, if you have one, to your position");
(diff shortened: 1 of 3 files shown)


Akkadius:
In general having an optional boolean flag to change the behavior of a function is a code smell and a sign that the functionality should be done out of band of the function itself. ([SRP](https://en.wikipedia.org/wiki/Single_responsibility_principle))

Because the other GMMove already does this, I'm fine with this change for now. But we really should be decoupling responsibilities and avoiding boolean overloading functions.

Good to have this behavior fixed

Akkadius (approved):

Kinglykrab:
> In general having an optional boolean flag to change the behavior of a function is a code smell and a sign that the functionality should be done out of band of the function itself. ([SRP](https://en.wikipedia.org/wiki/Single_responsibility_principle))
> 
> Because the other GMMove already does this, I'm fine with this change for now. But we really should be decoupling responsibilities and avoiding boolean overloading functions.
> 
> Good to have this behavior fixed

That's fine with me, was just going for parity with the other method.

PR #3781, opened 2023-12-18
[Quest API] Add GetNPCAggro() and SetNPCAggro() to Perl/Lua

Kinglykrab:
# Perl
- Add `$npc->GetNPCAggro()`.
- Add `$npc->SetNPCAggro(in_npc_aggro)`.

# Lua
- Add `npc:GetNPCAggro()`.
- Add `npc:SetNPCAggro(in_npc_aggro)`.

# Notes
- Allows operators to enable or disable an NPC's NPC aggro capability dynamically.

### DIFF
--- a/zone/api_service.cpp
+++ b/zone/api_service.cpp
@@ -245,7 +245,7 @@ Json::Value ApiGetNpcListDetail(EQ::Net::WebsocketServerConnection *connection, 
 		row["swarm_owner"]     = npc->GetSwarmOwner();
 		row["swarm_target"]    = npc->GetSwarmTarget();
 		row["waypoint_max"]    = npc->GetWaypointMax();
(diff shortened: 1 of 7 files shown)


zone/api_service.cpp
```diff
@@ -245,7 +245,7 @@ Json::Value ApiGetNpcListDetail(EQ::Net::WebsocketServerConnection *connection,
 		row["swarm_owner"]     = npc->GetSwarmOwner();
 		row["swarm_target"]    = npc->GetSwarmTarget();
 		row["waypoint_max"]    = npc->GetWaypointMax();
-		row["will_aggro_npcs"] = npc->WillAggroNPCs();
```
Akkadius, line 248: I don't think there's anything hard pressed to use this json key so we can change the json key too

Akkadius (approved):

PR #3783, opened 2023-12-18
[Bug Fix] Fix Bard Invisibility Songs breaking every 4 ticks

### LINKED ISSUE #2361: Bard Invisibility Breaks Every 4th Tick
**Description:** Bard invisibility song Shauri's Sonorous Clouding will not apply when already active, so it breaks every 4th tick.
image

**Steps to reproduce:** 
- Be a bard of level 19 or greater
- Purchase and memorise the song Shauri's Sonorous Clouding (PoK or Erudin Library)
- Purchase and equip a wooden flute (required)
- Play the song

**Expected:** Each tick the success message appears and the song effects continue as long the song is played ("A soft mist surrounds you")
**Actual:** On song start the effect is applied and the player becomes invisible. Each subsequent tick that the effect is still active, it is not refreshed. The message "\<player\> tries to cast an invisibility spell on you, but you are already invisible." is seen. Once the initial effect has worn off it is reapplied on the following tick, resulting in the player being visible for one tick.

Kinglykrab:
# Notes
- Allows Bard songs to bypass the invisibility restriction as they are sung constantly and otherwise would break.
- Fixes #2361.

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -3885,7 +3885,7 @@ bool Mob::SpellOnTarget(
 	// Prevent double invising, which made you uninvised
 	// Not sure if all 3 should be stacking
 	//This is not live like behavior (~Kayen confirmed 2/2/22)
-	if (!RuleB(Spells, AllowDoubleInvis)) {
+	if (!RuleB(Spells, AllowDoubleInvis) && !IsActiveBardSong(spell_id)) {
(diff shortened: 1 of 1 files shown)


KayenEQ (approved):
This should be fine as a solution to prevent the spam. There is very little reason anyone should be playing with this rule as false anyways.

Issue #699, opened 2018-01-15
Teleport bind shouldn't allow unkeyed toons to zone


mackal:
So if you can bind in a keyed/locked zone, teleport bind will allow you to bring unkeyed/flagged toons into the zone. I don't think this should work like this.

daerath:
But how are the non keyed people bound in a zone they cant enter?  Is it a group spell?

daerath:
But how are the non keyed people bound in a zone they cant enter?  Is the teleport a group spell?

mackal:
Teleport Bind is a group spell. There seems to be more to this since someone did try sebilis and it did block them. Another zone didn't.

daerath:
Which zone worked?

daerath:
I think I know the "problem".  The spell Teleport and the two Teleport Bind spell effects (5953, 16441) work properly.  They do not override the zone flags in the database allowing entry to a zone that requires flags when they don't have those flags.

However, the problem is that not all zones are properly marked with require flags.  I can bind my character in Veeshan's Peak, and then using 5953 or 16441, send a character there even if they are not keyed for that zone.

The problem is that the database doesn't have a flag specified for Veeshan's Peak.  You still need the key to enter it via Skyfire, but from the spell's perspective, the zone is not marked as requiring a flag.

Sebilis on the other hand does have a flag requirement in the database which is why it cannot be accessed with Teleport Bind.

So the only fix here is to update the database with the proper flag entries for any zone that requires a key to click into.

daerath:
I think the only three missing flag entries are Veeshan's Peak, Vex Thal and Charasis.  Well, up until OoW, but I'm not sure what level of implementation exists there yet.

mackal:
VP is a door that's scripted.

daerath:
so did some checking and the flags make more sense now.  Having the key, even if the key name matches the value of flag_needed doesn't actually get you into a zone.  The quest file for Emperor Ganak gives you the key, which gets you through the door, but simply having the key means nothing.  The quest file also adds the sebilis zone flag.  Also, the actual value of the flag_needed field can be anything you want.  As long as it isn't empty it triggers the lookup against your zone_flags, which is just a match against zone ID

And...  Vex Thal doesn't have a flag entry because none of the quest files set one.  It's also nobind = 0 so you can't bind there anyway, making the flag somewhat pointless since the only way in is via the door, which does have a key requirement.

[...]

PR #3784, opened 2023-12-18
[Bug Fix] Disallow ZoneToBindPoint sending to inaccessible zones

Kinglykrab:
# Notes
- #699 Describes this in more detail, this should keep people from zoning to bind if their bind is in a zone they can't enter.

### DIFF
--- a/zone/zoning.cpp
+++ b/zone/zoning.cpp
@@ -293,7 +293,6 @@ void Client::Handle_OP_ZoneChange(const EQApplicationPacket *app) {
 			target_y = m_pp.binds[0].y;
 			target_z = m_pp.binds[0].z;
 			target_heading = m_pp.binds[0].heading;
-			ignore_restrictions = 1;	//can always get to our bind point? seems exploitable
 			break;
 		case ZoneSolicited: //we told the client to zone somewhere, so we know where they are going.
 			//recycle zonesummon variables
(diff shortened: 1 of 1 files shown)


zone/zoning.cpp
```diff
@@ -832,6 +832,11 @@ void Client::ZonePC(uint32 zoneID, uint32 instance_id, float x, float y, float z
 
 		zone_mode = zm;
 		if (zm == ZoneToBindPoint) {
+			if (!CanEnterZone(pZoneName)) {
```
Akkadius, line 835: `instance_id` should be included here

Akkadius:
Testing done?

zone/zoning.cpp
Kinglykrab: `CanEnterZone` uses version. Do we mind making a database call to get version from instance ID?

Akkadius: It should be available in the `zd` object earlier in `ZonePC` 

```cpp
	auto zd = GetZoneVersionWithFallback(zoneID, zone->GetInstanceVersion());
	if (zd) {
		pZoneName = strcpy(new char[zd->long_name.length() + 1], zd->long_name.c_str());
	}
```

Kinglykrab: That doesn't use `instance_id` to find the version unfortunately.

Kinglykrab: Will have to change to something like this.
```cpp
const uint8 version = database.GetInstanceVersion(instance_id);

auto zd = GetZoneVersionWithFallback(zoneID, version);
if (zd) {
	pZoneName = strcpy(new char[zd->long_name.length() + 1], zd->long_name.c_str());
}```

Akkadius: Reading closer, `zd` version should be fetched **with** `instance_id` that's passed in, not the current zone objects instance_id

We will likely need to do a query regardless in order to make that lookup truly accurate before doing the correct lookup. It can then be used in both places.

We can't avoid a query in this case - can add lookup caching later

Issue #3678, opened 2023-11-05
Faction Inconsistencies with Live Client


joligario:
According to the live client, we do some faction things wrong.

*Base Faction Data*
There should be a `min_faction` and `max_faction` where they are considered lower and upper limits. These can be found in `FactionBaseData.txt`

For example: Brownies of Faydwer (65) have limits of -2000 to 2000 (65^-2000^2000^0^0^0^)

Note: The last 3 numbers aren't exactly clear yet, but they are either all zeros or 100/20/100. A couple dozen use this pattern which seems to place the base standing of 0 into a "Scowls" level. Perhaps a forced mod so you can never get above scowl.

*Faction Associations*
We may want to relook what we call "Faction Associations. The client uses the term "Faction Associations" where we currently use "Faction Mods". These can be found in `FactionAssociations.txt`. Most of ours seem correct from what I quickly checked. We need to add Drakkin mods. If not already, this should be built into `import_client_files` and `export_client_files`.

For example: Brownies of Faydwer (65) have the following mods:
- 65^51^-600^ // Human: -600
- 65^52^-600^ // Barbarian: -600
- 65^53^-600^ // Erudite: -600
- 65^54^-600^ // Wood Elf: -600
- 65^55^-600^ // High Elf: -600
- 65^56^-700^ // Dark Elf: -700
- 65^57^-600^ // Half Elf: -600
- 65^58^-600^ // Dwarf: -600
- 65^59^-750^ // Troll: -750
- 65^60^-750^ // Ogre: -750
- 65^61^-600^ // Halfling: -600
- 65^62^-600^ // Gnome: -600
- 65^178^-1000^ // Iksar: -1000
- 65^180^-600^ // Vah Shir: -600
- 65^661^-600^ // Guktan: -600
- 65^1106^-600^ // Drakkin: -600

![EQ000002](https://github.com/EQEmu/Server/assets/3617814/4e6e5da9-de53-44fd-9d2d-465c0fab720c)

![EQ000003](https://github.com/EQEmu/Server/assets/3617814/d7222cf4-267f-4bff-beba-032fd0e51cfc)

Each of our faction levels look spot on.
![image](https://github.com/EQEmu/Server/assets/3617814/55e51bd4-7a1c-41ec-bc87-f38036d31d3b)

joligario:
Closing for now. Looks like we have the data. Just not sure if we implement yet.

joligario:
[...]

PR #3799, opened 2023-12-22
[Database] Update faction mods with Live data

joligario:
Optional update to bring database up-to-date with Live data which includes Froglok and Drakkin data which was previously missing. Adds 999 new entries. Raw data can be found in the Live client at Resources\Faction\FactionAssociations.txt.

Creates a backup of the existing data at `faction_list_mod_backup_23122` which can be deleted if satisfied with the updates.

See discussion at #3678


Akkadius (approved):

noudess:
Looks good for stock databases.  If you've made changes or additions to the base table (Rolath has identified several new factions that are not exposed by the client) then you will want to change this to only overwrite core stuff.

noudess (approved):
Looks good for systems that have no custom mods or additions.  Those that do might want to modify sql to overwrite standard stuff.

Akkadius:
> Looks good for systems that have no custom mods or additions. Those that do might want to modify sql to overwrite standard stuff.
> 
> I would not want this added to required and be automated.

Required updates are only schema or DDL changes these days - no content changes

PR #3804, opened 2023-12-26
[Tasks] Add enabled column

Akkadius:
Adds an enabled column (default enabled) so we can import live task data framing and set them as disabled until fully implemented


Kinglykrab (approved):

Akkadius:
This still needs to be tested

hgtw (approved):

Akkadius:
**Testing**

```
  Zone |    Info    | LoadTasks Loaded [478] Tasks 
  Zone |    Info    | LoadTasks Loaded [2065] Task Activities 
```

joligario:
Need a new repository run?

Akkadius:
> Need a new repository run?

Great catch, I swore I had already ran it and committed it and distinctly remember it - but not here. Pushed up

joligario (approved):

PR #3813, opened 2023-12-28
[SQL] Fixed typo in manifest for task updates

joligario:


joligario:
Per @Akkadius , this is optional with single queries.

PR #3825, opened 2023-12-30
[Bug Fix] Client:SetBucket Overload Incorrectly Named

fryguy503:
There was an issue with Client:SetBucket when attempting to use an expiration date. Looks like a lua def was mistakenly named SetBucketExpires.


Kinglykrab (approved):

Kinglykrab:
Good eye.

PR #3826, opened 2023-12-30
[Commands] Add scoped buckets and editing to #databuckets

Kinglykrab:
# Notes
- Cleanup messages and logic.
- Utilize keyed buckets functionality optionally.
- Use repositories instead of a regular query.
- Added edit capability.

## Usage Menu
image

## Delete
image

## Edit
image
image

## View
image


Akkadius (changes requested):
Need to see testing screens. When we're changing commands we've had many trivial issues pop up that could be caught in simple manual testing

Kinglykrab:
Will upload more, but these are what I sent to show @fryguy503 in Discord.

# Edit
image
image

Akkadius (approved):

PR #3829, opened 2023-12-31
[Commands] Cleanup #petname Command

Kinglykrab:
# Notes
- Cleanup messages and logic.

### DIFF
--- a/zone/gm_commands/petname.cpp
+++ b/zone/gm_commands/petname.cpp
@@ -2,21 +2,44 @@
 
 void command_petname(Client *c, const Seperator *sep)
 {
-	Mob *target;
-	target = c->GetTarget();
+	Mob *t = nullptr;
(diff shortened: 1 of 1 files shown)


zone/gm_commands/petname.cpp
```diff
@@ -2,21 +2,44 @@
...
 {
-	Mob *target;
-	target = c->GetTarget();
+	Mob *t = nullptr;
+	if (c->GetTarget()) {
+		t = c->GetTarget();
+	}
 
-	if (!target || !target->IsPet()) {
-		c->Message(Chat::White, "Usage: #petname newname (requires a pet target)");
```
Akkadius, line 9: We should include a help message when no argument provided. Also mention that the name change is temporary because its not permanent

Akkadius (approved):

PR #3839, opened 2024-01-05
[Character] Convert Load/Save of Character Data to Repositories

Kinglykrab:
# Notes
- Converts `LoadCharacterData` and `SaveCharacterData` to repositories.

# Screenshots
## Before Relog
image
## After Relog
image

### DIFF
--- a/zone/zonedb.cpp
+++ b/zone/zonedb.cpp
@@ -1072,8 +1072,8 @@ bool ZoneDatabase::SaveCharacterData(
 
 	e.id                      = c->CharacterID();
 	e.account_id              = c->AccountID();
-	e.name                    = Strings::Escape(pp->name);
-	e.last_name               = Strings::Escape(pp->last_name);
+	e.name                    = pp->name;
+	e.last_name               = pp->last_name;
 	e.gender                  = pp->gender;
 	e.race                    = pp->race;
 	e.class_                  = pp->class_;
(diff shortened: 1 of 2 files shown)


zone/zonedb.cpp
```diff
@@ -1156,301 +1064,116 @@ bool ZoneDatabase::SaveCharacterData(
...
+
+	auto e = CharacterDataRepository::FindOne(database, c->CharacterID());
+	if (!e.id) {
+		return false;
+	}
+
+	e.id                      = c->CharacterID();
+	e.account_id              = c->AccountID();
+	e.name                    = Strings::Escape(pp->name);
+	e.last_name               = Strings::Escape(pp->last_name);
```
Akkadius, line 1076: We don't need to escape ourselves, this is a benefit we get for free with repositories. This actually would double escape and that will definitely create problems.

See repository code

```cpp
		v.push_back("'" + Strings::Escape(e.name) + "'");
		v.push_back("'" + Strings::Escape(e.last_name) + "'");
		v.push_back("'" + Strings::Escape(e.title) + "'");
		v.push_back("'" + Strings::Escape(e.suffix) + "'");
```

Akkadius (changes requested):
Looks good, minor comments.

Akkadius (approved):

PR #3850, opened 2024-01-06
[Character] Convert Delete/Load/Save of Character Disciplines to Repositories

Kinglykrab:
# Notes
- Convert `DeleteCharacterDiscipline` and `SaveCharacterDiscipline` to repositories.
- `LoadCharacterDiscipline` already used repositories, cleaned up the logic.

# Images
## Load
image
image
## Save
image
image
## Delete
image
image

### DIFF
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -10744,17 +10744,19 @@ void Client::SaveDisciplines()
 {
 	std::vector<CharacterDisciplinesRepository::CharacterDisciplines> v;
 
-	for (int index = 0; index < MAX_PP_DISCIPLINES; index++) {
-		if (IsValidSpell(m_pp.disciplines.values[index])) {
+	for (int slot_id = 0; slot_id < MAX_PP_DISCIPLINES; slot_id++) {
+		if (IsValidSpell(m_pp.disciplines.values[slot_id])) {
 			auto e = CharacterDisciplinesRepository::NewEntity();
 
 			e.id      = CharacterID();
-			e.slot_id = index;
-			e.disc_id = m_pp.disciplines.values[index];
+			e.slot_id = slot_id;
+			e.disc_id = m_pp.disciplines.values[slot_id];
 
 			v.emplace_back(e);
 		}
 	}
+
+	CharacterDisciplinesRepository::DeleteOne(database, CharacterID());
 
 	if (!v.empty()) {
 		CharacterDisciplinesRepository::ReplaceMany(database, v);
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -10756,8 +10756,8 @@ void Client::SaveDisciplines()
 		}
 	}
 
-	if (!l.empty()) {
-		CharacterDisciplinesRepository::ReplaceMany(database, l);
+	if (!v.empty()) {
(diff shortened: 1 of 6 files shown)


zone/zonedb.cpp
```diff
@@ -1071,11 +1074,15 @@ bool ZoneDatabase::SaveCharacterSkill(uint32 character_id, uint32 skill_id, uint
...
-	return true;
+bool ZoneDatabase::SaveCharacterDiscipline(uint32 character_id, uint32 slot_id, uint32 disc_id)
+{
+	auto e = CharacterDisciplinesRepository::NewEntity();
+
+	e.id      = character_id;
+	e.slot_id = slot_id;
+	e.disc_id = disc_id;
+
+	return CharacterDisciplinesRepository::ReplaceOne(*this, e);
```
Akkadius, line 1085: Inline

zone/zonedb.cpp
```diff
@@ -766,24 +766,27 @@ bool ZoneDatabase::LoadCharacterLeadershipAbilities(uint32 character_id, PlayerP
...
 		)
 	);
 
-	if (character_disciplines.empty()) {
+	if (l.empty()) {
 		return false;
 	}
 
-	/* Initialize Disciplines */
-	memset(pp->disciplines.values, 0, (sizeof(pp->disciplines.values[0]) * MAX_PP_DISCIPLINES));
```
Akkadius, line 781: We need this still

zone/client.cpp
```diff
@@ -10742,22 +10742,24 @@ void Client::SaveSpells()
...
+
+			e.id      = CharacterID();
+			e.slot_id = slot_id;
+			e.disc_id = m_pp.disciplines.values[slot_id];
+
+			v.emplace_back(e);
 		}
 	}
 
-	CharacterDisciplinesRepository::DeleteWhere(database, fmt::format("id = {}", CharacterID()));
```
Akkadius, line 10757: Why is this now DeleteOne? Shouldn't we remove this line altogether if we're using ReplaceMany ?

zone/zonedb.cpp
Kinglykrab: I figured this was the same thing?
```cpp
for (int slot_id = 0; slot_id < MAX_PP_DISCIPLINES; slot_id++) { // Initialize Disciplines
	pp->disciplines.values[slot_id] = 0;
}
```

zone/client.cpp
Kinglykrab: `DeleteOne` actually deletes everything for that character ID, similar to `DeleteWhere`, but without needing anything but the `CharacterID()` passed to it. And yeah, we can definitely remove it since we're replacing now.

Akkadius: `DeleteOne` makes the assumptions that `id` is an auto incr id usually which doesn't have more than one. It's fine though I see what you're saying

Kinglykrab: Yeah, in this case our primary key doesn't auto increment so some tables let us use `DeleteOne` as a short-hand.

Akkadius: It just doesn't read well

Akkadius (approved):

PR #3853, opened 2024-01-07
[Pets] Convert Load/Save of Pet Info to Repositories

Kinglykrab:
# Notes
- Convert `LoadPetInfo` to repositories.
- `SavePetInfo` already used repositories, cleanup logic.

# Images
## Load
image
image
## Save
image
image

### DIFF
--- a/zone/zonedb.cpp
+++ b/zone/zonedb.cpp
@@ -3121,7 +3121,7 @@ void ZoneDatabase::LoadAuras(Client *c)
 
 void ZoneDatabase::SavePetInfo(Client *client)
 {
-	PetInfo* c = nullptr;
+	PetInfo* p = nullptr;
 
 	std::vector<CharacterPetInfoRepository::CharacterPetInfo> pet_infos;
 	auto pet_info = CharacterPetInfoRepository::NewEntity();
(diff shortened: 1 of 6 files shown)


zone/zonedb.cpp
```diff
@@ -3275,109 +3287,106 @@ void ZoneDatabase::DeleteItemRecast(uint32 character_id, uint32 recast_type)
...
+	);
+
+	if (info.empty()) {
 		return;
 	}
 
-	PetInfo *pi;
-	for (auto& row = results.begin(); row != results.end(); ++row) {
-		uint16 pet = Strings::ToInt(row[0]);
+	PetInfo* c;
```
Akkadius, line 3309: I don't think we should use `c` since `Client *client` gets passed in already and arguably you could be shortening that to `c`

I was reading this as client before I looked up. At least have it be `p` or stay `pi`

zone/zonedb.cpp
```diff
@@ -3121,124 +3121,136 @@ void ZoneDatabase::LoadAuras(Client *c)
...
-	CharacterPetInventoryRepository::CharacterPetInventory item = {};
+	std::vector<CharacterPetInventoryRepository::CharacterPetInventory> inventory;
+	auto item = CharacterPetInventoryRepository::NewEntity();
 
-	// Loop through pet types
-	for (int pet = 0; pet < 2; pet++) {
-		petinfo = client->GetPetInfo(pet);
-		if (!petinfo) {
+	for (int pet_info_type = PetInfoType::Current; pet_info_type <= PetInfoType::Suspended; pet_info_type++) {
+		c = client->GetPetInfo(pet_info_type);
```
Akkadius, line 3136: Same here `p` or `pi` for receiver

Akkadius (approved):

PR #3862, opened 2024-01-07
[Feature] Break Trader if moved

fryguy503:
If the player moves while in trader mode, it will end the mode, this prevents traders from moving outside of the trader area after entering trader mode.

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -4889,7 +4889,7 @@ void Client::Handle_OP_ClientUpdate(const EQApplicationPacket *app) {
 
 	if (cy != m_Position.y || cx != m_Position.x) {
 		// End trader mode if we move
-		if(Trader) {
+		if (Trader) {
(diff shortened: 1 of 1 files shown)


Kinglykrab (approved):

zone/client_packet.cpp
```diff
@@ -4887,8 +4887,13 @@ void Client::Handle_OP_ClientUpdate(const EQApplicationPacket *app) {
 			CheckIncreaseSkill(EQ::skills::SkillTracking, nullptr, -20);
 	}
 
-	/* Break Hide if moving without sneaking and set rewind timer if moved */
 	if (cy != m_Position.y || cx != m_Position.x) {
+		// End trader mode if we move
+		if(Trader) {
```
Kinglykrab, line 4892: `if (Trader)`

Akkadius (approved):

PR #3863, opened 2024-01-07
[Rules] Resist Softcap rules

fryguy503:
This rule will allow you to adjust the Resist softcap to help tune resists as you need.


zone/spells.cpp
```diff
@@ -5187,8 +5187,17 @@ float Mob::ResistSpell(uint8 resist_type, uint16 spell_id, Mob *caster, bool use
 
 	//Add our level, resist and -spell resist modifier to our roll chance
 	resist_chance += level_mod;
-	resist_chance += resist_modifier;
-	resist_chance += target_resist;
+	int softCapLvlMod;
```
Kinglykrab, line 5190: `soft_cap_level_modifier`.

Kinglykrab (approved):

zone/spells.cpp
```diff
@@ -5221,7 +5230,7 @@ float Mob::ResistSpell(uint8 resist_type, uint16 spell_id, Mob *caster, bool use
 	}
 
 	//Finally our roll
-	int roll = zone->random.Int(0, 200);
```
Kinglykrab, line 5224: We were using `200` here, but rule is default `255`?

Kinglykrab (approved):

PR #3864, opened 2024-01-07
[Bug Fix] Class Trainers dont steal your money!

fryguy503:
Class trainers log an error when you have reached your max train in the specific skill. They will continue to "Appear" to eat money and practices however they will be returned on logout/in or zone.

### DIFF
--- a/zone/client_process.cpp
+++ b/zone/client_process.cpp
@@ -1552,19 +1552,24 @@ void Client::OPGMTraining(const EQApplicationPacket *app)
 
 	Mob* pTrainer = entity_list.GetMob(gmtrain->npcid);
 
-	if(!pTrainer || !pTrainer->IsNPC() || pTrainer->GetClass() < Class::WarriorGM || pTrainer->GetClass() > Class::BerserkerGM)
+	if(!pTrainer || !pTrainer->IsNPC() || pTrainer->GetClass() < Class::WarriorGM || pTrainer->GetClass() > Class::BerserkerGM) {
 		return;
+	}
 
 	//you can only use your own trainer, client enforces this, but why trust it
 	if (!RuleB(Character, AllowCrossClassTrainers)) {
 		int trains_class = pTrainer->GetClass() - (Class::WarriorGM - Class::Warrior);
-		if (GetClass() != trains_class)
+		if (GetClass() != trains_class) {
+			safe_delete(outapp);
 			return;
+		}
 	}
 
 	//you have to be somewhat close to a trainer to be properly using them
-	if(DistanceSquared(m_Position,pTrainer->GetPosition()) > USE_NPC_RANGE2)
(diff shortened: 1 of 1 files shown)


zone/client_process.cpp
```diff
@@ -1691,11 +1696,12 @@ void Client::OPGMTrainSkill(const EQApplicationPacket *app)
 
 		uint16 skilllevel = GetRawSkill(skill);
 
-		if(skilllevel == 0) {
+		if(skilllevel == 0 || (skilllevel == 254 && SkillTrainLevel(skill, GetClass()) <= GetLevel())) {
```
Kinglykrab, line 1699: What does checking for `254` do here exactly?

Kinglykrab:
What do the `SetSkill` calls do exactly?

zone/client_process.cpp
fryguy503: Thank you, this was a carry over from the older TAKP code our server used. I do not believe this portion is needed anymore.

fryguy503:
SetSkill is setting the current skill to the value before the attempted increase. SetSkill is sending the packet to the client which dumbly continues to increase the values beyond the max. This makes sure the client is in sync.

zone/client_process.cpp
Kinglykrab: Still need to remove this if we don't need this.

zone/client_process.cpp
```diff
@@ -1552,19 +1552,24 @@ void Client::OPGMTraining(const EQApplicationPacket *app)
...
-		if (GetClass() != trains_class)
+		if (GetClass() != trains_class) {
+			safe_delete(outapp);
 			return;
+		}
 	}
 
 	//you have to be somewhat close to a trainer to be properly using them
-	if(DistanceSquared(m_Position,pTrainer->GetPosition()) > USE_NPC_RANGE2)
+	if(DistanceSquared(m_Position,pTrainer->GetPosition()) > USE_NPC_RANGE2) {
```
Kinglykrab, line 1569: `if (`

zone/client_process.cpp
```diff
@@ -1552,19 +1552,24 @@ void Client::OPGMTraining(const EQApplicationPacket *app)
 
 	Mob* pTrainer = entity_list.GetMob(gmtrain->npcid);
 
-	if(!pTrainer || !pTrainer->IsNPC() || pTrainer->GetClass() < Class::WarriorGM || pTrainer->GetClass() > Class::BerserkerGM)
+	if(!pTrainer || !pTrainer->IsNPC() || pTrainer->GetClass() < Class::WarriorGM || pTrainer->GetClass() > Class::BerserkerGM) {
```
Kinglykrab, line 1555: `if (`

zone/client_process.cpp
Kinglykrab, line 1699: `if (`

Akkadius (approved):

PR #3887, opened 2024-01-07
[Bug Fix] Add locations where melee can be bound outside of a city.

fryguy503:


Kinglykrab (approved):

zone/zone.cpp
```diff
@@ -304,6 +304,44 @@ bool Zone::LoadZoneObjects()
 	return true;
 }
 
+bool Zone::IsSpecialBindLocation(const glm::vec4& location) {
```
Kinglykrab, line 307: Usually we put `{` on new line for methods.

PR #3889, opened 2024-01-07
[Bug Fix] Swimming rules adjustment and racial

fryguy503:
We need to set all character default swiming prior to working through racials. Added racial check so if starting rule is over the racial default it will not over-ride the starting rule.

### DIFF
--- a/world/client.cpp
+++ b/world/client.cpp
@@ -1691,14 +1691,15 @@ bool Client::OPCharCreate(char *name, CharCreate_Struct *cc)
 	pp.hunger_level = 6000;
 	pp.thirst_level = 6000;
 
+	/* Set default skills for everybody */
+	pp.skills[EQ::skills::SkillSwimming] = RuleI(Skills, SwimmingStartValue);
+	pp.skills[EQ::skills::SkillSenseHeading] = RuleI(Skills, SenseHeadingStartValue);
+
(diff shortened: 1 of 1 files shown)


world/client.cpp
```diff
@@ -2143,7 +2144,9 @@ void Client::SetRaceStartingSkills( PlayerProfile_Struct *pp )
 		}
 	case FROGLOK:
 		{
-			pp->skills[EQ::skills::SkillSwimming] = 125;
+			if (RuleI(Skills, SwimmingStartValue) > 125) {
```
Kinglykrab, line 2147: Should this be `<`?

world/client.cpp
```diff
@@ -2160,7 +2163,9 @@ void Client::SetRaceStartingSkills( PlayerProfile_Struct *pp )
 	case IKSAR:
 		{
 			pp->skills[EQ::skills::SkillForage] = 50;
-			pp->skills[EQ::skills::SkillSwimming] = 100;
+			if (RuleI(Skills, SwimmingStartValue) > 100) {
```
Kinglykrab, line 2166: Should this be `<`?

world/client.cpp
fryguy503: Yes, thank you

Kinglykrab (approved):

PR #3891, opened 2024-01-07
[Zones] Convert IDLE_WHEN_EMPTY to a Zone Column

Kinglykrab:
# Notes
- Converts `IDLE_WHEN_EMPTY` to `idle_when_empty` in `zone` table so that we can change this on the fly or on a zone-by-zone basis instead of having to recompile to do this.
- Especially helpful for those using release binaries that do not compile their own source.

# Images
image
image
image


fryguy503:
Would disabling idle when empty be just setting the rule to a very high number?

Kinglykrab:
> Would disabling idle when empty be just setting the rule to a very high number?

It’s no longer a rule. You’d just set the `idle_when_empty` in `zone` table to `0`.

noudess (commented):
I like the fexibility this provides.

zone/zone.h
```diff
@@ -107,7 +107,7 @@ class Zone {
 	AA::Ability *GetAlternateAdvancementAbilityByRank(int rank_id);
 	AA::Rank *GetAlternateAdvancementRank(int rank_id);
 	bool is_zone_time_localized;
-	bool process_mobs_while_empty;
+	bool m_idle_when_empty;
```
Akkadius, line 110: Keep in mind we use `m_` for `private` member variables. If we're going to prefix `m_` we should move it to private.

Then if we want to mutate it outside of the zone object, we use setters/getters

common/database/database_update_manifest.cpp
```diff
@@ -5175,6 +5175,17 @@ ALTER TABLE `task_activities`
...
+  ManifestEntry{
+    .version = 9251,
+		.description = "2024_01_07_zone_idle_when_empty.sql",
+		.check = "SHOW COLUMNS FROM `zone` LIKE 'idle_when_empty'",
+		.condition = "empty",
+		.match = "",
+		.sql = R"(
+ALTER TABLE `zone`
+ADD COLUMN `idle_when_empty` tinyint(1) UNSIGNED NOT NULL DEFAULT 1 AFTER `min_lava_damage`
+)"
```
Akkadius, line 5187: This should be `.content_schema_update = true`

common/ruletypes.h
```diff
@@ -338,7 +338,7 @@ RULE_BOOL(Zone, UseZoneController, true, "Enables the ability to use persistent
 RULE_BOOL(Zone, EnableZoneControllerGlobals, false, "Enables the ability to use quest globals with the zone controller NPC")
 RULE_INT(Zone, GlobalLootMultiplier, 1, "Sets Global Loot drop multiplier for database based drops, useful for double, triple loot etc")
 RULE_BOOL(Zone, KillProcessOnDynamicShutdown, true, "When process has booted a zone and has hit its zone shut down timer, it will hard kill the process to free memory back to the OS")
-RULE_INT(Zone, SecondsBeforeIdle, 60, "Seconds before IDLE_WHEN_EMPTY define kicks in")
```
Akkadius, line 341: This should probably be a zone field as well

zone/zone.cpp
```diff
@@ -1002,8 +1002,9 @@ Zone::Zone(uint32 in_zoneid, uint32 in_instanceid, const char* in_short_name)
 		m_safe_points.y = z->safe_y;
 		m_safe_points.z = z->safe_z;
 		m_safe_points.w = z->safe_heading;
-		m_graveyard_id = z->graveyard_id;
-		m_max_clients  = z->maxclients;
+		m_graveyard_id    = z->graveyard_id;
+		m_max_clients     = z->maxclients;
+		idle_when_empty = z->idle_when_empty;
```
Akkadius, line 1007: alignment

zone/zone.cpp
```diff
@@ -1389,6 +1390,7 @@ bool Zone::LoadZoneCFG(const char* filename, uint16 instance_version)
 	allow_mercs               = true;
 	m_graveyard_id            = z->graveyard_id;
 	m_max_clients             = z->maxclients;
+	idle_when_empty         = z->idle_when_empty;
```
Akkadius, line 1393: alignment

common/database/database_update_manifest.cpp
```diff
@@ -5186,7 +5186,19 @@ ALTER TABLE `instance_list`
...
+	ManifestEntry{
+		.version = 9252,
+		.description = "2024_01_07_zone_idle_when_empty.sql",
+		.check = "SHOW COLUMNS FROM `zone` LIKE 'idle_when_empty'",
+		.condition = "empty",
+		.match = "",
+		.sql = R"(
+ALTER TABLE `zone`
+ADD COLUMN `idle_when_empty` tinyint(1) UNSIGNED NOT NULL DEFAULT 1 AFTER `min_lava_damage`
+ADD COLUMN `seconds_before_idle` int(11) UNSIGNED NOT NULL DEFAULT 1 AFTER `idle_when_empty`
```
Akkadius, line 5198: DEFAULT 60

zone/zone.h
```diff
@@ -99,7 +99,8 @@ class Zone {
 	AA::Ability *GetAlternateAdvancementAbilityByRank(int rank_id);
 	AA::Rank *GetAlternateAdvancementRank(int rank_id);
 	bool is_zone_time_localized;
-	bool process_mobs_while_empty;
+	bool idle_when_empty;
+	uint32 seconds_before_idle;
```
Akkadius, line 103: We really should try and make these private, we can access them within the Zone class but outside such as `questmanager` we'll need a get/set method - easy to generate in Clion

common/repositories/base/base_zone_repository.h
```diff
@@ -549,48 +557,50 @@ class BaseZoneRepository {
...
 			e.content_flags             = row[91] ? row[91] : "";
 			e.content_flags_disabled    = row[92] ? row[92] : "";
-			e.underworld_teleport_index = static_cast<int32_t>(atoi(row[93]));
-			e.lava_damage               = static_cast<int32_t>(atoi(row[94]));
-			e.min_lava_damage           = static_cast<int32_t>(atoi(row[95]));
+			e.underworld_teleport_index = row[93] ? static_cast<int32_t>(atoi(row[93])) : 0;
+			e.lava_damage               = row[94] ? static_cast<int32_t>(atoi(row[94])) : 50;
+			e.min_lava_damage           = row[95] ? static_cast<int32_t>(atoi(row[95])) : 10;
+			e.idle_when_empty           = row[96] ? static_cast<uint8_t>(strtoul(row[96], nullptr, 10)) : 1;
+			e.seconds_before_idle       = row[97] ? static_cast<uint32_t>(strtoul(row[97], nullptr, 10)) : 1;
```
Akkadius, line 603: Make sure to regen this

Akkadius (approved):
Looks good, great work. Let's attach some testing to show things work as expected

zone/zone.cpp
```diff
@@ -1005,6 +1008,9 @@ Zone::Zone(uint32 in_zoneid, uint32 in_instanceid, const char* in_short_name)
 		m_graveyard_id = z->graveyard_id;
 		m_max_clients  = z->maxclients;
 
+		SetIdleWhenEmpty(z->idle_when_empty);
+		SetSecondsBeforeIdle(z->seconds_before_idle);
```
Akkadius, line 1012: This is fine to leave here but since you are the zone class you could have used them since they are private and belong to zone

common/repositories/base/base_zone_repository.h
```diff
@@ -459,6 +465,8 @@ class BaseZoneRepository {
 		e.underworld_teleport_index = 0;
 		e.lava_damage               = 50;
 		e.min_lava_damage           = 10;
+		e.idle_when_empty           = 1;
+		e.seconds_before_idle       = 1;
```
Akkadius, line 469: Default is still wrong in the repo

PR #3892, opened 2024-01-07
[Bug Fix] Clear Ramp when Clearing hate

fryguy503:
When clearing a client from the hatelist, also clear them from ramp

### DIFF
--- a/zone/entity.cpp
+++ b/zone/entity.cpp
@@ -1491,18 +1491,21 @@ void EntityList::RemoveFromTargets(Mob *mob, bool RemoveFromXTargets)
 		Mob *m = it->second;
 		++it;
 
-		if (!m)
+		if (!m) {
 			continue;
+		}
 
 		if (RemoveFromXTargets && mob) {
-			if (m->IsClient() && (mob->CheckAggro(m) || mob->IsOnFeignMemory(m)))
+			if (m->IsClient() && (mob->CheckAggro(m) || mob->IsOnFeignMemory(m))) {
 				m->CastToClient()->RemoveXTarget(mob, false);
 			// FadingMemories calls this function passing the client.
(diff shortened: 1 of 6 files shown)


zone/mob_ai.cpp
```diff
@@ -2062,6 +2062,33 @@ void Mob::ClearRampage()
 	RampageArray.clear();
 }
 
+// if force is false, it will not remove feigned clients
+void Mob::RemoveFromRampageList(Mob* mob, bool force)
```
Kinglykrab, line 2066: Should we name this `remove_feigned` if that's all it's used for?

Kinglykrab (approved):

zone/perl_npc.cpp
```diff
@@ -204,6 +204,7 @@ bool Perl_NPC_IsOnHatelist(NPC* self, Mob* mob) // @categories Hate and Aggro
 void Perl_NPC_RemoveFromHateList(NPC* self, Mob* mob) // @categories Hate and Aggro
 {
 	self->RemoveFromHateList(mob);
+	self->RemoveFromRampageList(mob);
```
Kinglykrab, line 207: I see you did Perl but not Lua?

zone/entity.cpp
```diff
@@ -3618,6 +3631,11 @@ void EntityList::ClearFeignAggro(Mob *targ)
 			}
 
 			it->second->RemoveFromHateList(targ);
+
+			if (it->second->GetSpecialAbility(SPECATK_RAMPAGE)) {
+						it->second->RemoveFromRampageList(targ, true);
```
Kinglykrab, line 3636: Looks like two tabs here?

zone/perl_npc.cpp
fryguy503: There is not a lua equivalent.  Lua has `RemoveFromHateLists` which has that built in perl has the singular `RemoveFromHateList`

PR #3893, opened 2024-01-07
[Bug Fix] Rune Invis Break

fryguy503:
Rune doesn't stop invis breaking


Kinglykrab (approved):

Kinglykrab:
> Rune doesn't stop invis breaking

Where does Rune come in to this? All I see is that one logic change?

PR #3898, opened 2024-01-07
[Bug Fix] Amplification should not benefit from instrument mods

fryguy503:


zone/client_mods.cpp
```diff
@@ -1455,7 +1455,7 @@ int32 Client::CalcATK()
 
 uint32 Mob::GetInstrumentMod(uint16 spell_id)
 {
-	if (GetClass() != Class::Bard) {
+	if (GetClass() != Class::Bard || spells[spell_id].is_discipline || spell_id == SPELL_AMPLIFICATION) {
```
Kinglykrab, line 1458: Disciplines don't benefit either I take it?

Kinglykrab (approved):

PR #3902, opened 2024-01-08
[Tasks] Add support for task window element groups

hgtw:
Elements that share a group are placed in the same list section and separated from other groups with a divider.

Live appears to only use this for optional elements in some tasks and when used each optional always gets its own group. This might indicate it's done automatically under certain criteria to ensure optionals are never grouped with non-optionals regardless of index.

Since groups are available in captures and there's very few tasks that use this, we don't need to worry about trying to replicate any automatic behavior since this allows more customization.

Before and after example with each optional having its own group:
image

### DIFF
--- a/common/database/database_update_manifest.cpp
+++ b/common/database/database_update_manifest.cpp
@@ -5162,7 +5162,19 @@ ALTER TABLE `tasks`
 ADD COLUMN `enabled` smallint NULL DEFAULT 1 AFTER `faction_amount`
 )",
 		.content_schema_update = true
-	}
+	},
+	ManifestEntry{
+		.version = 9250,
+		.description = "2023_01_06_task_activities_list_group.sql",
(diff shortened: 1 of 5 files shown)


joligario:
image

Kinglykrab (approved):

joligario:
So do optionals already split into their own group by default or do we need to do 1, 2, 3, etc?

hgtw:
Have to manually add them. I didn't want to try to do anything automatic because I'm not exact on what live's criteria is for doing it (if they even actually do it that way). I suspect they add each optional to its own group if that optional doesn't have a requirement to unlock it, but that's hard to confirm because there's some behavior with optionals that I haven't figured out yet (like them not always appearing if they're the last element until any other element has been done, or not appearing until zoning into a mission dz, etc)

joligario (approved):

PR #3908, opened 2024-01-08
[Rules] Backstab Damage Modifier

fryguy503:
Added NPCBackstabMod - Default 1.9 - Lower mod = higher backstab.

### DIFF
--- a/zone/special_attacks.cpp
+++ b/zone/special_attacks.cpp
@@ -183,7 +183,7 @@ int Mob::GetBaseSkillDamage(EQ::skills::SkillType skill, Mob *target)
 				}
 			} else if (IsNPC()) {
 				auto *npc = CastToNPC();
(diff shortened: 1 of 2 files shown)


zone/special_attacks.cpp
```diff
@@ -182,7 +183,7 @@ int Mob::GetBaseSkillDamage(EQ::skills::SkillType skill, Mob *target)
 				}
 			} else if (IsNPC()) {
 				auto *npc = CastToNPC();
-				base = std::max(base, npc->GetBaseDamage());
+				base = round((npc->GetMaxDMG() + npc->GetMinDMG()) / RuleR(NPC, NPCBackstabMod));
```
Akkadius, line 186: Where does this formula come from

fryguy503: https://github.com/EQEmu/Server/blob/master/zone/npc.cpp#L216

Akkadius (approved):

PR #3910, opened 2024-01-08
[Rule] Allow GMs to silently summon

fryguy503:
Disable to silently summon players with #summon.

Converted SummonPC to a MessageString

### DIFF
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -310,6 +310,7 @@ RULE_INT(World, MaximumQuestErrors, 30, "Changes the maximum number of quest err
 RULE_INT(World, BootHour, 0, "Sets the in-game hour world will set when it first boots. 0-24 are valid options, where 0 disables this rule")
 RULE_BOOL(World, UseItemLinksForKeyRing, false, "Uses item links for Key Ring Listing instead of item name")
 RULE_BOOL(World, UseOldShadowKnightClassExport, true, "Disable to have Shadowknight show as Shadow Knight (live-like)")
+RULE_BOOL(World, SendGMSummonMessageToClient, true, "Disable to allow gm's to silently #summon player.")
 RULE_CATEGORY_END()
 
 RULE_CATEGORY(Zone)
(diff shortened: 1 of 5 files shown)


common/ruletypes.h
```diff
@@ -310,6 +310,7 @@ RULE_INT(World, MaximumQuestErrors, 30, "Changes the maximum number of quest err
 RULE_INT(World, BootHour, 0, "Sets the in-game hour world will set when it first boots. 0-24 are valid options, where 0 disables this rule")
 RULE_BOOL(World, UseItemLinksForKeyRing, false, "Uses item links for Key Ring Listing instead of item name")
 RULE_BOOL(World, UseOldShadowKnightClassExport, true, "Disable to have Shadowknight show as Shadow Knight (live-like)")
+RULE_BOOL(World, SendGMSummonMessageToClient, true, "Disable to allow gm's to silently #summon player.")
```
Akkadius, line 313: Let's just remove the rule for this. While its nice to have rules/knobs - we should also try to minimize them where they're not really necessary and strive for sane defaults. In this case I don't see why we wouldn't want a message to display in this case

fryguy503: I know a few servers that do not allow MQ2 and utilize no messages to summon players and watch for bot snapping and returning logic. Many of these MQ2 bot scripts will look for the summon message and fully unload when seen.

Akkadius: If that's the case why don't we adjust how we handle summon as a GM. 

If the GM is hidden, send a different ZoneMode flag when we summon versus when the GM isn't hidden.

Add a `GMSummonHidden` enum and add that to the switch case and then in the `ProcessMovePC` function we simply handle it the same - just don't send the message to the client

```cpp
typedef enum {
	ZoneToSafeCoords, // Always send ZonePlayerToBind_Struct to client: Succor/Evac
	GMSummon, // Always send ZonePlayerToBind_Struct to client: Only a GM Summon
	ZoneToBindPoint, // Always send ZonePlayerToBind_Struct to client: Death Only
	ZoneSolicited, // Always send ZonePlayerToBind_Struct to client: Portal, Translocate, Evac spells that have a x y z coord in the spell data
	ZoneUnsolicited,
	GateToBindPoint, // Always send RequestClientZoneChange_Struct to client: Gate spell or Translocate To Bind Point spell
	SummonPC, // In-zone GMMove() always: Call of the Hero spell or some other type of in zone only summons
	Rewind, // Summon to /rewind location.
	EvacToSafeCoords
} ZoneMode;
```

fryguy503: I believe [`7bd0237` (#3910)](https://github.com/EQEmu/Server/pull/3910/commits/7bd02371891cfdf1264f9dcb6518538bc586f908) should cover that request.

Akkadius (approved):
Beautiful, thank you

PR #3911, opened 2024-01-08
[Bug] Allow Bard Group Dispell

fryguy503:


Akkadius:
Can you describe this change a little more for me please

PR #3918, opened 2024-01-08
[Bug Fix] Limit MeleeMitigationEffect to defender if only client.

fryguy503:


zone/attack.cpp
```diff
@@ -5360,24 +5360,26 @@ void Mob::DoRiposte(Mob *defender)
...
+		}
+
+		if (defender->IsOfClientBotMerc()) {
+			damage_bonus_mod += (defender->spellbonuses.MeleeMitigationEffect + itembonuses.MeleeMitigationEffect +
+				aabonuses.MeleeMitigationEffect);
 		}
-		// 168 defensive
-		dmgbonusmod += (defender->spellbonuses.MeleeMitigationEffect +
-		                defender->itembonuses.MeleeMitigationEffect +
-		                defender->aabonuses.MeleeMitigationEffect);
```
Akkadius, line 5377: So these were all defender before and now only one of them is - this doesn't seem right

Akkadius (approved):

PR #3920, opened 2024-01-08
[Crash Fix] Character Creation Class/Race out of Range.

fryguy503:
Known bug to crash the world server from character creation.

You can send a packet with a manual race or class entry below 0 or above 255 will cause world crash.

### DIFF
--- a/world/client.cpp
+++ b/world/client.cpp
@@ -539,12 +539,12 @@ bool Client::HandleNameApprovalPacket(const EQApplicationPacket *app)
 	uchar clas = app->pBuffer[68];
 
 	if (!IsPlayerRace(race)) {
-		LogInfo("Client::HandleNameApprovalPacket Race was less then zero or over 255");
+		LogInfo("Invalid Race ID.");
 		return false;
 	}
 
-	if (if (!EQ::ValueWithin(class, Class::Warrior, Class::Berserker))) {
-		LogInfo("Client::HandleNameApprovalPacket Class was less then zero or over 255");
+	if (!EQ::ValueWithin(class, Class::Warrior, Class::Berserker))) {
+		LogInfo("Invalid Class ID.");
 		return false;
 	}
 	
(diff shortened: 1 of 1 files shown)


world/client.cpp
```diff
@@ -538,6 +538,16 @@ bool Client::HandleNameApprovalPacket(const EQApplicationPacket *app)
 	uchar race = app->pBuffer[64];
 	uchar clas = app->pBuffer[68];
 
+	if (race < 0 || race > 255) {
+		LogInfo("Client::HandleNameApprovalPacket Race was less then zero or over 255");
+		return false;
+	}
+
+	if (clas < 0 || clas > 255) {
+		LogInfo("Client::HandleNameApprovalPacket Class was less then zero or over 255");
```
Akkadius, line 547: Drop `Client::HandleNameApprovalPacket` as it comes with logging for free now

world/client.cpp
```diff
@@ -538,6 +538,16 @@ bool Client::HandleNameApprovalPacket(const EQApplicationPacket *app)
 	uchar race = app->pBuffer[64];
 	uchar clas = app->pBuffer[68];
 
+	if (race < 0 || race > 255) {
+		LogInfo("Client::HandleNameApprovalPacket Race was less then zero or over 255");
```
Akkadius, line 542: Drop `Client::HandleNameApprovalPacket` as it comes with logging for free now

Akkadius (approved):
Minor comments otherwise lgtm

world/client.cpp
Kinglykrab: `if (!EQ::ValueWithin(Class::Warrior, Class::Berserker))`

world/client.cpp
Kinglykrab: Wouldn’t Froglok/Drakkin be higher than 255? I’d just do `if (!IsPlayerRace(race))`.

world/client.cpp
fryguy503: > Drop `Client::HandleNameApprovalPacket` as it comes with logging for free now

Not sure what you mean here

world/client.cpp
```diff
@@ -539,12 +539,12 @@ bool Client::HandleNameApprovalPacket(const EQApplicationPacket *app)
 	uchar clas = app->pBuffer[68];
```
Kinglykrab, line 540: `class`.

Kinglykrab (approved):

PR #3924, opened 2024-01-08
[Bug Fix] Attack Corpse Skill Up Fix

fryguy503:
Occasions where a corpse will allow skillups from attacks being turned on (even if no damage)

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -1463,7 +1463,7 @@ void Mob::DoAttack(Mob *other, DamageHitInfo &hit, ExtraAttackOptions *opts, boo
 //SYNC WITH: tune.cpp, mob.h TuneClientAttack
 bool Mob::Attack(Mob* other, int Hand, bool bRiposte, bool IsStrikethrough, bool IsFromSpell, ExtraAttackOptions *opts)
 {
-	if (!other || other->IsCorpse()) {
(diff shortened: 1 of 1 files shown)


zone/attack.cpp
```diff
@@ -1463,9 +1463,9 @@ void Mob::DoAttack(Mob *other, DamageHitInfo &hit, ExtraAttackOptions *opts, boo
 //SYNC WITH: tune.cpp, mob.h TuneClientAttack
 bool Mob::Attack(Mob* other, int Hand, bool bRiposte, bool IsStrikethrough, bool IsFromSpell, ExtraAttackOptions *opts)
 {
-	if (!other) {
```
Akkadius, line 1466: Should be 

```cpp
if (!other || (other && other->IsCorpse())) {
```

zone/attack.cpp
```diff
@@ -6299,8 +6299,9 @@ void NPC::SetAttackTimer()
 
 void Client::DoAttackRounds(Mob *target, int hand, bool IsFromSpell)
 {
-	if (!target)
+	if (!target || target->IsCorpse()) {
```
Akkadius, line 6302: Should be

```cpp
if (!target || (target && target->IsCorpse())) {
```

Akkadius (approved):

PR #3926, opened 2024-01-08
[Bug Fix] Undead Should Never Flee

fryguy503:
Added a hard return on Undead Body Types. They should never flee under any circumstance.

### DIFF
--- a/zone/fearpath.cpp
+++ b/zone/fearpath.cpp
@@ -49,6 +49,11 @@ void Mob::CheckFlee()
 		return;
 	}
 
+	// Undead do not flee
+	if(GetBodyType() == BT_Undead) {
+		return;
+	}
+
 	// Check if Flee Timer is cleared
 	if (!flee_timer.Check()) {
 		return;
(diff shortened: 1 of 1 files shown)


zone/fearpath.cpp
```diff
@@ -49,6 +49,11 @@ void Mob::CheckFlee()
 		return;
 	}
 
+	// Undead do not flee
+	if(GetBodyType() == BT_Undead) {
```
Akkadius, line 53: `if (` vs `if(`

Akkadius (approved):

joligario (commented):
I feel that the immune to flee flag could handle this. What if a custom server wants them to flee?

fryguy503:
I can rule it?

There are 5886 npc's that are BodyType Undead (3) but only 3948 of those include "21" in their special_abilities for immune to flee.

Akkadius:
I'd rather we lean into sane defaults for this one and have undead just not flee

joligario:
> I'd rather we lean into sane defaults for this one and have undead just not flee

It’s not a default if it is the only option

PR #3928, opened 2024-01-08
[Bug Fix] Increase Precision in CheckDoubleAttack

fryguy503:
On DA Checks where class was a non skill based DA Attacker (Bard or BST with granted DA) The DA check was too steep and should not have been dividing by 500 but rather 100. Also adjusted logic percision to use floats so loss of data does not occur.

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -3726,7 +3726,8 @@ bool Client::CheckDoubleAttack()
 		chance = (float(bonus_give_double_attack + bonus_double_attack) / 100.0f);
 	}
 
-	LogCombatDetail("skill = [{}] - bonus_give_double_attack = [{}] - bonus_double_attack = [{}] - chance = [{}]",
+	LogCombatDetail(
+		"skill [{}] bonus_give_double_attack [{}] bonus_double_attack [{}] chance [{}]",
 		skill,
 		bonus_give_double_attack,
 		bonus_double_attack,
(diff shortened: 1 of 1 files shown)


zone/attack.cpp
```diff
@@ -3702,22 +3702,38 @@ bool Mob::HasRangedProcs() const
...
+	}
+
+	//Use skill calculations otherwise, if you only have AA applied GiveDoubleAttack chance then use that value as the base.
+	if (skill) {
+		chance = (float(skill + GetLevel()) * (float(100.0f + bonus_double_attack + bonus_give_double_attack) / 100.0f)) / 500.0f;
+	} else {
+		chance = (float(bonus_give_double_attack + bonus_double_attack) / 100.0f);
+	}
+
+	LogCombatDetail("DA Skill = [{}] - GiveDoubleAttack = [{}] - DoubleAttackChance = [{}] - Final Double Attack Chance = [{}]",
```
Akkadius, line 3729: When we're printing vars just keep the names consistent with the vars and just brackets

`skill [{}] bonus_give_double_attack [{}] bonus_double_attack [{}] chance [{}]`

Akkadius, line 3729: When we're printing vars just keep the names consistent with the vars and just brackets

`skill [{}] bonus_give_double_attack [{}] bonus_double_attack [{}] chance [{}]`

Kinglykrab (approved):

Akkadius (approved):

PR #3931, opened 2024-01-08
[Rules] Classic Spell Data SPA Calc variability

fryguy503:
When  using lucy imports of older more classic data, they lack the limit values which provides Focus Random Effectiveness.

Example:

Bazu Plauge (6472) went from a 40 limit to 0 limit on the 2006 lucy import.

Lucy does have limit value data, but it appears the affected SPAs sony handled in source vs in spell data.

### DIFF
--- a/zone/spell_effects.cpp
+++ b/zone/spell_effects.cpp
@@ -5892,16 +5892,19 @@ int64 Mob::CalcFocusEffect(focusType type, uint16 focus_id, uint16 spell_id, boo
 
 			// handle effects
 			case SE_ImprovedDamage:
-				if (type == focusImprovedDamage) {
-					value = GetFocusRandomEffectivenessValue(focus_spell.base_value[i], focus_spell.limit_value[i], best_focus);
-				}
-				break;
-
 			case SE_ImprovedDamage2:
-				if (type == focusImprovedDamage2) {
-					value = GetFocusRandomEffectivenessValue(focus_spell.base_value[i], focus_spell.limit_value[i], best_focus);
+				if (!RuleB(Spells, UseClassicSpellFocus)) {
+					if (type == focusImprovedDamage || type == focusImprovedDamage2) {
+						value = GetFocusRandomEffectivenessValue(focus_spell.base_value[i], focus_spell.limit_value[i], best_focus);
+					}
+					break;
+				} else {
+					if (best_focus) {
+						value = focus_spell.base_value[i];
+					} else {
+						value = zone->random.Int(1, focus_spell.base_value[i]);
+					}
 				}
-				break;
 
 			case SE_Fc_Amplify_Mod:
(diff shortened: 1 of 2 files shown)


Kinglykrab (approved):

zone/spell_effects.cpp
```diff
@@ -5967,10 +5973,16 @@ int64 Mob::CalcFocusEffect(focusType type, uint16 focus_id, uint16 spell_id, boo
...
 
 			case SE_ReduceReagentCost:
-				if (type == focusReagentCost) {
-					value = GetFocusRandomEffectivenessValue(focus_spell.base_value[i], focus_spell.limit_value[i], best_focus);
+				if (!RuleB(Spells, UseClassicSpellFocus)) {
+					if (type == focusReagentCost) {
+						value = GetFocusRandomEffectivenessValue(focus_spell.base_value[i], focus_spell.limit_value[i], best_focus);
+					}
+					break;
+				} else {
```
Akkadius, line 5981: ` else {` isn't needed with the `break;` in the above `if`

We also need a `break` where the current logic that is in `else` lies

zone/spell_effects.cpp
```diff
@@ -5991,10 +6003,20 @@ int64 Mob::CalcFocusEffect(focusType type, uint16 focus_id, uint16 spell_id, boo
...
 
 			case SE_SpellHateMod:
-				if (type == focusSpellHateMod) {
-					value = GetFocusRandomEffectivenessValue(focus_spell.base_value[i], focus_spell.limit_value[i], best_focus);
+				if (!RuleB(Spells, UseClassicSpellFocus)) {
+					if (type == focusSpellHateMod) {
+						value = GetFocusRandomEffectivenessValue(focus_spell.base_value[i], focus_spell.limit_value[i], best_focus);
+					}
+					break;
+				} else {
```
Akkadius, line 6011: ` else {` isn't needed with the `break;` in the above `if`

We also need a `break` where the current logic that is in `else` lies

zone/spell_effects.cpp
```diff
@@ -5892,16 +5892,19 @@ int64 Mob::CalcFocusEffect(focusType type, uint16 focus_id, uint16 spell_id, boo
...
-
 			case SE_ImprovedDamage2:
-				if (type == focusImprovedDamage2) {
-					value = GetFocusRandomEffectivenessValue(focus_spell.base_value[i], focus_spell.limit_value[i], best_focus);
+				if (!RuleB(Spells, UseClassicSpellFocus)) {
+					if (type == focusImprovedDamage || type == focusImprovedDamage2) {
+						value = GetFocusRandomEffectivenessValue(focus_spell.base_value[i], focus_spell.limit_value[i], best_focus);
+					}
+					break;
+				} else {
```
Akkadius, line 5901: ` else {` isn't needed with the `break;` in the above `if`

We also need a `break` where the current logic that is in `else` lies

PR #3942, opened 2024-01-09
[Feature] Classic Taunt (Pre 2006) style of taunt

fryguy503:
RULE: ClassicTauntSystem

Completely seperate logic for each system. Could potentially be re-writte to be more inline but there are a good bit of minor differences.

### DIFF
--- a/zone/special_attacks.cpp
+++ b/zone/special_attacks.cpp
@@ -2128,6 +2128,19 @@ void Client::DoClassAttacks(Mob *ca_target, uint16 skill, bool IsRiposte)
 	}
 }
 
+	/* Classic Taunt Methodology
+	* This is not how Sony did it.  This is a guess that fits the very limited data available.
+	* Low level players with maxed taunt for their level taunted about 50% on white cons.
+	* A 65 ranger with 150 taunt skill (max) taunted about 50% on level 60 and under NPCs.
+	* A 65 warrior with maxed taunt (230) was taunting around 50% on SSeru NPCs.		*/
+
+	/* Rashere in 2006: "your taunt skill was irrelevant if you were above level 60 and taunting
+	* something that was also above level 60."
+	* Also: "The chance to taunt an NPC higher level than yourself dropped off at double the rate
+	* if you were above level 60 than if you were below level 60 making it very hard to taunt creature
+	* higher level than yourself if you were above level 60."
+	* 
+	* See http://www.elitegamerslounge.com/home/soearchive/viewtopic.php?t=81156 */
 void Mob::Taunt(NPC *who, bool always_succeed, int chance_bonus, bool from_spell, int32 bonus_hate)
 {
 	if (who == nullptr || DivineAura() || (!from_spell && !CombatRange(who)) || (IsNPC() && IsCharmed()) || !CombatRange(who)) {
@@ -2138,9 +2151,10 @@ void Mob::Taunt(NPC *who, bool always_succeed, int chance_bonus, bool from_spell
 		CastToClient()->CheckIncreaseSkill(EQ::skills::SkillTaunt, who, 10);
 	}
 
-	Mob *hate_top = who->GetHateMost();
-
+	Mob *hate_top        = who->GetHateMost();
 	int level_difference = GetLevel() - who->GetLevel();
+	bool success         = false;
+	int taunt_chance     = 0;
 
 	// Support for how taunt worked pre 2000 on LIVE - Can not taunt NPC over your level.
 	if (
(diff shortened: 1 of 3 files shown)


zone/special_attacks.cpp
```diff
@@ -2128,20 +2128,33 @@ void Client::DoClassAttacks(Mob *ca_target, uint16 skill, bool IsRiposte)
...
+	* something that was also above level 60."
+	* Also: "The chance to taunt an NPC higher level than yourself dropped off at double the rate
+	* if you were above level 60 than if you were below level 60 making it very hard to taunt creature
+	* higher level than yourself if you were above level 60."
+	* 
+	* See http://www.elitegamerslounge.com/home/soearchive/viewtopic.php?t=81156 */
 void Mob::Taunt(NPC *who, bool always_succeed, int chance_bonus, bool from_spell, int32 bonus_hate)
 {
-	if (!who || DivineAura() || (!from_spell && !CombatRange(who))) {
+	if (who == nullptr || DivineAura() || (!from_spell && !CombatRange(who)) || (IsNPC() && IsCharmed()) || !CombatRange(who)) {
```
Kinglykrab, line 2146: `!who`.

Kinglykrab (approved):

zone/special_attacks.cpp
```diff
@@ -2128,20 +2128,33 @@ void Client::DoClassAttacks(Mob *ca_target, uint16 skill, bool IsRiposte)
...
+	* something that was also above level 60."
+	* Also: "The chance to taunt an NPC higher level than yourself dropped off at double the rate
+	* if you were above level 60 than if you were below level 60 making it very hard to taunt creature
+	* higher level than yourself if you were above level 60."
+	* 
+	* See http://www.elitegamerslounge.com/home/soearchive/viewtopic.php?t=81156 */
 void Mob::Taunt(NPC *who, bool always_succeed, int chance_bonus, bool from_spell, int32 bonus_hate)
 {
-	if (!who || DivineAura() || (!from_spell && !CombatRange(who))) {
+	if (!who || DivineAura() || (!from_spell && !CombatRange(who)) || (IsNPC() && IsCharmed()) || !CombatRange(who)) {
```
Akkadius, line 2146: What's the difference between the `!CombatRange` from `(!from_spell && !CombatRange(who))` and the catch all `!CombatRange(who)`

zone/special_attacks.cpp
```diff
@@ -2153,104 +2166,133 @@ void Mob::Taunt(NPC *who, bool always_succeed, int chance_bonus, bool from_spell
...
+		// Taunt Chance Rule Bonus
+		taunt_chance += RuleI(Combat, TauntChanceBonus);
+	}
+
+	//success roll
+	success = zone->random.Roll(taunt_chance);
+
+	// Log result
+	LogHate(
+		"Taunter mob {} target npc {} taunt_chance [{}] success [{}] hate_top [{}]",
```
Akkadius, line 2240: Enclose all `{}` within brackets please `[{}]`

zone/special_attacks.cpp
```diff
@@ -2153,104 +2166,133 @@ void Mob::Taunt(NPC *who, bool always_succeed, int chance_bonus, bool from_spell
...
+
+		MessageString(Chat::Skills, TAUNT_SUCCESS, who->GetCleanName());
 	} else {
-		MessageString(Chat::SpellFailure, FAILED_TAUNT);
+		MessageString(Chat::Skills, FAILED_TAUNT);
 	}
 
-	TryCastOnSkillUse(who, EQ::skills::SkillTaunt);
+	// Modern Abilities
+	if (RuleB(Combat, ClassicTauntSystem)) {
```
Akkadius, line 2286: If this section is supposed to be for modern abilities, wouldn't this be `if (!RuleB(Combat, ClassicTauntSystem)) {` ?

fryguy503: Good Catch, thank you.

fryguy503: Wonder if I should rule this off separately or remove the rule all together. This is less for the taunting itself and more for additional features.

Akkadius: Generally - the more you can avoid deeply nested if/else chains the better. It quickly makes code harder to read and reason about. For this change specifically I would suggest the following

```cpp
	// classic taunt did not utilize modern abilities such as skill procs
	if (RuleB(Combat, ClassicTauntSystem)) {
		return;
	}
```

Akkadius: image

fryguy503: I agree with cleaning up nested ifs for sure.

What I am referencing is the `TrySkillProc` is less directly tied to taunt or classic taunt and is a separate system and I am thinking that it shouldn't be ruled out as part of the classic taunt system. If someone wants to have SkillProcs from Taunt and still use the classic taunt system they should be able to.

PR #3954, opened 2024-01-10
[NPCs] Convert Load of NPC Emotes to Repositories

Kinglykrab:
# Notes
- Convert `LoadNPCEmotes()` to repositories.

# Images
## Enter Combat
image
## Hail
image

### DIFF
--- a/common/emu_constants.h
+++ b/common/emu_constants.h
@@ -350,6 +350,7 @@ namespace EQ
 		};
 
 		enum EmoteTypes : uint8 {
+			Say,
 			Emote,
 			Shout,
 			Proximity
--- a/zone/zone.cpp
+++ b/zone/zone.cpp
@@ -62,6 +62,7 @@
 #include "../common/repositories/ldon_trap_entries_repository.h"
 #include "../common/repositories/ldon_trap_templates_repository.h"
 #include "../common/repositories/respawn_times_repository.h"
+#include "../common/repositories/npc_emotes_repository.h"
 #include "../common/serverinfo.h"
 
 #include <time.h>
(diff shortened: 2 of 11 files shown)


zone/zone.cpp
```diff
@@ -2583,27 +2584,29 @@ void Zone::DoAdventureActions()
...
-    {
-	    auto nes = new NPC_Emote_Struct;
-	    nes->emoteid = Strings::ToInt(row[0]);
-	    nes->event_ = Strings::ToInt(row[1]);
-	    nes->type = Strings::ToInt(row[2]);
-	    strn0cpy(nes->text, row[3], sizeof(nes->text));
-	    NPCEmoteList->push_back(nes);
-    }
+	for (const auto& e : l) {
+		auto n = new NPC_Emote_Struct;
```
Akkadius, line 2594: Do we really need to allocate `new` for this ?

Akkadius (changes requested):
Looks straightforward - but need testing

zone/zone.cpp
Kinglykrab: Not allocating new it seems to break all emotes.

common/emu_constants.h
```diff
@@ -350,6 +350,7 @@ namespace EQ
 		};
 
 		enum EmoteTypes : uint8 {
+			Say,
```
Akkadius, line 353: So putting this here is a big no-no. 

These are stored in the database in the order they are in this enum and you just threw `Say` (0) where `Emote` (0) is otherwise

This is stored as type and emote is (0). If you're going to add say add it to the end 

image

Kinglykrab: image
`Say` is `0`. The `enum` was wrong.

Kinglykrab: We only use types `0` to `3` and we didn't have `0` in the enum and we had `Emote`/`Shout`/`Proximity` as `0`/`1`/`2` incorrectly when they are `1`/`2`/`3`.

Akkadius: You're right, it was there all along. It looked like you added something to the source of truth enum for types

Akkadius (approved):

PR #3968, opened 2024-01-12
[Pets] Convert Pets to Repositories

Kinglykrab:
# Notes
- Convert `GetPoweredPetEntry()` to repositories.

# Images
## Summon
image

### DIFF
--- a/zone/pets.cpp
+++ b/zone/pets.cpp
@@ -20,6 +20,8 @@
 #include "../common/spdat.h"
 #include "../common/strings.h"
 
+#include "../common/repositories/pets_repository.h"
+
 #include "entity.h"
 #include "client.h"
 #include "mob.h"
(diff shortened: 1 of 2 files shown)


zone/pets.cpp
```diff
@@ -369,38 +371,33 @@ Pet::Pet(NPCType *type_data, Mob *owner, PetType type, uint16 spell_id, int16 po
...
-	auto row = results.begin();
-
-	into->npc_type = Strings::ToInt(row[0]);
-	into->temporary = Strings::ToInt(row[1]);
-	into->petpower = Strings::ToInt(row[2]);
-	into->petcontrol = Strings::ToInt(row[3]);
-	into->petnaming = Strings::ToInt(row[4]);
-	into->monsterflag = Strings::ToInt(row[5]);
-	into->equipmentset = Strings::ToInt(row[6]);
+	r->npc_type     = l[0].npcID;
```
Akkadius, line 394: Instead of directly using the first entry lets set another accessor

```cpp
auto p = l[0]; 
```

then just use p below

Akkadius:
Where's the testing

Akkadius (approved):

PR #3962, opened 2024-01-12
[Global Loot] Convert Global Loot to Repositories

Kinglykrab:
# Notes
- Convert `LoadGlobalLoot()` to repositories.

# Images
## Load
image
image

PR #3969, opened 2024-01-13
[Repository] Add null integer column support, instance_list notes migration, regenerate repositories

Akkadius:
**Repository Generator**

* Add support for nullable integer, float, decimal columns - not having support for this caused a crash with Global Loot in PR https://github.com/EQEmu/Server/pull/3962 - This will default to the column defined default value if value is null

**General**

* All base repositories generated - Many repositories missing newly added base methods, primary key id updates etc.
* Database migration added for `instance_list` to include `notes` column being distributed in PEQ database dumps but not present in current migrations to conform with repository schema


Kinglykrab (approved):

Akkadius:
**Testing**

Logged into game, world booted clean, zones booted clean, no query errors etc.

image

PR #3974, opened 2024-01-14
[Merchants] Change database structure for merchant slots

joligario:
The `slot` field should be unsigned int. The temp slot is currently limited to 127. The client will limit the visible slots. Fixes selling to merchants with inventory >127 as seen in:
```[**QueryErr**] **Zone** [**poknowledge**] MySQL Error (1264) [Out of range value for column 'slot' at row 1] Query [REPLACE INTO merchantlist_temp (npcid, slot, zone_id, instance_id, itemid, charges) VALUES(202173, 311, 202, 0, 26708, 1)]```

image

image


Kinglykrab (approved):

joligario:
image

image

image

PR #3979, opened 2024-01-14
[Bug Fix] Reversed logic on InLiquid

fryguy503:
Thank you @noudness for catching this. Reversed logic on InLiquid check for falling damage.


noudess (approved):
LGTM.

PR #3980, opened 2024-01-14
[Spawn2] Spawn condition value should default spawn_conditions value

noudess:
The spawn_conditions table has a default value for the condition.  This was not used at bootup.

I have zones where spawn2 locations have the same condition, but what spawns depends on the value 1, 2, 3 etc.

These spots are spawning with a bootup value of 0 instead of what was in the spawn_conditions table.

Yes, they fix themselves as either spawn_events or quests update the condition, but the bootup value should not always be 0.

I fixed the LogSpawn lines because the commas were confusing the formatting engine and ended up inside the [].

### DIFF
--- a/zone/spawn2.cpp
+++ b/zone/spawn2.cpp
@@ -966,8 +966,14 @@ bool SpawnConditionManager::LoadSpawnConditions(const char* zone_name, uint32 in
 			StrictCheck = true;
 
 		//If event is disabled, or we failed the strict check, set initial spawn_condition to default startup value from spawn_conditions.
-		if(!cevent.enabled || !StrictCheck)
-			SetCondition(zone->GetShortName(), zone->GetInstanceID(),cevent.condition_id,spawn_conditions[cevent.condition_id].value);
+		if(!cevent.enabled || !StrictCheck) {
+			SetCondition(
+					zone->GetShortName(),
+					zone->GetInstanceID(),
+					cevent.condition_id,
+					spawn_conditions[cevent.condition_id].value
+			);
+		}
 
 		if(!cevent.enabled)
             continue;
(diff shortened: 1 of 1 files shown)


zone/spawn2.cpp
```diff
@@ -965,9 +965,9 @@ bool SpawnConditionManager::LoadSpawnConditions(const char* zone_name, uint32 in
 			cevent.next.year == tod.year)
 			StrictCheck = true;
 
-		//If event is disabled, or we failed the strict check, set initial spawn_condition to 0.
+		//If event is disabled, or we failed the strict check, set initial spawn_condition to default startup value from spawn_conditions.
 		if(!cevent.enabled || !StrictCheck)
```
Kinglykrab, line 969:
```cpp
if () {
}
```

zone/spawn2.cpp
```diff
@@ -965,9 +965,9 @@ bool SpawnConditionManager::LoadSpawnConditions(const char* zone_name, uint32 in
...
-			SetCondition(zone->GetShortName(), zone->GetInstanceID(),cevent.condition_id,0);
+			SetCondition(zone->GetShortName(), zone->GetInstanceID(),cevent.condition_id,spawn_conditions[cevent.condition_id].value);
```
Kinglykrab, line 970: Formatting.
```cpp
SetCondition(
	zone->GetShortName(),
	zone->GetInstanceID(),
	cevent.condition_id,spawn_conditions[cevent.condition_id].value
);
```

zone/spawn2.cpp
noudess: Fixed.

zone/spawn2.cpp
noudess: Fixed

Kinglykrab (approved):

PR #3901, opened 2024-01-08
[Bug Fix] Disciplines should show when someone casts them.

fryguy503:

PR #3986, opened 2024-01-15
[Messages] Fix discipline message added by previous patch.

noudess:
Fix the new added message for disciplines in PR #[3901](https://github.com/EQEmu/Server/pull/3901).

The new message needed to use spelltar to retrieve the target of ths discipline.


fryguy503 (approved):
Thank you for catching this!

joligario (approved):

PR #3987, opened 2024-01-15
Push discpline messages under melee->disciplines

noudess:
To be able to move these messages from chat window to chat window, we need to send an actually packet with spell id.

All disciplines should:

- show in chat windows under melee->disciplines
- color changes in client done under special abilities


fryguy503 (approved):
Thank you for testing and refining this @noudess

noudess:
Hold off on merge fdr a day please.

noudess:
This looks good here.

PR #3989, opened 2024-01-16
[Zoning] Additional logs for zoning under instance checks

Akkadius:
This adds logging in cases where we validate instance checks. The client doesn't actually get these messages when we send a zone cancel and can be hard to pin down what is happening if you are getting rejected.


Kinglykrab (approved):

joligario (approved):
Recommend adding the target_instance_id to the log entry

PR #3992, opened 2024-01-17
[Objects] Add `fix_z` column to ground spawns

Kinglykrab:
# Notes
- Allows ground spawns to float without having `FixZ()` called.

# Images
## Ground spawn with `fix_z` of `0`
image
## Ground spawn with `fix_z` of `1`
image


Kinglykrab:
Tried making this work with regular objects but couldn't, don't see anywhere where we're doing any extra Z fixing or anything like that for objects. Z would not go above the floor no matter what.

Akkadius:
I love this change. 

I do think there is an opportunity to make sure our naming makes sense since we might apply this to other things going forward. 

`is_floating` is a little bit of a weird name to me - it implies that the object is intended to float in the air, which doesn't seem desirable and can be confusing first glance. 

What we're really implying is that we don't want the server to fix it's Z. It would make sense to me for us to make this `fix_z` and default it to `1` so that an operator can set it to `0` so it doesn't "Fix Z"

Akkadius (approved):

PR #3998, opened 2024-01-19
[Bot] ^mez command spell list fix

dariusuknuis:
This change will allow necros to cast mez spells when ^mez command is used.


Kinglykrab (approved):
@nytmyr has said this is a partial fix and should at the very least make the `^mez` command function.

PR #4000, opened 2024-01-20
[Bug Fix] Fix Mercenaries Buffs/Zoning Issues

Kinglykrab:
# Notes
- Mercenaries were disappearing on zoning because they were saving to slot `1` and being loaded from slot `0`.
- Mercenaries were not displaying properly in group window because of this as well.
- Mercenary buffs were saving even when they did not have any buffs.

### DIFF
--- a/zone/merc.cpp
+++ b/zone/merc.cpp
@@ -5575,20 +5575,17 @@ Merc* Client::GetMerc() {
 	return (tmp);
 }
 
-uint8 Client::GetNumberOfMercenaries() {
+uint8 Client::GetNumberOfMercenaries()
+{
+	uint8 count = 0;
 
-	uint8 numMercs = 0;
-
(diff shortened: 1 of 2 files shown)


zone/merc.cpp
```diff
@@ -5575,20 +5575,17 @@ Merc* Client::GetMerc() {
 	return (tmp);
 }
 
-uint8 Client::GetNumberOfMercenaries() {
-
-	uint8 numMercs = 0;
+uint8 Client::GetNumberOfMercenaries()
+{
```
joligario, line 5579: Moved bracket?

joligario (approved):
Just a minor comment

zone/merc.cpp
Kinglykrab: We prefer newline for function definitions from what I’ve seen in modern code. Inconsistent over the years.

PR #4010, opened 2024-01-24
[Bots] Convert Bot Commands to individual files

Kinglykrab:
# Notes
- Converts bot commands to individual files.
- Tested with random commands like create, spawn, face, hair color, etc.
- Some extra eyes by any bot users that self-compile would be great.


Kinglykrab:
I will revisit this once #4023 is merged.

Issue #4013, opened 2024-01-24
Cannot make a new froglok character on a fresh Akk-Stack server install - Repeatable 


bullr:
Setting up a new server using the latest Akk Script installer (server source as of 1/16/24) for Linux results in a working server however, froglok characters are not able to be created at the character select screen. This results in an immediate client disconnect from the server. All other races work fine. Confirmed database expansion settings in rules_values are correct. Drakkin race works so frogs should also work if the issue was an incorrect expansion setting.

On creation of a froglok toon and hitting Enter World the character will never be handed over to a zone server, and will also not be saved into the DB.

Tested on multiple fresh installs, as well as a long standing server (years) that was upgraded using the latest sources last week (1/16/24). The server stopped allowing froglok creation right after the update. Previous update was done in August, and frogs were working.

PR #4016, opened 2024-01-25
[Crash Fix] Fix crash when creating Frogloks/Drakkin

Kinglykrab:
# Notes
- https://github.com/EQEmu/Server/pull/3920 introduced an issue where we were using `uchar` for `race_selection` and `class_selection` which was not functioning properly within `IsPlayerRace()`.
- `uchar` is an `unsigned char` which has a max of `255`, so `330` (Froglok) and `522` (Drakkin) were causing unexpected outcomes.
- Fixes https://github.com/EQEmu/Server/issues/4013.

### DIFF
--- a/world/client.cpp
+++ b/world/client.cpp
@@ -535,67 +535,64 @@ bool Client::HandleNameApprovalPacket(const EQApplicationPacket *app)
 		return false;
 	}
 
-	auto length = snprintf(char_name, 64, "%s", (char*)app->pBuffer);
+	auto n = (NameApproval_Struct*) app->pBuffer;
 
-	uchar race_selection  = app->pBuffer[64];
-	uchar class_selection = app->pBuffer[68];
+	strn0cpy(char_name, n->name, sizeof(char_name));
 
-	if (!IsPlayerRace(race_selection)) {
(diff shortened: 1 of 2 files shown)


world/client.cpp
```diff
@@ -535,67 +535,64 @@ bool Client::HandleNameApprovalPacket(const EQApplicationPacket *app)
...
-	
-	LogInfo("Name approval request. Name=[{}], race_selection=[{}], class=[{}]", char_name, GetRaceIDName(race_selection), GetClassIDName(class_selection));
 
-	EQApplicationPacket *outapp;
-	outapp = new EQApplicationPacket;
-	outapp->SetOpcode(OP_ApproveName);
-	outapp->pBuffer = new uchar[1];
-	outapp->size = 1;
+	LogInfo(
+		"Name approval request. Name=[{}], race_selection=[{}], class=[{}]",
```
Akkadius, line 557: The function is included in logs now so the first part no longer makes sense

```cpp
LogInfo("name [{}] race_selection [{}] class [{}]");
```

Akkadius (approved):

PR #4018, opened 2024-01-26
[Bug Fix] Limit Pet Taunt Distance

fryguy503:
Previously this was not regulated and allowed players to exploit unlimited taunt distance.

Rule Name: PetTauntRange
Rule Default: 150

Calculation is Rule Squared.

### DIFF
--- a/zone/special_attacks.cpp
+++ b/zone/special_attacks.cpp
@@ -1819,7 +1819,8 @@ void NPC::DoClassAttacks(Mob *target) {
 		target->GetBodyType() != BT_Undead &&
 		taunt_time &&
 		type_of_pet &&
-		type_of_pet != petTargetLock
+		type_of_pet != petTargetLock &&
+		DistanceSquared(this->GetPosition(), target->GetPosition()) <= (RuleI(Pets, PetTauntRange)*RuleI(Pets, PetTauntRange))
 	) {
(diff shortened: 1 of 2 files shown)


zone/special_attacks.cpp
```diff
@@ -1819,7 +1819,8 @@ void NPC::DoClassAttacks(Mob *target) {
 		target->GetBodyType() != BT_Undead &&
 		taunt_time &&
 		type_of_pet &&
-		type_of_pet != petTargetLock
+		type_of_pet != petTargetLock &&
+		DistanceSquared(this->GetPosition(), target->GetPosition()) <= (RuleI(Pets, PetTauntRange)*RuleI(Pets, PetTauntRange))
```
Kinglykrab, line 1823: No need for `this`. Can also just create a variable to hold pet taunt range.
```cpp
const int pet_taunt_range = (RuleI(Pets, PetTauntRange) * RuleI(Pets, PetTauntRange));
```

fryguy503: Will remove `this` but is a variable needed when its only referenced once? e.g. is that general best practice going forward?

Kinglykrab: We can leave it as is.

Kinglykrab (approved):

PR #4019, opened 2024-01-26
[Bug Fix] Limit Player Taunt Distance

fryguy503:
Previously this was not regulated on the server side and allowed players to exploit situations where the client did not enforce Z distance checks.

Rule Name: MaximumTauntDistance
Rule Default: 150

Calculation is Rule Squared

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -14905,6 +14905,11 @@ void Client::Handle_OP_Taunt(const EQApplicationPacket *app)
 		return;
 	}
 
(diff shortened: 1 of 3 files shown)


zone/client_packet.cpp
```diff
@@ -14905,6 +14905,11 @@ void Client::Handle_OP_Taunt(const EQApplicationPacket *app)
 		return;
 	}
 
+	if (DistanceSquared(this->GetPosition(), GetTarget()->GetPosition()) > (RuleI(Skills, MaximumTauntDistance) * (RuleI(Skills, MaximumTauntDistance)))) {
```
Kinglykrab, line 14908: No need for `this`. Can also just create a variable to hold pet taunt range.
```cpp
const int player_taunt_range = (RuleI(Skills, MaximumTauntDistance) * RuleI(Skills, MaximumTauntDistance));
```

Kinglykrab (approved):

PR #4022, opened 2024-01-27
[Cleanup] Add Raid Loot Type Constants

Kinglykrab:
# Notes
- Add constants for Raid loot types.

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -2764,6 +2764,7 @@ bool NPC::Death(Mob* killer_mob, int64 damage, uint16 spell, EQ::skills::SkillTy
 						}
 
 						switch (r->GetLootType()) {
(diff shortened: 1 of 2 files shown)


zone/attack.cpp
```diff
@@ -2764,31 +2764,31 @@ bool NPC::Death(Mob* killer_mob, int64 damage, uint16 spell, EQ::skills::SkillTy
 						}
 
 						switch (r->GetLootType()) {
-						case 0:
```
Akkadius, line 2767: What happened to 0

Kinglykrab: It's unused, client doesn't have an option `0`, only `1`, `2`, `3`, and `4`.

Akkadius (approved):

PR #4023, opened 2024-01-27
 [Bots] Convert Bot Database Methods to Repositories

Kinglykrab:
# Notes
- Converts all bot database methods in `bot_database.cpp` to repositories.
- Removes foreign key constraints as they are no longer necessary.
- Adds `interval` to reserved words in repository generator.
- Cleans up some code where we were doing individual queries inside of loops instead of a mass insert/delete query.
- Removes unused methods.
- Removes unnecessary `error_message` pointer that was passed around in a few methods.
- Removes `fail` methods that were unnecessary.
- Removes `owner_id` parameter from methods that did not need `owner_id` for their queries.
- Regenerates all bot-based repositories.


Akkadius (approved):
This is bots so I'm less concerned about the details and trust you tested this

PR #4039, opened 2024-02-02
[Fixes for PR # 3999

noudess:
These are the naming changes I proposed to use correct naming for:

npc_faction_id - an ID that is an umbrella ID indexing into behavior, hits, primary faction, etc.
faction_id - an id that indexes an actual in game faction like Guards of Qeynos

Also restoring ShowStats to show in game faction, not the umbrella name.

Also I believe NpcFactionRepository::NpcFaction* Zone::GetNPCFaction(const uint32 npc_faction_id) was retrieving the incorrect entry based on naming confustion.  We're passing in an npc_faction_id and expecting back the umbrella data.  The comparison to primary_faction_id implies that the npc_faction_id we got in was a faction.

Restored std::string ZoneDatabase::GetFactionName(int32 faction_id) back to returning information about a faction, rather than an npc_faction_id.

Also added code to load the actual in game factions in addition to the umbralla data to the OP_Reload functionality,

I still need help understanding the GetNPCFactionEntries  code.

### DIFF
--- a/zone/zone_npc_factions.cpp
+++ b/zone/zone_npc_factions.cpp
@@ -119,10 +119,10 @@ void Zone::ReloadNPCFactions()
 	LoadNPCFactions(faction_ids);
 }
 
-NpcFactionRepository::NpcFaction* Zone::GetNPCFaction(const uint32 faction_id)
+NpcFactionRepository::NpcFaction* Zone::GetNPCFaction(const uint32 npc_faction_id)
 {
 	for (auto& e : m_npc_factions) {
-		if (e.primaryfaction == faction_id) {
+		if (e.id == npc_faction_id) {
 			return &e;
 		}
 	}
@@ -130,7 +130,7 @@ NpcFactionRepository::NpcFaction* Zone::GetNPCFaction(const uint32 faction_id)
 	return nullptr;
 }
 
-std::vector<NpcFactionEntriesRepository::NpcFactionEntries> Zone::GetNPCFactionEntries(const uint32 faction_id) const
+std::vector<NpcFactionEntriesRepository::NpcFactionEntries> Zone::GetNPCFactionEntries(const uint32 npc_faction_id) const
 {
 	std::vector<NpcFactionEntriesRepository::NpcFactionEntries> npc_faction_entries = { };
 
@@ -138,7 +138,7 @@ std::vector<NpcFactionEntriesRepository::NpcFactionEntries> Zone::GetNPCFactionE
 
 	for (auto e : m_npc_faction_entries) {
 		if (
-			e.npc_faction_id == faction_id &&
+			e.npc_faction_id == npc_faction_id &&
 			std::find(
 				faction_ids.begin(),
 				faction_ids.end(),
--- a/zone/zone_npc_factions.cpp
+++ b/zone/zone_npc_factions.cpp
@@ -3,9 +3,19 @@
 #include "../common/repositories/npc_faction_repository.h"
 #include "../common/repositories/npc_faction_entries_repository.h"
 
-void Zone::LoadNPCFactions(const std::vector<uint32> &faction_ids)
+void Zone::LoadNPCFactions(const std::vector<uint32> &npc_faction_ids)
 {
-	for (const auto& e : faction_ids) {
+	LogFaction(
+		"Load for Faction IDs [{}]",
+		Strings::Join(npc_faction_ids, ", ")
+	);
+
+	if (npc_faction_ids.empty()) {
+		LogFactionDetail("No NPC factions to load.");
+		return;
+	}
+
+	for (const auto& e : npc_faction_ids) {
 		for (const auto& f : m_npc_factions) {
 			if (e == f.id) {
 				LogFactionDetail("Faction [{}] already loaded.", e);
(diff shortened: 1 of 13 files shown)


noudess:
Added fix for primary_faction not being filled in when umbrella has no hits or aggro info.

noudess:
Ok, looks like factions from kills works now for both times.

I still need to do alot more testing on other faction issues.

Fixed this push:

- mods can be < 0
- more naming as needed where faction_id was using instead of npc_faction_id
- Removed a chunk of LoadNPCFactions that went out to get an npc_faction_id list, when that is was was being passed in
- Added LogFaction() code all over.  We can leave or take out when I am done.
- LoadFactionsAssociations needed to be rewritten.  There are up to 10 associations per one db fetch - all in the same line.
- The key for faction_associations is id, not id_1.  Fixed in several places and tested.

To test these yourself, you have to:

- Give an npc an npc_faction_id that is valid, has a primary_faction and no entries in npc_faction_entries, plus fill in the faction_amount field in npc_types for the base.
- Create a line in faction_associations for the primary_faction and some mods.

I was rebooting zone to test.,

Tomorrow I will try the reload and see if that all works with the above.

If that works, I'll move onto splash faction from quests.

noudess:
Did a new push.  

- LoadNPCFactions can now accept a vector with ids already loaded mixed with ids that have not been,  Adds as needed.
- LoadFactionAssociations can now accept a vector with ids already loaded mixed with ids that have not been,  Adds as needed.

zone/zone_npc_factions.cpp
```diff
@@ -15,28 +15,46 @@ void Zone::LoadNPCFactions(const std::vector<uint32> &npc_faction_ids)
...
+	// as the result of a new spawn.  Ignore those already loaded.
+
+	std::vector<uint32> new_npc_faction_ids = { };
+
 	for (const auto& e : npc_faction_ids) {
-		for (const auto& f : m_npc_factions) {
-			if (e == f.id) {
-				LogFactionDetail("Faction [{}] already loaded.", e);
-				return;
+		bool found=0;
```
Kinglykrab, line 24: `bool found = false;`

zone/zone_npc_factions.cpp
```diff
@@ -164,14 +173,30 @@ void Zone::LoadFactionAssociations(const std::vector<uint32>& npc_faction_ids)
 
 	for (const auto& e : npc_faction_ids) {
 		for (const auto& f : m_npc_factions) {
+			bool found=false;
```
Kinglykrab, line 176: `bool found = false;`

zone/zone_npc_factions.cpp
```diff
@@ -164,14 +173,30 @@ void Zone::LoadFactionAssociations(const std::vector<uint32>& npc_faction_ids)
...
 			if (e == f.id && f.primaryfaction > 0) {
-				faction_ids.emplace_back(f.primaryfaction);
+				for (auto a : m_faction_associations) {
+					if (a.id == f.primaryfaction) {
+						found=true;
```
Kinglykrab, line 180: `found = true;`

zone/zone_npc_factions.cpp
```diff
@@ -164,14 +173,30 @@ void Zone::LoadFactionAssociations(const std::vector<uint32>& npc_faction_ids)
...
+						LogError("Association [{}] already loaded", a.id);
+					break;
```
Kinglykrab, line 182: Weird indentation.

zone/zone_npc_factions.cpp
```diff
@@ -164,14 +173,30 @@ void Zone::LoadFactionAssociations(const std::vector<uint32>& npc_faction_ids)
 
 	for (const auto& e : npc_faction_ids) {
 		for (const auto& f : m_npc_factions) {
+			bool found=false;
 			if (e == f.id && f.primaryfaction > 0) {
-				faction_ids.emplace_back(f.primaryfaction);
+				for (auto a : m_faction_associations) {
```
Kinglykrab, line 178: `const auto& a`?

zone/zonedb.cpp
```diff
@@ -1980,7 +1980,7 @@ const NPCType *ZoneDatabase::LoadNPCTypesData(uint32 npc_type_id, bool bulk_load
 
 	DataBucket::BulkLoadEntities(DataBucketLoadType::NPC, npc_ids);
 
-	if (bulk_load) {
+	if (!npc_faction_ids.empty()) {
```
Kinglykrab, line 1983: Where do we use `bulk_load`?

noudess:
Added support for tasks using npc_faction_id for giving out faction.

zone/zonedb.cpp
noudess: I added that for awhile to isolate what was being loaded so I could test.  Not needed here.  I added the empty check just to reduce instrumentation and an unneeded call.

zone/zone_npc_factions.cpp
noudess: fixed

zone/zone_npc_factions.cpp
noudess: fixed

zone/zone_npc_factions.cpp
noudess: fixed

zone/zone_npc_factions.cpp
noudess: I don't seee this issue locally - tabs in github?

zone/zone_npc_factions.cpp
noudess: fixed.

noudess:
Tomorrow I'll add suppport for tasks and quests using splash faction rewards.

zone/zone_npc_factions.cpp
Kinglykrab: Probably tabs versus spaces.

noudess: Fixed in next push.

noudess:
Support for splash tasks and quests along with reformatting of some ill formatted stuff pushed.

PR #4044, opened 2024-02-04
[Quest API] Add GetAAEXPPercentage() and GetEXPPercentage() to Perl/Lua

Kinglykrab:
# Perl
- Add `$client->GetAAEXPPercentage()`.
- Add `$client->GetEXPPercentage()`.

# Lua
- Add `client:GetAAEXPPercentage()`.
- Add `client:GetEXPPercentage()`.

# Notes
- Allows operators to get current AA or regular experience percentages.
- Math is verified by @hgtw to match what the client does so we're not +/- 1%.

### DIFF
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -12004,3 +12004,23 @@ void Client::SetEXPModifier(uint32 zone_id, float exp_modifier, int16 instance_v
 
 	database.LoadCharacterEXPModifier(this);
 }
+
+float Client::GetAAEXPPercentage()
+{
+	const float current = GetAAXP();
+	const float max     = GetRequiredAAExperience();
+
+	const float percentage = ((current / max) * 100);
+
+	return std::round(percentage * 100) / 100;
+}
+
+float Client::GetEXPPercentage()
+{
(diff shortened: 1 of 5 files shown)


hgtw (approved):
Note to match the client tooltip these depend on matching the scaled values sent to client in exp packets. Also the math is slightly different than what the client does (they use a different double constant 0.303031 than 100.0 / 330.0 and the gauge value is multiplied by 10 when stored before it's rounded which could just be optimization).

This tested ok for all possible values when I tried it though. Not sure on use case vs returning a more precise float that quests can choose to round themselves or not but this seems low impact enough.

Kinglykrab:
> Note to match the client tooltip these depend on matching the scaled values sent to client in exp packets. Also the math is slightly different than what the client does (they use a different double constant 0.303031 than 100.0 / 330.0 and the gauge value is multiplied by 10 when stored before it's rounded which could just be optimization).
> 
> This tested ok for all possible values when I tried it though. Not sure on use case vs returning a more precise float that quests can choose to round themselves or not but this seems low impact enough.

Yeah, I'm fine with just having the rounded number since that's how it's displayed in the UI.

PR #4054, opened 2024-02-10
[Repositories] Convert database.cpp to Repositories

Kinglykrab:
# Notes
- Convert all `database.cpp` methods to repositories where possible.

# Videos of Copy Character
[Screencast from 02-19-2024 07:27:37 PM.webm](https://github.com/EQEmu/Server/assets/89047260/eb705208-6fcc-49d0-b1b4-028be813cd33)
[Screencast from 02-19-2024 07:28:20 PM.webm](https://github.com/EQEmu/Server/assets/89047260/a931cdd6-53af-41b2-ac09-248934fee1a1)


common/repositories/base/base_variables_repository.h
```diff
@@ -135,7 +135,7 @@ class BaseVariablesRepository {
 			e.varname     = row[1] ? row[1] : "";
 			e.value       = row[2] ? row[2] : "";
 			e.information = row[3] ? row[3] : "";
-			e.ts          = row[4] ? row[4] : std::time(nullptr);
+			e.ts          = row[4] ? reinterpret_cast<time_t>(row[4]) : std::time(nullptr);
```
Akkadius, line 138: ts is a string object in the struct

Kinglykrab: By default this had issues is why I added the `reinterpret_cast` to `variables` and `account_ip`.
image
image

Kinglykrab: We weren't using these repositories before so I guess it was never an issue.

utils/scripts/generators/repository-generator.pl
```diff
@@ -348,6 +348,10 @@
             $all_entries      .= sprintf("\t\t\te.%-${longest_column_length}s = row[%s] ? strtod(row[%s], nullptr) : %s;\n", $column_name_formatted, $index, $index, $default_value);
             $find_one_entries .= sprintf("\t\t\te.%-${longest_column_length}s = row[%s] ? strtod(row[%s], nullptr) : %s;\n", $column_name_formatted, $index, $index, $default_value);
         }
+        elsif ($data_type =~ /timestamp/i) {
+            $all_entries      .= sprintf("\t\t\te.%-${longest_column_length}s = row[%s] ? reinterpret_cast<time_t>(row[%s]) : %s;\n", $column_name_formatted, $index, $index, $default_value);
+            $find_one_entries .= sprintf("\t\t\te.%-${longest_column_length}s = row[%s] ? reinterpret_cast<time_t>(row[%s]) : %s;\n", $column_name_formatted, $index, $index, $default_value);
+        }
```
Akkadius, line 354: We don't really support timestamp (we support datetime) but this is only supporting one of the pieces in the generator. Where is the time object handling on the struct type

common/repositories/base/base_variables_repository.h
Akkadius: We need to implement the type on the struct `time_t` - for datetime we do translations between unix timestamp because that's what `time_t` represents and then we `FROM_UNIXTIME` when we save. See `datetime` handling in repository generator

common/database.cpp
```diff
@@ -58,396 +74,311 @@
...
-		"SELECT id, status FROM account WHERE `name` = '{}' AND ls_id = '{}' AND password is NOT NULL "
-		"AND length(password) > 0 AND (password = '{}' OR password = MD5('{}'))",
-		temporary_username,
-		Strings::Escape(loginserver),
-		temporary_password,
-		temporary_password
+	const auto& l = AccountRepository::GetWhere(
+		*this,
+		fmt::format(
+			"`name` = '{}' AND `ls_id` = {} AND `password` IS NOT NULL AND LENGTH(`password`) > 0 AND (`password` = '{}' OR `password` = MD5('{}'))",
```
Akkadius, line 127: `ls_id` value is quoted

common/database.cpp
```diff
@@ -813,292 +744,275 @@ uint32 Database::GetCharacterID(const char *name) {
...
 	}
 
-	auto query = fmt::format(
-		"SELECT `id`, `status`, `lsaccount_id` FROM `account` WHERE `name` = '{}' AND `ls_id` = '{}' LIMIT 1",
-		Strings::Escape(account_name),
-		Strings::Escape(loginserver)
+	const auto& l = AccountRepository::GetWhere(
+		*this,
+		fmt::format(
+			"`name` = '{}' AND `ls_id` = '{}'",
```
Akkadius, line 786: This should continue to be limit 1

common/database.cpp
```diff
@@ -813,292 +744,275 @@ uint32 Database::GetCharacterID(const char *name) {
...
+	} else if (Strings::EqualFold(field_name, "crc_eqgame")) {
+		e.crc_eqgame = checksum;
+	} else if (Strings::EqualFold(field_name, "crc_skillcaps")) {
+		e.crc_skillcaps = checksum;
+	}
+
+	AccountRepository::UpdateOne(*this, e);
+}
+
+bool Database::GetZoneGraveyard(
```
Akkadius, line 967: This function is a terrible access pattern, passing pointers and returning their values. This function is only called once so lets remove it. 

Do the repository call where this function is being called and just use it inline in the code

```cpp
bool GraveYardLoaded = content_db.GetZoneGraveyard(graveyard_id(), &pgraveyard_zoneid, &m_graveyard.x, &m_graveyard.y, &m_graveyard.z, &m_graveyard.w);
```

common/database.cpp
```diff
@@ -2134,137 +1919,149 @@ struct TimeOfDay_Struct Database::LoadTime(time_t &realtime)
...
+	return e.id;
 }
 
-/**
- * @param source_character_name
- * @param destination_character_name
- * @param destination_account_name
- * @return
- */
 bool Database::CopyCharacter(
```
Akkadius, line 2004: Test #copycharacter

common/database.cpp
```diff
@@ -58,396 +74,311 @@
...
-		query = StringFormat(
-			"INSERT INTO account SET name='%s', status=%i, ls_id='%s', lsaccount_id=%i, time_creation=UNIX_TIMESTAMP();",
-			name,
-			status,
-			loginserver,
-			lsaccount_id
-		);
+	e.name          = name;
+	e.status        = status;
+	e.lsaccount_id  = lsaccount_id;
```
Akkadius, line 220: `ls_id` is entirely missing from this and is very critical

common/database.cpp
```diff
@@ -58,396 +74,311 @@
...
+	e.time_creation = std::time(nullptr);
+
+	if (!Strings::EqualFold(password, std::to_string(0))) {
```
Akkadius, line 223: What is this doing? Why not just do a not empty check

common/database.cpp
```diff
@@ -1128,979 +1042,850 @@ bool Database::CheckNameFilter(std::string name, bool surname)
...
+	return CharacterDataRepository::UpdateOne(*this, e);
 }
 
-uint8 Database::GetSkillCap(uint8 skillid, uint8 in_race, uint8 in_class, uint16 in_level)
+uint32 Database::GetCharacterInfo(
+	const std::string& name,
+	uint32* account_id,
+	uint32* zone_id,
+	uint32* instance_id
+)
```
Akkadius, line 1193: This function is used in one place and should be removed and repository used inline where it is in the code. These pointer return loader functions are gross

Kinglykrab:
Will go through and resolve comments this afternoon.

Akkadius:
I'll go through this again in the next few days

Akkadius:
Please rebase this one so we can close the loop on this one soon as well. We'll need to do testing after another pass

Kinglykrab:
> Please rebase this one so we can close the loop on this one soon as well. We'll need to do testing after another pass

Rebased.

submodules/fmt
Akkadius, line 1: This submodule shouldn't have been changed

Kinglykrab: How do I revert this? Not sure why this changed.

Akkadius (approved):

Akkadius:
We're going to do testing on master with a few other PR's simultaneously

PR #4060, opened 2024-02-11
[Database] Fix default value for `time_of_death` in `character_corpses`

joligario:
As before with other cases, default `datetime` can no longer be 0s. Made it default to `CURRENT_TIMESTAMP` as it was originally defined as `NOT NULL`


Kinglykrab (approved):

joligario:
Checking to see if we need to rebase the repository for this change

joligario:
Bah, stand by

PR #4065, opened 2024-02-12
[Quest API] Add Getter/Setter for Guild Favor to Perl/Lua

Kinglykrab:
# Perl
- Add `quest::get_guild_favor(guild_id)`.
- Add `quest::get_guild_member_favor(guild_id, character_id)`.
- Add `quest::set_guild_favor(guild_id, favor)`.
- Add `quest::set_guild_member_favor(guild_id, character_id, favor)`.

# Lua
- Add `eq.get_guild_favor(guild_id)`.
- Add `eq.get_guild_member_favor(guild_id, character_id)`.
- Add `eq.set_guild_favor(guild_id, favor)`.
- Add `eq.set_guild_member_favor(guild_id, character_id, favor)`.

# Notes
- Allows operators to set guild and guild member favor from a quest script.
- Would be nice to expose a lot more of these methods or a guild object to scripts at some point.

### DIFF
--- a/zone/embparser_api.cpp
+++ b/zone/embparser_api.cpp
@@ -34,6 +34,7 @@
 #include "questmgr.h"
 #include "zone.h"
 #include "data_bucket.h"
+#include "guild_mgr.h"
 
 #include <cctype>
 
@@ -5762,6 +5763,26 @@ std::string Perl__convert_money_to_string(perl::hash table)
 	return Strings::Money(platinum, gold, silver, copper);
 }
 
+uint32 Perl__get_guild_favor(uint32 guild_id)
+{
+	return guild_mgr.GetGuildFavor(guild_id);
+}
+
+void Perl__set_guild_favor(uint32 guild_id, uint32 favor)
+{
+	guild_mgr.SetGuildFavor(guild_id, favor);
+}
+
+uint32 Perl__get_guild_member_favor(uint32 guild_id, uint32 character_id)
+{
+	return guild_mgr.GetGuildMemberFavor(guild_id, character_id);
+}
+
+void Perl__set_guild_member_favor(uint32 guild_id, uint32 character_id, uint32 favor)
+{
+	guild_mgr.SetGuildMemberFavor(guild_id, character_id, favor);
+}
+
 void perl_register_quest()
 {
 	perl::interpreter perl(PERL_GET_THX);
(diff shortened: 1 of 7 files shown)


Akkadius:
cc @neckkola

zone/embparser_api.cpp
```diff
@@ -5762,6 +5763,26 @@ std::string Perl__convert_money_to_string(perl::hash table)
...
 }
 
+uint32 Perl__get_guild_favor(uint32 guild_id)
+{
+	return guild_mgr.GetGuildFavor(guild_id);
+}
+
+void Perl__set_guild_favor(uint32 guild_id, uint32 favor)
+{
+	guild_mgr.SetGuildFavor(guild_id, favor);
```
neckkola, line 5773: While this directly updates the database, it will have to also update the in memory version, otherwise the change will be lost/overwritten when/if any player performs an actually donation.  We will also have to send the appropriate packets so that the entire guild is updated with the changed favor, by adding the difference between the existing db value and the set value.

zone/embparser_api.cpp
```diff
@@ -5762,6 +5763,26 @@ std::string Perl__convert_money_to_string(perl::hash table)
...
+}
+
+uint32 Perl__get_guild_member_favor(uint32 guild_id, uint32 character_id)
+{
+	return guild_mgr.GetGuildMemberFavor(guild_id, character_id);
+}
+
+void Perl__set_guild_member_favor(uint32 guild_id, uint32 character_id, uint32 favor)
+{
+	guild_mgr.SetGuildMemberFavor(guild_id, character_id, favor);
```
neckkola, line 5783: While this directly updates the database, it will have to also update the in memory version and then update the guild db and in memory version of guild, otherwise the change will be lost/overwritten when/if the player performs an actually donation.  We will also have to send the appropriate packets so that the entire guild is updated with the changed favor

zone/lua_general.cpp
```diff
@@ -5406,6 +5407,27 @@ std::string lua_convert_money_to_string(luabind::adl::object table)
...
 }
 
+uint32 lua_get_guild_favor(uint32 guild_id)
+{
+	return guild_mgr.GetGuildFavor(guild_id);
+}
+
+void lua_set_guild_favor(uint32 guild_id, uint32 favor)
+{
+	guild_mgr.SetGuildFavor(guild_id, favor);
```
neckkola, line 5417: Same as perl comment

zone/lua_general.cpp
```diff
@@ -5406,6 +5407,27 @@ std::string lua_convert_money_to_string(luabind::adl::object table)
...
+}
+
+uint32 lua_get_guild_member_favor(uint32 guild_id, uint32 character_id)
+{
+	return guild_mgr.GetGuildMemberFavor(guild_id, character_id);
+}
+
+void lua_set_guild_member_favor(uint32 guild_id, uint32 character_id, uint32 favor)
+{
+	guild_mgr.SetGuildMemberFavor(guild_id, character_id, favor);
```
neckkola, line 5427: Same as perl comment

Kinglykrab:
Closing this for the time being until I speak with @neckkola.

PR #4066, opened 2024-02-12
[Quest API] Add IsInAGuild() to Perl/Lua

Kinglykrab:
# Perl
- Add `$client->IsInAGuild()`.

# Lua
- Add `client:IsInAGuild()`.

# Notes
- Allows operators to more accurately tell if a player is in a group.
- `GuildID()` returns `uint32` max value if the player isn't in a guild so conditions using it must check for a value over a certain point, this is just a bool that simplifies that logic.


Akkadius:
cc @neckkola

neckkola (approved):

PR #4071, opened 2024-02-12
[Code] Update fmt to 10.2.1

hgtw:
This fixes MSVC compile warnings for stdext::checked_array_iterator which were deprecated in STL for VS 2022 17.8:

https://github.com/microsoft/STL/wiki/Changelog#vs-2022-178


Akkadius:
Seems to make no material impact on build time 

```
# before

1m43.522s
1m45.847s

# after

1m39.443s
1m45.524s
1m47.755s
```

Kinglykrab (approved):

Akkadius (approved):

PR #4076, opened 2024-02-14
[Bug Fix] Fix ObjectContentsRepository using content_db

Kinglykrab:
# Notes
- `*this` was using content database, causing this not to work on servers using sharded databases.

### DIFF
--- a/zone/zonedb.cpp
+++ b/zone/zonedb.cpp
@@ -214,7 +214,7 @@ void ZoneDatabase::LoadWorldContainer(uint32 parent_id, EQ::ItemInstance* contai
 	}
 
 	const auto& l = ObjectContentsRepository::GetWhere(
-		*this,
+		database,
 		fmt::format(
 			"`parentid` = {}",
 			parent_id
@@ -271,7 +271,7 @@ void ZoneDatabase::SaveWorldContainer(uint32 zone_id, uint32 parent_id, const EQ
 		}
 
 		ObjectContentsRepository::ReplaceOne(
-			*this,
+			database,
 			ObjectContentsRepository::ObjectContents{
 				.zoneid = zone_id,
 				.parentid = parent_id,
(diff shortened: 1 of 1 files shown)


joligario (approved):

zone/zonedb.cpp
```diff
@@ -214,7 +214,7 @@ void ZoneDatabase::LoadWorldContainer(uint32 parent_id, EQ::ItemInstance* contai
 	}
 
 	const auto& l = ObjectContentsRepository::GetWhere(
-		*this,
+		database,
```
hgtw, line 217: This is a bizarre pattern. Database functions internally should always use `*this` and the caller should specify which database class instance to use e.g., `database.LoadWorldContainer(...)` or `content_db.LoadWorldContainer(...)`. It looks like these are already using the state `database` in callers though, is there something else going on?

Kinglykrab: Not 100% sure what is going on there. I just know it was calling the content database.

hgtw: Object::PutItem
  calls database.SaveWorldContainer (ok)
  calls Save()
    Object::Save calls content_db.UpdateObject() or content_db.AddObject()
      ZoneDatabase::UpdateObject() and ZoneDatabase::AddObject() both call SaveWorldContainer() on their own content_db instance
      SaveWorldContainer() calls DeleteWorldContainer() on that same content_db instance

Probably the calls to SaveWorldContainer() in UpdateObject/AddObject could be changed to explicitly use
database.SaveWorldContainer() but I don't know what exactly is being saved here.

It looks like old code also did this so it's not immediately clear why this would start warning so often now unless Object::Save is now being called from somewhere it didn't used to be?

hgtw: Nevermind I see, the old code called `database.QueryDatabase(query)` explicitly in `SaveWorldContainer` and `DeleteWorldContainer`. I think this one should be reverted and the same thing done.

Kinglykrab: Care to put in a pull request to adjust that?

PR #4079, opened 2024-02-15
[Quest API] Add tracebacks to Lua error messages

hgtw:
This adds an error msg handler for lua calls to add a stack trace to error messages. Lua 5.1 does not have luaL_traceback like luajit and lua 5.2+ so debug.traceback() is used directly as the msg handler.

The traceback will add more detail to errors than just logging package and event names. Exceptions in C++ binds and luabind errors will now show script context instead of just the error message.

e.g., for a luabind overload error:
```
  No matching overload found, candidates:
  void signal(int,int,int)
  void signal(int,int)
  stack traceback:
    [C]: in function 'signal'
    quests/arena/player.lua:10: in function 'somefn'
    quests/arena/player.lua:27: in function <quests/arena/player.lua:17>
```

### DIFF
--- a/zone/lua_parser.cpp
+++ b/zone/lua_parser.cpp
@@ -190,6 +190,14 @@ std::map<std::string, std::list<lua_registered_event>> lua_encounter_events_regi
 std::map<std::string, bool> lua_encounters_loaded;
 std::map<std::string, Encounter *> lua_encounters;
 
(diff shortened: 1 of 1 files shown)


Kinglykrab (approved):
:fire:

Akkadius (approved):

PR #4081, opened 2024-02-15
[Character Creation] Improved Random Name Generator

catapultam-habeo:
Reliably produces valid names (sufficient characters, not already in use), and attempts to produce names which are vaguely pronounceable and aesthetically pleasing.

### DIFF
--- a/world/client.cpp
+++ b/world/client.cpp
@@ -598,82 +598,64 @@ bool Client::HandleNameApprovalPacket(const EQApplicationPacket *app)
 }
 
 bool Client::HandleGenerateRandomNamePacket(const EQApplicationPacket *app) {
-	// creates up to a 10 char name
-	char vowels[18]="aeiouyaeiouaeioe";
-	char cons[48]="bcdfghjklmnpqrstvwxzybcdgklmnprstvwbcdgkpstrkd";
-	char rndname[17]="\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0";
-	char paircons[33]="ngrkndstshthphsktrdrbrgrfrclcr";
-	int rndnum=emu_random.Int(0, 75),n=1;
-	bool dlc=false;
-	bool vwl=false;
-	bool dbl=false;
-	if (rndnum>63)
-	{	// rndnum is 0 - 75 where 64-75 is cons pair, 17-63 is cons, 0-16 is vowel
-		rndnum=(rndnum-61)*2;	// name can't start with "ng" "nd" or "rk"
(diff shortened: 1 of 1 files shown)


world/client.cpp
```diff
@@ -598,82 +598,66 @@ bool Client::HandleNameApprovalPacket(const EQApplicationPacket *app)
...
-	bool vwl=false;
-	bool dbl=false;
-	if (rndnum>63)
-	{	// rndnum is 0 - 75 where 64-75 is cons pair, 17-63 is cons, 0-16 is vowel
-		rndnum=(rndnum-61)*2;	// name can't start with "ng" "nd" or "rk"
-		rndname[0]=paircons[rndnum];
-		rndname[1]=paircons[rndnum+1];
-		n=2;
-	}
-	else if (rndnum>16)
```
fryguy503, line 617: } else if (cond) {

fryguy503:
General Formatting mixing spaces and tabs.

catapultam-habeo:
> General Formatting mixing spaces and tabs.

What is the standard around this?

fryguy503:
> > General Formatting mixing spaces and tabs.
> 
> What is the standard around this?

https://github.com/EQEmu/Server/blob/master/.editorconfig has it noted as tabs.

If you are using vscode or other IDE with addon support, installing the editorconfig plugin should enforce the formatting standards, at least for indenting and eof.

Akkadius:
Has anyone done testing with this and screens / video they could share

catapultam-habeo:
image
image
image
image
image

Five sequential names. THJ Development Server is online running this code.

joligario (commented):
Ran it through roughly 225 times. Here are the outputs. Not sure what constitutes improved.

[random_names_test.log](https://github.com/EQEmu/Server/files/14325027/random_names_test.log)

Akkadius:
This seems like somewhat of an improvement. 

I took a bat at this a few years ago but never went back to it after a crash. I'd love to see a stab at someone reincorporating it https://github.com/EQEmu/Server/pull/1007

Akkadius:
> > This seems like somewhat of an improvement.
> > I took a bat at this a few years ago but never went back to it after a crash. I'd love to see a stab at someone reincorporating it #1007
> 
> I'll take a look at this when I finish my current to-do list. It looks significantly more robust than my solution, which was never intended to be much more than a marginal improvement that at least didn't fail to generate names which wouldn't be accepted.

Please take a look at it as a follow up if you wouldn't mind. We can at least get this merged in since it does improve things.

Akkadius (approved):

PR #4085, opened 2024-02-16
[Bug Fix] Update database_schema to add guild_permissions

neckkola:
The guild_permissions table was excluded from database_schema.h creating issues with peq db dumps and new installs.


common/database_schema.h
```diff
@@ -148,6 +148,7 @@ namespace DatabaseSchema {
 			"guild_members",
 			"guild_ranks",
 			"guild_relations",
+			"guild_permissions",
```
Kinglykrab, line 151: Should be right after `guild_members` for alphabetical order.

neckkola: :-)  Updated.

Kinglykrab (approved):

PR #4086, opened 2024-02-16
[Bug Fix] Fix Sympathetic Procs for Bots/Mercenariess

Kinglykrab:
# Notes
- Bots and Mercenaries were being rejected at the top of this method despite us allowing them to use this method where it's called.


MortimerGreenwald:
Works well on Karana. No issues, bots firing correctly.

PR #4087, opened 2024-02-16
[Auras] Clear all character auras on save

joligario:
Certain classes can have multiple auras - code may not support this yet. When saving character auras, we should clear all auras for the character.

https://forums.daybreakgames.com/eq/index.php?threads/auras.289675/
https://everquest.fanra.info/wiki/Auras

EQ000000
EQ000001
EQ000002

### DIFF
--- a/zone/zonedb.cpp
+++ b/zone/zonedb.cpp
@@ -3087,7 +3087,13 @@ void ZoneDatabase::LoadBuffs(Client *client)
 
 void ZoneDatabase::SaveAuras(Client *c)
 {
-	CharacterAurasRepository::DeleteOne(database, c->CharacterID());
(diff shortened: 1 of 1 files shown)


Kinglykrab (commented):
In this case the `DeleteOne` does the same as the `DeleteWhere` because the primary key is the one you're using as the identifier to delete.

Kinglykrab:
The queries are identical.

joligario:
image
Multi-key primary

Kinglykrab:
> image Multi-key primary

Repositories only use the first.

https://github.com/EQEmu/Server/blob/e6df5be1ed5c1d05abdcb09287f0ed13d3770054/common/repositories/base/base_character_auras_repository.h#L27

PR #4090, opened 2024-02-17
[Expansions] Fix content filter criteria for max_expansions

Akkadius:
We should not be loading rows that are **greater** than our current expansion.

This was never caught because no one is doing anything with expansion filtering to the extent we have with ProjectEQ Expansions lately.


Kinglykrab (approved):

Akkadius:
This was actually correct, I need sleep

PR #4097, opened 2024-02-19
[Bug Fix] Fix issue with IsEntOnHateList

Kinglykrab:
# Notes
- We were not validating pointer in `IsEntOnHateList()`.
- This has been running on Karana for several days now and they were having this crash a lot, but since the fix they have not had this crash.


MortimerGreenwald:
Confirmed!

Akkadius (approved):

Issue #4098, opened 2024-02-19
Lich-type spells not treated correctly


joligario:
Lich-type spells should not break invis. Also, the damage should not be mitigated by spell shield type effects.

Kinglykrab:
I’d imagine the intention was to check for rune spell effect.

PR #4101, opened 2024-02-19
[Bug Fix] Revert 2df7d19

joligario:
The original change breaks lich-type spells as seen in #4098. Per comment on original change, also don't see where Runes come into play.

Current behavior immediately after casting invis:
image

Reverted behavior:
image


joligario (commented):
Looks good!

Akkadius (approved):

PR #4108, opened 2024-02-22
[Quest API] Add Bot Special Attacks for Immune Aggro/Damage

Kinglykrab:
# Notes
- Adds `IMMUNE_AGGRO_BOT` and `IMMUNE_DAMAGE_BOT` for uses in special abilities.

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -3078,15 +3078,15 @@ void Mob::AddToHateList(Mob* other, int64 hate /*= 0*/, int64 damage /*= 0*/, bo
 		return;
 	}
 
-	if (GetSpecialAbility(IMMUNE_AGGRO_BOT) && other->IsBot()) {
+	if (other->IsBot() && GetSpecialAbility(IMMUNE_AGGRO_BOT)) {
 		return;
 	}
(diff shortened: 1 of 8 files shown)


zone/attack.cpp
```diff
@@ -3070,26 +3070,37 @@ void Mob::AddToHateList(Mob* other, int64 hate /*= 0*/, int64 damage /*= 0*/, bo
...
 		return;
+	}
 
-	if (IsFamiliar() || GetSpecialAbility(IMMUNE_AGGRO))
+	if (IsFamiliar() || GetSpecialAbility(IMMUNE_AGGRO)) {
 		return;
+	}
 
-	if (GetSpecialAbility(IMMUNE_AGGRO_NPC) && other->IsNPC())
+	if (GetSpecialAbility(IMMUNE_AGGRO_BOT) && other->IsBot()) {
```
Akkadius, line 3081: Shouldn't we check if other bot first before trying to process ability

Kinglykrab: True enough.

Akkadius (approved):

PR #4122, opened 2024-02-25
[Bug Fix] Cleanup NPC Mana Tap Logic

fryguy503:
Mana Tap rule logic was invalid - Cleaned up and simplified, either we care about npc mana or we dont.


noudess:
So my question:

Should a melee mob always produce the message regardless of the rule? Live does.

noudess (approved):
I think this works.  At first it was a mind twister as all (4) conditions need to be && together, but it does seem to do what we want.

LGTM. 

Tested?

noudess:
I still have an issue with this.  I'll explain later today.

noudess:
Shoot - I meant to PR off this PR - My new PR is off main but includes your changes.

noudess:
I meant to PR off your PR, but I PRed main over [here](https://github.com/EQEmu/Server/pull/4134)

PR #4124, opened 2024-02-25
[Spells] Add Unblockable Spell Table and Spell Blocked Event

Kinglykrab:
# Perl
- Add `EVENT_SPELL_BLOCKED`, exports `$blocking_spell_id`, `$cast_spell_id`, `$blocking_spell`, and `$cast_spell`.

# Lua
- Add `event_spell_blocked`, exports `e.blocking_spell_id`, `e.cast_spell_id`, `e.blocking_spell`, and `e.cast_spell`.

# Database
- Adds `spells_unblockable` table with a `spell_id` and `is_unblockable` column.
- This table will need to be populated based on known spells that should be unblockable.

# Notes
- This event will allow operators to perform events when spells are blocked.


joligario:
What would stop us from using unresistable?
image

Kinglykrab:
> What would stop us from using unresistable? image

A spell may be unresistable, but that doesn't necessarily mean we want it to always override buffs like an `unblockable` spell would be used for. @fryguy503 can speak more to the use case here.

Kinglykrab:
@fryguy503 has been using this for a bit, waiting on his feedback.

fryguy503 (approved):
So far things seem good. But I will know best in a real world test on tuesday when raid guilds start hitting the zone thats using this.

Kinglykrab:
> So far things seem good. But I will know best in a real world test on tuesday when raid guilds start hitting the zone thats using this.

No rush, we can wait until then.

Akkadius:
rebase

Akkadius:
This feels like a lot for what should be existing functionality or could be used from the existing `spells_new` data model.

Could we not make a combination of `resist` = unresistable and another meta field say (resistadj = -1) that sets this same exact behavior without all the extra code and new table ?

Kinglykrab:
> resistadj

My only issue here is this requires editing the spell file and there's no specific format to the spells that will be using this functionality.

Leaving them as they are on Live and not requiring a spell file edit for these to work would be preferable.

zone/spells.cpp
```diff
@@ -3520,28 +3534,103 @@ int Mob::AddBuff(Mob *caster, uint16 spell_id, int duration, int32 level_overrid
...
+				curbuf.spellid,
+				curbuf.casterlevel,
+				spell_id,
+				caster_level,
+				entity_list.GetMobID(curbuf.casterid),
+				caster,
+				buffslot
+			);
+
+			if (ret == -1) { // stop the spell
```
Akkadius, line 3547: Constant?

Akkadius:
> I agree as someone who imports raw lucy spell dumps, the separate table makes life easier as its defined separately.
> 
> Also so far testing has gone well on production, no issues to report.

I'm not entirely opposed to this, but I do need additional context. Can you please speak to the use case, that context is entirely missing. 

If this is something that is implemented off of Live spell data, I'm confused why we need something topical to define spell behavior when most of it is already defined on the spell row itself. Are we not identifying a combination of fields that drive this behavior on live?

All I'm trying to encourage here is simpler design, less bloat, if this is needed and there's justification then I'm fine with it.

KayenEQ:
I was asked to look into this. After discussing this with Trust and the learning what the underlying problem was that caused him to request this feature. It was fairly obvious what the problem was and turns out this is a source code bug for issue that was fixed on live in 2018. 

The real answer is we need to update some buff stacking related code. Basically spells with Spell effect id 148 which blocks other buffs based on certain values, such as those found on HP/AC buffs to make sure other HP/AC buffs with lower values don't override them. However, it looks like there is no exceptions for detrimental debuffs to HP/AC (negative values ect) so it would block those also. Thus making things that should not be blocked, be blocked. It was fixed on live not to block them, which just seems like an obvious bug that got overlooked.  

The only question I have in how to properly code the behavior is confirming if the Buff is replaced/overwritten with the debuff or if you keep your buff and it just adds the debuff to a new buff slot.

Using a table to fix this really is not the ideal solution.

Kinglykrab:
> I was asked to look into this. After discussing this with Trust and the learning what the underlying problem was that caused him to request this feature. It was fairly obvious what the problem was and turns out this is a source code bug for issue that was fixed on live in 2018.
> 
> The real answer is we need to update some buff stacking related code. Basically spells with Spell effect id 148 which blocks other buffs based on certain values, such as those found on HP/AC buffs to make sure other HP/AC buffs with lower values don't override them. However, it looks like there is no exceptions for detrimental debuffs to HP/AC (negative values ect) so it would block those also. Thus making things that should not be blocked, be blocked. It was fixed on live not to block them, which just seems like an obvious bug that got overlooked.
> 
> The only question I have in how to properly code the behavior is confirming if the Buff is replaced/overwritten with the debuff or if you keep your buff and it just adds the debuff to a new buff slot.
> 
> Using a table to fix this really is not the ideal solution.

Appreciate the input, so would the preference be to close this for now and wait until a proper fix can be done?

PR #4130, opened 2024-02-28
[Hotfix] Fix forage returning first result from table

nytmyr:
Forage was not properly incrementing the total chance of items when more than one was found and would result in the final chance roll being based off the last item found rather than the total. This would cause the first item returned to be chosen in most cases when the chances are the same in the table.


Kinglykrab (approved):

joligario (approved):

joligario:
image

Before:
image
~45 rolls

After:
image
~25 rolls

PR #4136, opened 2024-03-01
[Commands] Add #fish Command

Kinglykrab:
# Notes
- Adds `#fish` command.
- Allows operators to simulate fishing to see what they would get, consumes no bait.


joligario (commented):
Recommend `use_bait` to avoid double-negative

Kinglykrab:
> Recommend `use_bait` to avoid double-negative

Gotcha.

joligario (approved):

PR #4137, opened 2024-03-01
[Feature] Exempt a zone from IP-limit checks.

catapultam-habeo:
Allows server operators to select a zone to be skipped for checking IP limits. An example use-case would be to allow the Bazaar to be populated by traders without needing to build a scripted IP-exemption infrastructure.

### DIFF
--- a/common/rulesys.cpp
+++ b/common/rulesys.cpp
@@ -34,8 +34,8 @@ const char *RuleManager::s_categoryNames[_CatCount + 1] = {
 	"InvalidCategory"
 };
 
-const RuleManager::RuleInfo RuleManager::s_RuleInfo[IntRuleCount + RealRuleCount + BoolRuleCount + 1] = {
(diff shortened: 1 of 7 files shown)


Kinglykrab:
I think this would be better served as a column in the `zone` table so this isn’t limited to one zone.

catapultam-habeo:
Alternatively - the rule could be a delimited list.

Kinglykrab:
> Alternatively - the rule could be a delimited list.
> 
> I would argue that rule is a better fit, because all existing data regarding IP limits are rules, and looking for them in one place would be more intuitive.

Adding support for string-based rules now. That way things like a delimited list are possible in rules.

Kinglykrab:
Support added [here](https://github.com/EQEmu/Server/pull/4138).

Kinglykrab (approved):

PR #4141, opened 2024-03-02
[Bug Fix] Prevent NPE when creating DZ using ad-hoc version IDs

catapultam-habeo:
It is otherwise possible to create a DZ using an arbitrary version ID, to do things like create non-respawning versions of zones by duplicating their `spawn2` rows with modifications. However, if you try to instantiate them through the Shared Task system, this throws an NPE due to not being able to retrieve information from `zone` about that version.

I'm not certain if this is the best way to fix this, but it won't crash and does a reasonable job finding sane data to put here.

### DIFF
--- a/zone/task_client_state.cpp
+++ b/zone/task_client_state.cpp
@@ -2290,8 +2290,20 @@ void ClientTaskState::CreateTaskDynamicZone(Client* client, int task_id, Dynamic
 	}
 
 	// dz should be named the version-based zone name (used in choose zone window and dz window on live)
-	auto zone_info = zone_store.GetZone(dz_request.GetZoneID(), dz_request.GetZoneVersion());
-	dz_request.SetName(zone_info->long_name.empty() ? task->title : zone_info->long_name);
+    auto zone_info = zone_store.GetZone(dz_request.GetZoneID(), dz_request.GetZoneVersion());
+
+	// use version 0 name for this if there is no zone with the specified version
+    if (!zone_info) {
+      zone_info = zone_store.GetZone(dz_request.GetZoneID());
+    }
+
+	// fall back to task title if that doesn't exist somehow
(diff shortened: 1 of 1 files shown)


hgtw (commented):
Just a single line change should be needed to fix the crash:
`dz_request.SetName(zone_info && !zone_info->long_name.empty() ? zone_info->long_name : task->title);`

The dz name for tasks isn't that important with emu clients so should just fallback to using the task title if a specific zone version isn't available instead of using version 0's long name.

catapultam-habeo:
This is your baby, but I'd like to throw out there that getting the task title in there makes doing stuff like this more complicated;

``sub HasDynamicZoneAssigned {
    my $client  = shift;
    my $zone    = shift;
    my $dbh     = plugin::LoadMysql();

    my $character_id = $client->CharacterID();

    my $query = "SELECT COUNT(*) FROM dynamic_zone_members, dynamic_zones WHERE dynamic_zones.id = dynamic_zone_members.dynamic_zone_id AND character_id = ? AND name LIKE ?";
    my $sth = $dbh->prepare($query);
    $sth->execute($character_id, $zone);

    my $count = $sth->fetchrow();
    $sth->finish();
    $dbh->disconnect();

    return $count > 0 ? 1 : 0;
}`

hgtw:
The fallback can stay if it's desired (it's probably technically correct to do it that way anyway). It looks like there's already an overload to get a zone version 0 fallback though so this can all be shortened to

```cpp
auto zone_info = zone_store.GetZoneWithFallback(dz_request.GetZoneID(), dz_request.GetZoneVersion());
dz_request.SetName(zone_info && !zone_info->long_name.empty() ? zone_info->long_name : task->title);
```

hgtw (approved):

Kinglykrab (approved):

PR #4144, opened 2024-03-02
[Crash Fix] Update to location of qGlobals initialization

neckkola:
Error log from Spire id 19626 fingerprint a7bba against 22.45.0

Setting the rule Character:PerCharacterQglobalMaxLevel to true would cause zone.exe to crash on character login.

qGlobals was not being initialized before it was being used in the routine resulting in the nullptr check to fail.
Relocated the initialization to before the routine.


Kinglykrab (approved):

Akkadius (approved):

Akkadius:
Love to see crash fixes!

PR #4025, opened 2024-01-27
[Quests] Cleanup Quest Parser Logic

Kinglykrab:
# Notes
- Consolidate duplicated code into loops to make it easier to read and easier to add on to.

# Images
## NPC `EVENT_SAY` Example
image
## Global Player `EVENT_SAY` Example
image
## Global Player `EVENT_TARGET_CHANGE` Example
image
## Zone Player `EVENT_EXP_GAIN` Example
image

PR #4149, opened 2024-03-03
[Bug Fix] Fix Spawns Not Parsing Quest on Zone Bootup

Kinglykrab:
# Notes
- Spawns were not parsing quests on initial zone bootup, causing stuff that relied on this to not work.

### DIFF
--- a/zone/zone.cpp
+++ b/zone/zone.cpp
@@ -158,6 +158,8 @@ bool Zone::Bootup(uint32 iZoneID, uint32 iInstanceID, bool is_static) {
 	LogInfo("Zone bootup type [{}] short_name [{}] zone_id [{}] instance_id [{}]",
 		(is_static) ? "Static" : "Dynamic", zonename, iZoneID, iInstanceID);
 	parse->Init();
+	parse->ReloadQuests();
+	zone->Repop();
 	UpdateWindowTitle(nullptr);
 
 	// Dynamic zones need to Sync here.
--- a/zone/zone.cpp
+++ b/zone/zone.cpp
@@ -158,8 +158,6 @@ bool Zone::Bootup(uint32 iZoneID, uint32 iInstanceID, bool is_static) {
 	LogInfo("Zone bootup type [{}] short_name [{}] zone_id [{}] instance_id [{}]",
 		(is_static) ? "Static" : "Dynamic", zonename, iZoneID, iInstanceID);
 	parse->Init();
-	parse->ReloadQuests();
-	zone->Repop();
 	UpdateWindowTitle(nullptr);
 
 	// Dynamic zones need to Sync here.
--- a/zone/quest_parser_collection.cpp
+++ b/zone/quest_parser_collection.cpp
@@ -805,7 +805,7 @@ int QuestParserCollection::EventBotGlobal(
 
 QuestInterface* QuestParserCollection::GetQIByNPCQuest(uint32 npc_id, std::string& filename)
 {
-	if (!zone || !zone->IsLoaded()) {
+	if (!zone) {
 		return nullptr;
 	}
 
@@ -871,7 +871,7 @@ QuestInterface* QuestParserCollection::GetQIByNPCQuest(uint32 npc_id, std::strin
 
 QuestInterface* QuestParserCollection::GetQIByPlayerQuest(std::string& filename)
 {
-	if (!zone || !zone->IsLoaded()) {
+	if (!zone) {
 		return nullptr;
 	}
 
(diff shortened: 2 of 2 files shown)


Natedog2012:
Tested Bloodfields to see if this would break encounters and it seems to be just fine, if anyone has a better idea to fix the flow of how the zone_controller is spawned let me know

noudess:
@Akkadius ?

noudess:
> # Notes
> * Spawns were not parsing quests on initial zone bootup, causing stuff that relied on this to not work.

Can you give me an example that I can reproduce locally of the bug?

crotonrz:
I originally reported the bug. For us (EZ Server) it was specifically causing issues with the zone controller, which was not loading scripts properly, or at all depending on the script. 

From a shut down zone, with attached zone controller script (or similar), nothing will happen with zone controller. Following a #reload world 1 or 2, zone controller will function correctly, until zone shuts down again on inactivity, at which point it quits working again.

[zone_controller.txt](https://github.com/EQEmu/Server/files/14473113/zone_controller.txt)

noudess:
> I originally reported the bug. For us (EZ Server) it was specifically causing issues with the zone controller, which was not loading scripts properly, or at all depending on the script.
> 
> From a shut down zone, with attached zone controller script (or similar), nothing will happen with zone controller. Following a #reload world 1 or 2, zone controller will function correctly, until zone shuts down again on inactivity, at which point it quits working again.
> 
> [zone_controller.txt](https://github.com/EQEmu/Server/files/14473113/zone_controller.txt)

Def gonna defer to @Akkadius  to start as he added zone controller back in the day.

zone/zone.cpp
```diff
@@ -158,6 +158,8 @@ bool Zone::Bootup(uint32 iZoneID, uint32 iInstanceID, bool is_static) {
 	LogInfo("Zone bootup type [{}] short_name [{}] zone_id [{}] instance_id [{}]",
 		(is_static) ? "Static" : "Dynamic", zonename, iZoneID, iInstanceID);
 	parse->Init();
+	parse->ReloadQuests();
+	zone->Repop();
```
hgtw, line 162: The call to `parse->Init()` already calls ReloadQuests for Lua so this would unncessarily reload quests twice. It does not currently call it for Perl since https://github.com/EQEmu/Server/pull/3648 was backed out for breaking something with perl (I never followed up on the report, it may be a similar regression). This would also essentially repop twice since Repop does the same things as  zone->Init. 

The original issue with zone controller on EZ Server is likely caused by a regression in #4025 which added calls to check `zone->IsLoaded` in places that it previously didn't. Even before `parse->Init()` is called, the call to `zone->Init` triggers quest events for zone controller. These dispatches were disabled with that patch since `is_zone_loaded` is set to true only after the call to `zone->Init`.

Kinglykrab:
What is your recommendation here @hgtw as far as a proper fix? This was intended as a band-aid fix so conversation like this could happen and find a real fix that has no regressions.

hgtw:
Zone::Init spawns the zone controller and calls AddNPC which triggers its EVENT_SPAWN (this is currently what is broken by #4025).

The changes in #3648 and #3333 likely broke this since any initializations done in that event in perl were reset by quest reload and EVENT_SPAWN would then require another #repop to trigger. Since lua always reloaded quests in Init, event_spawn for zone controller in lua has probably just never been used (which may cause some inconvenience and is probably one of the issues with using it in lua for initialization).

Ideally we would want to reimplement #3648 to fix the issue of stale perl plugin changes, then spawn zone controller after the `parse->Init()` so EVENT_SPAWN can trigger in both lua/perl after the latest quests are loaded. This would require some refactoring since we wouldn't want to spawn zone controller in Zone::Init's call to PopulateZoneSpawnList just to then spawn it again after parse->Init. Especially since that function is also used for repop.

Note that parse->ReloadQuests() is also called in main() before Zone::Bootup. I suspect this could be removed in favor of the single parse->Init() loading quests as long as nothing else uses the quest parser/pushes events until after parse->Init() occurs. That would have to be looked at closer and extensively tested to make sure it doesn't break anything with script_init/encounters/plugins/modules/globals though.

As far as getting a quick fix out for the regression, I would recommend reverting the #4025 additions of checking for `zone->IsLoaded()` where they didn't exist before unless there was some secondary reason those were added.

Natedog2012:
No secondary reason, it was just a quick bandaid that did resolve the issues reported on EZ. So if you your idea is better I'm all for it

Akkadius:
Was this tested with the latest changes ?

Kinglykrab:
> Was this tested with the latest changes ?

It has not been. Didn’t have time last night. Will test this afternoon.

Kinglykrab:
Works just fine. This is upon zoning in to the zone after it just booted.
```pl
sub EVENT_SPAWN {
	quest::debug("Zone controller spawned.");
	quest::settimer("Test", 10);
}

sub EVENT_TIMER {
	if ($timer eq "Test") {
		quest::stoptimer("Test");
		quest::debug("Test timer fired for zone controller.");
	}
}
```
image

hgtw (approved):

PR #4155, opened 2024-03-03
[Bug Fix] Add id to the guild_bank table

neckkola:
Added the id field to the guild_bank table to resolve a repo query error.  Also set as the primary key.

### DIFF
--- a/common/database/database_update_manifest.cpp
+++ b/common/database/database_update_manifest.cpp
@@ -5422,7 +5422,6 @@ ALTER TABLE `guild_bank`
 ADD COLUMN `id` INT UNSIGNED NOT NULL AUTO_INCREMENT FIRST,
 ADD PRIMARY KEY (`id`);
 		)",
-		.content_schema_update = true
 	}
 // -- template; copy/paste this when you need to create a new entry
 //	ManifestEntry{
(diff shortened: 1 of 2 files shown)


common/database/database_update_manifest.cpp
```diff
@@ -5410,6 +5410,19 @@ ADD COLUMN `augment_five` int(11) UNSIGNED NOT NULL DEFAULT 0 AFTER `augment_fou
...
+		.description = "2024_03_03_add_id_to_guild_bank.sql",
+		.check = "SHOW COLUMNS FROM `guild_bank` LIKE 'id'",
+		.condition = "empty",
+		.match = "",
+		.sql = R"(
+ALTER TABLE `guild_bank`
+ADD COLUMN `id` INT UNSIGNED NOT NULL AUTO_INCREMENT FIRST,
+ADD PRIMARY KEY (`id`);
+		)",
+		.content_schema_update = true
```
joligario, line 5425: Guild banks won't be part of the content database

joligario (changes requested):
One change

common/database/database_update_manifest.cpp
neckkola: Removed.  Thanks JJ.

Kinglykrab (approved):

joligario (approved):

PR #4165, opened 2024-03-06
[Bug Fix] Fix Default Value in `rule_values` table

Kinglykrab:
# Notes
- Some versions of SQL do not allow a default value for text fields.


joligario (commented):
I think you just modify the original. The previous would fail on those systems in the first place. If it didn’t fail, it won’t make a difference.

PR #4168, opened 2024-03-07
[Bug Fix] Using %T in channel messages on fresh corpse yields mob, not corpse name.

noudess:
The problem:

Kill a PC or NPC.  Target the corpse.  The client shows the corpse name as xyz's_corpse.  But the result of using %T in channels is just the npc or player name without the ''s corpse' addition.  If you zone out and back in, then %T works as expected.

I broke down testing this to the two different cases:

MOB:

The only way I could make the client honor %T was to do a corpse->Spawn() after we do AddCorpse to the entity list.  AddCorpse ends up setting the new name, but the client doesn't see this without me adding corpse->Spawn().  We do not get (2) corpses, it just seems to fix the (1).  

Concerns:

- If we don't use a backtick in the corpse name, but rather use a tick - the client doesnt fix it. Seems really odd to me.  I tested if it was just a matter of changing any character.  Nothing I did worked other that using a backtick instead of a tick.  Maybe somewhere at some packet levels the tick is getting stripped?

PC:

[ Could not get this to function.  The rename worked but caused multiple side effects on the client.  Not worth it for just displaying correct name in macros. ]

@neckkola credited for the rename fix.  I had uses a spawn packet instead.

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -2870,17 +2870,17 @@ bool NPC::Death(Mob* killer_mob, int64 damage, uint16 spell, EQ::skills::SkillTy
 		entity_list.AddCorpse(corpse, GetID());
 
 		auto out2 = new EQApplicationPacket(OP_MobRename, sizeof(MobRename_Struct));
-        auto data = (MobRename_Struct *)out2->pBuffer;
-        out2->priority = 6;
+		auto data = (MobRename_Struct *)out2->pBuffer;
+		out2->priority = 6;
 
-        strn0cpy(data->old_name, temp_name.c_str(), sizeof(data->old_name));
-        strn0cpy(data->old_name_again, data->old_name, sizeof(data->old_name_again));
-        strn0cpy(data->new_name, corpse->GetCleanName(), sizeof(data->new_name));
-        data->unknown192 = 0;
-        data->unknown196 = 1;
+		strn0cpy(data->old_name, temp_name.c_str(), sizeof(data->old_name));
+		strn0cpy(data->old_name_again, data->old_name, sizeof(data->old_name_again));
+		strn0cpy(data->new_name, corpse->GetCleanName(), sizeof(data->new_name));
+		data->unknown192 = 0;
+		data->unknown196 = 1;
 
-        entity_list.QueueClients(killer_mob, out2, false);
-        safe_delete(out2);
+		entity_list.QueueClients(killer_mob, out2, false);
+		safe_delete(out2);
 
 		entity_list.UnMarkNPC(GetID());
 		entity_list.RemoveNPC(GetID());
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -2012,9 +2012,10 @@ bool Client::Death(Mob* killer_mob, int64 damage, uint16 spell, EQ::skills::Skil
 			}
 
 			entity_list.AddCorpse(new_corpse, GetID());
-			new_corpse->Spawn();
 			SetID(0);
 
+			//send the become corpse packet to everybody else in the zone.
+			entity_list.QueueClients(this, &app2, true);
 			ApplyIllusionToCorpse(illusion_spell_id, new_corpse);
 			leave_corpse = true;
 		}
@@ -2506,8 +2507,9 @@ bool NPC::Death(Mob* killer_mob, int64 damage, uint16 spell, EQ::skills::SkillTy
 	d->spell_id     = UINT32_MAX;
 	d->attack_skill = SkillDamageTypes[attack_skill];
 	d->damage       = damage;
+	d->corpseid		= GetID();
 
-	app->priority = 6;
+	app->priority = 1;
 
 	entity_list.QueueClients(killer_mob, app, false);
 
@@ -2859,8 +2861,27 @@ bool NPC::Death(Mob* killer_mob, int64 damage, uint16 spell, EQ::skills::SkillTy
 		}
 
 		entity_list.LimitRemoveNPC(this);
+		
+		// Rename corpse on client so %T works for those in zone already
+		std::string temp_name {};
+		temp_name = fmt::format("{}'s_corpse{}",
+			EntityList::RemoveNumbers((char *)corpse->GetName()), GetID());
+
 		entity_list.AddCorpse(corpse, GetID());
-		corpse->Spawn();
+
+		auto out2 = new EQApplicationPacket(OP_MobRename, sizeof(MobRename_Struct));
+        auto data = (MobRename_Struct *)out2->pBuffer;
+        out2->priority = 6;
+
+        strn0cpy(data->old_name, temp_name.c_str(), sizeof(data->old_name));
+        strn0cpy(data->old_name_again, data->old_name, sizeof(data->old_name_again));
+        strn0cpy(data->new_name, corpse->GetCleanName(), sizeof(data->new_name));
+        data->unknown192 = 0;
+        data->unknown196 = 1;
+
+        entity_list.QueueClients(killer_mob, out2, false);
+        safe_delete(out2);
+
 		entity_list.UnMarkNPC(GetID());
 		entity_list.RemoveNPC(GetID());
 
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -2867,7 +2867,7 @@ bool NPC::Death(Mob* killer_mob, int64 damage, uint16 spell, EQ::skills::SkillTy
 		// The client sees NPC corpses as name's_corpse.  The server uses
 		// name`s_corpse so that %T works on corpses (client workaround)
 		// Rename the new corpse on client side.
-		std::string old_name = Strings::Replace(corpse->GetName(), "`", "'");
+		std::string old_name = Strings::Replace(corpse->GetName(), "`s_corpse", "'s_corpse");
 		SendRename(killer_mob, old_name.c_str(), corpse->GetCleanName());
 
 		entity_list.UnMarkNPC(GetID());
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -2868,7 +2868,7 @@ bool NPC::Death(Mob* killer_mob, int64 damage, uint16 spell, EQ::skills::SkillTy
 		// name`s_corpse so that %T works on corpses (client workaround)
 		// Rename the new corpse on client side.
 		std::string old_name = Strings::Replace(corpse->GetName(), "`s_corpse", "'s_corpse");
-		SendRename(killer_mob, old_name.c_str(), corpse->GetCleanName());
+		SendRename(killer_mob, old_name.c_str(), corpse->GetName());
 
 		entity_list.UnMarkNPC(GetID());
 		entity_list.RemoveNPC(GetID());
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -2861,26 +2861,14 @@ bool NPC::Death(Mob* killer_mob, int64 damage, uint16 spell, EQ::skills::SkillTy
 		}
 
 		entity_list.LimitRemoveNPC(this);
-		
-		// Rename corpse on client so %T works for those in zone already
-		std::string temp_name {};
-		temp_name = fmt::format("{}'s_corpse{}",
-			EntityList::RemoveNumbers((char *)corpse->GetName()), GetID());
 
 		entity_list.AddCorpse(corpse, GetID());
 
-		auto out2 = new EQApplicationPacket(OP_MobRename, sizeof(MobRename_Struct));
-		auto data = (MobRename_Struct *)out2->pBuffer;
-		out2->priority = 6;
-
-		strn0cpy(data->old_name, temp_name.c_str(), sizeof(data->old_name));
-		strn0cpy(data->old_name_again, data->old_name, sizeof(data->old_name_again));
-		strn0cpy(data->new_name, corpse->GetCleanName(), sizeof(data->new_name));
-		data->unknown192 = 0;
-		data->unknown196 = 1;
-
-		entity_list.QueueClients(killer_mob, out2, false);
-		safe_delete(out2);
+		// The client sees NPC corpses as name's_corpse.  The server uses
+		// name`s_corpse so that %T works on corpses (client workaround)
+		// Rename the new corpse on client side.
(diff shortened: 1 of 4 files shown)


zone/corpse.cpp
```diff
@@ -622,7 +622,7 @@ void Corpse::CalcCorpseName()
 	EntityList::RemoveNumbers(name);
 	char tmp[64];
 	if (m_is_player_corpse) {
-		snprintf(tmp, sizeof(tmp), "'s corpse%d", GetID());
+		snprintf(tmp, sizeof(tmp), "`s corpse%d", GetID());
```
fryguy503, line 625: Do we even need the if (m_is_player_corpse) condition if we make them both backtics?

neckkola:
I can take a look in the next few days. See what I can add.

zone/corpse.cpp
noudess: For sure.  Holding off until we can see if we need this backtick hack at all, but yes.

hgtw:
Using %T on live behaves the same way. Even if you re-enter the zone it only outputs the name without "corpse" appended. If it used to work I'd suspect some unnecessary spawn packets were removed at some point.

As far as client behavior, %t uses the target's `DisplayedName` (mq variable naming) which is a cached stripped/cleaned name. Other things like Hail, UI labels, etc trims the raw `Name` of underscore/numbers and uses that. OP_Death appends `'s corpse` to the raw name when it does its work of setting the entity to corpse type etc. The corpse string doesn't get appended to `DisplayedName` here since it breaks the loop when it encounters an apostrophe.

Also for what it's worth live uses an apostrophe e.g., `name's_corpse0` (always 0 even if there's multiple corpses of the same name) for the spawn packet when entering a zone that has corpses. This is likely why even re-entering a zone on live doesn't change %T behavior since setting `DisplayedName` trims anything after encountering the apostrophe.

Sending a spawn packet just to change the name is a bit heavy so it's probably not the right thing to do. If this is really desired I'd recommend adding a rule (since this wouldn't be live like) and just sending a mob rename packet for the corpse. The rule would have to switch server side names to use backticks since apostrophes are trimmed. The spawn packet essentially only does a rename anyway if it detects a player (spawn entity) id already exists in the zone and returns early after setting the name. This wouldn't be as heavy as sending a full spawn packet.

noudess:
> Using %T on live behaves the same way. Even if you re-enter the zone it only outputs the name without "corpse" appended. If it used to work I'd suspect some unnecessary spawn packets were removed at some point.
> 
> As far as client behavior, %t uses the target's `DisplayedName` (mq variable naming) which is a cached stripped/cleaned name. Other things like Hail, UI labels, etc trims the raw `Name` of underscore/numbers and uses that. OP_Death appends `'s corpse` to the raw name when it does its work of setting the entity to corpse type etc. The corpse string doesn't get appended to `DisplayedName` here since it breaks the loop when it encounters an apostrophe.
> 
> Also for what it's worth live uses an apostrophe e.g., `name's_corpse0` (always 0 even if there's multiple corpses of the same name) for the spawn packet when entering a zone that has corpses. This is likely why even re-entering a zone on live doesn't change %T behavior since setting `DisplayedName` trims anything after encountering the apostrophe.
> 
> Sending a spawn packet just to change the name is a bit heavy so it's probably not the right thing to do. If this is really desired I'd recommend adding a rule (since this wouldn't be live like) and just sending a mob rename packet for the corpse. The rule would have to switch server side names to use backticks since apostrophes are trimmed. The spawn packet essentially only does a rename anyway if it detects a player (spawn entity) id already exists in the zone and returns early after setting the name. This wouldn't be as heavy as sending a full spawn packet.

I'm going to tey the rename as suggested.

Regarding the rule, are we sure we want that?  I know this worked at some point on live, and on eqemu.  I know we strive to be live like, but who would "choose" to not be able to use %T on a corpse for commands and have them work?

I greatly appreciate you looking into this!  It explains why the backtick is needed to get the change to take place.

neckkola:
I have ran a few tests and have had some success in sending a name update to correct the problem.  Will test further and report back.  The rename is linked the corpseid in the deathpacket.  Travelling today, though will try and get something up for you to test @noudess soon.

hgtw:
Are we sure live ever output the corpse suffix with %t? Grepping through some old logs from both 2006 and 2014 I'm seeing cases where a heal hotkey was clearly pressed on someone that just died and it just shows their name without the corpse.

It also looks like an apostrophe has always been used for corpses from hail messages not a backtick so this would make sense if the name has always been stripped of everything after an apostrophe (at least looking at 2003 and 2005 clients). The only thing I see in older logs with a backtick is `Venril`s Corpse` which is probably some old venril sathir`s remains quest npc spawn?

noudess:
> I have ran a few tests and have had some success in sending a name update to correct the problem. Will test further and report back. The rename is linked the corpseid in the deathpacket. Travelling today, though will try and get something up for you to test @noudess soon.


I saw that the mob version of Kill_Struct did not fill in corpseid, felt odd, but didnt mess with that aspect.

noudess:
> Are we sure live ever output the corpse suffix with %t? Grepping through some old logs from both 2006 and 2014 I'm seeing cases where a heal hotkey was clearly pressed on someone that just died and it just shows their name without the corpse.
> 
> It also looks like an apostrophe has always been used for corpses from hail messages not a backtick so this would make sense if the name has always been stripped of everything after an apostrophe (at least looking at 2003 and 2005 clients). The only thing I see in older logs with a backtick is "Venril`s Corpse" which is probably some old venril sathir`s remains quest npc spawn?
> 
> edit: It doesn't bother me if it needs to be a rule or not, someone else can decide. It will probably come down to a difference in using backticks vs apostrophes depending on how the rename testing works out.

My recollections from live come from being having dozens of %T macros.  Its just instilled in me that I can tell someone else about a corpse in /gui, and I would have noticed if the tells used the real name not the corpse's name.  But as well all know memory can be selective, so I only say that I'd bet $1000 on it, but I could lose.

noudess:
Is there any evidence in the client code that BodyType was checked and changed client behaviour (it not stripping off corpse for corpses)?

hgtw:
> Is there any evidence in the client code that BodyType was checked and changed client behaviour (ie not stripping off corpse for corpses)?

Only the spawn type is checked to make sure the entity id is not already a corpse when handling the death packet. It also checks for npc vs pc to form the string a little different (if npc it checks if the class is an object and just appends an id instead of "corpse", it also does something with setting any numbers to 0 in the suffix).

I don't see anything else that would not strip off `'s_corpse` for %T. `DisplayedName` is just a cached stripped name in rof2. Older clients called `stripName` when handling %T while rof2 just caches it when changing the name in the death packet and returns the cached version for %T.

There's also a couple places where the client searches entity names for the substr `'s corpse` as a fallback if it doesn't find by actual name in some group/raid member code and another place it generates locale eqstr messages if the string contains `'s corpse`. If client corpses on emu have always been sent with a backtick then whatever these fallbacks are for have probably never worked anyway though. I think they're for some role/leadership stuff.

neckkola:
@noudess , I put together this branch for one version of a potential fix.  https://github.com/neckkola/Server/tree/%25T_Updates

It uses the MobRename packet to adjust the stored named used in the %T after the death packet is done.  Have not done any regression testing so not sure if there are side effects.  Interested in your thoughts and approach.

noudess:
> @noudess , I put together this branch for one version of a potential fix. https://github.com/neckkola/Server/tree/%25T_Updates
> 
> It uses the MobRename packet to adjust the stored named used in the %T after the death packet is done. Have not done any regression testing so not sure if there are side effects. Interested in your thoughts and approach.

I'll test this locally when I have some time tomorrow.  In the meantime, thoughts on these:

1. If your solution works, doesn't it invalidate earlier comments that say any corpse name with a tick in it will be stripped out for use with %T?
2. Should the creation of name be moved out of CalcCorpseName (which requires a corpse object) or should we create the corpse earlier and use corpse->GetName() to avoid the name code being in three places?
3. What about when you zone in?  That currently uses the corpse_list and that code has backticks to make it work.

noudess:
@hgtw The patch @neckkola posted here works for fresh corpses without using the backtick.  I don't understand why.  Do you?

@neckkola How can we apply the same concept to zoning into a zone and seeing a corpse.  That only works now because CalcCorpseName uses a backtick.  If I change that to a normal tick we get the same problem

joligario:
image

Akkadius:
@hgtw @noudess @neckkola @joligario what's the status on this guy?

joligario:
On Live, when I use a %t it shows the NPC's clean name.

noudess:
> @hgtw @noudess @neckkola @joligario what's the status on this guy?

@neckkola is still looking for time to resolve side effect on looting PC corpses.

If it can't be resolved cleanly, we will either discard this PR or go with the "fix" for NPC corpses only.

zone/corpse.cpp
noudess: Removed this aspect (PC corpses) from the fix entirely as it created too many side effects.

noudess:
> image Currently on PEQ yet image

I believe this to be a separate issue.  Maybe @neckkola can answer this one?

noudess:
@neckkola  and I both saw too many issues with PC corpse rename.  This works fine for NPC corpses, aside from the name looking a little funky.  That said, it already was in place for looking at corpses after you zone in (the corpse list uses the backtick already) so this just makes corpses look the same to those already in the zone as opposed to those new people zoning in.

noudess:
@neckkola You ok with this PR?  Basically your code.  Its been working locally.

hgtw (commented):
Since server side is currently already using a backtick for npc corpses it probably does make sense to update the corpse name here on clients.

I still think we should rule this off at some point to use apostrophes just in the interest of live/client accuracy (and less packets) but if it's always been like this then I don't see it hurting anything.

zone/attack.cpp
```diff
@@ -2507,8 +2507,9 @@ bool NPC::Death(Mob* killer_mob, int64 damage, uint16 spell, EQ::skills::SkillTy
 	d->spell_id     = UINT32_MAX;
 	d->attack_skill = SkillDamageTypes[attack_skill];
 	d->damage       = damage;
+	d->corpseid		= GetID();
 
-	app->priority = 6;
+	app->priority = 1;
```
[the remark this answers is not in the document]
neckkola: In my testing I noted that the priority was negatively influencing the order that the packets arrived at the client.  I made the change to ensure that they arrived in the correct order so that the rename would take effect.

zone/attack.cpp
```diff
@@ -2860,8 +2861,27 @@ bool NPC::Death(Mob* killer_mob, int64 damage, uint16 spell, EQ::skills::SkillTy
 		}
 
 		entity_list.LimitRemoveNPC(this);
+		
+		// Rename corpse on client so %T works for those in zone already
+		std::string temp_name {};
+		temp_name = fmt::format("{}'s_corpse{}",
+			EntityList::RemoveNumbers((char *)corpse->GetName()), GetID());
```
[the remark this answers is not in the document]
noudess: I'll take a look again later this week.

zone/attack.cpp
```diff
@@ -2860,8 +2861,27 @@ bool NPC::Death(Mob* killer_mob, int64 damage, uint16 spell, EQ::skills::SkillTy
...
+		out2->priority = 6;
+
+		strn0cpy(data->old_name, temp_name.c_str(), sizeof(data->old_name));
+		strn0cpy(data->old_name_again, data->old_name, sizeof(data->old_name_again));
+		strn0cpy(data->new_name, corpse->GetCleanName(), sizeof(data->new_name));
+		data->unknown192 = 0;
+		data->unknown196 = 1;
+
+		entity_list.QueueClients(killer_mob, out2, false);
+		safe_delete(out2);
```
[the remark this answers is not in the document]
noudess: I'll look at that also.

zone/attack.cpp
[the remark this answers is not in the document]
noudess: Fixed to use your suggestion.  Tested just fine on client that kills, client observing and client entering zone.

zone/attack.cpp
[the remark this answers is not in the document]
noudess: Done.

noudess:
I think this is ready for merge.

zone/attack.cpp
```diff
@@ -2860,8 +2861,15 @@ bool NPC::Death(Mob* killer_mob, int64 damage, uint16 spell, EQ::skills::SkillTy
...
 
 		entity_list.LimitRemoveNPC(this);
+
 		entity_list.AddCorpse(corpse, GetID());
 
+		// The client sees NPC corpses as name's_corpse.  The server uses
+		// name`s_corpse so that %T works on corpses (client workaround)
+		// Rename the new corpse on client side.
+		std::string old_name = Strings::Replace(corpse->GetName(), "`s_corpse", "'s_corpse");
+		SendRename(killer_mob, old_name.c_str(), corpse->GetCleanName());
```
hgtw, line 2871: This should set the new name to `corpse->GetName()` instead of the clean name. The new name on the client should still be in the raw ``name`s_corpse##`` format

noudess: I did exactly what @neckkola did - sending the name without and it works without a hitch.  Are you sure?  @neckkola ?

I saw it was different from what you suggested, but my solution (before we moved it into a function) and @neckkola both use GetCleanName(). 

Let me know.

hgtw: Yes, we want to just replace `Some_Name's_corpse01` with ``Some_Name`s_corpse01`` on the client here. Using clean name would just replace it with ``Some Name`s corpse``. The client will handle trimming and stripping underscores and numbers for any UI/%t output.

There's places in the client that get spawns by name and this helps distinguish them (for example using a non-mq `/target Some_Name's_corpse01` vs `/target Some_Name's_corpse02`). This would also allow scripts to use `Mob::TempName` on the corpse afterwards since it renames using the server side raw name which the client uses to find the right spawn in case there's multiples with the same name.

noudess: Got it - will fix in the next couple of days.  Tomorrow is crazy and I of course want to retest.

noudess: Fixed.  Thanks.  Was dumb of me not to see that.  I was so tired of then many iterations of this that I was sloppy. 

This is by far the best solution so far, so thank you.

hgtw (approved):

PR #4173, opened 2024-03-08
[Bug Fix] Fix Crash in ClientList::GetCLEIP

Kinglykrab:
# Crashes
https://spire.akkadius.com/dev/release/22.46.1?id=19955
https://spire.akkadius.com/dev/release/22.46.1?id=19948
https://spire.akkadius.com/dev/release/22.46.0?id=19945

### DIFF
--- a/world/clientlist.cpp
+++ b/world/clientlist.cpp
@@ -96,7 +96,11 @@ void ClientList::GetCLEIP(uint32 in_ip) {
 
 	while (iterator.MoreElements()) {
 		cle = iterator.GetData();
-		
+
+		if (!cle) {
+			continue;
+		}
+
 		const auto zones = Strings::Split(RuleS(World, IPExemptionZones), ",");
 		for (const auto &z : zones) {
 			if (Strings::ToUnsignedInt(z) == cle->zone()) {
--- a/world/clientlist.cpp
+++ b/world/clientlist.cpp
@@ -94,18 +94,17 @@ void ClientList::GetCLEIP(uint32 in_ip) {
 	int count = 0;
 	iterator.Reset();
 
+	const auto& zones = Strings::Split(RuleS(World, IPExemptionZones), ",");
+
 	while (iterator.MoreElements()) {
 		cle = iterator.GetData();
 
-		if (!cle) {
-			continue;
-		}
-
-		const auto zones = Strings::Split(RuleS(World, IPExemptionZones), ",");
-		for (const auto &z : zones) {
-			if (Strings::ToUnsignedInt(z) == cle->zone()) {
-				iterator.Advance();
-				continue;
+		if (!zones.empty() && cle->zone()) {
+			for (const auto& z : zones) {
+				if (Strings::ToUnsignedInt(z) == cle->zone()) {
+					iterator.Advance();
+					continue;
+				}
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


world/clientlist.cpp
```diff
@@ -96,7 +96,11 @@ void ClientList::GetCLEIP(uint32 in_ip) {
 
 	while (iterator.MoreElements()) {
 		cle = iterator.GetData();
-		
+
+		if (!cle) {
+			continue;
+		}
+
```
hgtw, line 103: This isn't the cause of those crashes since they occur in the middle of the function below in calls to `RemoveCurrent` after `cle` has been accessed a bunch. I also think with these forward iterators the pointer from `GetData` won't be null since it's wrapping `clientlist`.

There's a couple bugs with the `IPExemptionZones` loop below this that would lead to crashes:
- If the rule is empty it's still passing the `== cle->zone()` comparison condition and calling `iterator.Advance();` and `continue`
- The `continue` is occuring on the inner `for (const auto &z : zones)` loop so this will advance multiple times if the zone rule is populated (but always at least once for an empty rule)

When the zone rule loop is done it will always proceed to the rest of the function even if it advanced the iterator. This means the main loop condition is never checked after advancing it. If the loop advanced the iterator to the end of the clientlist then the iterator element would be null. If anything then calls `iterator.RemoveCurrent()` that function dereferences the nullptr `current_element` and crashes.

Also as an optimization the `const auto zones = Strings::Split(RuleS(World, IPExemptionZones), ",");` should be called above and outside the main iterator loop so it's not split for every client.

STR for debugging:
1) Have a default empty `IPExemptionZones` rule
2) Set `World:MaxClientsPerIP` to 1 to trigger an `iterator.RemoveCurrent`
3) Log 1 client in and enter world
4) Log a 2nd client in and enter world
5) Crash in one of the calls to `RemoveCurrent` (line 181)

Kinglykrab: That makes sense. So what should we do to resolve this?

hgtw: The zone check below needs to be rewritten. It needs to guard against `cle->zone()` being 0 which is allowing the condition to pass when the rule has an empty string and the `continue` after advancing the iterator needs to occur for the main loop not the inner zone string loop.

Kinglykrab: Adjusted the code if you want to take another look.

world/clientlist.cpp
```diff
@@ -94,14 +94,17 @@ void ClientList::GetCLEIP(uint32 in_ip) {
...
-		for (const auto &z : zones) {
-			if (Strings::ToUnsignedInt(z) == cle->zone()) {
-				iterator.Advance();
-				continue;
+
+		if (!zones.empty() && cle->zone()) {
+			for (const auto& z : zones) {
+				if (Strings::ToUnsignedInt(z) == cle->zone()) {
+					iterator.Advance();
+					continue;
```
hgtw, line 106: The `Advance()` and `continue` are still within the inner `for (const auto& z : zones)` loop. These should only be called once after the loop if a match was found to continue the main `while (iterator.MoreElements())` loop. This would also lead to crashes again if the same zone was listed twice in the rule and matched cle->zone.

It should be breaking from the zones loop when a zone match is found and then only calling advance + continue outside it. I'd suggest something like:

```cpp
auto it = std::ranges::find_if(zones,
		[cle](const auto& z) { return Strings::ToUnsignedInt(z) == cle->zone(); });

if (it != zones.end())
{
		iterator.Advance();
		continue;
}
```

Kinglykrab: Hopefully what I just changed is what you mean.

hgtw (approved):

PR #4184, opened 2024-03-10
[Bug Fix] Fix GetLeaderName() for Groups

Kinglykrab:
# Notes
- We were getting bogus data in this.
- Made it its own method.
- This works across zones as well because it calls the database.

# Image
image

### DIFF
--- a/zone/expedition_request.cpp
+++ b/zone/expedition_request.cpp
@@ -125,7 +125,7 @@ bool ExpeditionRequest::CanGroupRequest(Group* group)
 	}
 
 	// Group::GetLeaderName() is broken if group formed across zones, ask database instead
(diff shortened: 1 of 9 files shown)


zone/expedition_request.cpp
```diff
@@ -125,7 +125,7 @@ bool ExpeditionRequest::CanGroupRequest(Group* group)
 	}
 
 	// Group::GetLeaderName() is broken if group formed across zones, ask database instead
-	m_leader_name = m_leader ? m_leader->GetName() : GetGroupLeaderName(group->GetID()); // group->GetLeaderName();
+	m_leader_name = m_leader ? m_leader->GetName() : group->GetLeaderName();
```
hgtw, line 128: Delete the `ExpeditionRequest::GetGroupLeaderName` function with this

hgtw (approved):

PR #4185, opened 2024-03-10
[Bug Fix] Fix EVENT_KILLED_MERIT firing before NPC removal

Kinglykrab:
# Perl
- Add `$client->GetGroupOrRaidOrSelf()`.
- Add `$client->GetGroupOrRaidOrSelf(clients_only)`.

# Lua
- Add `client:GetGroupOrRaidOrSelf()`.
- Add `client:GetGroupOrRaidOrSelf(clients_only)`.

# Notes
- NPCs were parsing this event too early and anything that checked if they were still alive in `EVENT_KILLED_MERIT` would show them still alive because of this.

# Image
image

### DIFF
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -12330,7 +12330,7 @@ std::vector<Mob*> Client::GetRaidOrGroupOrSelf(bool clients_only)
 
 		if (r) {
 			for (const auto& m : r->members) {
-				if (m.member && (!clients_only || !m.is_bot)) {
+				if (m.member && (!m.is_bot || !clients_only)) {
 					v.emplace_back(m.member);
 				}
 			}
@@ -12340,7 +12340,7 @@ std::vector<Mob*> Client::GetRaidOrGroupOrSelf(bool clients_only)
 
 		if (g) {
 			for (const auto& m : g->members) {
-				if (m && (!clients_only || !m->IsBot())) {
+				if (m && (m->IsClient() || !clients_only)) {
 					v.emplace_back(m);
 				}
 			}
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -12320,3 +12320,34 @@ int Client::GetEXPPercentage()
 
 	return static_cast<int>(std::round(scaled * 100.0 / 330.0)); // unscaled pct
 }
+
+std::vector<Mob*> Client::GetRaidOrGroupOrSelf(bool clients_only)
+{
+	std::vector<Mob*> v;
+
+	if (IsRaidGrouped()) {
+		Raid* r = GetRaid();
+
+		if (r) {
+			for (const auto& m : r->members) {
+				if (m.member && (!clients_only || !m.is_bot)) {
+					v.emplace_back(m.member);
+				}
+			}
+		}
+	} else if (IsGrouped()) {
+		Group* g = GetGroup();
+
+		if (g) {
+			for (const auto& m : g->members) {
+				if (m && (!clients_only || !m->IsBot())) {
+					v.emplace_back(m);
+				}
+			}
+		}
+	} else {
+		v.emplace_back(this);
+	}
+
+	return v;
+}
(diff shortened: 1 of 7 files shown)


fryguy503:
was @Kinglykrab
[Add GetRaidOrGroupOrSelf() to Perl/Lua](https://github.com/EQEmu/Server/pull/4185/commits/08e6d4a51cffaf6e9da79bbb4e8c1e8cb55ef501) meant to be added to this PR?

Akkadius:
> was @Kinglykrab [Add GetRaidOrGroupOrSelf() to Perl/Lua](https://github.com/EQEmu/Server/pull/4185/commits/08e6d4a51cffaf6e9da79bbb4e8c1e8cb55ef501) meant to be added to this PR?

Yes because we loop this logic multiple times in Death

fryguy503:
Gotcha, I know KK wanted feedback on this, I have the version without the lua/perl defs and so far its working great

zone/client.cpp
```diff
@@ -12258,3 +12258,34 @@ int Client::GetEXPPercentage()
...
+					v.emplace_back(m.member);
+				}
+			}
+		}
+	} else if (IsGrouped()) {
+		Group* g = GetGroup();
+
+		if (g) {
+			for (const auto& m : g->members) {
+				if (m && (!clients_only || !m->IsBot())) {
```
hgtw, line 12281: Can mercs also be in groups on emu? Maybe these should just check for `|| IsClient()` instead of !bot

Kinglykrab: > Can mercs also be in groups on emu? Maybe these should just check for `|| IsClient()` instead of !bot

Ah yeah. Forgot about Mercenaries.

zone/lua_client.cpp
```diff
@@ -20,6 +20,10 @@
 
 struct InventoryWhere { };
 
+struct Lua_Mob_List {
+	std::vector<Lua_Mob> entries;
```
hgtw, line 24: I think using the luabind stl iterators for new stuff should be avoided when possible and just a table returned. Their only use case looks like returning access to containers that already exists on exported classes. This (and entity_list usages) just construct and copy everything to a new vector anyway for no benefit. Luabind also ends up making another copy of the vector when it copies the exported class instance for ownership.

I guess an argument for using them could be that they match current entity_list apis but I still think just returning a table would be better.

Akkadius: Not sure that I understand fully what you’re suggesting. 

The goal is to just return a list of mobs to be able to iterate through in a loop like any other entity list call. 

Can you give Kingly (and myself) an example of an alternative, why it’d be better and how it’d look with usage

hgtw: I'm suggesting changing the return to a lua table instead of a Lua_Mob_List that wraps a vector. Then it can just be iterated in a conventional lua way:
```lua
local party = client:GetGroupOrRaidOrSelf()
eq.debug("party count: " .. #party)
for i, mob in pairs(party)
  -- ...
end
```

see https://github.com/EQEmu/Server/blob/82aa6a1587477e642e7ac44e7902f4568bb8563e/zone/lua_mob.cpp#L3307

zone/lua_client.cpp
```diff
@@ -3308,6 +3308,38 @@ bool Lua_Client::RemoveAlternateCurrencyValue(uint32 currency_id, uint32 amount)
...
 
+luabind::object Lua_Client::GetRaidOrGroupOrSelf(lua_State* L)
+{
+	auto t = luabind::newtable(L);
+	if (d_) {
+		auto self = reinterpret_cast<NativeType*>(d_);
+		auto l = self->GetRaidOrGroupOrSelf();
+		int i = 1;
+		for (const auto& e : l) {
+			t[i] = e;
```
hgtw, line 3319: You'll probably need to be explicit on these and push `t[i] = Lua_Mob(e)`. Otherwise it'll just push the raw Mob pointer type and not the exported quest class. Might need to double check that previous buff patch for that too

hgtw (approved):

PR #4189, opened 2024-03-14
[Bug Fix] Fix Proximity Say

Kinglykrab:
# Notes
- Without setting `HaveProximitySays` to `true` along with `n->proximity->say` NPCs will not respond to proximity say stuff.
- Prior to this we were required to enable say in `quest;:set_proximity` or `quest::set_proximity_range` then use `quest::enable_proximity_say` as well.


noudess:
Seems like a good addition to me.  I've seen many quests that don't know that they need to call the extra function and I don't know why you should have to?

noudess (approved):
LGTM.

PR #4191, opened 2024-03-17
[Process] Windows preprocessor define in crash.cpp

joligario:
Originally changed in #2632


Akkadius:
What's the reason for this change though ?

joligario:
Originally or now?

Akkadius (approved):

PR #4192, opened 2024-03-17
[Loot] Fix issue with nested data being loaded multiple times

Akkadius:
Fixed an issue where nested data was getting loaded into memory multiple times. Added checks in nested loot data to ensure that nothing gets double added to memory.

### DIFF
--- a/zone/zone.h
+++ b/zone/zone.h
@@ -437,7 +437,7 @@ public:
 
 	// loot
 	void LoadLootTable(const uint32 loottable_id);
-	void LoadLootTables(const std::vector<uint32>& loottable_ids);
+	void LoadLootTables(const std::vector<uint32> in_loottable_ids);
 	void ClearLootTables();
 	void ReloadLootTables();
 	LoottableRepository::Loottable *GetLootTable(const uint32 loottable_id);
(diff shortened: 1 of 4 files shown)


fryguy503:
So far this is looking massively better!

zone/zone.h
```diff
@@ -437,7 +437,7 @@ class Zone {
 
 	// loot
 	void LoadLootTable(const uint32 loottable_id);
-	void LoadLootTables(const std::vector<uint32>& loottable_ids);
```
KimLS, line 440: Why did we remove this?

zone/zone_loot.cpp
```diff
@@ -5,20 +5,36 @@
 #include "../common/repositories/lootdrop_repository.h"
 #include "../common/repositories/lootdrop_entries_repository.h"
 
-void Zone::LoadLootTables(const std::vector<uint32> &loottable_ids)
```
KimLS, line 8: And this reference, why did we remove it?

KimLS (changes requested):
Don't see any issues otherwise but those dropped references forces a copy for no apparent reason.

zone/zone_loot.cpp
Akkadius: We modify it by filtering on it below, it wouldn't work on the reference so I made a copy. Very low overhead operation to do the copy anyhow

zone/zone.h
Akkadius: Same reference as below, answered

KimLS (approved):

Akkadius:
We spoke on this one

PR #4194, opened 2024-03-20
[Rules] Add World:Rules Rule

Kinglykrab:
# Notes
- Adds `World:Rules` rule to take the place of the variables table value.


common/ruletypes.h
```diff
@@ -329,6 +329,7 @@ RULE_BOOL(World, UseItemLinksForKeyRing, false, "Uses item links for Key Ring Li
 RULE_BOOL(World, UseOldShadowKnightClassExport, true, "Disable to have Shadowknight show as Shadow Knight (live-like)")
 RULE_STRING(World, IPExemptionZones, "", "Comma-delimited list of zones to exclude from IP-limit checks. Empty string to disable.")
 RULE_STRING(World, MOTD, "", "Server MOTD sent on login, change from empty to have this be used instead of variables table 'motd' value")
+RULE_STRING(World, Rules, "", "Server Rules, change from empty to have this be used instead of variables table 'rules' value, lines are pipe (|) separated, example: A|B|C")
```
Akkadius, line 332: Newlines should be terminated by something a little more standard such as `\n`, but I understand that `|` was in place here prior

Akkadius (approved):

PR #4195, opened 2024-03-22
[Bug Fix] Radiant/Ebon Crystals should only extract to 1000

fryguy503:
Creating a stack larger than 1000 can cause potential dupe issues further down the stream if other tasks are performed on the stack.

### DIFF
--- a/common/eq_packet_structs.h
+++ b/common/eq_packet_structs.h
@@ -130,6 +130,11 @@ enum CrystalReclaimTypes
 	Radiant = 4,
 };
 
+namespace ItemStackSizeConstraint {
+	constexpr int16 Minimum = 1;
+	constexpr int16 Maximum = 1000;
+}
+
 ///////////////////////////////////////////////////////////////////////////////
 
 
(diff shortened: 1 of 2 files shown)


joligario (commented):
I think RoF(2) allows up to 1K. I think Tit peaks at ~32. Thinking we could test and limit per client?

Kinglykrab:
> I think RoF(2) allows up to 1K. I think Tit peaks at 20? Thinking we could test and limit per client?

Are we sure that Titanium only stacks to 1,000? I don't have it to test.

joligario:
Hence thinking we need to test. I have titanium and I will check tonight.

fryguy503:
fwiw

```Live | 2007-03-22 14:00 | Changed stacksize from '1000' to '232'```

Kinglykrab:
Confirmed on Titanium we can stack up to 1,000. Decided to use the item's stacksize instead if the item is stackable, if not we default to a max of 1 reclaimed at a time. We clamp a stackable item's reclaim amount between 1 and 1,000, even if the stacksize goes beyond 1,000.

joligario:
Perfect! I like that much better. I can test later if you haven't already.

Kinglykrab:
I have not tested this yet.

Akkadius:
Pending testing it looks like

joligario:
Looks like Titanium can handle 1K charges.

image

Akkadius (approved):

PR #4069, opened 2024-02-12
[Skill Caps] Remove from shared memory and simplify

Kinglykrab:
# Notes
- Removes Skill Caps loading from shared memory and puts it into zone.
- Adds `id` column to `skill_caps`.
- Remove primary keys and use `id` as primary key.
- Add unique index using `skill_id`, `class_id`, `level`, and `cap`.
- Renames `class` to `class_id` in `skill_caps` table.
- Renames `skillID` to `skill_id` in `skill_caps` table.
- Regenerates Skill Caps repository.
- Adds `#reload skill_caps` to reload skill caps in real time.
- Resolves #4009.

# Video
[Screencast from 03-23-2024 04:01:17 PM.webm](https://github.com/EQEmu/Server/assets/89047260/a85d473b-b450-4ac4-a27d-e3bd3e3aa3e2)

PR #4205, opened 2024-03-24
[Skill Caps] Further improvements

Akkadius:
This further improves upon https://github.com/EQEmu/Server/pull/4069

We were generating tens of thousands of queries still for fetching skill caps. Not everything was happening entirely in memory. We are now.

**Changes**

* Create a `SkillCaps` class that can be used globally to pull skill cap references in both **world** and **zone**
* The `SkillCaps` class is in **common** so it can be used by both **world** and **zone**
* All references to skill caps now happen purely in memory using the functions that @Kinglykrab created, they are just now living in the shared `SkillCaps` class

**Testing**

Skills seem to be working

```
 World |    Info    | LoadSkillCaps Loaded [58359] Skill Cap Entries
  Zone |    Info    | LoadSkillCaps Loaded [58359] Skill Cap Entries 
```

image

Additional testing will need to be done

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -83,6 +83,7 @@ SET(common_sources
     shared_tasks.cpp
     shareddb.cpp
     skills.cpp
+    skill_caps.cpp
     spdat.cpp
     strings.cpp
     struct_strategy.cpp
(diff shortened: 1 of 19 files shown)


Kinglykrab (approved):

Kinglykrab:
Looks good to me. Will do so more thorough testing here shortly before merge.

PR #4208, opened 2024-03-25
[Feature] Additive Spell Focus from Worn slot with Limit Checks

KayenEQ:
New rule **UseAdditiveFocusFromWornSlotWithLimits**  (False by default)

Allows you place to focus effects in item worn slots which will then apply the focus additively and perform normal limit checks on those focus. This differs from regular focus behavior that only takes highest value. This will stack with all regular focus effects.

Example: Focus effect "Vampiric Fire" id 7849 which is a 60% fire focus. With the rule turned on. If you put this focus effect spell in the worn slot of 3 different items and equipped all three items. You would receive a 180% spell focus to fire spells.

This rule allows for vastly expanded itemization options for custom servers.

This is a new version of an old rule I wrote, **UseAdditiveFocusFromWornSlot** which allowed similar behavior but ignored focus limits. Thus hindering its full potential for itemization. **This rule is still functional but you can only choose to have one of these rules be true at the same time.**


Kinglykrab (approved):

zone/spell_effects.cpp
```diff
@@ -6801,9 +6801,53 @@ int64 Mob::GetFocusEffect(focusType type, uint16 spell_id, Mob *caster, bool fro
...
 
-	//Non-Live like feature to allow for an additive focus bonus to be applied from foci that are placed in worn slot. (No limit checks)
+
 	int32 worneffect_bonus = 0;
-	if (RuleB(Spells, UseAdditiveFocusFromWornSlot)) {
+	//Non-Live like feature to allow for an additive focus bonus to be applied from foci that are placed in worn slot. (Limit Checks)
+	if (RuleB(Spells, UseAdditiveFocusFromWornSlotWithLimits)) {
+		//Check if item focus effect exists for the mob.
+		if (itembonuses.FocusEffectsWornWithLimits[type]) {
+			const EQ::ItemData* TempItem = nullptr;
```
Kinglykrab, line 6810: `temp_item`

zone/spell_effects.cpp
```diff
@@ -6801,9 +6801,53 @@ int64 Mob::GetFocusEffect(focusType type, uint16 spell_id, Mob *caster, bool fro
...
+			const EQ::ItemData* UsedItem = nullptr;
```
Kinglykrab, line 6811: `used_item`

zone/spell_effects.cpp
```diff
@@ -6801,9 +6801,53 @@ int64 Mob::GetFocusEffect(focusType type, uint16 spell_id, Mob *caster, bool fro
...
+					else {
+						worneffect_bonus += CalcFocusEffect(type, TempItem->Worn.Effect, spell_id);
+					}
+				}
+
+				for (int y = EQ::invaug::SOCKET_BEGIN; y <= EQ::invaug::SOCKET_END; ++y) {
+					EQ::ItemInstance* aug = nullptr;
+					aug = ins->GetAugment(y);
+					if (aug) {
+						const EQ::ItemData* TempItemAug = aug->GetItem();
```
Kinglykrab, line 6835: `temp_item_aug`

Kinglykrab:
Looks good, a couple comments.

PR #4211, opened 2024-03-27
[LUA] Add Zone and Language Constants

fryguy503:
This will add Zone:

```lua
if eq.get_zone_id() == Zone.qeynos then
   foo()
end
```

It will also add Language:

```lua
if e.language == Language.ElderElvish and e.other:GetLanguageSkill(Language.ElderElvish) >= 100 then
   e.self:Say("You know my language!", Language.ElderElvish);
end
```

These changes should help avoid magic numbers in quests and aide in readability without the need for -- comments to clarify.


Kinglykrab:
Let's do all zone constants as lowercase so it's consistent.

Kinglykrab (approved):

PR #4213, opened 2024-03-28
[Bug Fix] Fix Auto Login Issue

Kinglykrab:
# Notes
- We were setting `live_name` value regardless of if we were zoning, causing us to automatically log in to the last character we'd logged in to before.
- Thanks @neckkola for this.


joligario (commented):
Not sure you need the change to client.cpp anymore after removing reference type

Ah, disregard. Dropped the const. Didn't notice that on my phone.

joligario (approved):

PR #4216, opened 2024-03-28
[Feature] Add Character Auto Login

Kinglykrab:
# Perl
- Add `$client->SetAutoLoginCharacterName()`.
- Add `$client->SetAutoLoginCharacterName(character_name)`.
- Add `$client->GetAutoLoginCharacterName()`.
- Add `quest::GetAutoLoginCharacterNameByAccountID(account_id)`.
- Add `quest::SetAutoLoginCharacterNameByAccountID(account_id, character_name)`.

# Lua
- Add `client:SetAutoLoginCharacterName()`.
- Add `client:SetAutoLoginCharacterName(character_name)`.
- Add `client:GetAutoLoginCharacterName()`.
- Add `eq.get_auto_login_character_name_by_account_id(account_id)`.
- Add `eq.set_auto_login_character_name_by_account_id(account_id, character_name)`.

# Commands
- Add `#set auto_login [0|1]` to enable or disable auto login on the your or your target's account, when using `1` it sets to current character name.
- Add `#show auto_login` to show your or your target's auto login character name.
- Both will say the system is disabled if `World:EnableAutoLogin` is `false`.

# Rules
- Add `World:EnableAutoLogin`, default is `false`, allows operators to enable this functionality.

# Notes
- Allows operators to give scripted or command access to the ability to set an auto login character name on an account.
- This will allow servers that support boxing of large numbers (6, 12, 18, etc) to allow players to set their accounts to auto login specific characters.
- Could also be used by GMs to auto login to their account on a specific GM character.


Akkadius:
A person can simply camp out initially and log back in another character right? Does that work?

Kinglykrab:
> A person can simply camp out initially and log back in another character right? Does that work?

Yessir.

world/client.cpp
```diff
@@ -194,14 +194,19 @@ bool Client::CanTradeFVNoDropItem()
...
+
 	if (is_player_zoning) {
 		live_name = database.GetLiveChar(GetAccountID());
-		if(database.GetAccountIDByChar(live_name) != GetAccountID()) {
+		if (database.GetAccountIDByChar(live_name) != GetAccountID()) {
 			eqs->Close();
 			return;
 		} else {
 			LogInfo("Telling client to continue session");
 		}
```
Akkadius, line 205: We return early in the `if` - the else should not be needed as we fall through

world/client.cpp
```diff
@@ -194,14 +194,19 @@ bool Client::CanTradeFVNoDropItem()
...
+	} else {
+		if (RuleB(World, EnableAutoLogin)) {
+			live_name = AccountRepository::GetAutoLoginCharacterNameByAccountID(database, GetAccountID());
+		}
```
Akkadius, line 209: Return early in the `if` of `is_player_zoning` and remove the `else`

world/client.cpp
```diff
@@ -194,14 +194,19 @@ bool Client::CanTradeFVNoDropItem()
...
-		if(database.GetAccountIDByChar(live_name) != GetAccountID()) {
+		if (database.GetAccountIDByChar(live_name) != GetAccountID()) {
 			eqs->Close();
 			return;
 		} else {
 			LogInfo("Telling client to continue session");
 		}
+	} else {
+		if (RuleB(World, EnableAutoLogin)) {
+			live_name = AccountRepository::GetAutoLoginCharacterNameByAccountID(database, GetAccountID());
```
Akkadius, line 208: For debugging purposes, it would be good to know if `live_name` ever ends up blank because we failed to find anything in the database. We assume happy path currently

Akkadius:
Rebase and perhaps attach testing videos illustrating toggling between characters

```
Login -> Character select (Select character A) -> World (enable auto login)
Login -> World (Character A) -> Character select (select character B)
Login -> World (Character B)
```

Also a separate video showing that once you've disabled it after having it initially enabled it works as expected.

Once there's that, it should be good to merge in my book

Kinglykrab:
Been busy with work, will get to this sometime in the next few days.

Kinglykrab:
Adding testing video.

Akkadius (approved):

PR #4223, opened 2024-03-29
[Cleanup] More reference types and initializers in `zone` project

joligario:
Would appreciate some extra eyes for sanity checks in case I missed something.


joligario:
Trying to do too much with this. Will come back to this some day.

PR #4224, opened 2024-03-30
[Spells] Implemented SPA 463 SE_SHIELD_TARGET

KayenEQ:
Implemented SPA 463 SE_SHIELD_TARGET

Live description: "Shields your target, taking a percentage of their damage".

Only example spell on live is an NPC who uses it during a raid event "Laurion's Song" expansion. SPA 54492 'Guardian Stance' Described as 100% Melee Shielding. Therefore implemented to the best of my judgement how I think it works.

Base value: percentage of melee damage shielded on target, the caster of the buff takes a percentage of the shielded damage.

Example of mechanic. Base value = 70. Caster puts buff on target. Each melee hit Buff Target takes 70% less damage, Buff Caster receives 30% of the melee damage. 

Added mechanic to cause buff to fade if target or caster are separated by a distance greater than the casting range of the spell. This allows similar mechanics to the /shield ability, without a range removal mechanic it would be too easy to abuse if put on a player spell. *can not confirm live does this currently but it makes sense

Can not be cast on self.


Kinglykrab (approved):

Valorith:
Very cool! Awesome to see @KayenEQ  getting back into some SPA work :D Thx!

PR #4225, opened 2024-03-30
[Feature] Add Zone Reload Quests

xackery:
I noticed that when a sleeping zone starts up, it was not reloading quests as I expected.

Since I wasn't sure if this was an intended effect, I put the behavior I desired behind a rule.

This rule, when enabled, will reload quests any time a zone goes through the bootup function and is not set to static.

This repairs a symptom I was experiencing where global quests that were changed while a zone was sleeping, and a `#reload world` is called, would still retain the old global quest data it had during initial zone init.

### DIFF
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -359,6 +359,7 @@ RULE_INT(Zone, FishingChance, 399, "Chance of fishing from zone table vs global 
 RULE_BOOL(Zone, AllowCrossZoneSpellsOnBots, false, "Set to true to allow cross zone spells (cast/remove) to affect bots")
 RULE_BOOL(Zone, AllowCrossZoneSpellsOnMercs, false, "Set to true to allow cross zone spells (cast/remove) to affect mercenaries")
 RULE_BOOL(Zone, AllowCrossZoneSpellsOnPets, false, "Set to true to allow cross zone spells (cast/remove) to affect pets")
+RULE_BOOL(Zone, QuestsReloadOnBootup, false, "Reloads all quests when a non-static zone boots up")
 RULE_CATEGORY_END()
 
 RULE_CATEGORY(Map)
--- a/zone/zone.cpp
+++ b/zone/zone.cpp
@@ -144,6 +144,11 @@ bool Zone::Bootup(uint32 iZoneID, uint32 iInstanceID, bool is_static) {
 		}
 	}
 
+	if (!is_static && RuleB(Zone, QuestsReloadOnBootup)) {
+		LogInfo("Reloading quests");
+		parse->ReloadQuests(RuleB(HotReload, QuestsResetTimersWithReload));
+	}
+
 	is_zone_loaded = true;
 
 	worldserver.SetZoneData(iZoneID, iInstanceID);


Valorith:
I've experienced this as well! This will be nice to have resolved. I really think think this should be default behavior. If a rule is used, hopefully it is default true

xackery:
> I've experienced this as well! This will be nice to have resolved. I really think this should be default behavior. If a rule is used, hopefully it is default true

I made the PR with the assumption current design was expected flow but we can instead always:
- sleeping zones honor `#reload world`Server_OP packets, similar to how `#reload rules` work

That is arguably more expected behavior, or yes, enable my new rule by default.

Akkadius:
This should be default behavior and should not need an additional rule. I don't know what the current state of debugging and reliability of things are but there's been several PR's attempts at this issue with lots of historical context involving different folks.

I know when we try to do this for both language types, it broke Lua encounters. 

There was an attempt to have it just do it for Perl, but then something else broke that I don't remember. 

https://github.com/EQEmu/Server/pull/3333
https://github.com/EQEmu/Server/pull/3648
https://github.com/EQEmu/Server/pull/4149

We should try to get the Perl reload re-introduced and tested https://github.com/EQEmu/Server/pull/3648

xackery:
I'll be using this on my server until we do decide how we want to address it, I just know that even with the above merged PR's, the issue still exists. And this appears to fix it. I didn't test all the weird edge cases like timer resets/encounters/etc, I will keep an eye out for that and test until something more formal is decided to remedy the situation..

Akkadius:
> I'll be using this on my server until we do decide how we want to address it, I just know that even with the above merged PR's, the issue still exists. And this appears to fix it. I didn't test all the weird edge cases like timer resets/encounters/etc, I will keep an eye out for that and test until something more formal is decided to remedy the situation..

If you look closer, many of the PR's have been backed out or reverted because of similar attempts that maybe didn't test everything or ran into issues. 

This PR as it stands would likely break Lua encounters as it repeats some of the other PR attempts such as https://github.com/EQEmu/Server/pull/3333

We need to make sure we're considering all cases and stability, that's why we do code review and have others flag things we've seen before and have a chance to chime in.

I think we all want the desired outcome of your PR and agree it should be addressed

xackery:
> This PR as it stands would likely break Lua encounters as it repeats some of the other PR attempts such as https://github.com/EQEmu/Server/pull/3333
But, it's opt in and default false.
I'll learn more about encounters and how to test against it, and see if I can resolve side effects.

Thanks.

Valorith:
Really hoping this one gets flushed out and doesn't die on the vine. This has been a thorn for a while.

Akkadius:
> Really hoping this one gets flushed out and doesn't die on the vine. This has been a thorn for a while.

I don't think it will die, it does need to get resolved. Needs more than surface level testing to validate timers, encounters, initial loads, zone controllers all work as expected.

hg — Today at 6:55 AM
Dunno, it needs to be fixed correctly but requires some refactoring and debugging to be sure things don't break
I assume this is a perl issue since lua already reloads

xackery:
> hg — Today at 6:55 AM Dunno, it needs to be fixed correctly but requires some refactoring and debugging to be sure things don't break I assume this is a perl issue since lua already reloads

Just noting on this comment, that lua is not reloading when a #reload world is called while a zone was sleeping. That was my motivator for this PR initially.

Give me a week or two to bump my server to newer expansions with encounters and I'll loop back on this to try to create test scenarios and deep dive what's going on with them/timers in perl/lua.

Akkadius:
I'll put up a separate PR for this

PR #4226, opened 2024-03-31
[Feature] Add CommonDamage HealDamage hooks

xackery:
This introduces two new lua hooks (expect a lot coming from me in the near future), CommonDamage and HealDamage.

This allows you to modify single spells or all spells, damage, and other variations of damage right before they are finally concluded.

This is a cleaner hook point compared to ones previously exposed for custom server modding.


xackery:
Going to redo with a master-synced branch

PR #4227, opened 2024-03-31
[Feature] Add LuaMod functions for CommonDamage and HealDamage

xackery:
This introduces two new lua hooks (expect a lot coming from me in the near future), CommonDamage and HealDamage.

This allows you to modify single spells or all spells, damage, and other variations of damage right before they are finally concluded.

This is a cleaner hook point compared to ones previously exposed for custom server modding.

PR #4228, opened 2024-03-31
[Feature] Fix ignore_default on lua mod damage

xackery:
This changes the new lua mods to require ignore_default to be set to true in order for a return value to be applied to the following logic.

Normally, ignore_default returns from a function, but due to how these functions work, that can't be done elegantly so this is the compromise.

I'll update the documentation and make an example accordingly.

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -3968,12 +3968,8 @@ void Mob::CommonDamage(Mob* attacker, int64 &damage, const uint16 spell_id, cons
 	int64 lua_ret = 0;
 	bool ignore_default = false;
 	lua_ret = LuaParser::Instance()->CommonDamage(this, attacker, damage, spell_id, static_cast<int>(skill_used), avoidable, buffslot, iBuffTic, static_cast<int>(special), ignore_default);
-	if (lua_ret != 0) {
-		damage = lua_ret;
-	}
-
 	if (ignore_default) {
-		//return lua_ret;
+		damage = lua_ret;
 	}
 #endif
 	// This method is called with skill_used=ABJURE for Damage Shield damage.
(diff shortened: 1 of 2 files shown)


Kinglykrab (approved):

Akkadius:
Original PR, merged early https://github.com/EQEmu/Server/pull/4227

Akkadius:
I think it's great we are investing more into the modding system. 

There might be an opportunity to think about composition as we start adding more mods so I'm looping @KimLS and @hgtw for possible input.

The change itself is straight forward otherwise. Thanks for correcting what was missing

KimLS (approved):
Looks good to me. Aorry Akk asked me to review it earlier but was off discord most of the day.

xackery:
While waiting on the reviews, I'm noticing that HealDamage is commonly called without context, so the caster is null and the spell_id is 65535 (unknown)..
This makes it difficult to do smart checks. 
I'm testing currently adding said context to HealDamage. 
- Noticed that raid healing is actually just using SetHP and bypassing HealDamage
- HealDamage will send a message to you and the caster based on context, so when I do add this in, there may be more heal messages coming in than before

hgtw (approved):
I'm not really familiar with lua mods but would it make more sense to just not check ignore_default for damage/heal at all and consider them non-stoppable? I'd expect ignore_default==false would still want to use the return from mods (unless there's issues with damage==0)

Otherwise I don't have any context so looks fine to me

xackery:
> I'm not really familiar with lua mods but would it make more sense to just not check ignore_default for damage/heal at all and consider them non-stoppable? I'd expect ignore_default==false would still want to use the return from mods (unless there's issues with damage==0)

Yeah, this is why I was hesitant to do the change I did for this.. I think for this case, both routes work fine, but it'd nice to create a standard and maybe create a new argument type for the behavior this function needs...

The challenge with a nonzero check is in cases the default value may not be 0.. It may work great for this function, but may be a case in the future this assumption is bad.

Maybe something like:

- `ignore_default` to ignore logic that follows a hook and return early
- `return_value` for values that return from a mod
- `override_default` to override the default value with returned value

So in this case, override_default needs to be set to true to make the return value override the default?

I'm open to debate. I think the only person to really touch lua mods before my PRs is KLS's, unsure if there was a deep thought on this regard prior.

PR #4234, opened 2024-04-01
[Spells] Implemented SPA 122 SE_ReduceSkill

KayenEQ:
Implemented SPA 122 - ReduceSkill

This spell effect acts as a debuff that reduces a clients skill value by percentage.

Example Live spell Curse of 1H Blunt (39667) "Decrease 1H Blunt Skill by 45%"
Base value: Skill ID, Limit: none, Max: none, **Formula value** = % reduction
For above example. Base=0 (1H Blunt skill id), Formula = 45

So if your 1H Blunt skills is 200, and you apply a 45% debuff you now have a 1H Blunt skill of 110 for the duration of the debuff.

The client GetSkill function I know looks a bit ugly, but I coded it in a way to be consistent with the current format and use the least amount of additional checks/calculations, since we do check this very often in the code. Essentially it just adds one check if you have the debuff, then will do all calculations to account for that, otherwise it just checks skills in the way it did before.


common/spdat.h
```diff
@@ -1321,6 +1321,7 @@ typedef enum {
 #define SE_Duration_HP_Pct				524 // implemented - Decrease Current Hit Points by % of Total Hit Points per Tick, up to a MAX per tick
 #define SE_Duration_Mana_Pct			525 // implemented - Decrease Current Mana by % of Total Mana per Tick, up to a MAX per tick
 #define SE_Duration_Endurance_Pct		526 // implemented - Decrease Current Endurance by % of Total Hit Points per Tick, up to a MAX per tick
+#define SE_SpellDamageConversion		527 // implemented - Add to a damage spell to return a % of damage done as a resource. base: % = value/10, limit: 0=HP,1=Mana,2=Endur, max: max resource returned
```
Kinglykrab, line 1324: Is this supposed to be here?

KayenEQ: Nope. That was from something else I thought I deleted it.

zone/client.h
```diff
@@ -801,7 +801,7 @@ class Client : public Mob
...
+	virtual uint16 GetSkill(EQ::skills::SkillType skill_id) const { if (skill_id <= EQ::skills::HIGHEST_SKILL) { if (spellbonuses.ReduceSkill[skill_id] > 0) { return(itembonuses.skillmod[skill_id] > 0 ? (itembonuses.skillmodmax[skill_id] > 0 ? std::min(m_pp.skills[skill_id] + itembonuses.skillmodmax[skill_id], m_pp.skills[skill_id] + ((m_pp.skills[skill_id] * (itembonuses.skillmod[skill_id] - spellbonuses.ReduceSkill[skill_id])) / 100)) : m_pp.skills[skill_id] + ((m_pp.skills[skill_id] * (itembonuses.skillmod[skill_id] - spellbonuses.ReduceSkill[skill_id])) / 100)) : m_pp.skills[skill_id] - ((m_pp.skills[skill_id] * spellbonuses.ReduceSkill[skill_id]) / 100)); }return(itembonuses.skillmod[skill_id] > 0 ? (itembonuses.skillmodmax[skill_id] > 0 ? std::min(m_pp.skills[skill_id] + itembonuses.skillmodmax[skill_id], m_pp.skills[skill_id] * (100 + itembonuses.skillmod[skill_id]) / 100) : m_pp.skills[skill_id] * (100 + itembonuses.skillmod[skill_id]) / 100) : m_pp.skills[skill_id]); }return 0; }
```
Akkadius, line 804: Can we break this out in multiple lines now, this is insane to read

zone/client.h
```diff
@@ -801,7 +801,25 @@ class Client : public Mob
 
 	void IncreaseSkill(int skill_id, int value = 1) { if (skill_id <= EQ::skills::HIGHEST_SKILL) { m_pp.skills[skill_id] += value; } }
 	void IncreaseLanguageSkill(uint8 language_id, uint8 increase = 1);
-	virtual uint16 GetSkill(EQ::skills::SkillType skill_id) const { if (skill_id <= EQ::skills::HIGHEST_SKILL) { return(itembonuses.skillmod[skill_id] > 0 ? (itembonuses.skillmodmax[skill_id] > 0 ? std::min(m_pp.skills[skill_id] + itembonuses.skillmodmax[skill_id], m_pp.skills[skill_id] * (100 + itembonuses.skillmod[skill_id]) / 100) : m_pp.skills[skill_id] * (100 + itembonuses.skillmod[skill_id]) / 100) : m_pp.skills[skill_id]); } return 0; }
+	virtual uint16 GetSkill(EQ::skills::SkillType skill_id) const
```
Akkadius, line 804: We need to make this more readable, this is insane even when I auto formatted it.

Let's break things out into readable variables that describe what things do. We have nested ternaries and it's so hard to make sense of this.

We're at a point as well where this should just be a whole function in `.cpp` not stuff in the header

zone/client.cpp
```diff
@@ -12393,3 +12393,34 @@ std::vector<Mob*> Client::GetRaidOrGroupOrSelf(bool clients_only)
...
+				return m_pp.skills[skill_id] + ((m_pp.skills[skill_id] * (itembonuses.skillmod[skill_id] - spellbonuses.ReduceSkill[skill_id])) / 100);
+			}
+			return m_pp.skills[skill_id] - ((m_pp.skills[skill_id] * spellbonuses.ReduceSkill[skill_id]) / 100);
+		}
+
+		if (itembonuses.skillmod[skill_id] > 0) {
+
+			if (itembonuses.skillmodmax[skill_id] > 0){
+				
+				if (itembonuses.skillmodmax[skill_id] > 0) {
```
Kinglykrab, line 12417: These are the same conditions, is this intentional or copy and paste error?

Akkadius (approved):
Still insane to read, maybe can be improved at a later date

PR #4238, opened 2024-04-01
[Bug Fix] SPA214 SE_MaxHPChange calculation errors corrected.

KayenEQ:
Bug: SPA 214 which changes hit points by percentage was not displaying correctly on the client. This was causing differences between client displayed hit points and server calculations.

Solution: The formula we were using to apply the spell and item bonuses using SPA 214 to client max hit points was incorrect. It should not have been applying it after the AA version (Natural Durability) and after Spell HP buffs (Ie. Virtue). The correct calculation was it should be applied at the same time as the AA version.

Client and server now display the same values within 5 hps of each other.

Fixed for Bots and Mercs as well.

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -4645,7 +4645,7 @@ bool Mob::SpellOnTarget(
 		However due to server thinking your healed, you are unable to correct it by healing.
 		Solution: You need to resend the HP update after buff completed and action packet resent.
 	*/
-	if (IsEffectInSpell(spell_id, SE_TotalHP) && (IsEffectInSpell(spell_id, SE_CurrentHPOnce) || IsEffectInSpell(spell_id, SE_CurrentHP))) {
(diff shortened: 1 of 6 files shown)


zone/spells.cpp
```diff
@@ -4639,6 +4639,16 @@ bool Mob::SpellOnTarget(
...
 
+	/*
+		Bug: When an HP buff with a heal effect is applied for first time, the heal portion of the effect heals the client and
+		updates HPs correctly server side, but client side the HP bar does not register it as a heal thus you display as less than full HP.
+		However due to server thinking your healed, you are unable to correct it by healing.
+		Solution: You need to resend the HP update after buff completed and action packet resent.
+	*/
+	if ((IsEffectInSpell(spell_id, SE_TotalHP) || IsEffectInSpell(spell_id, SE_MaxHPChange)) && (IsEffectInSpell(spell_id, SE_CurrentHPOnce) || IsEffectInSpell(spell_id, SE_CurrentHP))) {
+		SendHPUpdate(true);
+	}
```
Kinglykrab, line 4650: Seems to be the same as #4237.

KayenEQ: Ah I can remove it. I needed to properly test this. I will remove it so they can be separate PR's.

Akkadius (approved):
Use the PR template next time please

PR #4239, opened 2024-04-02
[Lua] Add SetEXP and SetAAEXP Lua Mods

xackery:
# Description

This adds two new lua hooks, SetEXP and SetAAEXP.

Due to the nature of said functions, I also added a new constant called ExpSource, where each call of said functions pass a context on what source the exp gain or loss comes from. This was important because without it, it was difficult to identify why an EXP change occurred. This also causes this PR to touch a lot of files to add ExpSource context. 

An existing lua mod exists called `GetExperienceForKill`, but it is limited to only kills. This new lua mod allows you to recognize and change EXP from all origins. This granularity was needed for a server vision I had, and I believe this PR will be hugely beneficial for future custom servers as they want to add experience modifications via lua.

The new ExpSources I decided on were:
Unknown, - A fallback default value, so new entries that SetExp have a flag to note it is an unknown origin
Quest, - For any quest provided EXP tweaks
GM, - For when a GM command executs EXP tweaks
Kill, - For when a mob is killed
Death, - For when a client dies
Resurrection, - For when a client gets resurrected
LDoNChest, - For when an LDoN chest rewards exp
Task, - For when a task rewards exp. This is arguably similar to quest and some tasks may give exp via quest
Sacrifice - For when the Sacrifice AA skill is used (possibly necro skill too?), I made this unique since it's a pretty unique mechanic

Fixes # (issue)

No fixes
## Type of change

# Testing

This is still ongoing. I created the PR as a draft to allow others give feedback but please do not put this into master until I'm given more time to test it locally.

Clients tested: 

RoF2. Other clients likely won't be affected as this is a server side change.

# Checklist

### DIFF
--- a/common/eq_constants.h
+++ b/common/eq_constants.h
@@ -1131,4 +1131,17 @@ namespace LeadershipAbilitySlot {
 	constexpr uint16 HealthOfTargetsTarget = 14;
 }
 
+enum ExpSource
+{
+	Unknown,
+	Quest,
+	GM,
+	Kill,
+	Death,
+	Resurrection,
+	LDoNChest,
+	Task,
+	Sacrifice
+};
+
 #endif /*COMMON_EQ_CONSTANTS_H*/
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -1946,7 +1946,7 @@ bool Client::Death(Mob* killer_mob, int64 damage, uint16 spell, EQ::skills::Skil
 			} else {
 				newexp -= exploss;
 			}
-			SetEXP(newexp, GetAAXP());
+			SetEXP(newexp, GetAAXP(), false, ExpSource::Death);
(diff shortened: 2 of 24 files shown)


xackery:
Windows build appears to dislike the Unknown enum, I'll wait for another developer to review this and give advise on preferred way to resolve this. (I was thinking of changing Unknown to ExpUnknown? Add prefix? unsure best approach or standards, I can review more later)

joligario:
Stupid windows using `Unknown` for disk types...

Do we need an `Unknown` experience type? Perhaps handle unknowns in the switch as a default or perhaps use `Other`?

xackery:
I'm going to likely revise this a bit, the default value unknown is more a negative thing.

xackery:
Closing in favor of a new PR #4292 4292

PR #4245, opened 2024-04-04
[Pets] Fix errors in sync with client of sit/stand

noudess:
# Description

Pet sit was problematic in a couple of situations.

1. On live, if you tell a pet to sit, and later engage, the pet returns to the mode the pet was in before sitting.  So if we was guarding, he went back to guarding.  If he was following, he went back to following.  EQEMU always sent the pet back to Follow.
2. The majority of pet commands on eqemu were sending the new animation, but not setting it with SetAppearance,  This resulted in sit/stand constantly getting out of sync because the cashed appearance variable on server side wasnt getting set.

## Type of change

# Testing

This is a before [video](http://norrath.no-ip.org/PetBefore.mp4).

This is an after the patch [video](http://norrath.no-ip.org/PetAfter.mp4).

_This PR changes bots in theory.  Bots already had the concept of storing the previous_pet_state, which we needed for this fix.  Rather than have a separate variable, I opted to using the same variable name that bots used, and moved the variable to mob.h.   Otherwise there would be both a mob and bot version of the variable.

I don't see how this could have harmed any bot functionality, but I'd love a volunteer to test if they understand bots and how bots use this variable._

Clients tested: 

RoF2

# Checklist

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -4050,7 +4050,7 @@ void Mob::CommonDamage(Mob* attacker, int64 &damage, const uint16 spell_id, cons
 				Mob *owner = GetOwner();
 				if (owner && owner->IsClient()) {
 					if (GetPetOrder() == SPO_Sit) {
-						SetPetOrder(SPO_Follow);
+						SetPetOrder(GetPreviousPetOrder());
 					}
 					// fix GUI sit button to be unpressed and stop sitting regen
 					owner->CastToClient()->SetPetCommandState(PET_BUTTON_SIT, 0);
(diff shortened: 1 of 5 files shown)


MortimerGreenwald:
Can someone Merge this one so I can test?

MortimerGreenwald:
Confirmed the fix using bots, using ^guard and ^guard clear on Karana

noudess:
> Confirmed the fix using bots, using ^guard and ^guard clear on Karana. Looks good.

Thanks!!

Akkadius (approved):

PR #4246, opened 2024-04-05
[Bug Fix] Hero forge armor bug on login and show helm toggle.

noudess:
# Description

1. Other clients witnessing a new client that is wearing Hero Forge armor would not see the armor until a zone or re-equip.
2. A character wearing hero forge armor on head would lose the model when toggling show helm.

HELP!!
**This fix potentially removes the need for the heroforge_wearchange_timer which is set in Client::CompleteConnect().  Changing that timer to be longer also fixes the problem, but was not the root cause.  I would love some expert opinion on if taking that timer out completely is advisable.**

@KayenEQ 

https://github.com/EQEmu/Server/pull/1883

HELP!!

## Type of change

# Testing

Video showing bug [here](http://norrath.no-ip.org/herob4.mp4).
Video showing fix [here](http://norrath.no-ip.org/heroafter.mp4).

Clients tested:  ROF2

# Checklist

### DIFF
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -278,8 +278,6 @@ Client::Client(EQStreamInterface *ieqs) : Mob(
 	if (RuleI(World, PVPMinLevel) > 0 && level >= RuleI(World, PVPMinLevel) && m_pp.pvp == 0) SetPVP(true, false);
 	dynamiczone_removal_timer.Disable();
 
-	heroforge_wearchange_timer.Disable();
-
 	//for good measure:
 	memset(&m_pp, 0, sizeof(m_pp));
 	memset(&m_epp, 0, sizeof(m_epp));
--- a/zone/client.h
+++ b/zone/client.h
@@ -1953,8 +1953,6 @@ private:
 	Timer task_request_timer;
 	Timer pick_lock_timer;
 
-	Timer heroforge_wearchange_timer;
-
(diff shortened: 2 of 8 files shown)


KayenEQ:
If that change solves the root issue and it looks like it does then by all means get rid of the timer. I only coded it like that cause I couldn't figure out the underlying the problem at the time and by delaying the packet a few milliseconds it seemed to resolve the issue for me.

KayenEQ:
I tested the change myself. I couldn't find any issues with other clients seeing heroforge armor on zone in or on item swaps.
I reverted the timer changes and put the wearchange send back in client connect where it originally was. With this fix it appears that sending the wearchange on client connect doesn't affect heroforge visibility. However, I don't know if there were other situations that require that code to be present, so to avoid other potential issues thus I recommended just keeping it in place.

noudess:
> I tested the change myself. I couldn't find any issues with other clients seeing heroforge armor on zone in or on item swaps. I reverted the timer changes and put the wearchange send back in client connect where it originally was. With this fix it appears that sending the wearchange on client connect doesn't affect heroforge visibility. However, I don't know if there were other situations that require that code to be present, so to avoid other potential issues thus I recommended just keeping it in place.

Did as you asked.

KayenEQ (approved):

Akkadius (approved):

PR #4271, opened 2024-04-19
[Commands] Adjust #kill/#list Commands

Kinglykrab:
# Description
- Adds the ability to use `#kill` with a supplied entity ID, i.e. `#kill 123`.
- Adds bot support to the `#list` command and adds `#kill` saylinks to the `#list` data if the client can use the `#kill` command.

## Type of Change

# Testing
image
image

# Checklist

### DIFF
--- a/zone/gm_commands/kill.cpp
+++ b/zone/gm_commands/kill.cpp
@@ -2,27 +2,40 @@
 
 void command_kill(Client *c, const Seperator *sep)
 {
-	auto target = c->GetTarget();
-	if (!target) {
-		c->Message(Chat::White, "You must have a target to use this command.");
+	if (!sep->IsNumber(1) && !c->GetTarget()) {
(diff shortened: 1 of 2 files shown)


Akkadius:
> I'm not looking for a change in this PR, but I think in the future we should split changes. 1 PR for enhancing #kill for entity IDs and another for adding bot support. The more singular we make PRs, the easier it is to review quickly, as well as rollback 1 change while leaving the other.

Agreed

Akkadius (approved):

Akkadius (changes requested):
Your test images show `NPCss` so we're getting double plurality due to some of the changes you made

Kinglykrab:
> Your test images show `NPCss` so we're getting double plurality due to some of the changes you made

Resolved.

Akkadius (approved):

PR #4273, opened 2024-04-19
[Bug Fix] All NPC classes could stun with kick/only warriors could be stunned

noudess:
# Description

The intention of the code prior to this patch was that all NPC warriors that had kick skill would also have a chance to stun, while only level 56 and above PC warriors did.   The ACTUAL result of the current implementation was that all classes that had kick could stun, only warriors could be stunned.

This PR fixes the code to operate as intended.  However, since all servers currently only see warriors getting stunned, I expect many more people will be interested in being able to limit the level at which NPC Warriors get a chance to stun.  I therefore added in two rules for PC and NPC levels at which they can have a chance of stun when they kick.  These levels havce been set to the existing code behavior (1 for NPC, 56 for PC).

This bug has been in eqemu 8 years!

## Type of change

# Testing

I tested these cases:

- only warrior attackers were given a chance to stun (was wrong before PR)
- all  PCs being attacked could be stunned by a kick (was wrong before PR)
- altering either rule results in the correct level threshold for a stun chance for both NPCs and PCs.

Clients tested: RoF2

# Checklist


zone/attack.cpp
```diff
@@ -4302,8 +4302,15 @@ void Mob::CommonDamage(Mob* attacker, int64 &damage, const uint16 spell_id, cons
...
-				(attacker->GetLevel() > 55 || attacker->IsNPC()) && GetClass() == Class::Warrior) {
-				can_stun = true;
+					attacker->GetClass() == Class::Warrior) {
+				int stun_level;
+				if (attacker->IsClient()) {
+					stun_level = RuleI(Combat, PCKickStunLevel);
+				}
+				else {
+					stun_level = RuleI(Combat, NPCKickStunLevel);
+				}
```
Akkadius, line 4312: Remove the `else` and have this be the default case

```cpp
int stun_level = RuleI(Combat, NPCKickStunLevel);
if (attacker->IsClient()) {
	stun_level = RuleI(Combat, PCKickStunLevel);
}
```

noudess: Not sure why, as now clients get 2 assignments instead of 1.  But I'm ok with it.  Committing now.

Kinglykrab (approved):

PR #4279, opened 2024-04-20
[Lua] Add UpdatePersonalFaction Lua Mod

xackery:
# Description

This introduces a new way to increase faction via lua mods. This allows players to create personalized, custom lua-driven ways to give faction boosts.

Here's an example of a froglok in seb with a 4x positive modifier with debug messages showing math:
image

Fixes # (issue)

## Type of change

- [x] This change requires a documentation update

# Testing

Here's a [sample file](https://gist.github.com/xackery/9d70f4108b441bbeaec46bd43cfec47c) where I have a mighty system to boost exp and faction by a modifier. With what is exposed via this method, it's also possible to fine-tune a specific faction so a player can have it boosted when they achieve something on a server.

No snippet of above, as the sample file should show a decent example of how it works.

Clients tested: RoF2

# Checklist

### DIFF
--- a/zone/lua_mod.cpp
+++ b/zone/lua_mod.cpp
@@ -37,6 +37,7 @@ void LuaMod::Init()
 	m_has_get_experience_for_kill = parser_->HasFunction("GetExperienceForKill", package_name_);
 	m_has_common_outgoing_hit_success = parser_->HasFunction("CommonOutgoingHitSuccess", package_name_);
 	m_has_calc_spell_effect_value_formula = parser_->HasFunction("CalcSpellEffectValue_formula", package_name_);
+	m_has_update_personal_faction = parser_->HasFunction("UpdatePersonalFaction", package_name_);
 	m_has_register_bug = parser_->HasFunction("RegisterBug", package_name_);
 	m_has_common_damage = parser_->HasFunction("CommonDamage", package_name_);
 	m_has_heal_damage = parser_->HasFunction("HealDamage", package_name_);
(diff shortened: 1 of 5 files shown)


KimLS (approved):

Kinglykrab (approved):

zone/lua_mod.cpp
```diff
@@ -177,6 +178,61 @@ void GetExtraAttackOptions(luabind::adl::object &ret, ExtraAttackOptions *opts)
...
 }
 
+void LuaMod::UpdatePersonalFaction(Mob *self, int32 npc_value, int32 faction_id, int32 current_value, int32 temp, int32 this_faction_min, int32 this_faction_max, int32 &return_value, bool &ignore_default)
+{
+	int start = lua_gettop(L);
+
+	try {
+		if (!m_has_update_personal_faction) {
+			return;
+		}
```
Akkadius, line 188: Should we be calling this check even earlier for all of our mod calls? We're needlessly calling `lua_gettop` in potentially very hot paths for functionality that is not enabled for most users

Akkadius: General question - doesn't have to block this PR - but we should probably return earlier for the rest of our mods

Akkadius (approved):

zone/lua_mod.cpp
xackery: I'll work on a PR to test this situation, and see if we can exit earlier with no negatives

xackery:
Documentation updated at https://docs.eqemu.io/server/scripting/lua-mods/

PR #4281, opened 2024-04-23
[Bug Fix] MariaDB 10.5.x error with parcel function GetParcelCountAndCharacterName

neckkola:
# Description

As experienced by N0ctnrl, parcel merchants and the #parcel add command would respond that the player could not be found, even when the player name was correct.  The problem was traced to the query within the  GetParcelCountAndCharacterName function returning incorrect results.  It was further traced to mariadb 10.5.22 as the problem was not seen on 10.4.13 or 10.10.3. The problem was reproduced on 10.5.24.

Fixes # (issue)
Updated the query with the keyword LEFT to establish the JOIN.  Mariadb 10.5.22 and 10.5.24 then functioned properly.  This was retested with 10.4.13 and 10.10.3 without issue.

## Type of change
Bug fix

# Testing
The result was tested against 10.4.13, 10.5.24 and 10.10.3.  N0ctnrl also tested with their install of 10.5.22 and reported that the problem was fixed.

Clients tested: 
RoF2

# Checklist

### DIFF
--- a/common/repositories/character_parcels_repository.h
+++ b/common/repositories/character_parcels_repository.h
@@ -58,7 +58,7 @@ public:
 		auto results = db.QueryDatabase(
 			fmt::format(
 				"SELECT c.name, COUNT(p.id), c.id FROM character_data c "
-				"JOIN character_parcels p ON p.char_id = c.id "
+				"LEFT JOIN character_parcels p ON p.char_id = c.id "
 				"WHERE c.name = '{}' "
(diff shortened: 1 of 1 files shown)


Kinglykrab (approved):

Kinglykrab:
Let’s change the name of this pull request to something more descriptive and make sure it merges with the descriptive name so when we’re looking for this in the future we can more easily find it.

PR #4288, opened 2024-04-25
[Doors] Replace magic # in code with a field in doors table for when door closes.

noudess:
# Description

Existing code uses a hard coded 5000ms for a timer that tries to match the client in regards to when doors on the client close themselves.  This PR allows each door to have it's own value, as I have found that doors in runnyeye run on 10000ms timer.  Perhaps if we gather enough data in the db over time, as we repair values in the new field in doors, we can draw some conclusions that would allow a more general rule.  But if seems like a small price to pay (one field) to be able to control the server to match client behavior per door.

This PR defaults the database to 5000ms, as per existing behaviour.

image

## Type of change

# Testing

Tested doors wil various values and added a temporary log entry displaying door timer setting.

Clients tested: 

# Checklist

### DIFF
--- a/common/database/database_update_manifest.cpp
+++ b/common/database/database_update_manifest.cpp
@@ -5545,7 +5545,7 @@ ALTER TABLE `character_parcels`
 		.match = "",
 		.sql = R"(
 ALTER TABLE `doors`
-ADD COLUMN `close_timer_ms` smallint(8) DEFAULT 5000 AFTER `is_ldon_door`;
+ADD COLUMN `close_timer_ms` smallint(8) UNSIGNED NOT NULL DEFAULT 5000 AFTER `is_ldon_door`;
 )",
 		.content_schema_update = true
 	}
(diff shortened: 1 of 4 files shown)


common/database/database_update_manifest.cpp
```diff
@@ -5536,6 +5536,18 @@ ALTER TABLE `character_parcels`
...
+	},
+	ManifestEntry{
+		.version = 9273,
+		.description = "2024_04_24_door_close_timer.sql",
+		.check = "SHOW COLUMNS FROM `doors` LIKE 'close_timer'",
+		.condition = "empty",
+		.match = "",
+		.sql = R"(
+ALTER TABLE `doors`
+ADD COLUMN `close_timer_ms` smallint(8) DEFAULT 5000 AFTER `is_ldon_door`;
```
Kinglykrab, line 5548: Can we make this `NOT NULL` and `UNSIGNED` as well?

Kinglykrab (approved):
Small comment, otherwise looks good to me.

common/database/database_update_manifest.cpp
noudess: Done.

PR #4292, opened 2024-05-01
[Lua] Add SetExp/SetAAExp lua mods

xackery:
# Description

(Note: this is a second PR, original #4239 since master moved so much since)

Highlights from previous PR:
- Added ExpSource to first line of each call instead of last, to not conflict with optional fields
- Removed Unknown type, requiring type to be specified when called.

This adds two new lua hooks, SetEXP and SetAAEXP.

Due to the nature of said functions, I also added a new constant called ExpSource, where each call of said functions pass a context on what source the exp gain or loss comes from. This was important because without it, it was difficult to identify why an EXP change occurred. This also causes this PR to touch a lot of files to add ExpSource context.

An existing lua mod exists called GetExperienceForKill, but it is limited to only kills. This new lua mod allows you to recognize and change EXP from all origins. This granularity was needed for a server vision I had, and I believe this PR will be hugely beneficial for future custom servers as they want to add experience modifications via lua.

The new ExpSources I decided on were:
Quest, - For any quest provided EXP tweaks
GM, - For when a GM command executs EXP tweaks
Kill, - For when a mob is killed
Death, - For when a client dies
Resurrection, - For when a client gets resurrected
LDoNChest, - For when an LDoN chest rewards exp
Task, - For when a task rewards exp. This is arguably similar to quest and some tasks may give exp via quest
Sacrifice - For when the Sacrifice AA skill is used (possibly necro skill too?), I made this unique since it's a pretty unique mechanic

Fixes # (issue)

No fixes
## Type of change

# Testing

Tested on my server using [this logic](<https://gist.github.com/xackery/8ffbce249da611e164c37f3ceae6f181>), basically I have two opt in features to boost exp.

Clients tested: 

RoF2. Other clients likely won't be affected as this is a server side change.

# Checklist


Akkadius (approved):

xackery:
Documentation updated at https://docs.eqemu.io/server/scripting/lua-mods/

PR #4302, opened 2024-05-08
[Quests] Hot Reload Changes

Akkadius:
# Description

This PR addresses a few minor issues with quest hot reloading that were found when porting hot reloading over to Spire natively away from Occulus itself.

**Fixes**

* An issue where if global auto reload was turned off, you'd see as many messages to reload as there were zones
* An issue where double messaging can be queued and sent, we use debouncing for this now

## Type of change

# Testing

**Problem before**

We see as many messages as we do zones booted

image

**After**

One line, as expected

image

# Checklist

### DIFF
--- a/world/console.cpp
+++ b/world/console.cpp
@@ -918,6 +918,8 @@ void ConsoleReloadWorld(
 	safe_delete(pack);
 }
 
+auto debounceReload = std::chrono::system_clock::now();
+
 /**
  * @param connection
  * @param command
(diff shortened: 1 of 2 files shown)


Kinglykrab (approved):

world/console.cpp
```diff
@@ -918,6 +918,8 @@ void ConsoleReloadWorld(
 	safe_delete(pack);
 }
 
+auto debounceReload = std::chrono::system_clock::now();
```
Kinglykrab, line 921: `debounce_reload`?

PR #4314, opened 2024-05-18
[Bug Fix] Fix issue with #suspend

Kinglykrab:
# Description
- Described by Dracaena of EZ Server [here](https://discord.com/channels/212663220849213441/1240979738366054451).

## Type of Change
- [X] Bug fix

# Testing
image
image
image
image

# Testing (Post-Repository Change)
image
image
image
image

# Checklist

### DIFF
--- a/common/database.cpp
+++ b/common/database.cpp
@@ -206,9 +206,19 @@ void Database::LoginIP(uint32 account_id, const std::string& login_ip)
 	QueryDatabase(query);
 }
 
-int16 Database::CheckStatus(uint32 account_id)
+int16 Database::GetAccountStatus(uint32 account_id)
 {
-	return AccountRepository::GetAccountStatus(*this, account_id);
+	auto e = AccountRepository::FindOne(*this, account_id);
+
+	if (e.suspendeduntil > 0 && e.suspendeduntil < std::time(nullptr)) {
+		e.status         = 0;
+		e.suspendeduntil = 0;
+		e.suspend_reason = "";
+
+		AccountRepository::UpdateOne(*this, e);
+	}
+
+	return e.status;
 }
 
 uint32 Database::CreateAccount(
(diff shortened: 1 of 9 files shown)


common/repositories/character_data_repository.h
```diff
@@ -64,6 +64,134 @@ class CharacterDataRepository: public BaseCharacterDataRepository {
...
+			e.zone_id                 = row[6] ? static_cast<uint32_t>(strtoul(row[6], nullptr, 10)) : 0;
+			e.zone_instance           = row[7] ? static_cast<uint32_t>(strtoul(row[7], nullptr, 10)) : 0;
+			e.y                       = row[8] ? strtof(row[8], nullptr) : 0;
+			e.x                       = row[9] ? strtof(row[9], nullptr) : 0;
+			e.z                       = row[10] ? strtof(row[10], nullptr) : 0;
+			e.heading                 = row[11] ? strtof(row[11], nullptr) : 0;
+			e.gender                  = row[12] ? static_cast<uint8_t>(strtoul(row[12], nullptr, 10)) : 0;
+			e.race                    = row[13] ? static_cast<uint16_t>(strtoul(row[13], nullptr, 10)) : 0;
+			e.class_                  = row[14] ? static_cast<uint8_t>(strtoul(row[14], nullptr, 10)) : 0;
+			e.level                   = row[15] ? static_cast<uint32_t>(strtoul(row[15], nullptr, 10)) : 0;
```
Akkadius, line 100: We don't want to be defining all of this manually and should be leaving it to code generation.

We'd be better off having a `GetWhere` here, return a vector/array of struct entries and just return the first entry and remove all of this

Akkadius:
Minor comment should be addressed, will need some quick testing otherwise looks good

common/database.cpp
```diff
@@ -2094,3 +2094,28 @@ void Database::ClearGuildOnlineStatus()
...
+	}
+
+	const int replaced = AccountRepository::ReplaceMany(*this, l);
+
+	if (!replaced) {
+		LogError(
+			"Failed to clear expired suspensions for [{}] account{}",
+			l.size(),
+			l.size() != 1 ? "s" : ""
+		);
```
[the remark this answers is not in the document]
Kinglykrab: True enough. I’ll remove it.

world/main.cpp
```diff
@@ -179,6 +179,8 @@ int main(int argc, char **argv)
 	EQTimeTimer.Start(600000);
 	Timer parcel_prune_timer(86400000);
 	parcel_prune_timer.Start(86400000);
+	Timer expired_suspensions_prune_timer(60000);
+	expired_suspensions_prune_timer.Start(60000);
```
Akkadius, line 183: Does this need to run every minute ?

Once an hour is fine. 

We'll also need to make sure when we're querying for the account during login we are filtering by suspension date as well. That way suspensions are cleared the moment time passes, not when we null out the field.

Kinglykrab: Removed timer per our discussion.

Akkadius (approved):

PR #4332, opened 2024-05-26
[Rule] Added MeleeMitigation Level Difference Roll Adjusted for level diffs

fryguy503:
# Description

This patch allows server ops to adjust the MeleeMitigation of NPC's based on the level difference and tune the amount adjusted. This will give greater diffs an advantage/disadvantage based on the values. -1 value should disabled and not change any current functionality.

## Type of change

# Checklist

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -1043,18 +1043,22 @@ void Mob::MeleeMitigation(Mob *attacker, DamageHitInfo &hit, ExtraAttackOptions 
 	auto roll = RollD20(hit.offense, mitigation);
 
 	// Add bonus to roll if level difference is sufficient
-	int level_diff = attacker->GetLevel() - GetLevel();
-	int level_diff_roll_check = RuleI(Combat, LevelDifferenceRollCheck);
+	const int level_diff            = attacker->GetLevel() - GetLevel();
+	const int level_diff_roll_check = RuleI(Combat, LevelDifferenceRollCheck);
 
-	if (level_diff_roll_check >= 0 && level_diff > level_diff_roll_check) {
-		roll += RuleR(Combat, LevelDifferenceRollBonus);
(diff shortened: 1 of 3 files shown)


zone/attack.cpp
```diff
@@ -1042,6 +1042,20 @@ void Mob::MeleeMitigation(Mob *attacker, DamageHitInfo &hit, ExtraAttackOptions
 
 	auto roll = RollD20(hit.offense, mitigation);
 
+	// Add bonus to roll if level difference is sufficient
+	signed lvldiff = attacker->GetLevel() - this->GetLevel();
```
Kinglykrab, line 1046: `signed`? Why not just use `int`? Also don't need `this`. Can name the variable `level_diff`.

zone/attack.cpp
```diff
@@ -1042,6 +1042,20 @@ void Mob::MeleeMitigation(Mob *attacker, DamageHitInfo &hit, ExtraAttackOptions
...
+		roll += RuleR(Combat, LevelDifferenceRollBonus);
+		if (roll > 2.0) {
+			roll = 2.0;
+		}
+	} else if (RuleI(Combat, LevelDifferenceRollCheck) >= 0 && lvldiff < (-RuleI(Combat, LevelDifferenceRollCheck))) {
+		roll -= RuleR(Combat, LevelDifferenceRollBonus);
+		if (roll < 0.1) {
+			roll = 0.1;
+		}
+	}
```
Kinglykrab, line 1057: In this case since the types that we're using for the rule and a variable are consistent, it'd save code duplication to use a variable for all of these.
```cpp
int level_diff_roll_check = RuleI(Combat, LevelDifferenceRollCheck);
```

Kinglykrab:
May need to adjust Tune method(s).

Kinglykrab (approved):

PR #4334, opened 2024-05-26
[Bug Fix] Raid Targets should not be Blindable as this will break all spell casting AI.

fryguy503:
# Description

Raid Targets should not be Blindable as this will break all spell casting AI.

## Type of change

# Checklist


Kinglykrab:
Is this Live-like?

fryguy503:
My understanding is it has always been this way that raid targets should not be blindable as it can completely trivialize many encounters. (Thanks Blind Whips). I can rule this if we think it should be.

Kinglykrab (approved):

Kinglykrab:
> My understanding is it has always been this way that raid targets should not be blindable as it can completely trivialize many encounters. (Thanks Blind Whips). I can rule this if we think it should be.

I’ll leave that up to you.

Kinglykrab:
> My understanding is it has always been this way that raid targets should not be blindable as it can completely trivialize many encounters. (Thanks Blind Whips). I can rule this if we think it should be.

I think since it changes the way combat works, it should be ruled and by default it should be enabled if that's Live-like.

PR #4343, opened 2024-05-26
[Scripts] Fix zone data load ordering issue

Akkadius:
# Description

This is to adjust some behavior change caused from https://github.com/EQEmu/Server/pull/4298

The script reload change is not necessarily an issue by itself but the order in which it is called causes issues with scripts depend on some of those things being initialized such as dynamic zones.

This attempts to move some load ordering around. Still needs testing

## Type of change

# Testing

Scripts still load properly prior to zone bootup

```perl
sub EVENT_SPAWN {
    quest::settimerMS("test", 10);
}

sub EVENT_TIMER {
    quest::say("Trigger");
}
```

image

# Checklist

### DIFF
--- a/zone/zone.cpp
+++ b/zone/zone.cpp
@@ -1157,9 +1157,17 @@ bool Zone::Init(bool is_static) {
 	DynamicZone::CacheAllFromDatabase();
 	Expedition::CacheAllFromDatabase();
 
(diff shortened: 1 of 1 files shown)


Kinglykrab (approved):

fryguy503 (approved):
lgtm - Thank you Akk!

PR #4348, opened 2024-05-26
[Cleanup] Cleanup Client File Exporting

Kinglykrab:
# Description
- Prior to this we were either performing raw queries and putting it together manually, or using a mix of this and repositories.
- To alleviate this I created repository methods that grab a `^` separated value for each row of these tables, allowing us an easy way to write this string to the file without building it out ourselves or being error prone.

## Type of Change
- [X] New feature

# Files
- [BaseData.txt](https://gist.github.com/Kinglykrab/26664938cd2c1ca1d587039e937d2780)
- [dbstr_us.txt](https://gist.github.com/Kinglykrab/2e32f550105dcfa06e13d2c1b7e63c9e)
- [SkillCaps.txt](https://gist.github.com/Kinglykrab/95c5139439fdd06661d4ebd9638b5db8)
- [spells_us.txt](https://gist.github.com/Kinglykrab/c7965d7f285432b3425690d3d1c5d213)

# Testing
- Exported all 4 file types after these changes to confirm everything works properly, logged in game, etc.
## Exporting
image
## Base Data
image
image
image
## Skill Caps
- Having issues here, need to test more and figure out what's going on with Inventory Skills Window.
## Spells
image

# Checklist

### DIFF
--- a/client_files/export/main.cpp
+++ b/client_files/export/main.cpp
@@ -159,13 +159,7 @@ void ExportSkillCaps(SharedDatabase* db)
 		return;
 	}
 
-	const uint8 skill_cap_max_level = (
-		RuleI(Character, SkillCapMaxLevel) > 0 ?
-		RuleI(Character, SkillCapMaxLevel) :
-		RuleI(Character, MaxLevel)
-	);
-
-	const auto& lines = SkillCapsRepository::GetSkillCapFileLines(*db, skill_cap_max_level);
+	const auto& lines = SkillCapsRepository::GetSkillCapFileLines(*db);
 
 	const std::string& file_string = Strings::Implode("\n", lines);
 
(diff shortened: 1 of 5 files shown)


Akkadius:
Can we actually test using the exported files in game? Attach screens that skills, spells all function as expected. Db strings aren't jacked etc.

Akkadius:
> Would like some eyes on why a valid skill caps file and skill caps table would not work properly.
> 
> Here is the [Skill Caps file](https://github.com/EQEmu/Server/files/15473270/SkillCaps.txt) I am using.
> 
> 1. Maxing my skills works fine.
>    image
> 2. Visually the skills window does not show the skills.
>    image
> 3. Database shows proper skill values for character.
>    image
>    image
>    image
> 4. Skill caps table has proper data as well.
>    image

Should be resolved with the other PR

Akkadius:
This has had a lot of changes since initial draft, we should some more thorough testing of all the export types

Akkadius:
I can look at this more later in the week, we should get some others to help out with testing on this to be safe

joligario (approved):

PR #4350, opened 2024-05-26
[Performance] Improve SkillCaps::GetTrainLevel() Efficiency

Kinglykrab:
# Description
- Being we're using `.find` with an integer-based keying, the outer loop is unnecessary now.
- The loop initially existed prior to the integer-based keying and was not removed, as a result of this we are needlessly looping `m_skill_caps` and performing extra `.find` calls.

# Perl
- Add `$client->GetSkillTrainLevel(skill_id)`.

## Perl Script
```pl
sub EVENT_SAY {
	if ($text=~/#a/i) {
		foreach my $skill_id (1, 2) {
			my $skill_name = quest::getskillname($skill_id);
			my $train_level = $client->GetSkillTrainLevel($skill_id);
			quest::message(315, "$skill_name ($skill_id) trains at level $train_level.");
		}
	}
}
```

# Lua
- Add `client:GetSkillTrainLevel(skill_id)`.

## Lua Script
```lua
function event_say(e)
	if e.message:findi("#a") then
		for skill_id = 1, 2 do
			local skill_name = eq.get_skill_name(skill_id);
			local train_level = e.self:GetSkillTrainLevel(skill_id);
			eq.message(315, string.format("%s (%d) trains at level %d.", skill_name, skill_id, train_level));
		end
	end
end
```

# Before Benchmark
image

# After Benchmark
image

## Type of Change
- [X] Bug fix

# Checklist

### DIFF
--- a/common/skill_caps.cpp
+++ b/common/skill_caps.cpp
@@ -1,4 +1,5 @@
 #include "skill_caps.h"
+#include "timer.h"
 
 SkillCaps *SkillCaps::SetContentDatabase(Database *db)
 {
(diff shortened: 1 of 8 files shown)


Akkadius:
Tested myself

Skills menu is populated, skills increase

image

Akkadius (approved):

PR #4351, opened 2024-05-27
[Feature] Extend #parcels Command

Kinglykrab:
# Description
- Add support for item links and augments to existing `#parcels add` subcommand.
- Remove references to an unimplemented `#parcels details` subcommand.

## Type of Change
- [X] New feature

# Testing
image
image
image
image
image

# Checklist


neckkola:
Good evening @Kinglykrab 

I ran a few tests and would highlight a few items for your consideration.

Item #1 - As seen below The Ak'Anon Report Vol 9. No 10. shows a quantity of 0.  Should be 1.
Item #2 - Should probably add (copper) to the end of money.
Item #3 - The item link for money shows (1c) even though it is 1000c/1p
parcelerror1

Item #4 - #parcels 99990 1000 32 does not produce the note.  It appears that the note must contain a non-numeric value.  #parcels 99990 1000 32s works for example.

Item #5 - #parcels add Rola ItemLink 1 note returns the #parcel syntax and does not add the parcel.  Command is noted below.
parcelerror2

Item #6 - #parcels add Rola 6941 1 85490 85490 Note sends the parcel and shows the Fine Steel Jute with the augment in slot 4, and in slot 20 even though the augment is not for those slots.  Likely need to do some checks to ensure we don't end up with erroneous items.  The resulting item is useable until the augment is removed.

I would suggest some more testing.

zone/gm_commands/parcels.cpp
```diff
@@ -4,304 +4,488 @@
...
+			SendParcelsSubCommands(c);
 			return;
 		}
 
-		auto player_id = CharacterParcelsRepository::GetParcelCountAndCharacterName(database, player_name);
-		if (!player_id.at(0).char_id) {
-			c->MessageString(Chat::Yellow, CANT_FIND_PLAYER, player_name.c_str());
-			return;
-		}
+		const std::string& to_name = sep->arg[2];
```
neckkola, line 30: Is the & required here?

zone/gm_commands/parcels.cpp
```diff
@@ -4,304 +4,488 @@
...
-			fmt::format("char_id = '{}' ORDER BY slot_id ASC", player_id.at(0).char_id)
-		);
+		std::vector<uint32> augment_ids = { 0, 0, 0, 0, 0, 0 };
 
-		if (results.empty()) {
-			c->Message(Chat::White, fmt::format("No parcels could be found for {}", player_name).c_str());
-			return;
-		}
+		if (!sep->IsNumber(3)) {
+			const std::string& cmd_msg = sep->msg;
```
neckkola, line 37: Same thing here, though I am not positive.  You likely know better than I.

zone/gm_commands/parcels.cpp
```diff
@@ -4,304 +4,488 @@
...
+			fmt::format(
+				" with a note of '{}'",
+				note
+			) :
+			""
+		);
+
 		if (item_id == PARCEL_MONEY_ITEM_ID) {
-			if (quantity > INT32_MAX) {
+			if (quantity > std::numeric_limits<int>::max()) {
```
neckkola, line 114: Kingly, is there an advantage of using std::numeric_limit over the define in limits.h of INT32_MAX?  Curious if there is an advantage from one over the other.  Perhaps the compiler replaces them both anyways?

zone/gm_commands/parcels.cpp
```diff
@@ -4,304 +4,488 @@
...
+					).c_str()
 				);
-				quantity = INT32_MAX;
+				quantity = std::numeric_limits<int>::max();
 			}
 
 			auto item = database.GetItem(PARCEL_MONEY_ITEM_ID);
 			if (!item) {
-				c->Message(Chat::Yellow, "Could not find item with id {}", item_id);
+				c->Message(Chat::White, "Item ID {} does not exist.", item_id);
```
neckkola, line 128: Are there standards as to which color is for what?  Yellow vs white?  And the syntax of the messages?

zone/gm_commands/parcels.cpp
```diff
@@ -4,304 +4,488 @@
...
+			auto item = database.CreateItem(
+				item_id,
+				quantity,
+				augment_ids[0],
+				augment_ids[1],
+				augment_ids[2],
+				augment_ids[3],
+				augment_ids[4],
+				augment_ids[5]
+			);
```
neckkola, line 200: With the change to create item, item will need to be delete at all the right places or it will cause a leak I believe.  I would suggest using a unique_ptr so that you don't need to worry about the deletes.

zone/gm_commands/parcels.cpp
```diff
@@ -4,304 +4,488 @@
...
+				"Found {} parcel{} for {}.",
+				results.size(),
+				results.size() != 1 ? "s" : "",
+				character_name
+			).c_str()
+		);
+
+		for (auto const& p: results) {
+			const auto* item = database.CreateItem(
+				p.item_id,
```
neckkola, line 340: Same delete/unique_ptr concern

zone/gm_commands/parcels.cpp
```diff
@@ -4,304 +4,488 @@
...
+				"Found {} parcel{} for {}.",
+				parcels.size(),
+				parcels.size() != 1 ? "s" : "",
+				character_name
+			).c_str()
+		);
+
+		for (auto const& p : parcels) {
+			const auto* item = database.CreateItem(
+				p.second.item_id,
```
neckkola, line 420: Delete/ptr concern

neckkola (commented):
@Kinglykrab Provided a few comments for your consideration.  I did not highlight all the string& items as I am not sure it makes a difference.  Perhaps someone else can comment on those.  Thanks for the chance to review.  I can re-test after you have a chance to consider the comments, just let me know.

Akkadius:
Still has pending feedback

PR #4352, opened 2024-05-27
[Fix] Correct issue with bazaar purchase via parcels

neckkola:
# Description

Correct an issue with bazaar purchases via parcel delivery.

If parcel delivery was selected for a bazaar purchase, under certain circumstances the item quantity was incorrectly updating in the database, resulting in items with 0 quantity.  This would create erroneous results in the bazaar search window. 

Fixes # (issue)
This fixes a report by a Server Operator.

## Type of change

Clients tested: 
RoF2

# Checklist


fryguy503 (approved):
Thank you for a quick turn-around on this!

PR #4361, opened 2024-05-29
[Cleanup] Cleanup Class Code

Kinglykrab:
# Description
- Introduces an `EQ::classes` namespace for stuff like `IsPlayer` so that we're not confusing this for a mob method or something of the like.
- Cleans up a lot of the logic in finding a class, checking for a valid class, getting bitmask, etc.

## Type of Change
- [X] New feature

# Testing
- Logged in game, ran around, didn't notice any issues, can post screenshots or videos as necessary if there is specific functionality we need tested.

# Checklist

### DIFF
--- a/common/shareddb.cpp
+++ b/common/shareddb.cpp
@@ -1773,7 +1773,7 @@ void SharedDatabase::LoadSpells(void *data, int max_spells) {
 		sp[tempid].environment_type=Strings::ToInt(row[102]);
 		sp[tempid].time_of_day=Strings::ToInt(row[103]);
 
-		for(y=0; y < Class::PLAYER_CLASS_COUNT;y++)
+		for(y=0; y < Class::Berserker;y++)
 			sp[tempid].classes[y]=Strings::ToInt(row[104+y]);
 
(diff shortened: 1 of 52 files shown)


common/shareddb.cpp
```diff
@@ -1773,7 +1773,7 @@ void SharedDatabase::LoadSpells(void *data, int max_spells) {
 		sp[tempid].environment_type=Strings::ToInt(row[102]);
 		sp[tempid].time_of_day=Strings::ToInt(row[103]);
 
-		for(y=0; y < Class::PLAYER_CLASS_COUNT;y++)
+		for(y=0; y < Class::Berserker;y++)
```
Akkadius, line 1776: Still would like to reference a `Class:Max or Class::PlayerClassCount` that points to Berserker instead of pointing to a class itself

Akkadius (changes requested):
Few minor comments

Kinglykrab:
Going to hold on this until other stuff is merged so I'm not fighting merge conflicts over and over.

PR #4362, opened 2024-05-29
[Cleanup] Cleanup Race Code

Kinglykrab:
# Description
- Introduces an `EQ::races` namespace for stuff like `IsPlayerRace` so that we're not confusing this for a mob method or something of the like.
- Cleans up a lot of the logic in finding a race, checking for a valid race, getting bitmask, etc.
- Removes the last of the straggler race constants not defined in the `Race` namespace.

## Type of Change
- [X] New feature

# Testing
- Logged in game, ran around, didn't notice any issues, can post screenshots or videos as necessary if there is specific functionality we need tested.

# Checklist


common/races.h
```diff
@@ -885,7 +772,534 @@ namespace Race {
...
+	{ Race::Human,                 { 7.0f,   7.0f }},
+	{ Race::Barbarian,             { 5.0f,   5.0f }},
+	{ Race::Erudite,               { 5.0f,   5.0f }},
+	{ Race::WoodElf,               { 5.5f,   5.5f }},
+	{ Race::HighElf,               { 4.0f,   4.0f }},
+	{ Race::DarkElf,               { 8.0f,   8.0f }},
+	{ Race::HalfElf,               { 9.0f,   9.0f }},
+	{ Race::Dwarf,                 { 3.5f,   3.5f }},
+	{ Race::Troll,                 { 3.0f,   3.0f }},
+	{ Race::Ogre,                  { 2.0f,   2.0f }},
```
Akkadius, line 882: Can we show our work to show that this list is 100% accurate? This can create all kinds of headache if we get this wrong

Akkadius: I wrote a script to compare old / new. But without even using the script I see that the first set of entries is wrong

```
	static float male_height[] = {
	    6.0f,  6.0f,  7.0f,   6.0f,   5.0f,  6.0f,  5.0f,   5.5f,  4.0f,  8.0f,  9.0f,  3.5f,  3.0f,  6.0f,   6.0f,
```

```
	static float female_height[] = {
	    6.0f,  6.0f,  7.0f,   6.0f,   5.0f,  6.0f,  5.0f,   5.5f,  4.0f,  8.0f,  9.0f,  3.5f,  3.0f,  6.0f,   6.0f,
```

In the new array we are 7,5,5

Not a good sign

Akkadius (changes requested):
height arrays look entirely off

common/races.h
Kinglykrab: My script I used did not put a unique case for `6.0` since it is the default male and female size for races, so any race that had `6.0` is not specified in the new data and instead uses the default.

Kinglykrab: I can post the code and explain when at home.

Kinglykrab:
[This](https://gist.github.com/Kinglykrab/41a8d2a8a14960b74cb94d548730fd26) is the script I wrote.

Basically I took the initial data that we had stored and ripped it apart and put it in these two files.
[female_sizes_before.txt](https://github.com/user-attachments/files/15524538/female_sizes_before.txt)
[male_sizes_before.txt](https://github.com/user-attachments/files/15524539/male_sizes_before.txt)

After this, I ran the script and got this.
[sizes_after.txt](https://github.com/user-attachments/files/15524541/sizes_after.txt)

All of this is consistent with the pre-existing values we have and shows why I seem to have less data due to my removal of any race that has a `6.0` default for both male and female from the values that we define.

If a male **or** female gender of a race is a non `6.0` value we store that, even if the opposite gender's size is `6.0`, this allows us to still keep these races as their default, but allow the male or female counterparts to have their separate size as well.

These are the lines that skip `6.0`.
```pl
if ($current_line =~ /^6.0$/ig) {
	next;
}
```

Inside the loop I then do this.
```pl
my $female_size_modifier = exists $size_data{"female"}{$race_id} ? $size_data{"female"}{$race_id} : "6.0f";
my $male_size_modifier = exists $size_data{"male"}{$race_id} ? $size_data{"male"}{$race_id} : "6.0f";
```

This defines a default of `6.0f` if the male or female counterpart exists as a non `6.0f` value and we have not already defined our current gender in `%size_data`.

- Then we have `Gender::GetRaceDefaultHeight` which returns the male height for male or neuter gender, and the female height for females.
- If we do not have a male or female height specified, meaning both male and female are `6.0f` height, we just return `6.0f`.
```cpp
float Gender::GetRaceDefaultHeight(uint16 race_id, uint8 gender_id)
{
	const auto& e = race_sizes.find(race_id);
	if (e != race_sizes.end()) {
		const uint8 gender_index = gender_id != Gender::Female ? 0 : 1;
		return e->second[gender_index];
	}

	return 6.0f;
}
```

Akkadius:
This still doesn't explain why for the very first race, the values are off, same for the 2nd and the 3rd. If we're only storing races in the new map that are non-6.0.

Assume we never use the first entry (race id 0) 

Human is 6 (yet we're storing 7)
Barbarian is 7 (new array is 5)
Erudite is 6 (new array is 5)

```cpp
static float male_height[] = {
	    6.0f,  6.0f,  7.0f,   6.0f, 
```

```cpp
	static float female_height[] = {
	    6.0f,  6.0f,  7.0f,   6.0f,
```

```cpp
static const std::map<uint16, std::vector<float>> race_sizes = {
	{ Race::Human,                 { 7.0f,   7.0f }},
	{ Race::Barbarian,             { 5.0f,   5.0f }},
	{ Race::Erudite,               { 5.0f,   5.0f }},
```

PR #4369, opened 2024-05-30
[Quests] Fix Lua encounter double register

Akkadius:
# Description

This fixes a scenario where Lua encounters get registered / fired twice during zone initialization. There's no need to call `parse->Init()` when `parse->ReloadQuests()` does largely the same functionality.

## Type of change

# Testing

Awaiting @fryguy503 to do hands on testing

# Checklist


Kinglykrab (approved):

fryguy503 (approved):
Solved issues so far. Approving and Merging

PR #4375, opened 2024-06-01
[Cleanup] Cleanup Object Type Code

Kinglykrab:
# Description
- Introduces a `ObjectType` namespace for object types instead of an enum like we had before.
- Cleans up a lot of the logic in finding an object type, checking for a valid object type, etc.
- Adds `#find object_type` subcommand.

## Type of Change
- [X] New feature

# Testing
image

# Checklist


common/emu_constants.cpp
```diff
@@ -736,3 +658,13 @@ std::string EQ::constants::GetConsiderColorName(uint32 consider_color)
...
+
+std::string ObjectType::GetName(uint32 object_type)
+{
+	return IsValid(object_type) ? object_types[object_type] : "UNKNOWN OBJECT TYPE";
+}
+
+bool ObjectType::IsValid(uint32 object_type)
+{
+	return object_types.find(object_type) != object_types.end();
+}
```
Akkadius, line 670: This can go to object.cpp/.h

Akkadius (approved):

PR #4377, opened 2024-06-01
[Skills] Fix caps out of bounds issue

Akkadius:
# Description

This fixes out of bounds data issues with skill caps where if an NPC or a Clients level is outside of the max defined skill levels defined in the database, we return 0 values as reported by @fryguy503 

This restores logic we had previously https://github.com/EQEmu/Server/blob/e06c7d7735184d6e1807f7aa12aba1b71634de4f/common/shareddb.cpp#L1837-L1865

This also adds logic to cache skill cap max values per class / skill id pair during load.

To get max value, we anchor to max that is available in the database. We always honor what's defined in rules as max.

In the situation Trust had, his skill caps only went to level 70 and his rule max was 75. NPC's were returning 0 above level 70.

## Type of change

# Testing

Awaiting test confirmation from @fryguy503 

Skills show proper values when beyond defined level of 100, they anchor to 100

image

# Checklist

### DIFF
--- a/client_files/export/main.cpp
+++ b/client_files/export/main.cpp
@@ -32,6 +32,7 @@
 #include "../../common/repositories/skill_caps_repository.h"
 #include "../../common/file.h"
 #include "../../common/events/player_event_logs.h"
+#include "../../common/skill_caps.h"
 
 EQEmuLogSys         LogSys;
 WorldContentService content_service;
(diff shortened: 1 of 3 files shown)


fryguy503:
On a server with a level 70 max skill cap and no skill cap entries in the database beyond level 70, we ended up with mobs that had significatly lower defense than anticipated.

Level 75 mob:

image


image


Level 70 Mob:

image

image


After fix (Initial Testing) - Level 75 mob:

image

image

Kinglykrab (approved):
Looks good to me.
image

PR #4387, opened 2024-06-05
[Bug Fix] Fix potential crash when updating spawn events

joligario:
# Description

As seen [here](https://spire.akkadius.com/dev/release/22.52.0?id=24562), we need to check that the event actually exists before attempting to update it.

## Type of change

# Testing

Clients tested: RoF2

# Checklist


Kinglykrab (approved):

Akkadius:
This crash looks like the database pointer is actually null

Akkadius:
Safe to merge regardless, not sure it fixes mentioned crash

Akkadius (approved):

PR #4396, opened 2024-06-15
Revert "[Bug Fix] Fix potential trader crash when serialized item not found"

joligario:
Reverts EQEmu/Server#4386


joligario:
Not the problem. Neckkola is looking into the actual problem.

PR #4397, opened 2024-06-16
[Bug Fix] Fix trader mode

joligario:
# Description

With latest changes, we are unable to start in trader mode because it can't find valid items to trade due to serialized items. We were iterating through slots based on the first serial number instead of the number of allowed bazaar slots. We also weren't initializing the item struct causing the search function to falsely identify items and serial numbers. Finally, forced a trader table truncate on startup to ensure an empty table in the event of a server crash or unclean shutdown which could leave invalid data in the table.

## Type of change

# Testing
Starting trader mode:
image

Stopping/starting trader mode, removing item from trade, and changing prices:
image

Clients tested: RoF2

# Checklist

### DIFF
--- a/common/database.cpp
+++ b/common/database.cpp
@@ -77,6 +77,7 @@
 #include "zone_store.h"
 #include "repositories/merchantlist_temp_repository.h"
 #include "repositories/bot_data_repository.h"
+#include "repositories/trader_repository.h"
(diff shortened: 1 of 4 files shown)


neckkola (approved):
Tested and looks good.

PR #4402, opened 2024-07-05
[Bug Fix] Fix for random disconnects when a large number of guild members zone or disconnect

neckkola:
# Description

Reports from multiple servers indicate that when a large number of guild members zone, or /ex to disconnect, some guild members will go linkdead and disconnect.  Reports from both PEQ and EZ

It appears that the zone process is becoming overwhelmed with packet processing regarding sending guild member status (zone, time, etc).  When an impacted client de-guilds, the problem is no longer experienced.

# Fix

The destructor for Client was sending a guild member status packet with a zone_id = 0 to indicate that the client was offline.  This packet would then be sent to all zone servers to notify guild players.  This was originally established to notify clients when a player logged.  However, if the client was simply zoning, the packet was still sent, even though the CompleteConnect function would immediately send another packet to indicate the new zone of the player.  

Therefore, the destructor packet was modified to only be sent when the player is not zoning.  This reduces the packet load by 50%, a significant reduction for large guilds.  

I also confirmed that the db update for online status is occurring as designed and should not be an issue. (first logon, and a disconnect).

I believe that this will resolve the random nature of this issue.

## Type of change

# Testing
I monitored packets being sent when a guild member is logged in, and another guild member zones.  When zoning, player A would receive 2 packets, one with a zone id of 0, another with the proper zone id.  With the fix, player A receives a single packet with the updated zone id.

I logged clients into the arena, bazaar and crescent reach and monitored packets and the guild window to ensure it continues to function as expected.

Clients tested: 
RoF2

# Checklist

### DIFF
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -215,6 +215,7 @@ Client::Client(EQStreamInterface *ieqs) : Mob(
 	guild_id = GUILD_NONE;
 	guildrank = 0;
 	guild_tribute_opt_in = 0;
+	SetGuildListDirty(false);
 	GuildBanker = false;
 	memset(lskey, 0, sizeof(lskey));
 	strcpy(account_name, "");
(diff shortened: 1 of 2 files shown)


zone/client.cpp
```diff
@@ -410,8 +411,12 @@ Client::~Client() {
 		zone->ClearEXPModifier(this);
 	}
 
-	if(IsInAGuild())
-		guild_mgr.SendGuildMemberUpdateToWorld(GetName(), GuildID(), 0, time(nullptr));
+	if (!IsZoning()) {
```
Kinglykrab, line 414: Any reason to not do `if (!IsZoning() && IsInAGuild())` to consolidate the conditions?

neckkola: I was just thinking that there will be more items that could be moved within based on the !zoning criteria, though at the moment no other reason really.

Kinglykrab (approved):

Akkadius (approved):

PR #4407, opened 2024-07-14
[Improvement] Flee Overhaul

fryguy503:
# Description

There was a number of issues with the current flee code that caused some heartburn. I merged the TAKP logic with our rule and extended flee checking logic to create this.

- Mobs will properly check Race and Faction of nearby mobs
- Mobs will check flee status on HP change (damage or heal/regen) to determine if they should continue to flee (Issue on current code).
- Mobs will appropriately slowdown/speedup when their HP drops/increases.
- Mob pathing is greatly increased to not push them outside of mesh/player accessible areas.

Adds #fleeinfo for mob flee status and potential

Credits go to https://github.com/eqmacemu/server for a very large portion of this code.

## Type of change

# Checklist


zone/aggro.cpp
```diff
@@ -564,6 +564,55 @@ bool Mob::CheckWillAggro(Mob *mob) {
...
+		// Square it because we will be using DistNoRoot
+		AggroRange *= AggroRange;
+
+		if (DistanceSquared(mob->GetPosition(), exclude->GetPosition()) > AggroRange) {
+			continue;
+		}
+
+		// If exclude doesn't have a faction, check for buddies based on race. Also exclude common factions such as noob monsters, indifferent, kos, kos animal
+		if(exclude->GetPrimaryFaction() != 0 &&
+			exclude->GetPrimaryFaction() != 5023 && exclude->GetPrimaryFaction() != 5032 && exclude->GetPrimaryFaction() != 5013 && exclude->GetPrimaryFaction() != 5014) {
```
Kinglykrab, line 598: Should we be _really_ be hard coding these?

fryguy503: Nope, but I am not sure how else to reference these specially as different servers could have different ids.

fryguy503: Make them rules? e.g.
RuleI(Aggro, IndifferentFactionID)?

Kinglykrab: We can use a string-based rule so we can have it pipe-separated like `1|2|3` or something it they want multiple factions.

fryguy503:
This needs full testing before being merged. 1 other server reported issues with fleeing causing mobs to pop underground, but I am not sure how much that was existing issues with mesh and z axis and how much may have been caused by this PR.

fryguy503:
Recent adjustment helps with mobs following furthest Z mesh when multiple Z planes of near Z mesh are on top of each other.

zone/attack.cpp
```diff
@@ -2186,6 +2186,19 @@ bool Client::Death(Mob* killer_mob, int64 damage, uint16 spell, EQ::skills::Skil
 	return true;
 }
 
+bool Client::CheckDeath()
```
KimLS, line 2189: This has a bit of a confusing name for what it's actually doing.  Maybe something like CheckIfAlreadyDead instead of CheckDeath.

KimLS: CheckDeath kind of implies you're checking if the client is dead instead of checking if they have triggered the dying logic before.

zone/fearpath.cpp
```diff
@@ -29,17 +29,52 @@ extern Zone* zone;
...
+	if (other != nullptr) {
+		hate_top = other;
+	}
+
+	if (!hate_top) {
+		return 0;
+	}
+
+	// If no special flee_percent check for Gray or Other con rates
+	if (GetLevelCon(hate_top->GetLevel(), GetLevel()) == ConsiderColor::Gray && flee_ratio == 0 && RuleB(Combat, FleeGray) &&
```
KimLS, line 46: Since these both require flee_ratio to be 0 before triggering I think it makes sense to enclose them in a if(flee_ratio) block.

As it is right now, you're always checking the GetLevelCon stuff even if the flee ratio != 0.

KimLS (dismissed):
Overall, I don't see a ton wrong with this.  I marked a few spots that potentially have minor changes to make but even if we didn't make them it would be fine.

catapultam-habeo:
This has been working much better since latest changes

joligario (approved):

Akkadius (approved):

PR #4412, opened 2024-07-15
[Bug Fix] Searching in the bazaar was not honoring the minimum cost

neckkola:
# Description

If a player searched in the bazaar the returned results would not honor the minimum cost field.  This was due to the client using platinum only and my code was not adjusting the min cost correctly.

## Type of change

# Testing
image

Clients tested: 
RoF2

# Checklist


fryguy503 (approved):
lgtm - Thank you!

PR #4414, opened 2024-07-15
[Bug Fix] Personal Tributes for bard items not being applied correctly

neckkola:
# Description

It was reported that the person tribute 'Harmony of Drums/Winds/etc' were not applying their modifications correctly.  I was able to reproduce.  This fix should resolve the issue.

## Type of change
- [X] Bug fix

# Testing
https://www.youtube.com/watch?v=qscETMzr7Hs

Clients tested: 
RoF2

# Checklist

- [X] I have tested my changes
- [X] I have performed a self-review of my code. Ensuring variables, functions and methods are named in a human-readable way, comments are added only where naming of variables, functions and methods can't give enough context.
- [X] I own the changes of my code and take responsibility for the potential issues that occur

PR #4415, opened 2024-07-16
Revert "[Bug Fix] Personal Tributes for bard items not being applied correctly"

neckkola:
Reverts EQEmu/Server#4414

I discovered an issue with Bots.  Reverting until I have a chance to correct it.


fryguy503 (approved):
Pushing approval

PR #4419, opened 2024-07-20
[Cleanup] Mask GM Show Buff message behind EntityVariable

nytmyr:
# Description

Removes the spam of "Your GM flag allows you to always see your targets' buffs." for GMs every time a buff lands on a target.

It will now lock to an Entity Variable and only show once per zone.

# Testing

Zoned multiple times and allows multiple spells to landed on targeted NPCs and verified it only displays once per zone and resets each zone.

# Checklist

### DIFF
--- a/zone/entity.h
+++ b/zone/entity.h
@@ -60,6 +60,8 @@ class Bot;
 
 extern EntityList entity_list;
 
+constexpr std::string SEE_BUFFS_FLAG = "see_buffs_flag";
+
 class Entity
 {
 public:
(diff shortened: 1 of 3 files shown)


Kinglykrab (approved):

Kinglykrab:
One small change, let’s use a `constexpr std::string` for the variable name so we’re not hard-coding the string in both places in case we ever change it.

zone/entity.h
```diff
@@ -60,6 +60,8 @@ class Bot;
 
 extern EntityList entity_list;
 
+constexpr std::string SEE_BUFFS_FLAG = "see_buffs_flag";
```
Kinglykrab, line 63: Sorry, I misspoke, this should actually be `constexpr const char*`. I'd edit it myself, but you're on a different repository.

PR #4428, opened 2024-07-23
Akkadius/zone routing algorithm improvements

Akkadius:
# Description

This PR is necessary to fix a few reported extreme edge cases. It simplifies the routing logic quite a bit. This logic has otherwise been battle tested for a few months now with little to no issues.

* Using #zoneinstance to a public static instance and being able to not zone
* Rare chance that the wrong zone data gets matched with an instance route
* We no longer route players when they matched a zone but didn't match an instance, this is something that should be handled by the rest of the code and normally be returned as zone_id 0 so zoning can continue as normal. This also created the possibility that the wrong zone data could be used in the route.

The code now ensures that if the zone in question is an instance, that we match explicitly the instance_id when we are iterating through public static instances. Before, there was a rare chance that the wrong instance could be matched if multiples of the same were in the database.

```cpp
WorldContentService::FindZoneResult WorldContentService::FindZone(uint32 zone_id, uint32 instance_id)
{
	for (const auto &z: zone_store.GetZones()) {
		for (auto &i: m_zone_static_instances) {
			if (
				z.zoneidnumber == zone_id &&
				DoesZonePassContentFiltering(z) &&
				i.zone == zone_id &&
				i.version == z.version) {

				if (instance_id > 0 && i.id != instance_id) {
					continue;
				}

				LogInfo(
					"Routed player to public static instance [{}] of zone [{}] ({}) version [{}] long_name [{}] notes [{}]",
					i.id,
					z.short_name,
					z.zoneidnumber,
					z.version,
					z.long_name,
					i.notes
				);

				return WorldContentService::FindZoneResult{
					.zone_id = static_cast<uint32>(z.zoneidnumber),
					.instance = i,
					.zone = z
				};
			}
		}
	}

	return WorldContentService::FindZoneResult{.zone_id = 0};
}
```

## Type of change

# Testing

Tested zoning locally

# Checklist

### DIFF
--- a/common/content/world_content_service.cpp
+++ b/common/content/world_content_service.cpp
@@ -6,6 +6,7 @@
 #include "../rulesys.h"
 #include "../eqemu_logsys.h"
 #include "../repositories/instance_list_repository.h"
+#include "../zone_store.h"
 
 
 WorldContentService::WorldContentService()
@@ -183,8 +184,8 @@ void WorldContentService::ReloadContentFlags()
 	}
 
 	SetContentFlags(set_content_flags);
-	LoadZones();
 	LoadStaticGlobalZoneInstances();
+	zone_store.LoadZones(*m_content_database);
 }
 
 Database *WorldContentService::GetDatabase() const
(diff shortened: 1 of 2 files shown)


Akkadius:
@fryguy503 will have this sitting on his server soon too to tesrt

neckkola (approved):

fryguy503:
Coupled with PR #4429  This has been working well.

PR #4432, opened 2024-07-29
[Bug Fix] Stop DOSing ourselves with OP_WearChange

catapultam-habeo:
# Description

This modifies `Handle_OP_WearChange` to skip sending the packet back to the sender. The client appears to echo this packet back to us under this circumstance, which generates a packet storm which significantly impacts client performance. This manifests as increased rate of random disconnects under many circumstances.

I'm opening this as a draft because there may be a better way to handle this; I'm not certain that we should be spraying this to all clients in the Handler method at all, but I don't understand either the purpose of this packet or the intended architecture here well enough to make that call, and would like the input from more experienced contributors.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Testing

Prior to implementing this change, it could be observed that any populated zone would be absolutely inundated with OP_WearChange C->S AND S->C traffic (The entire client chat buffer would fill with either filter of this traffic, occuluding ANY other output). After this change, that amount of traffic is much more reasonable.

Clients tested: 
ROF2

# Checklist

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -16035,7 +16035,8 @@ void Client::Handle_OP_WearChange(const EQApplicationPacket *app)
 		wc->hero_forge_model = GetHerosForgeModel(wc->wear_slot_id);
 
 	// we could maybe ignore this and just send our own from moveitem
-	entity_list.QueueClients(this, app, false);
+	// We probably need to skip this entirely when it is send as an ack, but not sure how to ID that.
+	entity_list.QueueClients(this, app, true);
 }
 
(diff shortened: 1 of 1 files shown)


catapultam-habeo:
This PR currently functionally breaks Ornaments and HF. Might be worth including in next release anyway, as imho constant LDs are worse than Ornaments being broken.

I am still working on fixing this properly, though. Simply reverting the originating PR doesn't seem correct.

fryguy503 (approved):
So far no adverse effects outside of the noted ornament regression, however the packet count has dropped dramatically.

PR #4433, opened 2024-07-30
[Bug Fix] Fix issue with quest::echo and quest::me

Kinglykrab:
# Description
- Fixes an issue where we were using `initiator` instead of `owner` for these events, causing edge case crashes when used in an event where we do not have an `initiator` such as in an NPC `EVENT_TIMER`.

## Type of change
- [X] Bug fix

# Checklist


fryguy503 (approved):
lgtm thank you KK

PR #4438, opened 2024-08-02
[Feature] Add Character:DefaultGuildRank Rule

Kinglykrab:
# Description
- Adds `Character:DefaultGuildRank` rule that sets members of the `Character:DefaultGuild` to the rank specified, `0` is the default for the lowest possibly member rank.

## Type of change
- [X] New feature

# Checklist

### DIFF
--- a/common/database.cpp
+++ b/common/database.cpp
@@ -66,6 +66,7 @@
 #endif
 
 #include "database.h"
+#include "data_verification.h"
 #include "eq_packet_structs.h"
 #include "extprofile.h"
 #include "strings.h"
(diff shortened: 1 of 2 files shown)


neckkola:
Good morning

We should set the default to 8 so that it loads as a recruit instead of 0 (leader)

May also want a bounds check to ensure it is a valid number 1 to 8. 

Thanks for doing this Kingly. I have been migrating dev pcs.

joligario (approved):

joligario (approved):

PR #4441, opened 2024-08-04
[Feature] Extend Spell Buckets Functionality

Kinglykrab:
# Description
- Extends spell bucket functionality to allow operators to use bucket comparison similar to merchant buckets.

## Type of change
- [X] New feature

# Testing
image
`1512` is `Scent of Shadow` and is set with a bucket of `Test` and comparison `1` which is `Not Equal` with a value of `1`.
`440` is `Animate Dead` and is set with a bucket of `Test` and comparison `0` which is `Equal` with a value of `1`.
In this screenshot you can see `Scent of Shadow` being scribed because I do not have a bucket `Test` with a value of `1`, but we are not scribing `Animate Dead` because I do not have a bucket `Test` with a value of `1`.
image

## Before
image
## Update
image
## After
image

# Checklist

### DIFF
--- a/common/database/database_update_manifest.cpp
+++ b/common/database/database_update_manifest.cpp
@@ -5718,6 +5718,22 @@ COLLATE='latin1_swedish_ci'
 ENGINE=InnoDB
 AUTO_INCREMENT=1;
 )"
+	},
+	ManifestEntry{
+		.version = 9282,
+		.description = "2024_08_02_spell_buckets_comparison.sql",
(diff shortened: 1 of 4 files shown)


Akkadius (approved):
LGTM. Make sure the manifest is tested with a database pre-migration

Kinglykrab:
> LGTM. Make sure the manifest is tested with a database pre-migration

Will do this once I’m off of work.

PR #4445, opened 2024-08-09
[Bug Fix] Fix AddCrystals() in Perl/Lua

Kinglykrab:
# Description
- Fixes an issue where when using `AddCrystals`, if you were to provide a greater than `0` amount of Ebon Crystals the method would return early regardless of the supplied Radiant Crystals amount.

## Type of change
- [X] Bug fix

# Checklist


fryguy503 (approved):
lgtm

PR #4456, opened 2024-08-20
[Quest API] Add AreTasksCompleted() to Perl/Lua.

Kinglykrab:
# Description
- Adds a method to allow operators to check if multiple tasks are complete without stringing together multiple `IsTaskCompleted` calls.

# Perl
- Add `$client->AreTasksCompleted(task_ids)`.
- Add `quest::aretaskscompleted(task_ids)`.

## Perl Example
```pl
sub EVENT_SAY {
	if ($text=~/#a/i) {
		my @task_ids = (5166, 5745);
		my $are_completed = quest::aretaskscompleted(@task_ids);
		quest::message(315, "Perl: $are_completed");
	}
}
```

## Perl Screenshot
image

# Lua
- Add `client:AreTasksCompleted(task_ids)`.
- Add `eq.are_tasks_completed(task_ids)`.

## Lua Example
```lua
function event_say(e)
	if e.message:findi("#a") then
		local task_ids = { 5166, 5745 }
		local are_completed = eq.are_tasks_completed(task_ids);
		eq.message(315, "Lua: " .. tostring(are_completed))
	end
end
```

## Lua Screenshot
image

### DIFF
--- a/zone/task_client_state.cpp
+++ b/zone/task_client_state.cpp
@@ -1599,6 +1599,23 @@ int ClientTaskState::IsTaskCompleted(int task_id)
 	return 0;
 }
 
+int ClientTaskState::AreTasksCompleted(const std::vector<int>& task_ids)
+{
+	if (!(RuleB(TaskSystem, RecordCompletedTasks))) {
+		return -1;
+	}
+
+	int tasks_completed = 1;
+
+	for (const auto& task_id : task_ids) {
+		if (!IsTaskCompleted(task_id)) {
+			tasks_completed = 0;
+		}
+	}
+
+	return tasks_completed;
+}
+
 bool ClientTaskState::TaskOutOfTime(TaskType task_type, int index)
 {
 	// Returns true if the Task in the specified slot has a time limit that has been exceeded.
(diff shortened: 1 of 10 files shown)


zone/task_client_state.cpp
```diff
@@ -1599,6 +1599,23 @@ int ClientTaskState::IsTaskCompleted(int task_id)
 	return 0;
 }
 
+int ClientTaskState::AreTasksCompleted(const std::vector<int>& task_ids)
+{
+	if (!(RuleB(TaskSystem, RecordCompletedTasks))) {
+		return -1;
+	}
+
+	int tasks_completed = 1;
```
Akkadius, line 1608: Strange we didn’t just make this a bool along with the return type

Akkadius (approved):
Minor comment

zone/task_client_state.cpp
Kinglykrab: Yeah, from what I see they used `-1` for disabled, `0` for fail, `1` for success.

Kinglykrab: I can change this and the other method to a bool if we’d prefer. I doubt anyone even knows the `-1` is a thing or has ever used it in a script because why would they check for that value with the task methods if they have task system disabled?

Kinglykrab: Changed.

PR #4459, opened 2024-08-26
[Feature] Add Rule to Limit Task Update Messages

Kinglykrab:
# Description
- Adds a rule `TaskSystem:MaxUpdateMessages` that allows operators to limit sending a message for every `1` count.
- Issue was that if you have a task that requires 500/1,000/10,000/50,000 or whatever large number if you use `#task update task_id activity_id count` and the `count` is a large number it will send 500/1,000/10,000/50,000 messages, possibly causing your character to disconnect to character select.

## Type of change
- [X] New feature

# Checklist


Kinglykrab:
> Use a rule for the max number of quest messages instead. Then servers can just change it to 1 if they want while by default it won't lag the zone/client out.
> 
> A max default of 50 would probably be reasonable. The most update messages I've seen with an item stack turn in on live is 32 (with 38 total messages from the trade via filling other slots) so I don't know if live has any upper limit for throttling. They probably don't have any quests where it would cause an issue.
> 
> This will also make it more readable over a multiline conditional operator/ternary
> 
> ```c++
>     int msg_count = 1;
>     if (activity_type != TaskActivityType::GiveCash)
>     {
>       msg_count = std::min(count, RuleI(TaskSystem, MaxUpdateMessages));
>     }
> ```

No problem, I'll change this now. 👍

hgtw (approved):

PR #4453, opened 2024-08-15
[Feature] Lazy Load Bank Contents

catapultam-habeo:
# Description

This provides a rule which causes the contents of the Bank to not load until the player is within proximity of a Banker. This has has been observed to dramatically improve zoning times and zoning stability for characters with well-populated banks while using custom code to expand bag sizes, but is likely to produce similar effects with normal bag sizes.

## Type of change

Please delete options that are not relevant.

- [X] New feature (non-breaking change which adds functionality)

# Testing

This has been in usage on Retribution for approximately two months.

Clients tested: 

# Checklist

- [X] I have tested my changes
- [X] I have performed a self-review of my code. Ensuring variables, functions and methods are named in a human-readable way, comments are added only where naming of variables, functions and methods can't give enough context.
- [X] I own the changes of my code and take responsibility for the potential issues that occur

PR #4462, opened 2024-08-27
[Code] Add IsCloseToBanker method

Akkadius:
# Description

This is a very small PR that adds a helper method in assistance to another PR https://github.com/EQEmu/Server/pull/4453

## Type of change

# Checklist

### DIFF
--- a/zone/mob.cpp
+++ b/zone/mob.cpp
@@ -8616,7 +8616,7 @@ bool Mob::IsCloseToBanker()
 {
 	for (auto &e: entity_list.GetCloseMobList(this)) {
 		auto mob = e.second;
-		if (mob->IsNPC() && mob->GetClass() == Class::Banker) {
+		if (mob && mob->IsNPC() && mob->GetClass() == Class::Banker) {
 			return true;
 		}
 	}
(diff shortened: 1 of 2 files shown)


Kinglykrab (approved):

Kinglykrab:
Are we going to change the code he submitted to use this? Or just add the method itself?

zone/mob.cpp
```diff
@@ -8611,3 +8611,15 @@ void Mob::SetExtraHaste(int haste, bool need_to_save)
 		}
 	}
 }
+
+bool Mob::IsCloseToBanker()
+{
+	for (auto &e: entity_list.GetCloseMobList(this)) {
+		auto mob = e.second;
+		if (mob->IsNPC() && mob->GetClass() == Class::Banker) {
```
Kinglykrab, line 8619: Should we validate `mob` here as well?

Akkadius: Shouldn't have to but it doesn't hurt

Akkadius:
> Are we going to change the code he submitted to use this? Or just add the method itself?

I'm just adding the method in this instance. Will follow up with the PR separately.

Kinglykrab:
> > Are we going to change the code he submitted to use this? Or just add the method itself?
> 
> I'm just adding the method in this instance. Will follow up with the PR separately.

10-4.

PR #4475, opened 2024-09-15
[Bug Fix] Fix 'Teleport Doors' from being blocked by GM flag

catapultam-habeo:
# Description

GM flag causes teleport doors to always act as if you do not have the key, instead of acting as if you do have the key

## Type of change

# Testing

This is a trivial fix

Clients tested: 

# Checklist

### DIFF
--- a/zone/doors.cpp
+++ b/zone/doors.cpp
@@ -542,7 +542,7 @@ void Doors::HandleClick(Client *sender, uint8 trigger)
 	if (EQ::ValueWithin(m_open_type, 57, 58) && HasDestinationZone()) {
 		bool has_key_required = (required_key_item && required_key_item == player_key);
 
-		if (sender->GetGM() && has_key_required) {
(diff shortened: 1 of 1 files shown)


zone/doors.cpp
```diff
@@ -543,7 +543,7 @@ void Doors::HandleClick(Client *sender, uint8 trigger)
 		bool has_key_required = (required_key_item && required_key_item == player_key);
 
 		if (sender->GetGM() && has_key_required) {
-			has_key_required = false;
+			has_key_required = true;
```
Kinglykrab, line 546: If `has_key_required` is already `true` based on the condition this is inside of, can’t we just remove this line altogether?

catapultam-habeo: Fixed it to predicate on `!has_key_required` so that the user is notified if their GM status changed the result of the interaction.

Kinglykrab (approved):

PR #4476, opened 2024-09-15
Login Fatal Error Spamming 

KimLS:
# Description

When connecting to Legacy Login Servers with bad login info you will get a fatal error back from the server.  It will then disconnect you triggering the reconnect code to trigger.  This changes the logic slightly to prevent this from happening.  If we get a fatal error from the server, we disconnect on the world side instead of waiting for login to do it and then do not attempt to reconnect.

Fixes # (issue)

## Type of change

# Testing

I've run it locally there aren't many cases to test.

Clients tested: 

This does not affect clients it is between world -> login.

# Checklist

### DIFF
--- a/world/login_server.cpp
+++ b/world/login_server.cpp
@@ -315,6 +315,13 @@ void LoginServer::ProcessLSFatalError(uint16_t opcode, EQ::Net::Packet &p)
 		error,
 		reason
 	);
+
+	if (m_legacy_client) {
+		m_legacy_client.release();
+	}
+	else if (m_client) {
+		m_client.release();
+	}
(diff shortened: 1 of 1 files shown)


KimLS:
There was some question of approach here, Akkadius wanted a back off timer but the more I thought about it the more it seemed like if we're getting fatal errors from login: do we even want to try reconnecting? My thought was no; if we still want to put the work into the exponential backoff instead we can but it's not simple.

Akkadius:
Looks good but I need to test this yet

Akkadius (approved):

PR #4477, opened 2024-09-16
[Bug Fix] Fix Issue with Database::ReserveName

Kinglykrab:
# Description
- If a bot or NPC is created prior to a character being created, they may end up with the same name.
- This causes issues when a bot is spawned because it will crash any player with that same name in the zone.
- NPCs with the same name will get the `000` identifier instead of the player, which could cause issues.

## Type of change
- [X] Bug fix

# Checklist


Akkadius (approved):

catapultam-habeo:
This PR fails to check content database in a multi-tenant deployment.

PR #4488, opened 2024-09-28
[Bug Fix] Fix Targeted AOE Max Targets Rule

Kinglykrab:
# Description
- Fixes an issue where we were using the rule improperly, causing spells that should not be limited to be limited instead of only limiting specific spell types.
- We were also using the rule as `Range:AOEMaxTargets` when it was `Spells:AOEMaxTargets`.
- I fixed this by removing the old rule since it does nothing and adding a new one with a default of `0` for unlimited.

## Type of change
- [X] Bug fix

# Testing
image

# Checklist

### DIFF
--- a/zone/effects.cpp
+++ b/zone/effects.cpp
@@ -1096,7 +1096,7 @@ void EntityList::AESpell(
 		max_targets = nullptr;
 	}
 
-	int max_targets_allowed = RuleI(Range, AOEMaxTargets); // unlimited
+	int max_targets_allowed = 4;
 	if (max_targets) { // rains pass this in since they need to preserve the count through waves
 		max_targets_allowed = *max_targets;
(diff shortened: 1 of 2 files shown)


zone/effects.cpp
```diff
@@ -1096,7 +1096,7 @@ void EntityList::AESpell(
 		max_targets = nullptr;
 	}
 
-	int max_targets_allowed = RuleI(Range, AOEMaxTargets); // unlimited
+	int max_targets_allowed = 4;
```
Akkadius, line 1099: Why is this fixed now ?

Kinglykrab: This is the default for Mesmerize and Lull spells. Whereas before we were doing this in reverse, which could have caused issues if the rule actually functioned.

Akkadius (approved):

PR #4504, opened 2024-10-10
[Bug] FindBestZ selecting false zone floor as bestz - Results in roambox failures

fryguy503:
# Description

Added underworld checks per the EQMac project. Thank you TAKP for being a wonderful source of resolutions to issues that you have tackled over the last 10 years.

## Type of change

# Testing

Prior:

Mobs on roamboxes would bunch up in dips/holes/lowspots in some olderzones.
image

In these cases, it was due to a below world requested_z from the false zone floor in the map file (In this example its -198 is the false floor)

image

Post:

Same location after 20 minutes, normally filling with "stuck" mobs:

image

snippet showing the requested_z no longer being the -198 consistently

image

# Checklist

### DIFF
--- a/zone/map.cpp
+++ b/zone/map.cpp
@@ -31,12 +31,14 @@ Map::~Map() {
 }
 
 float Map::FindBestZ(glm::vec3 &start, glm::vec3 *result) const {
-	if (!imp)
+	if (!imp) {
 		return BEST_Z_INVALID;
+	}
 
 	glm::vec3 tmp;
-	if(!result)
+	if (!result) {
 		result = &tmp;
+	}
 
 	start.z += RuleI(Map, FindBestZHeightAdjust);
 	glm::vec3 from(start.x, start.y, start.z);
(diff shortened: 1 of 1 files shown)


Kinglykrab (approved):

Kinglykrab:
Looks good to me, would like other eyes on this though before merging.

Akkadius:
I have concerns with this, but not enough to keep this from going through.

This relies on the `underworld` field in `zone` to be 100% accurate. 

We are bound to potentially see new bugs pop up where we don't find a Z because we're in a circumstance where NPC's are pathing underneath `underworld`.

Again not a criticism but just raising a concern.

Let's push it through and give it a whirl and observe. Thanks for putting it up

Akkadius (approved):

jdestefx:
Unfortunately this did not fix the issue for me.  I still have mobs bunching up in low-level areas.
image

PR #4520, opened 2024-10-22
[Bots] Fix pet buffs from saving duplicates every save

nytmyr:
# Description

Previously we were not checking the pet index properly when clearing buffs in the database before saving which resulted in no prior data being deleted.

This corrects the logic for the save and also will clean up any buffs for pets that don't exist in the table.

## Type of change

# Testing

Tested saving before and after to validate no dupes. Verified query only clears buffs for pets that don't exist in the database

Clients tested: 

RoF2

# Checklist

- [x] I have made corresponding changes to the documentation (if applicable, if not delete this line)


Akkadius (approved):
As a general rule of thumb, only things that change schema should be in the manifest.For the sake of time and cutting a release I'll do the suggested changes for you.

nytmyr:
> As a general rule of thumb, only things that change schema should be in the manifest.For the sake of time and cutting a release I'll do the suggested changes for you.

Appreciate it. I figured I would throw it in the manifest because it was a one off to catch all the dupes from the issue, going forward we won't need to purge the table because it deletes as needed on SavePet.

PR #4527, opened 2024-10-27
[Rule] Add Rule for restricting client versions to world server

knervous:
# Description

It's a common scenario for custom servers to implicitly restrict clients they support for various reasons (mainly supporting custom files/content which locks them into a certain version). Currently there's nothing preventing users from logging into that server from another client, resulting in a mismatch in expected behavior for what the server is trying to achieve. Since indications and functionality from a custom file perspective can only be presented or applied if the correct version and patch is installed for a custom server, it's a cart before the horse situation where the user won't know they're using the wrong client until there is undefined behavior.

This change adds a new rule option to restrict clients to a comma-separated list (default empty string, which is no restriction, i.e. no changes until someone opts in to use this). If the client's version doesn't fit into the list, they're presented with a dummy character on char select with no option to create another characters or log in but present the list of supported clients.

Fixes # (issue)

## Type of change

# Testing

Attach images and describe testing done to validate functionality.
Validated between two clients. I set the rule value to Titanium,Underfoot. Here is the result logging in with RoF2 and attempt to log in via button:

image

And with the same rule logging in with Titanium:

image

Changing the rule to RoF2 and logging in with Titanium:

image

Clients tested: 
RoF2
Titanium

# Checklist

- [ ] I have made corresponding changes to the documentation (if applicable, if not delete this line) -- Makes sense to update documentation after PRs are approved?

### DIFF
--- a/world/client.cpp
+++ b/world/client.cpp
@@ -526,9 +526,31 @@ bool Client::HandleSendLoginInfoPacket(const EQApplicationPacket *app)
 		SendEnterWorld(cle->name());
 		SendPostEnterWorld();
 		if (!is_player_zoning) {
-			SendExpansionInfo();
-			SendCharInfo();
-			database.LoginIP(cle->AccountID(), long2ip(GetIP()));
+			const auto supported_clients = RuleS(World, SupportedClients);
+			bool skip_char_info = false;
+			if (!supported_clients.empty()) {
+				auto name = EQ::versions::ClientVersionName(m_ClientVersion);
+				bool supported = false;
+				std::stringstream ss(supported_clients);
+				std::string item;
+				while (std::getline(ss, item, ',')) {
+					if (item == name) {
+						supported = true;
+						break;
+					}
+				}
+				if (!supported) {
+					std::string message = "Client Not In Supported List [" + supported_clients + "]";
+					SendUnsupportedClientPacket(message);
+					skip_char_info = true;
+				}
+			}
+
+			if (!skip_char_info) {
+				SendExpansionInfo();
+				SendCharInfo();
+				database.LoginIP(cle->AccountID(), long2ip(GetIP()));
+			}
 		}
 
 		cle->SetIP(GetIP());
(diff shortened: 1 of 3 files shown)


world/client.cpp
```diff
@@ -526,9 +526,31 @@ bool Client::HandleSendLoginInfoPacket(const EQApplicationPacket *app)
...
+						supported = true;
+						break;
+					}
+				}
+				if (!supported) {
+					std::string message = "Client Not In Supported List [" + supported_clients + "]";
+					SendUnsupportedClientPacket(message);
+					skip_char_info = true;
+				}
+			}
```
Kinglykrab, line 547: I believe we could use `Strings::Split` here and use `std::find` to simplify this a bit.

Something like this maybe?
```cpp
if (!supported_clients.empty()) {
	const std::string& name = EQ::versions::ClientVersionName(m_ClientVersion);

	const auto& clients = Strings::Split(supported_clients, ",");

	if (std::find(clients.begin(), clients.end(), name) == clients.end()) {
		std::string message = "Client Not In Supported List [" + supported_clients + "]";
		SendUnsupportedClientPacket(
			fmt::format(
				"Client Not In Supported List [{}]",
				supported_clients
			)
		);
		skip_char_info = true;
	}
}
```

world/client.cpp
```diff
@@ -2453,3 +2475,35 @@ void Client::SendGuildTributeOptInToggle(const GuildTributeMemberToggle *in)
...
+
+	unsigned char *buff_ptr = packet.pBuffer;
+	CharacterSelect_Struct *cs = (CharacterSelect_Struct *) buff_ptr;
+
+	cs->CharCount = character_count;
+	cs->TotalChars = character_limit;
+
+	buff_ptr += sizeof(CharacterSelect_Struct);
+
+	CharacterSelectEntry_Struct *entry = (CharacterSelectEntry_Struct *) buff_ptr;
```
Kinglykrab, line 2494: `auto e = (CharacterSelectEntry_Struct*) buff_ptr;`

world/client.cpp
```diff
@@ -2453,3 +2475,35 @@ void Client::SendGuildTributeOptInToggle(const GuildTributeMemberToggle *in)
...
+	strcpy(entry->Name, message.c_str());
+
+	entry->Race = HUMAN;
+	entry->Class = 1;
```
Kinglykrab, line 2498: `entry->Class = Class::Warrior`

world/client.cpp
```diff
@@ -2453,3 +2475,35 @@ void Client::SendGuildTributeOptInToggle(const GuildTributeMemberToggle *in)
...
+	entry->Level = 1;
+	entry->ShroudClass = entry->Class;
+	entry->ShroudRace = entry->Race;
+	entry->Zone = -1;
+	entry->Instance = 0;
+	entry->Gender = 0;
```
Kinglykrab, line 2504: `entry->Gender = Gender::Male;`

world/client.cpp
```diff
@@ -2453,3 +2475,35 @@ void Client::SendGuildTributeOptInToggle(const GuildTributeMemberToggle *in)
...
+
+	cs->CharCount = character_count;
+	cs->TotalChars = character_limit;
+
+	buff_ptr += sizeof(CharacterSelect_Struct);
+
+	CharacterSelectEntry_Struct *entry = (CharacterSelectEntry_Struct *) buff_ptr;
+	strcpy(entry->Name, message.c_str());
+
+	entry->Race = HUMAN;
```
Kinglykrab, line 2497: `entry->Race = Race::Human;`

Kinglykrab:
```cpp
void Client::SendUnsupportedClientPacket(const std::string& message)
{
	EQApplicationPacket packet(OP_SendCharInfo, sizeof(CharacterSelect_Struct) + sizeof(CharacterSelectEntry_Struct));

	unsigned char* buff_ptr = packet.pBuffer;
	auto cs = (CharacterSelect_Struct*) buff_ptr;

	cs->CharCount  = 1;
	cs->TotalChars = 1;

	buff_ptr += sizeof(CharacterSelect_Struct);

	auto e = (CharacterSelectEntry_Struct*) buff_ptr;

	strcpy(e->Name, message.c_str());

	e->Race        = Race::Human;
	e->Class       = Class::Warrior;
	e->Level       = 1;
	e->ShroudClass = e->Class;
	e->ShroudRace  = e->Race;
	e->Zone        = -1;
	e->Instance    = 0;
	e->Gender      = Gender::Male;
	e->GoHome      = 0;
	e->Tutorial    = 0;
	e->Enabled     = 0;

	QueuePacket(&packet);
}```

world/client.cpp
knervous: Yep looks good, updated

Akkadius (approved):
Thanks for putting this up and having the patience and understanding for us being slow to respond for feedback. I had some things going on and I know Kingly is in the middle of a move.

knervous:
> Thanks for putting this up and having the patience and understanding for us being slow to respond for feedback. I had some things going on and I know Kingly is in the middle of a move.

Hey this is my first PR into EQEmu after all this time, I'm patient! Happy to be able to contribute

Kinglykrab (approved):
🥇

PR #4543, opened 2024-11-15
[Rules] Add Rule to Disable NPCs Facing Target

Kinglykrab:
# Description
- Adds `NPC:ExcludedFaceTargetRaces` which is a comma-separated list of race IDs that should not turn to face target when hailed.

## Type of change
- [X] New feature

# Checklist


Akkadius:
I know this change looks fairly straightforward but would be good to attach demo / testing

Kinglykrab:
> I know this change looks fairly straightforward but would be good to attach demo / testing

Attached a video.

PR #4548, opened 2024-11-22
[Feature] Implement "Big Bags"

Kinglykrab:
# Description
- Implements bags up to `200` slots for ROF2 clients.

# Disclaimer
- Servers using older clients (pre-ROF2) will need to use the following query to limit all their bags to `10` bag slots so their servers still function properly.
```sql
UPDATE `items` SET `bagslots` = 10 WHERE `bagslots` > 10;
```

## Type of change
- [X] Breaking change

# Before and After Slot IDs
image

# Testing
image
image
image
image
image

Clients tested: ROF2

# Checklist
- [X] I have tested the changes on a local database and updated `version.h` `CURRENT_BINARY_DATABASE_VERSION` to the new version.

## Credits
- Thanks to @catapultam-habeo and @SecretsOTheP for sharing their bag implementations and allowing me to finalize a version for EQEmulator as a whole.


crotonrz:
Thousands of players who don't follow server code, just got a semi...and they don't yet know why.

MortimerGreenwald:
While testing got a query error see below:
Placing an item and removing an item in the Bank / Shared Bank Slot.
image
{90333CB6-0E4A-49AE-9345-32C1B8591171}

common/shareddb.cpp
```diff
@@ -190,242 +191,268 @@ SharedDatabase::MailKeys SharedDatabase::GetMailKey(int character_id)
...
+	if (!EQ::InventoryProfile::SupportsContainers(slot_id)) {
+		return true;
+	}
+
+	const int16 base_slot_id = EQ::InventoryProfile::CalcSlotId(slot_id, EQ::invbag::SLOT_BEGIN);
+
+	return SharedbankRepository::DeleteWhere(
+		*this,
+		fmt::format(
+			"`account_id` = {} AND `slotid` BETWEEN {} AND {}",
```
MortimerGreenwald, line 450: should this be 'slot_id'?

Kinglykrab: Yes. I’ll fix it when I get home.

MortimerGreenwald:
Looking good now.

neckkola:
I have also been doing some testing and wonder if the swapitem, or actually the UpdateInventorySlot should be reworked given the increase bag size.  It currently pushes out a db query for each slot in the bag, everytime a bag is selected.  For example, if you have a 100 slot bag, and simply click it to place it on the cursor, there will be 100 db queries.  100 more when you place it back in an inventory slot.  I wonder if we should batch that instead.  This is how master functions, even with a 10 bag slot.  I will continue to do some testing.  Well done Kingly.  Will be a welcome update.

Akkadius:
> I have also been doing some testing and wonder if the swapitem, or actually the UpdateInventorySlot should be reworked given the increase bag size. It currently pushes out a db query for each slot in the bag, everytime a bag is selected. For example, if you have a 100 slot bag, and simply click it to place it on the cursor, there will be 100 db queries. 100 more when you place it back in an inventory slot. I wonder if we should batch that instead. This is how master functions, even with a 10 bag slot. I will continue to do some testing. Well done Kingly. Will be a welcome update.

Every query interaction will definitely need to be done in bulk. This feature lends itself to performance issues by nature so we will need to make sure its solid out of the gate.

MortimerGreenwald:
getting this air when someone is looting. I think it's iterating too many slots.
E6048048-D19A-4995-8347-377C80324F23

MortimerGreenwald:
A couple of reports:
Screenshot_20250122-085848

Also:
When opening multiple large-capacity bags and attempting to left-click items to move them to the bank bag, the following issues occur:
Approximately half of the bag slots disappear
The X button to close bags becomes non-functional
Items become invisible when attempting to move them with left-click (items still exist but cannot be seen)
Note: The bags function normally when only a few are opened.

PR #4556, opened 2024-11-26
[Bug Fix] Allow Items in ROF2 to Stack to 32,767

Kinglykrab:
# Description
- ROF2 actually allows items to stack to 32,767 instead of just to 1,000

## Type of change
- [X] New feature

# Testing
image

# Checklist

### DIFF
--- a/common/patches/rof2.cpp
+++ b/common/patches/rof2.cpp
@@ -6358,9 +6358,19 @@ namespace RoF2
 		//sprintf(hdr.unknown000, "06e0002Y1W00");
 		strn0cpy(hdr.unknown000, fmt::format("{:016}\0", inst->GetSerialNumber()).c_str(),sizeof(hdr.unknown000));
 
-		hdr.stacksize =
-			item->ID == PARCEL_MONEY_ITEM_ID ? inst->GetPrice() : (inst->IsStackable() ? ((inst->GetCharges() > 1000)
-				? 0xFFFFFFFF : inst->GetCharges()) : 1);
+		hdr.stacksize = (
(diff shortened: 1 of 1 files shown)


crotonrz:
@Akkadius hit coder buttons and approve this! :)

fryguy503:
Curious, what happens to items when users switch between clients rof2->titanium

common/patches/rof2.cpp
```diff
@@ -6358,9 +6358,19 @@ namespace RoF2
 		//sprintf(hdr.unknown000, "06e0002Y1W00");
 		strn0cpy(hdr.unknown000, fmt::format("{:016}\0", inst->GetSerialNumber()).c_str(),sizeof(hdr.unknown000));
 
-		hdr.stacksize =
-			item->ID == PARCEL_MONEY_ITEM_ID ? inst->GetPrice() : (inst->IsStackable() ? ((inst->GetCharges() > 1000)
-				? 0xFFFFFFFF : inst->GetCharges()) : 1);
+		hdr.stacksize = (
```
Akkadius, line 6361: This is not the most readable doing nested ternaries like this and is very hard to follow.

Suggest using something like this 

```cpp
hdr.stacksize = 1;
if (item->ID == PARCEL_MONEY_ITEM_ID) {
    hdr.stacksize = inst->GetPrice();
} else if (inst->IsStackable()) {
    if (inst->GetCharges() > std::numeric_limits<int16>::max()) {
        hdr.stacksize = std::numeric_limits<uint32>::max();
    } else {
        hdr.stacksize = inst->GetCharges();
    }
}
```

Akkadius (approved):

PR #4560, opened 2024-11-28
[Feature] Enable bazaar window 'Find Trader' functionality

neckkola:
Enables the 'Find Trader' button on the Bazaar search window.  Tested against multi-instanced bazaars as well.  Also updated the 'local/all' trader functions.

Update: 2024-12-03
Decided to move the SendBulkTraders out of the zoning routines to reduce zone in traffic.  Instead moved to when the /bazaar window is first opened using a specific action type.  I had done this with the buyer functionality was added and it has worked well.  This will save (78 * num_traders)bytes from being sent on each zone, for each client.
 
Also updated the SendBulkTraders routine to be instance aware.  I had missed that previously.

## Type of change

# Testing
https://www.youtube.com/watch?v=gwMwxlvyzek

Clients tested: 
RoF2

# Checklist

### DIFF
--- a/common/bazaar.cpp
+++ b/common/bazaar.cpp
@@ -8,13 +8,14 @@ std::vector<BazaarSearchResultsFromDB_Struct>
 Bazaar::GetSearchResults(
 	SharedDatabase &db,
 	BazaarSearchCriteria_Struct search,
-	uint32 char_zone_id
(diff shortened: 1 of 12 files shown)


Akkadius (approved):
Awesome work! Love the testing video as well

PR #4561, opened 2024-11-28
[Rules] Add rules for requiring custom files from client

knervous:
# Description

These rules are for custom servers that require custom files, giving them an option to lock out players that don't have them installed while presenting a URL to download those files and get up to date. This prevents stale patch files and undefined behavior for clients logging in without the appropriate server files.

Here is an example of a custom client binary sending a custom payload:

https://github.com/knervous/eqnexus/blob/main/core/server_options.ixx

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Testing

Example of what client sees when logging into a custom server without the correct custom files.

image

Clients tested: RoF2, Titanium

# Checklist


fryguy503:
Could you add a status level bypass for the enforcement of the client restrictions.

knervous:
> Could you add a status level bypass for the enforcement of the client restrictions.

Added, figure anything below ApprenticeGuide makes sense to enforce?

fryguy503:
> > Could you add a status level bypass for the enforcement of the client restrictions.
> 
> Added, figure anything below ApprenticeGuide makes sense to enforce?

Just make it a rule and call it easy to adjust.

knervous:
> > > Could you add a status level bypass for the enforcement of the client restrictions.
> > 
> > 
> > Added, figure anything below ApprenticeGuide makes sense to enforce?
> 
> Just make it a rule and call it easy to adjust.

Added a rule here with default of apprentice guides not having this enforced

Akkadius (approved):

PR #4581, opened 2024-12-31
[Bug Fix] Client 'Hitch Bug' Fix due to large guild membership

neckkola:
When the guild membership was large (1k+) the client would studder for half a sec when a guild member would login or logout.  This was reproduceable with a guild size of 2k members though the floor would be client dependent most likely.  I did not work out the lower limit though some operators noticed it above 1k guild members.

## Type of change

# Testing
I tested with a guild membership of 2000, with Player A running in the Arena while Player B from the same guild logged in and logged out.  The result is that Player A would 'hitch' or 'pause' for a brief moment.  It was very noticeable.
Post the change, the hitch was no longer present.

This also been running several days within WFH without any side effects.

Clients tested: 
RoF2

# Checklist

### DIFF
--- a/zone/client_packet.cpp
+++ b/zone/client_packet.cpp
@@ -860,7 +860,7 @@ void Client::CompleteConnect()
 	if (IsInAGuild()) {
 		if (firstlogon == 1) {
 			guild_mgr.UpdateDbMemberOnline(CharacterID(), true);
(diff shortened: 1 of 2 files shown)


Akkadius (approved):
Awesome

PR #4585, opened 2025-01-05
[Fix] Repair a memory leak in GuildsList

neckkola:
# Description

When the guildslist was encoded for RoF2, the inbound pBuffer was not being deleted correctly, resulting in a memory leak. 

## Type of change

# Testing
Validated the memory leak and the resolution with Valgrind

Clients tested: 
RoF2

# Checklist


Akkadius (approved):
Awesome, nice find!

PR #4586, opened 2025-01-05
[Fix] Repair a LoadNPCEmote MemoryLeak

neckkola:
# Description

Two issues were identified.

1. When a zone was shutdown, the npc_emote_list was cleared, however its contents were established with the 'new' keyword.  The entries were not deleted before the clear.
2. When reloading the NPCEmoteList, for example using #reload npc_emotes, the list was cleared, without the delete.

## Type of change

# Testing
Validated the memory leak and resolution with Valgrind.

# Checklist


Akkadius (approved):
Awesome, nice find!

PR #4595, opened 2025-01-11
[Fix] Update trader add/remove packets to limits for RoF2

neckkola:
# Description

As previously noted, the RoF2 limit for the number of traders is 600.  If there are already 600 traders within the client and a new trader is enabled, the 601st would crash the client.

This fix tracks that condition and does not send the packet.

## Type of change

# Testing
Tested with a client that had 600 traders, and then enabled a new trader.  Prior the client would crash.  Post, the client would not crash.

Clients tested: 
RoF2 against linux and windows builds

# Checklist

### DIFF
--- a/zone/client.h
+++ b/zone/client.h
@@ -1128,6 +1128,10 @@ public:
 	void SetTrader(bool status) { trader = status; }
 	uint16 GetTraderID() { return trader_id; }
 	void SetTraderID(uint16 id) { trader_id = id; }
+	void SetNoOfTraders(uint32 no) { m_no_traders = no; }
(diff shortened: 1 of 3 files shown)


Kinglykrab:
Looks great, appreciate the contribution.

Minor nitpick, could we change the names? Like `GetTraderCount`, `SetTraderCount`, `IncrementTraderCount`, `DecrementTraderCount`, and `m_trader_count`.

neckkola:
Thanks Kingly.  Updated accordingly.

Akkadius (approved):

PR #4598, opened 2025-01-17
[Feature] Add Account and Character Based Expansion Settings

Kinglykrab:
# Description
- Adds the ability to limit expansions on an account or character basis.

## Type of change
- [X] New feature

# Checklist

- [ ] I have tested my changes
- [ ] If my changes make database schema changes, I have tested the changes on a local database (attach image). Updated version.h CURRENT_BINARY_DATABASE_VERSION to the new version. (Delete this if not applicable)

### DIFF
--- a/common/database/database_update_manifest.cpp
+++ b/common/database/database_update_manifest.cpp
@@ -5805,6 +5805,20 @@ ALTER TABLE `zone`
 ADD COLUMN `shard_at_player_count` int(11) NULL DEFAULT 0 AFTER `seconds_before_idle`;
 )",
 		.content_schema_update = true
+	},
+	ManifestEntry{
+		.version = 9289,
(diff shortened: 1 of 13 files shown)


Akkadius:
Have a few things for this in the next few days

Akkadius:
I think for this we should use data buckets. 

We already have account / client flagging built into caching so I am far more comfortable for us leveraging buckets for this. We also already have quest hooks for all of this so we get many things for free.

https://github.com/EQEmu/Server/pull/4603

Kinglykrab:
> I think for this we should use data buckets.
> 
> We already have account / client flagging built into caching so I am far more comfortable for us leveraging buckets for this. We also already have quest hooks for all of this so we get many things for free.
> 
> #4603

Works for me.

PR #4599, opened 2025-01-18
[Feature] Add Support for Item Previews

Kinglykrab:
# Description
- Adds support for the Item Preview window in merchants.
- Credit to @dannuic.

## Type of change
- [X] New feature

# Testing
image
image
image

# Checklist


zone/client_packet.cpp
```diff
@@ -9299,6 +9300,29 @@ void Client::Handle_OP_ItemPreview(const EQApplicationPacket *app)
...
 }
 
+void Client::Handle_OP_ItemPreviewRequest(const EQApplicationPacket* app)
+{
+	VERIFY_PACKET_LENGTH(OP_ItemPreviewRequest, app, ItemPreview_Struct);
+	auto ips  = (ItemPreview_Struct*) app->pBuffer;
+	const EQ::ItemData* item = database.GetItem(ips->itemid);
+
+	if (item) {
+		EQ::ItemInstance* inst = database.CreateItem(item);
```
neckkola, line 9310: inst will need to be deleted I believe.

neckkola:
Good morning.  Kingly, I wonder if we should move the struct verification into the decode/encoders so that when we ever get the other clients online, we can have source be generic, and do the packet differences there?  It looks like we are doing most of that, at least on the decode side within client_packet.  I also see RoF changes though no RoF opcode.  Do we need to add to patch_rof?

Akkadius:
> Good morning. Kingly, I wonder if we should move the struct verification into the decode/encoders so that when we ever get the other clients online, we can have source be generic, and do the packet differences there? It looks like we are doing most of that, at least on the decode side within client_packet. I also see RoF changes though no RoF opcode. Do we need to add to patch_rof?

@neckkola mentioned he will add a separate PR for this later

Akkadius (approved):

PR #4628, opened 2025-01-28
[Bug] Item Purchase Offset when multiple buyers are buying at the same time.

fryguy503:
# Description

- Much of the code lifted from TAKP/P2002 and adapted
- Edge case exists where both players have a purchase qty sub-window open, the bug still persists.
- This should help prevent situations where multiple people are purchasing items from a merchant and both attempt to purchase the same temp merchant or limited item. Purchasing will send a packet update and immediately remove the item from the list.
- Updated strings

Fixes # (issue)

## Type of change

# Testing

Clients tested: 

# Checklist

### DIFF
--- a/common/database/database_update_manifest.cpp
+++ b/common/database/database_update_manifest.cpp
@@ -6394,7 +6394,7 @@ ALTER TABLE `trader`
 	ManifestEntry{
 		.version = 9297,
 		.description = "2024_01_22_sharedbank_guid_primary_key.sql",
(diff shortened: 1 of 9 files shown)


Kinglykrab (approved):

Kinglykrab:
Did some cleanup, fixed merge conflicts.

For future reference, when using an incrementor like `it`, it does not self increment, so if you throw a `continue;` in and do not increment prior to this, you will perform an infinite loop because it will never progress the incrementor.

PR #4647, opened 2025-02-06
[Loginserver] Modernize codebase

Akkadius:
# Description

This cleanup modernizes the loginserver codebase.

* Convert all database interactions to use repositories
* Remove "Database" (Loginserver) class, as we use purely repository methods now
* Simplify logic, remove duplicate logic
* Remove options that should be sane defaults
* Have clean inputs and outputs for function signatures and not "return" things by passing them by reference and modifying them in functions
* Consistent member variable `m_` naming
* Modernize list loops
* Use single letter receiver variables for complex types
* Reduced function signatures that have many duplicate and similar inputs to use "Context" objects which heavily simplifies the data that gets passed around
* Convert fmt logging numbered indexes to just use `{}` instead of `{0} {1} {2}` etc.
* Remove dependency on `login_server_list_types` table as it is purely constants. Constants are used in code instead
* Removed unnecessary doc blocks as this is a typed language and everything is described in the function name and signature
* Remove needless setters/getters

## Type of change

- [x] Code cleanup

# Testing

- Tested logging in
- Tested `/q` back to character select
- Tested unregistered server logging in
- Tested registered server logging in and listing as specific status
- Tested majority of CLI commands

Will list more test cases

# Checklist


world/cliententry.cpp
```diff
@@ -164,16 +164,17 @@ void ClientListEntry::LSUpdate(ZoneServer *iZS)
 	if (WorldConfig::get()->UpdateStats) {
 		auto pack = new ServerPacket;
 		pack->opcode  = ServerOP_LSZoneInfo;
-		pack->size    = sizeof(ZoneInfo_Struct);
+		pack->size    = sizeof(LoginserverZoneInfoUpdate);
 		pack->pBuffer = new uchar[pack->size];
-		ZoneInfo_Struct *zone = (ZoneInfo_Struct *) pack->pBuffer;
+		LoginserverZoneInfoUpdate *zone = (LoginserverZoneInfoUpdate *) pack->pBuffer;
```
Kinglykrab, line 169: Any reason to not use `auto` here?

Kinglykrab (approved):

Issue #4636, opened 2025-02-03
Discovery on Evolve


Veruca1:
A successful item evolve did not show up on discord discovery channel. Item 11010 evolves in to item 644, 11010 disappears, 664 shows up on cursor and no discovery was made.

Kinglykrab:
Resolved by https://github.com/EQEmu/Server/pull/4663.

PR #4663, opened 2025-02-09
[Bug Fix] Fix Item Discovery

Kinglykrab:
# Description
- Fixes an issue where items were not being properly discovered when picked up from the ground, evolved, fished, foraged, awarded by tasks, or as a result of a tradeskill recipe.
- Resolves https://github.com/EQEmu/Server/issues/4636.
- Adds caching to discovered items.

## Type of change
- [X] Bug fix

# Checklist

### DIFF
--- a/zone/client_evolving_items.cpp
+++ b/zone/client_evolving_items.cpp
@@ -260,6 +260,25 @@ bool Client::DoEvolveCheckProgression(const EQ::ItemInstance &inst)
 
 	std::unique_ptr<EQ::ItemInstance> const new_inst(database.CreateItem(new_item_id));
 
+	if (!new_inst) {
+		return false;
+	}
+
+	if (RuleB(Character, EnableDiscoveredItems) && !IsDiscovered(new_inst->GetItem()->ID)) {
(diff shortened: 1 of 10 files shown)


zone/client_evolving_items.cpp
```diff
@@ -260,6 +260,25 @@ bool Client::DoEvolveCheckProgression(const EQ::ItemInstance &inst)
...
+		} else {
+			const std::string& item_link = database.CreateItemLink(new_inst->GetItem()->ID);
+			Message(
+				Chat::White,
+				fmt::format(
+					"Your GM flag prevents {} from being added to discovered items.",
+					item_link
+				).c_str()
+			);
+		}
```
Akkadius, line 279: Can we add this code into a helper and call it once instead of duplicating these 10-15 or so lines

Kinglykrab: Sure.

Akkadius (approved):

Kinglykrab:
Going to add caching to `IsDiscovered` pre-merge to alleviate a query per item addition to player inventory.

PR #4666, opened 2025-02-10
[Commands] [Hotfix] Fix Illusion Block

nytmyr:
# Description

- This wasn't properly being set or saved if bots were not enabled.

## Type of change

# Testing

https://github.com/user-attachments/assets/ce6f0b02-784f-4096-9998-a8733a9a9bbb
- Verified proper loading on new and existing characters.
- Verified proper save and load between zoning.
- Verified NPC spells still bypass check and land as a whole.
- Verified via Project Illusion that illusions by other players honor the targets' setting.
- Verified bots respect rule for players.

Clients tested: RoF2

# Checklist

### DIFF
--- a/common/database/database_update_manifest.cpp
+++ b/common/database/database_update_manifest.cpp
@@ -6757,6 +6757,16 @@ ENGINE=InnoDB
 		.sql = R"(
 ALTER TABLE `character_data`
 	ADD COLUMN `illusion_block` TINYINT(11) UNSIGNED NOT NULL DEFAULT 0 AFTER `deleted_at`;
+
+UPDATE `command_settings`
+SET `aliases`=
+	CASE
+	WHEN LENGTH(`aliases`) > 0
+		THEN CONCAT(`aliases`, '|ib')
+	ELSE 'ib'
+END
+WHERE `command`='illusionblock'
+AND `aliases` NOT LIKE '%ib%';
 )",
 	}
 // -- template; copy/paste this when you need to create a new entry
--- a/common/database/database_update_manifest.cpp
+++ b/common/database/database_update_manifest.cpp
@@ -6747,6 +6747,17 @@ COLLATE='latin1_swedish_ci'
 ENGINE=InnoDB
 ;
 )"
+	},
+	ManifestEntry{
+		.version = 9302,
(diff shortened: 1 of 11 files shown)


common/database/database_update_manifest.cpp
```diff
@@ -6747,6 +6747,27 @@ COLLATE='latin1_swedish_ci'
...
+		.version = 9302,
+		.description = "2025_02_09_illusion_block.sql",
+		.check = "SHOW COLUMNS FROM `character_data` LIKE 'illusion_block'",
+		.condition = "empty",
+		.match = "",
+		.sql = R"(
+ALTER TABLE `character_data`
+	ADD COLUMN `illusion_block` TINYINT(11) UNSIGNED NOT NULL DEFAULT 0 AFTER `deleted_at`;
+
+UPDATE `command_settings`
```
Akkadius, line 6761: This needs to be made idempotent - every migration should be idempotent

This is an example

```
UPDATE `command_settings`
SET `aliases` =
    CASE
        WHEN LENGTH(`aliases`) > 0 AND `aliases` NOT LIKE '%|ib%'
            THEN CONCAT(`aliases`, '|ib')
        WHEN LENGTH(`aliases`) = 0
            THEN 'ib'
        ELSE `aliases`
    END
WHERE `command` = 'illusionblock'
AND `aliases` NOT LIKE '%ib%';
```

zone/client.cpp
```diff
@@ -13597,3 +13597,18 @@ void Client::CheckItemDiscoverability(uint32 item_id)
 
 	DiscoverItem(item_id);
 }
+
+void Client::SetIllusionBlock(bool value, bool save) {
```
Akkadius, line 13601: I'm not really a fan of the optional bool `save` pattern. A function should generally have a single responsibility and one reason to change. ([SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle))

Characters will hit Save() periodically and also when when they zone/camp/save so this will persist anyways. Let's strip this to simplify

Akkadius: Similarly, this field needs to be added in `ZoneDatabase::SaveCharacterData`

Akkadius (approved):

PR #4672, opened 2025-02-13
[Expeditions] Move expedition code into DynamicZone

hgtw:
This removes the separate Expedition class and moves lockout code and /dz command handlers into DynamicZone classes. It also refactors some code to reduce bloat and some database usage.

This completes the effort of moving everything to DynamicZone that started when implementing shared tasks. It also makes sense to do this since expeditions are just dynamic zones internally despite dzs being used for other types. Expedition specific things are just handled with dz type checks.

Functionally nothing should change. This is mainly internal refactoring and moving code around along with some bug fixes and reduced database usage.

Main changes:

 - The `expeditions` database table has been removed

 - Expeditions no longer use a separate id, the expedition id is just the dz id

 - Expedition lock state and replay timer option were moved to the `dynamic_zones` table

 - Expeditions no longer have a separate cache from dynamic zones

 - **Expedition creation no longer has every zone query the database to cache it**

 - Expedition internal lockouts are now stored on DynamicZone

 - The `expedition_lockouts` table has been renamed to `dynamic_zone_lockouts`

 - Fixed a small bug with the UpdateLockoutDuration api where the internal lockout would get the time added twice in memory in the initiating zone (this api is likely rarely used)

 - Fixed an issue where use of the group/raid DoesAnyMemberHaveExpeditionLockout api would query once for every out of zone character.

   - This api now checks all members in the current zone first and only performs a single bulk query for out of zone members if that check is exhausted

 - Deprecated the max_check_count param of DoesAnyMemberHaveExpeditionLockout, the quest api still exists to avoid api break but a passed arg has no effect

Every change was tested but this is a large refactor and it was left to bitrot for a bit so it could benefit from being tested on a server with heavier dz usage (both expeditions and shared tasks) to be sure there are no regressions.

## Type of change

# Checklist

- [x] I have made corresponding changes to the documentation (if applicable, if not delete this line)


Akkadius:
Awesome, I think I'm going to separate this from the huge amount of changes we already have to go out but also don't want this one waiting forever either

Akkadius (approved):

PR #4697, opened 2025-02-20
[Client Mod] Adds a hacked fast camp rule for GMs.

KimLS:
# Description

Adds a rule that disables the GM "disconnect" on /camp.  This allows you to use a hacked client either through injection or actually modifying the eqgame.exe.

## Type of change

# Testing

Tested locally with a hacked client, it should also be testable on a normal client in which case it would take the full 30s to logout on a GM.

Clients tested: RoF2

# Checklist


KimLS:
If you want to actually try the almost instant logout go grab any delta3 program of your choosing.  I like https://github.com/marco-calautti/DeltaPatcher.

You can then apply my mod file for RoF2. [eqgame-camp-mod.zip](https://github.com/user-attachments/files/18880270/eqgame-camp-mod.zip)

Akkadius (approved):

PR #4703, opened 2025-02-21
Fix infinite loop in mob:StopTimer().

zimp-wow:
Stumbled across this one when I had a zone spin up to 100% and lock up when I was writing content.


Kinglykrab (approved):

Kinglykrab:
Good catch.

PR #4706, opened 2025-02-21
[Bots] Line of Sight and Mez optimizations and cleanup

nytmyr:
# Description

- Renames `Map:CheckForLoSCheat` to `Map:CheckForDoorLoSCheat` to better reflect what it does.
- Renames `Map:RangeCheckForLoSCheat` to `Map:RangeCheckForDoorLoSCheat` to better reflect what it does.
- Adds the rule `Pets:PetsRequireLoS` to determine whether or not commanded pet attacks require an additional layer of LoS checks for edge-cases.
- Adds the rule `Bots:BotsRequireLoS` to determine whether or not bots require LoS to `^attack`, `^pull` and `^precombat`.
- Adds the rule `Map:ZonesToCheckDoorCheat` to control which, if any, zones will be checked.
- Corrects, removes and adds LoS checks where necessary and eliminated unnecessary duplicate checks.
- Improves door checking logic for locked or triggered doors that could be blocking LoS.
- Cleans up false positives for door cheat checks.
- Adds `drawbox` option to `#door` command. This will spawn points at the center and each corner of the door's "box". It will also spawn points at your and your target's location.
- Improves Mez and AE Mez logic
- Adds more details to the rule `Bots:EpicPetSpellName`
- Door logic can be further improved upon at a later date to account for different opening types if needed.
 - This is door cheat is really only intended to be used for old zones where the client vs server see some doors differently and the server will pass line of sight checks through old doors for certain commanded things.

## Type of change

# Testing

- Many LoS scenario checks involving pets, bots, combat, casting and AE logic.

Clients tested: RoF2

# Checklist

### DIFF
--- a/zone/botspellsai.cpp
+++ b/zone/botspellsai.cpp
@@ -61,10 +61,7 @@ bool Bot::AICastSpell(Mob* tar, uint8 chance, uint16 spell_type, uint16 sub_targ
 	bot_spell.SpellIndex = 0;
 	bot_spell.ManaCost = 0;
 
-	if (BotSpellTypeRequiresLoS(spell_type) && tar != this) {
-		SetHasLoS(DoLosChecks(tar));
-	}
-	else {
+	if (!BotSpellTypeRequiresLoS(spell_type) || tar == this) {
 		SetHasLoS(true);
 	}
 
@@ -219,7 +216,7 @@ bool Bot::AICastSpell(Mob* tar, uint8 chance, uint16 spell_type, uint16 sub_targ
 
 	for (const auto& s : bot_spell_list) {
 
-		if (!IsValidSpellAndLoS(s.SpellId, HasLoS())) {
+		if (!IsValidSpellAndLoS(s.SpellId, (!BotRequiresLoSToCast(spell_type, s.SpellId) ? true : HasLoS()))) {
 			continue;
 		}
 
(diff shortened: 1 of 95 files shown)


nytmyr:
> Renaming Rules can have adverse effects across the community as the cleanup will remove the old rule and what ever it was set for and set the new default. Just be aware of what can happen by doing so.

Thought about that with this one. Figured there's likely nobody actually using this rule ATM being freshly in the release and defaulted to false. Even if they were, it wasn't a reliable check in its previous form and had false positives. I can switch it to false by default or add a DB update to check and insert the rule based off current settings. Didn't feel the need to add a manifest update for this edge case rule but can if desired.

nytmyr:
Marked as draft till decision on rule change is made

zone/botspellsai.cpp
```diff
@@ -219,7 +216,7 @@ bool Bot::AICastSpell(Mob* tar, uint8 chance, uint16 spell_type, uint16 sub_targ
 
 	for (const auto& s : bot_spell_list) {
 
-		if (!IsValidSpellAndLoS(s.SpellId, HasLoS())) {
+		if (!IsValidSpellAndLoS(s.SpellId, (!BotRequiresLoSToCast(spell_type, s.SpellId) ? true : HasLoS()))) {
```
Akkadius, line 219: This is extremely gnarly to read

`IsValidSpellAndLoS` should just validate a valid spell, not also check LOS

There should be a separate check that handles LOS requirements of a spell which can do ternary stuff inside it and make things a little easier to read. This is duplicated in a few other spots.

zone/mob.cpp
```diff
@@ -8726,27 +8740,24 @@ bool Mob::CheckLosCheat(Mob* other) {
 	return true;
 }
 
-bool Mob::CheckLosCheatExempt(Mob* other)
```
Akkadius, line 8729: This is all commented out and just returns false ? Should we just remove this code ?

nytmyr:
Oh god, this got fucked. Closing and adding a new PR

PR #4716, opened 2025-02-24
[Feature] Add a rule for spells to bypass stacking rules

catapultam-habeo:
# Description

Adds `Spells:AlwaysStackSpells` to allow for a comma-separated list of SpellIDs which will always stack with every other spell, excluding themselves. 

# Testing
Case 1:
 -> Added 3467 (Virtue) to list, along with 3 other arbitrary IDs
 -> Cast Temperance (3692) and Virtue.
 -> Observed stacking.
 -> Tested each cast order of these two spells
 -> Cast Aegolism (1447)
 -> Observe stacks with Virtue, overwrites Temperance

Case 2:
 -> Added 3467 (Virtue) to list, along with 3 other arbitrary ID 
 -> Cast Protection of Cabbage, then Virtue
 -> Observe stacking
 -> Removed Virtue, Removed 3467 from rule list
 -> Cast Virtue, observed no stack.

Clients tested: 
ROF2

# Checklist

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -3135,13 +3135,13 @@ int Mob::CheckStackConflict(uint16 spellid1, int caster_level1, uint16 spellid2,
 		}
 	}
 
-	const std::string& always_stack_str = RuleS(Spells, AlwaysStackSpells);
-	if (spellid1 != spellid2 && !always_stack_str.empty()) {
-		const auto& v = Strings::Split(always_stack_str, ",");
-		if (std::find(v.begin(), v.end(), std::to_string(spellid1)) != v.end()) {
+	const std::string& always_stack_spells = RuleS(Spells, AlwaysStackSpells);
+	if (spellid1 != spellid2 && !always_stack_spells.empty()) {
+		const auto& v = Strings::Split(always_stack_spells, ",");
+		if (Strings::Contains(v, std::to_string(spellid1))) {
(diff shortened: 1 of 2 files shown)


zone/spells.cpp
```diff
@@ -3135,6 +3135,17 @@ int Mob::CheckStackConflict(uint16 spellid1, int caster_level1, uint16 spellid2,
 		}
 	}
 
+	const std::string& always_stack_str = RuleS(Spells, AlwaysStackSpells);
+	if (spellid1 != spellid2 && !always_stack_str.empty()) {
+		const auto& v = Strings::Split(always_stack_str, ",");
+		if (std::find(v.begin(), v.end(), std::to_string(spellid1)) != v.end()) {
```
Akkadius, line 3141: We have an in-house `Strings` utility to check if a string is in a vector of strings

```cpp
const std::string& always_stack_str = RuleS(Spells, AlwaysStackSpells);
if (spellid1 != spellid2 && !always_stack_str.empty()) {
	const auto& v = Strings::Split(always_stack_str, ",");
	if (Strings::Contains(v, std::to_string(spellid1))) {
		return 0;
	}
	if (Strings::Contains(v, std::to_string(spellid2))) {
		return 0;
	}
}
```

Akkadius (approved):

PR #4739, opened 2025-03-02
[Bug Fix] Cleanup zone buckets on instance purge.

zimp-wow:
When I tested it I found that the zone-scoped data buckets were persisting after the instance disappeared from the instance_list table due to hitting its expiration time.  Appeared to be due to this bit of missing logic in 'PurgeExpiredInstances'.

Test Steps:

1. Create an instance and set a zone data bucket.
2. Leave the DZ.
3. Update the start_time of the instance in instance_list to some value in the distant past and restart server.

Expected:

Zone data buckets with the same instance id disappear when instance is removed from instance_list.


Akkadius:
Good catch. These sorts of things ideally should all funnel through one spot but unfortunately we have two places where instances get deleted and expeditions have gone a little bit outside how instances are typically handled.

It was covered in **DeleteInstance** in the original zone bucket PR  https://github.com/EQEmu/Server/blob/27274397ecad6c920210011a61b48b1e41fcc8f2/common/database_instances.cpp#L482

Akkadius (approved):

PR #4746, opened 2025-03-03
[Bots] Line of Sight and Mez optimizations and cleanup

nytmyr:
# Description

This a is a resubmit of me derping https://github.com/EQEmu/Server/pull/4706

- Renames `Map:CheckForLoSCheat` to `Map:CheckForDoorLoSCheat` to better reflect what it does.
- Renames `Map:RangeCheckForLoSCheat` to `Map:RangeCheckForDoorLoSCheat` to better reflect what it does.
- Adds the rule `Pets:PetsRequireLoS` to determine whether or not commanded pet attacks require an additional layer of LoS checks for edge-cases.
- Adds the rule `Bots:BotsRequireLoS` to determine whether or not bots require LoS to `^attack`, `^pull` and `^precombat`.
- Adds the rule `Map:ZonesToCheckDoorCheat` to control which, if any, zones will be checked.
- Corrects, removes and adds LoS checks where necessary and eliminated unnecessary duplicate checks.
- Improves door checking logic for locked or triggered doors that could be blocking LoS.
- Cleans up false positives for door cheat checks.
- Adds `drawbox` option to `#door` command. This will spawn points at the center and each corner of the door's "box". It will also spawn points at your and your target's location.
- Improves Mez and AE Mez logic
- Adds more details to the rule `Bots:EpicPetSpellName`
- Door logic can be further improved upon at a later date to account for different opening types if needed.
 - This is door cheat is really only intended to be used for old zones where the client vs server see some doors differently and the server will pass line of sight checks through old doors for certain commanded things.

## Type of change

# Testing

- Many LoS scenario checks involving pets, bots, combat, casting and AE logic.

Clients tested: RoF2

# Checklist


nytmyr:
This a is a resubmit of me derping https://github.com/EQEmu/Server/pull/4706

zone/mob.cpp
```diff
@@ -8668,56 +8668,70 @@ bool Mob::IsInGroupOrRaid(Mob* other, bool same_raid_group) {
...
+							fmt::format(
+								"#goto {} {} {} {}",
+								d->GetX(),
+								d->GetY(),
+								d->GetZ(),
+								d->GetHeading()
+							)
+						)
+					)
+				); //deleteme
```
Kinglykrab, line 8731: ??

zone/mob.cpp
```diff
@@ -8668,56 +8668,54 @@ bool Mob::IsInGroupOrRaid(Mob* other, bool same_raid_group) {
...
+
+	const std::string& zones_to_check = RuleS(Map, ZonesToCheckDoorCheat);
+
+	if (zones_to_check.empty()) {
+		return true;
+	}
+
+	const auto& v = Strings::Split(zones_to_check, ",");
+
+	if (zones_to_check == "all" || std::find(v.begin(), v.end(), std::to_string(zone->GetZoneID())) != v.end()) {
```
Akkadius, line 8698: We have a utility for this

```cpp
	const auto& v = Strings::Split(zones_to_check, ",");
	if (zones_to_check == "all" || Strings::Contains(v, std::to_string(zone->GetZoneID()))) {
```

zone/bot_commands/precombat.cpp
```diff
@@ -18,7 +18,7 @@ void bot_command_precombat(Client* c, const Seperator* sep)
 		return;
 	}
 
-	if (!c->DoLosChecks(c->GetTarget())) {
+	if (RuleB(Bots, BotsRequireLoS) && !c->DoLosChecks(c->GetTarget())) {
```
Akkadius, line 21: In a lot of the places you're checking for DoLosChecks you're also adding this RuleB. We could simplify this by doing 

```cpp
bool Mob::DoLosChecks(Mob* other) {
	if (IsBot() && !RuleB(Bots, BotsRequireLoS)) {
		return false;
	}
```

zone/bot_commands/pull.cpp
```diff
@@ -48,7 +48,7 @@ void bot_command_pull(Client *c, const Seperator *sep)
 		return;
 	}
 
-	if (!c->DoLosChecks(target_mob)) {
+	if (RuleB(Bots, BotsRequireLoS) && !c->DoLosChecks(target_mob)) {
```
Akkadius, line 51: See https://github.com/EQEmu/Server/pull/4746/files#r2004455621

zone/botspellsai.cpp
```diff
@@ -1448,29 +1477,29 @@ Mob* Bot::GetFirstIncomingMobToMez(Bot* caster, int16 spell_id, uint16 spell_typ
 				continue;
 			}
 
+			if (is_pbae_spell) {
```
Akkadius, line 1480: This could probably be simplified down

```cpp
if ((is_pbae_spell ? spell_ae_range : spell_range) < Distance(caster->GetPosition(), npc->GetPosition())) {
    continue;
}
```

zone/bot_commands/cast.cpp
```diff
@@ -543,6 +546,9 @@ void bot_command_cast(Client* c, const Seperator* sep)
 				tar = bot_iter;
 			}
 
+			bool los_required = bot_iter != tar && !IsAnyHealSpell(chosen_spell_id) && !IsPBAESpell(chosen_spell_id);
```
Akkadius, line 549: Didn't we just add a helper to simplify if a spell requires LOS ? It'd be good to centralize that into a readable function

zone/bot_commands/attack.cpp
```diff
@@ -22,7 +22,7 @@ void bot_command_attack(Client *c, const Seperator *sep)
 		return;
 	}
 	
-	if (!c->DoLosChecks(target_mob)) {
+	if (RuleB(Bots, BotsRequireLoS) && !c->DoLosChecks(target_mob)) {
```
Akkadius, line 25: See https://github.com/EQEmu/Server/pull/4746/files#r2004455621

zone/bot.cpp
```diff
@@ -3215,14 +3215,17 @@ bool Bot::IsValidTarget(
 		return false;
 	}
 
+	SetHasLoS(DoLosChecks(tar));
```
Akkadius, line 3218: I'm not sure why we're adding a member variable to keep hold of LOS as LOS is prone to change at a moments notice and could become frail to other parts of the code. It's something you want to be able to calculate at that point in the code versus caching and hoping its accurate by the time you use it somewhere else.

If the goal is to reduce CPU cycles we should consider other avenues as this adds frailties to the code.

zone/bot_commands/cast.cpp
Akkadius: https://github.com/EQEmu/Server/pull/4746/files#diff-417633ac61d6e8de2dd9af8c0f76b6e7d0cf394bcb8b9171484527361c98455fR528

zone/bot.cpp
```diff
@@ -11443,7 +11446,7 @@ bool Bot::AttemptForcedCastSpell(Mob* tar, uint16 spell_id, bool is_disc) {
 		return false;
 	}
 
-	if (!DoLosChecks(tar)) {
+	if (!HasLoS() && !DoLosChecks(tar)) {
```
Akkadius, line 11449: Seems strange to not be able to rely on the member variable for HasLoS and also check / calculate anyways here. We should just simplify down to using DoLosChecks and do other things to measure how heavy LOS checks are with bot code paths and look for other means to optimize costs

zone/bot_commands/cast.cpp
nytmyr: Current helper requires a spell_type as well, created one just for spell_id.

Akkadius (approved):

Issue #4260, opened 2024-04-15
[Request] /changename support


fryguy503:
# Description

/changename (and the associated server side flag that should trigger the name change window), as well as questapi calls to trigger, would be useful for server owners to flag and enable players to manage name changes.

Packet sent from the client to server is OP_GMNameChange, but is smaller than the expected packet size and is discarded.

4/15/2024 4:59:56 PM Wrong size: OP_GMNameChange, size=[132], expected [196]
4/15/2024 4:59:56 PM [OP_GMNameChange] [0x035f] Size [134] 
   0: 54 72 75 73 74 74 65 73 - 74 00 00 00 00 00 00 00  | Trusttest.......
  16: 00 00 00 00 00 00 00 00 - 00 00 00 00 00 00 00 00  | ................
  32: 00 00 00 00 00 00 00 00 - 00 00 00 00 00 00 00 00  | ................
  48: 00 00 00 00 00 00 00 00 - 00 00 00 00 00 00 00 00  | ................
  64: 54 72 75 73 74 00 00 00 - 00 00 00 00 00 00 00 00  | Trust...........
  80: 00 00 00 00 00 00 00 00 - 00 00 00 00 00 00 00 00  | ................
  96: 00 00 00 00 00 00 00 00 - 00 00 00 00 00 00 00 00  | ................
 112: 00 00 00 00 00 00 00 00 - 00 00 00 00 00 00 00 00  | ................
 128: 00 00 00 00                                        | ....

You can see the desired name and the current name. This falls into the current structures 64 and 64 bytes.

```cpp
struct GMName_Struct {
	char oldname[64];
	char gmname[64];
	char newname[64];
	uint8 badname;
	uint8 unknown[3];
};
```

However it looks like it is giving them in a different order.

# Screenshots

Window manually opened.
![image](https://github.com/EQEmu/Server/assets/5864295/231c1651-45cf-4dd6-bc9a-929fb37e2688)

Window manually opened and a request submitted.
![image](https://github.com/EQEmu/Server/assets/5864295/85a349f5-a57f-404a-b926-e2ccef81a90a)

PR #4770, opened 2025-03-10
[Feature] Implement /changename & related script bindings. Clean up #set name

catapultam-habeo:
# Description

Implements #4260. 

* Implements the following script bindings;
`Client::GrantNameChange()` -> Immediately invokes the Name Change window, and allows usability of the manually invoked /changename command. Name chosen is run through the character creation name filters, the vanity pet name filters, and the profanity filters and rejected as appropriate. (A client sharing a name with an NPC seems to cause clientside memory corruption). While usability is enabled, the window will be presented upon initial login for the character, and dismissing it with the 'Don't Remind Me Again' button will cause this behavior to be suppressed.
`Client::IsNameChangeAllowed()` -> Returns true if name change is currently enabled, false if disabled.
`Client::ClearNameChange()` -> Removes pending name change usability. Returns true if succeeded, false if there was no usability permitted.

* Implements the rule `Character::AlwaysAllowNameChange` which allows /changename to always work without needing to be enabled by `Client::GrantNameChange()`.

* Refactors the `#set name` GM-command to use the same code paths as the user-facing system. The notable improvement here is that it no longer requires the player to be kicked to character select.

* Refactors `quest::rename(...)` to use the same code path as the user-facing system. Likewise no longer requires the player to be kicked to character select.

* Deprecates `Client::ChangeFirstName(in_firstname, gmname)`, as the `OP_GMNameChange` no longer functions in ROF2 as it apparently did in the past. There is more room here to refine the handler for this packet, as the only way this packet is processed is as this struct in RoF2;
```
struct AltChangeName_Struct 
{
  /*00*/ char new_name[64];
  /*40*/ char old_name[64];
  /*80*/ int  response_code;
};
```

## Type of change

# Testing

```
-- #Arias.lua
function event_say(e)
 e.self:Say("Name Change is Currently : " ..tostring(e.other:IsNameChangeAllowed()))
 e.other:GrantNameChange()
 e.self:Say("Name Change is now... :" ..tostring(e.other:IsNameChangeAllowed()))
end
```
image

image
image
image

Clients tested: 

RoF2 (required, older clients unsupported)

# Checklist


zimp-wow:
### Test Environment

Akk-stack running cata's branch
Base PEQ snapshot
Latest master branch of PEQ quests
Unmodified ROF2 client

### Test Case : Result

#set name
- No argument: Displays help usage
- Existing name: Error displayed about name already taken
- ' ' or " ": Error about name already taken
- Curse word: Error about name already taken
- Really long name: Error about name already taken
- Two words separated by space: Only first word is taken as the input
- Pet name in 'character_pet_name': Already taken
- Successful change: Character name updates in real-time

'/changename'
- Use without grant: Popup and chat message about not eligible.
- empty name: Submit button does nothing
- ' ' or " ": Input not allowed
- Curse word: Invalid name error popup and chat message
- Really long name: Input stopped at a max length
- Two words: Input not allowed
- Attempting second name change without re-grant: Not eligible
- Pet name: Already taken
- Successful change: Character name updates in real-time

API
- Lua:GrantChangeName: Working
- Lua:IsNameChangedAllowed: Working
- Lua:ClearNameChange: Working
- Lua:rename: Working, but does still say the 'kicking to character select' message.
- Perl:GrantChangeName: Working
- Perl:IsNameChangeAllowed: Working
- **Perl:ClearNameChange: Missing?**
- Rule: Working

### Issues/Observations

- The error message about the name being already taken appears to display stale data about the current name of the player.  Seems to require a relog or a zone change to update.  Seems very purely cosmetic and doesn't seem like it would be encountered in most use cases.
- GrantChangeName did not seem to open the window on the client which doesn't match the description in the PR.  Similarly the window does not appear on login.  Could only get it open with the /changename command.
- As mentioned about 'ClearNameChange' appeared to be missing in the perl api changes.

catapultam-habeo:
In reference to @zimp-wow test, I have added the missing `ClearNameChange` to perl api, adjusted `GrantNameChange` to be of type void instead of type bool (and allow re-initiating it even if the rule allows always name change or a credit is already issued).

The popup on login intentionally is not provoked when the rule to always allow changes is enabled.


I have attached some basic testing demos.

catapultam-habeo:
@Akkadius if you have any idea why the windows build failed here, I would appreciate the insight.

Akkadius:
> @Akkadius if you have any idea why the windows build failed here, I would appreciate the insight.

It's in the build output

```
C:\Windows\Temp\drone-dq8FW86mGv7FfgQQ\drone\src\zone\client.cpp(4809,2): error C2562: 'Client::GrantNameChange': 'void' function returning a value [C:\Windows\Temp\drone-dq8FW86mGv7FfgQQ\drone\src\win-build-x64\zone\zone.vcxproj]
C:\Windows\Temp\drone-dq8FW86mGv7FfgQQ\drone\src\zone\client.cpp(4800): message : see declaration of 'Client::GrantNameChange' [C:\Windows\Temp\drone-dq8FW86mGv7FfgQQ\drone\src\win-build-x64\zone\zone.vcxproj]
```

zimp-wow:
I wasn't seeing the windows pop up because my opcodes got patched.  Was testing on my older akk-stack install and it was probably still enabled from before I understood that its generally a bad thing to do.  So that clears up that bit of confusion.

zone/questmgr.cpp
```diff
@@ -1292,15 +1292,14 @@ void QuestManager::rename(std::string name) {
...
+		if (initiator->ChangeFirstName(name.c_str())) {
 			initiator->Message(
 				Chat::White,
 				fmt::format(
-					"Successfully renamed to {}, kicking to character select.",
+					"Successfully renamed to {}.",
 					name
 				).c_str()
 			);
-			initiator->Kick("Name was changed.");
```
Akkadius, line 1303: Are you sure there aren't unexpected consequences of not issuing a kick ? We had this in here previously for a reason

zone/perl_client.cpp
```diff
@@ -3393,6 +3408,9 @@ void perl_register_client()
 	package.add("CashReward", &Perl_Client_CashReward);
 	package.add("ChangeLastName", &Perl_Client_ChangeLastName);
 	package.add("GrantPetNameChange", &Perl_Client_GrantPetNameChange);
+	package.add("GrantNameChange", (void(*)(Client*))&Perl_Client_GrantNameChange);
+	package.add("IsNameChangeAllowed", (bool(*)(Client*))&Perl_Client_IsNameChangeAllowed);
```
Akkadius, line 3412: We are trying to keep this list alpha ordered, I'd move IsNameChangeAllowed down

zone/lua_client.cpp
```diff
@@ -3635,6 +3653,9 @@ luabind::scope lua_register_client() {
 	.def("CashReward", &Lua_Client::CashReward)
 	.def("ChangeLastName", (void(Lua_Client::*)(std::string))&Lua_Client::ChangeLastName)
 	.def("GrantPetNameChange", &Lua_Client::GrantPetNameChange)
+	.def("GrantNameChange", &Lua_Client::GrantNameChange)
+	.def("IsNameChangeAllowed", &Lua_Client::IsNameChangeAllowed)
```
Akkadius, line 3657: Alpha order

zone/gm_commands/set/name.cpp
```diff
@@ -24,17 +24,13 @@ void SetName(Client *c, const Seperator *sep)
 				).c_str()
 			);
 
-			c->Message(Chat::White, "Sending player to char select.");
-
-			t->Kick("Name was changed");
```
Akkadius, line 29: Same comment about not kicking and consequences

zone/client.cpp
```diff
@@ -4737,6 +4756,57 @@ bool Client::KeyRingRemove(uint32 item_id)
...
+
+	auto outapp = new EQApplicationPacket(packet_op, 0);
+	QueuePacket(outapp);
+	safe_delete(outapp);
+}
+
+void Client::GrantNameChange() {
+
+	auto k = GetScopedBucketKeys();
+	k.key = "NameChangesAllowed";
```
Akkadius, line 4803: Let's keep bucket keys uniform to `snake_case` or slugged lower case

zone/client.cpp
```diff
@@ -2579,10 +2568,40 @@ bool Client::ChangeFirstName(const char* in_firstname, const char* gmname)
 	entity_list.QueueClients(this, outapp, false);
 	safe_delete(outapp);
 
+	// success
+	return true;
+}
+
+bool Client::ChangeFirstName(const char* in_firstname)
```
Akkadius, line 2575: Any reason we didn't just use std::string here ? We are using modern string types wherever possible

Akkadius:
Few minor comments otherwise it looks great!

zone/questmgr.cpp
catapultam-habeo: Yes, I cleaned up the serverside ones, and the ROF2 client performs the updates needed between the packets issued as part of this PR.

zone/client.cpp
catapultam-habeo: old method signature I think? want me to refactor to that?

zone/questmgr.cpp
catapultam-habeo: Either way, if it turns out to be an issue, I'll fix it. I namechanged and went around and did a bunch of stuff and had no issues (zoning, combat, char state saves, etc etc), but we all know how wild the edge cases can be. Would you prefer this behavior be rule gated?

catapultam-habeo:
string handling revisions are limited in scope, I don't want to go down the rabbit hole of changing outside of scope atm

Akkadius (approved):
@zimp-wow is going to do one last round of testing

common/database.cpp
```diff
@@ -955,6 +955,29 @@ bool Database::UpdateName(const std::string& old_name, const std::string& new_na
...
 }
 
+bool Database::UpdateNameByID(const int character_id, const std::string& new_name)
+{
+	LogInfo("Renaming [{}] to [{}]", character_id, new_name);
+
+	auto l = CharacterDataRepository::GetWhere(
+		*this,
+		fmt::format(
+			"`id` = '{}'",
```
Akkadius, line 965: Integers don't need to be quoted in queries but this will still work. It can sometimes throw off query planners though if you quote integers and can make slow queries

PR #4785, opened 2025-03-19
[Fix] Zone State Entity Variable Load Pre-Spawn

Akkadius:
# Description

This fixes another rare edge case with zone state restore where entity variables are not present during an EVENT_SPAWN, they are restored shortly after. This fixes that and loads them prior to the NPC being added to the NPC list and firing quest events.

Originally reported by @zimp-wow in https://github.com/EQEmu/Server/pull/4783

## Type of change

# Testing

Tested basic functionality

image

# Checklist

### DIFF
--- a/zone/spawn2.cpp
+++ b/zone/spawn2.cpp
@@ -277,6 +277,12 @@ bool Spawn2::Process() {
 
 		npcthis = npc;
 
+		if (!m_entity_variables.empty()) {
+			for (auto &var : m_entity_variables) {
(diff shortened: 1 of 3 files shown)


Kinglykrab:
This also contains the position fix, just so you’re aware.

Kinglykrab (approved):

Akkadius:
> This also contains the position fix, just so you’re aware.

Good catch and thanks

PR #4794, opened 2025-03-22
[Performance] Reduce LFGuild Chatter to World

Akkadius:
# Description

This is a performance adjustment for **World** where 3k players and 1,800 zoneservers showed strain on World. A grand majority of the CPU load (98%) this was shown to be TCP chatter (server to server).

I ran S<->S packet logging in world and found the following data. This change will be one of several found from these results. https://gist.github.com/Akkadius/12b57fadbbe21ea9a46b854c62cc3a65#file-world-s2s-1m-md

For this PR - this targets the following packets and should reduce chatter by upwards of 100% in the case that most servers don't use this feature, don't run queryserv to support it and it creates a lot of packet chatter for almost nothing.

* **ServerOP_QueryServGeneric** (520 packets in a minute - 8.6 packets/s received from zones)

## Type of change

- [x] Optimization

# Testing

Rule disabled, packet doesn't send

# Checklist


KimLS (changes requested):
Default to previous functionality which would be on.  People who need it off can make that choice.

KimLS (approved):

PR #4796, opened 2025-03-22
[Performance] Have World Send Smarter Guild Updates

Akkadius:
# Description

This is a performance adjustment for **World** where 3k players and 1,800 zoneservers showed strain on World. A grand majority of the CPU load (98%) this was shown to be TCP chatter (server to server).

I ran S<->S packet logging in world and found the following data. This change will be one of several found from these results. https://gist.github.com/Akkadius/12b57fadbbe21ea9a46b854c62cc3a65#file-world-s2s-1m-md

---

For this PR, this addresses a lot of places where we blast packets to all zoneservers regardless if the zone needs it or not. This is not free especially at the scale of THJ. World is absolutely getting crushed having to allocate memory to send packets to X zoneservers (in this case lets say 1,800) when only say a dozen care about the packet at all.

We need to be smarter about how we relay packets because it adds up tremendously

## Type of change

- [x] Optimization

# Testing

Tested basic guild updates, create guild, zoning, adding, deleting

I added debugging in my packet sending function which only grabs zone server's that have clients with the interested Guild ID attached.

```
Zone 1
Zone 2
Sending packet to zone server_id 1
Sending packet to zone server_id 2
Zone 1
Zone 2
Sending packet to zone server_id 1
Sending packet to zone server_id 2
```

# Checklist

### DIFF
--- a/world/clientlist.cpp
+++ b/world/clientlist.cpp
@@ -1850,3 +1850,39 @@ std::map<uint32, ClientListEntry *> ClientList::GetGuildClientsWithTributeOptIn(
 	}
 	return guild_members;
 }
+
+#include <unordered_set>
+
+std::vector<uint32_t> ClientList::GetGuildZoneServers(uint32 guild_id)
+{
+	std::vector<uint32_t>        zone_server_ids;
+	std::unordered_set<uint32_t> seen_ids;
+
+	LinkedListIterator<ClientListEntry *> iterator(clientlist);
+
+	iterator.Reset();
+	while (iterator.MoreElements()) {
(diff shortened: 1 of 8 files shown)


Akkadius:
I'd like @neckkola to take a look at this one and make sure we are safe for all of the packet types to do this with. I am hoping he can run it through his testing as well.

neckkola (approved):
Don't see anything that should cause problems.

neckkola:
THJ are experiencing some issues that have been traced to this PR.  I can reproduce and will look at updating before this gets merged.

PR #4804, opened 2025-03-24
[Bugfix] Shutdown zone instances when they are purged from instance_list.

zimp-wow:
This may be more of a conversation starter than an actual solution.  Not entirely convinced this is the best approach (nor am I certain I did it right).

# Description

A change was made downstream that removed the 24 hour buffer on purging old instances from the instance_list table.  This combined with other settings like DynamicZone:EmptyShutdownDelaySeconds and the zone shutdown delay value can result in a scenario where the instance is purged from instance_list while the zone process is still running.

When this happens if a player were to then start a new instance and get that same instance id and attempt to zone in they will be connected to that existing process.  This may result in odd behaviors like zoning into the wrong zone, or the zone they zone into is in an odd state.

HOWEVER

There's a lot of factors that goes into this that really restricts the chances for this to happen.  As long as the current EQEmu source retains the 24 hour buffer in common/database_instances.cpp this problem effectively doesnt exist by default unless someone tweaks zone shutdown delays to some value > 24 hours.  Secondly, the calculation for purging the instance ids is:

start time + EmptyShutdownDelaySeconds + time player spent in dz + 24 hours (or 0 if on THJ).

So the issue can be further mitigated through things like increasing the EmptyShutdownDelaySeconds or by ensuring zone shutdown delays are below that value or by the fact the players will likely spend 10+ mins in an instance on average.

The only reason I even noticed this was because I had EmptyShutdownDelaySeconds set to 1 from testing zone-state saving and my zone shutdown delays are still quite high from the default PEQ database snapshot.

# Testing

akk-stack + RoF2 client

Test Steps
- Spun up DZ
- Zoned in
- Left
- Quit DZ

Reproduction
- Wait for instance to drop from instance_list, observe that the zone process is still running according to spire.
- Request a new DZ, preferrably for a different zone than the first
- Observe that zoning into the new DZ takes you to the old zone and not the one you requested.

Fixed
- Wait for instance to drop from instance_list, observe that the zone process immediately disappears with it

image

# Concerns

My outstanding concerns is that at scale on something like THJ there would still be a large enough window for players to end up trying to connect to that zone process that is in the process of shutting down.  I'm not sure what the resulting behavior will be in that case as its a bit hard to test.

### DIFF
--- a/common/database.h
+++ b/common/database.h
@@ -159,7 +159,7 @@ public:
 	void FlagInstanceByGroupLeader(uint32 zone_id, int16 version, uint32 character_id, uint32 group_id);
 	void FlagInstanceByRaidLeader(uint32 zone_id, int16 version, uint32 character_id, uint32 raid_id);
 	void GetCharactersInInstance(uint16 instance_id, std::list<uint32>& character_ids);
-	void PurgeExpiredInstances();
+	std::vector<int32_t> PurgeExpiredInstances();
 	void SetInstanceDuration(uint16 instance_id, uint32 new_duration);
 	void CleanupInstanceCorpses();
 
--- a/common/database_instances.cpp
+++ b/common/database_instances.cpp
@@ -535,8 +535,10 @@ void Database::GetCharactersInInstance(uint16 instance_id, std::list<uint32> &ch
 	}
 }
 
-void Database::PurgeExpiredInstances()
+std::vector<int32_t> Database::PurgeExpiredInstances()
 {
+	std::vector<int32_t> retval;
+
 	/**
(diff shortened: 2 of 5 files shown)


Akkadius:
Thank you so much @zimp-wow - you are an absolute ace.

I love that you traced this down. I added this [PR](https://github.com/EQEmu/Server/pull/4820) which adjusts a few things I needed to adjust anyhow and somewhat pulls back the change we made downstream to cull zones more aggressively.

This will still offset expiration purges by an hour to allow zone processes to churn gracefully and prevent the issue you found from occurring.

PR #4836, opened 2025-04-03
[Performance] Store Player Title Sets in Client Memory

Kinglykrab:
# Description
- Stores player title sets in client memory, loading them when a player zones in, instead of having an individual query per title.
- Testing shows no queries being ran, but server is aware the player has the proper title sets.

## Type of change
- [X] New feature

# Testing
image
image
image

# Checklist

### DIFF
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -1100,6 +1100,7 @@ bool Client::Save(uint8 iCommitNow) {
 	database.SaveCharacterData(this, &m_pp, &m_epp); /* Save Character Data */
 
 	database.SaveCharacterEXPModifier(this);
+	database.SaveCharacterTitleSets(this);
 
 	if (RuleB(Bots, Enabled)) {
 		database.botdb.SaveBotSettings(this);
--- a/zone/client.cpp
+++ b/zone/client.cpp
@@ -1100,7 +1100,6 @@ bool Client::Save(uint8 iCommitNow) {
 	database.SaveCharacterData(this, &m_pp, &m_epp); /* Save Character Data */
 
 	database.SaveCharacterEXPModifier(this);
-	database.SaveCharacterTitleSets(this);
 
 	if (RuleB(Bots, Enabled)) {
 		database.botdb.SaveBotSettings(this);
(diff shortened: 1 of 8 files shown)


zone/zonedb.cpp
```diff
@@ -4271,3 +4271,37 @@ void ZoneDatabase::SaveCharacterEXPModifier(Client* c)
...
+	const uint32 character_id = c->CharacterID();
+
+	for (const auto& e : l) {
+		c->AddTitle(e);
+	}
+}
+
+void ZoneDatabase::SaveCharacterTitleSets(Client* c)
+{
+	if (!zone || !c) {
```
Akkadius, line 4302: This need to be || c->GetTitles().empty()

zone/client.cpp
```diff
@@ -1100,6 +1100,7 @@ bool Client::Save(uint8 iCommitNow) {
 	database.SaveCharacterData(this, &m_pp, &m_epp); /* Save Character Data */
 
 	database.SaveCharacterEXPModifier(this);
+	database.SaveCharacterTitleSets(this);
```
Akkadius, line 1103: This PR looks great, generally I want us to be moving away from shoving so many things into Client::Save and actually work to save things when we need to save them selectively rather than brute-forcing saving a ton of stuff all of the time. 

Is there a better place where we could put this ?

Akkadius:
So I think this is most of the way there, your work on titles has been amazing. The titles code was extremely convoluted and now it is much better.

**Client::AddTitle** and **Client::EnableTitle** seems functionally the exact same, it would be good to consolidate them both.

`SaveCharacterTitleSets` also seems no longer needed.

After those adjustments and another round of testing we should be good to go <3

Kinglykrab:
> So I think this is most of the way there, your work on titles has been amazing. The titles code was extremely convoluted and now it is much better.
> 
> **Client::AddTitle** and **Client::EnableTitle** seems functionally the exact same, it would be good to consolidate them both.
> 
> `SaveCharacterTitleSets` also seems no longer needed.
> 
> After those adjustments and another round of testing we should be good to go <3

Reran testing, seems to be just fine with the changes.

Akkadius (approved):

PR #4841, opened 2025-04-04
[Bugfix] Prevent depops from blocking new spawns.

zimp-wow:
# Description

Was goofing around with potimeb and noticed that #repops and quest hot reloads were resulting in situations where the zone_status/zone_emote npcs were not spawned.

Digging into the problem showed that we were calling Depop on the NPC and then continuing with the rest of our reload logic.  The reload logic would then fail to spawn the replacement because spawn_limit was configured and the Depop had not yet been processed (which is what is responsible for removing the npc from the limit_list in entity.cpp).

This change goes ahead and removes the npc from the limit_list to prevent it from counting against the spawn_limit while it is still being processed for the depop.

Now normally in this scenario the respawn timer would start for the spawn but in the case of Non-Respawning instances on THJ this would be some high value and never actually happen.  This makes the behavior a bit easier to observe because on a normal zone process the npc would eventually show back up depending on its natural respawn timer, further muddying the water.

## Type of change

# Testing

Latest master branch with rof2 client.

#repop without change
image

#repop with change
image

# Checklist

### DIFF
--- a/zone/entity.cpp
+++ b/zone/entity.cpp
@@ -3159,6 +3159,13 @@ void EntityList::Depop(bool StartSpawnTimer)
 				UpdateFindableNPCState(pnpc, true);
 			}
 
+			// Depop below will eventually remove this npc from the entity list
+			// but that can happen AFTER we've already tried to spawn its replacement.
+			// So go ahead and remove it from the limits so it doesn't count.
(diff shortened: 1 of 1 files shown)


Akkadius (approved):

Akkadius:
This has been working on THJ

PR #4848, opened 2025-04-06
[Spells] Fear resistance effects edge case fixes and support for SPA 102 as an AA

KayenEQ:
Update to how we calculate fear resist chance from bonuses. This resolves a few rare edge cases that could potentially pop up.

Problem: SPA 181 FearResistChance  (% chance to outright resist fear) and SPA 102 Fearless (on live used only on pets for  full fear immunity if presents, with value of 1 = 100% immunity)
The way we were calculating the bonuses was mixing the variables we used for each together in a way that could cause conflicts when combined with a negate spell effect (SPA 382).

Solution: SPA 181 and SPA 102 each use their own variable consistently.

This update also adds support for SPA 102 to be used in AA.

While doing some live testing was able to confirm SPA102 when cast on pets will terminate the fear on the next tick which is consistent with how we have the mechanic coded.

## Type of change

# Checklist

### DIFF
--- a/zone/spells.cpp
+++ b/zone/spells.cpp
@@ -5667,14 +5667,9 @@ int16 Mob::CalcResistChanceBonus()
 
 int16 Mob::CalcFearResistChance()
 {
-	int resistchance = spellbonuses.ResistFearChance + itembonuses.ResistFearChance;
-	if (IsOfClientBot()) {
-		resistchance += aabonuses.ResistFearChance;
-		if (aabonuses.Fearless == true) {
-			resistchance = 100;
-		}
-	}
-	if (spellbonuses.Fearless == true || itembonuses.Fearless == true) {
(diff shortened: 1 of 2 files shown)


zone/spells.cpp
```diff
@@ -5667,14 +5667,9 @@ int16 Mob::CalcResistChanceBonus()
...
-	if (IsOfClientBot()) {
-		resistchance += aabonuses.ResistFearChance;
-		if (aabonuses.Fearless == true) {
-			resistchance = 100;
-		}
-	}
-	if (spellbonuses.Fearless == true || itembonuses.Fearless == true) {
+	int resistchance = spellbonuses.ResistFearChance + itembonuses.ResistFearChance + aabonuses.ResistFearChance;
+
+	if (spellbonuses.Fearless == true || itembonuses.Fearless == true || aabonuses.Fearless == true) {
```
Akkadius, line 5672: `spellbonuses.Fearless == true` and `spellbonuses.Fearless` are the same

Akkadius (approved):

PR #4861, opened 2025-04-14
[Fix] Fix Object Name Init, User Refs, and Client Sync on Close

zimp-wow:
# Description

Couple minor fixes to object behaviors I found when working with the object API.

### Issue 1:  Fix uninitialized char* in object.cpp

Objects created via the API would show trash data in their name.  This change memsets the char array to zero in the constructors.

### Issue 2: Clear object user and user tradeskill object on reset.

A call to Object:Delete(true) would uninitialize a lot of fields in the object but leave lingering references to the user who was using it as well as a reference back to that object from the user.  This would cause a zone crash if the user continued to interact with the object.  I've added a 'Close' call to the top of ResetState to clear these references.

### Issue 3: Send clear object packet on Object::Close()

Calling Object:Close() would move any items in the object's inventory back into the user's inventory but not actually update the client that the object's inventory was now empty, creating the appearance that there were two items.

This change sends a ClearObject packet to the active user when close is called.

## Type of change

# Testing

Akk-stack with rof2 client.

Test script
```
function player_say(e)
        if e.message:findi("object") then
                local id = eq.create_ground_object_from_model("IT70_ACTORDEF", e.self:GetX() + 10, e.self:GetY() + 10, e.self:GetZ(), e.self:GetHeading(), 19);
                eq.debug("Object id: "..id);
        end

        if e.message:findi("objreset") then
                local id_str = string.gsub(e.message, "objreset ", "");
                local id = tonumber(id_str) or -1;
                eq.debug("Resetting obj "..id);
                if id > 0 then
                        local obj = eq.get_entity_list():GetObjectByID(id);
                        obj:Delete(true);
                end
        end

        if e.message:findi("objclose") then
                local id_str = string.gsub(e.message, "objclose ", "");
                local id = tonumber(id_str) or -1;
                eq.debug("Close obj "..id);
                if id > 0 then
                        local obj = eq.get_entity_list():GetObjectByID(id);
                        obj:Close();
                end
        end
end
```

Issue 1:

Before
image

After
image

Issue 2:

Before
image

After
image

Issue 3:

Original stack trace
image

Crash did not reproduce after fix applied.

# Checklist

### DIFF
--- a/zone/object.cpp
+++ b/zone/object.cpp
@@ -75,6 +75,8 @@ decay_timer(300000)
 		decay_timer.Disable();
 	}
 
+	memset(m_display_name, 0, sizeof(m_display_name));
+
 	respawn_timer.Disable();
 
 	// Set drop_id to zero - it will be set when added to zone with SetID()
@@ -122,6 +124,8 @@ decay_timer(300000)
 	// Set as much struct data as we can
 	memset(&m_data, 0, sizeof(Object_Struct));
 
+	memset(m_display_name, 0, sizeof(m_display_name));
+
 	m_data.heading = heading;
 	m_data.z       = z;
 	m_data.zone_id = zone->GetZoneID();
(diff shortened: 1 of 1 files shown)


Akkadius (approved):
Thank you sir

PR #4867, opened 2025-04-21
[Feature] Add Character:TradeskillUpMinChance rule

zrix-eq:
# Description

The existing Character:TradeskillUp* rules can be misleading.  There's two checks to see whether you increase your skill in a particular trade, and the second check does not have any sort of rule associated with it.

This PR adds Character:TradeskillUpMinChance, which is the minimum chance to gain a tradeskill, and it affects both the first and second stage checks.  This would allow servers (specifically THJ) to make it easier on players to do tradeskills.

I originally submitted this PR to THJ directly and was asked to consider upstreaming it.

## Type of change

# Testing

Clients tested: akkstack/ROF2

# Checklist

### DIFF
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -156,6 +156,7 @@ RULE_REAL(Character, TradeskillUpPottery, 4.0, "Pottery skillup rate adjustment.
 RULE_REAL(Character, TradeskillUpResearch, 1.0, "Research skillup rate adjustment. Lower is faster")
 RULE_REAL(Character, TradeskillUpTinkering, 2.0, "Tinkering skillup rate adjustment. Lower is faster")
 RULE_REAL(Character, TradeskillUpTailoring, 2.0, "Tailoring skillup rate adjustment. Lower is faster")
+RULE_REAL(Character, TradeskillUpMinChance, 2.5, "Determines the minimum percentage chance to gain a skill increase from a tradeskill.  Cannot go below 2.5")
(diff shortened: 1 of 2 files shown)


common/ruletypes.h
```diff
@@ -156,6 +156,7 @@ RULE_REAL(Character, TradeskillUpPottery, 4.0, "Pottery skillup rate adjustment.
 RULE_REAL(Character, TradeskillUpResearch, 1.0, "Research skillup rate adjustment. Lower is faster")
 RULE_REAL(Character, TradeskillUpTinkering, 2.0, "Tinkering skillup rate adjustment. Lower is faster")
 RULE_REAL(Character, TradeskillUpTailoring, 2.0, "Tailoring skillup rate adjustment. Lower is faster")
+RULE_REAL(Character, TradeskillUpMinChance, 2.5, "Determines the minimum percentage chance to gain a skill increase from a tradeskill.  Cannot go below 2.5")
```
zrix-eq, line 159: I was told that adding a RULE_REAL adds a migration automatically, but I'm not seeing that behavior locally for some reason.  If this is not the case, let me know and I can add a migration and bump the DB version.

zrix-eq: I also chose a default of 2.5 to mimic the previous minimum skill up chance, so all other existing servers would be unaffected.

Akkadius (approved):

PR #4879, opened 2025-05-08
[Fix] Prevent Ranged Attack from being triggered at arbitrary rate

catapultam-habeo:
# Description

If `OP_CombatAbility` triggers a Ranged Attack, it can bypass the ranged attack timer. This is normally rate limited by client, though client does calculate the speed incorrectly at high haste values - allowing the 'ranged attack' button to be pressed faster than the autofire timer. However, this could be more seriously exploited by a MQ2 plugin which sends the appropriate packet at any arbitrary rate.

## Type of change

# Testing

Deployed to THJ

Clients tested: 
ROF2

# Checklist

### DIFF
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -6677,7 +6677,9 @@ void Client::SetAttackTimer()
 			else
 				speed = static_cast<int>(speed + ((hhe / 100.0f) * delay));
 		}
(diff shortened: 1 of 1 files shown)


Akkadius (approved):
LGTM. Thank you sir

PR #4903, opened 2025-05-20
[Performance] Auto Idle / AFK

Akkadius:
# Description

This is something that has been rolled out on THJ for 2 months to reduce significant packet load that are largely needless especially in hub / idle zones. 

**Why this Change?** Part of what we see in a critical hub zone such as Bazaar, is extremely high CPU, lag, causing a bad experience for other players trying to zone into the zone and function at all. This change aims to drastically reduce packet spam in the zone, keeping the zone lean and leading to a much better player experience when using the Bazaar. When a player with 4 pets is running in circles in the zone, it sends enormous amounts of position updates to X number of players in the zone, even those who are idle and don't need all of these updates.

When clients go AFK they stop receiving packet updates for movement. When they go un-AFK they get synchronized with all the clients around them.

---

Consider this - the EverQuest client requires that you send a clients position update to other clients every 12 seconds otherwise clients disappear. This is very inefficient. For this reason we send client positions every 12 seconds at most unless they are moving then we obviously send far more frequent updates. Whereas, NPC's don't need this kind of updating - if an NPC didn't change positions - we never send another position update for the NPC to any client beyond the initial spawn position packet creation.

## Quick Math Example

If you take a 200 player zone -

200 players sending updates every 12 seconds to 200 clients.

200 players * 5 (updates a minute minimum) / 60 (1 minute) = 16.6 * 200 clients = 3,333 packets per second in the **best** case scenario. 

When a client goes into idle mode, they no longer receive position updates and no longer get tons of needless chatter, this reduces strain on the zoneserver and ultimately makes a better experience for everyone else.

With this change - simply - if 90% of your clients are idling, you reduce all of the packet strain by 90%.

---

**Changelog**

The **Idle** and **AFK** systems have been separated to improve player experience and reduce unnecessary network load.

- In non-combat, players become idle after 1 minute and AFK after 10 minutes.
- In combat, players become idle after 10 minutes and AFK after 30 minutes.

### Key Changes:
- **Idle** now reduces client position updates to improve performance in crowded zones like the Bazaar.
- **AFK** is now fully independent of idle and manual `/afk`. Automatic AFK will not override or be overridden by manual AFK status.
- Manual `/afk` remains active until toggled off and is unaffected by automatic behavior.
- Automatic AFK messages no longer interfere with manual or automatic AFK or idle states.
- The following actions now reset idle and AFK timers:
  - Moving items
  - Casting spells
  - Using `/autoattack`
  - Receiving messages
- **Trade** (start or finish) now forces a position sync for both parties.
- **Buyers** and **Traders** are excluded from automatic AFK entirely.

### New Rules

```cpp
RULE_BOOL(Character, EnableAutoAFK, true, "Enable or disable the auto AFK feature, cuts down on packet spam")
RULE_BOOL(Character, AutoIdleFilterPackets, true, "Enable or disable filtering packets when auto AFK is enabled, heavily cuts down on packet spam in zones with lots of players")
RULE_INT(Character, SecondsBeforeIdleCombatZone, 600, "Seconds before a player is considered idle in combat zones (600 = 10 minutes)")
RULE_INT(Character, SecondsBeforeIdleNonCombatZone, 60, "Seconds before a player is considered idle in non-combat zones (60 = 1 minute)")
RULE_INT(Character, SecondsBeforeAFKCombatZone, 1800, "Seconds before a player is considered AFK in combat zones (1800 = 30 minutes)")
RULE_INT(Character, SecondsBeforeAFKNonCombatZone, 600, "Seconds before a player is considered AFK in non-combat zones (600 = 10 minutes)")
```

## Type of change

# Testing

Been testing on THJ for about 2 months now with over 4,000 players. Feature is stable.

# Checklist


zone/client.cpp
```diff
@@ -12657,7 +12683,7 @@ void Client::SendTopLevelInventory()
 	}
 }
 
-void Client::CheckSendBulkNpcPositions()
+void Client::CheckSendBulkNpcPositions(bool force)
```
joligario, line 12686: Recommend force_update

joligario (approved):
Looks great!

PR #4909, opened 2025-06-02
[Feature] Load contents of script_constants table into global lua state.

zimp-wow:
# Description

New content often relies on constant values that may change from initial development to deployment.  Things such as npc type ids, spell ids, etc.  This feature introduces a 'script_constants' table that will have its contents exposed into the Lua global namespace as a 'DBConstants' table.

At a minimum this helps makes the content more portable but the end-goal here is to allow ambitious content writers to create setup routines that can utilize the scripting DB interface to seed the database with new data and then store the resulting constant values in this table to get their content working with no manual effort.

## Type of change

# Testing

image

Based on the table contents above the following constants should be present in 'Blackburrow' version 0
DBConstants.test.value1 = 'hello world'
DBConstants.test2.value1 = 'hello world 2'
DBConstants.test4.value1 = 'hello world 4'

Actual:
image

'Blackburrow' version 255 expected:
DBConstants.test.value1 = 'hello world'
DBConstants.test2.value1 = 'hello world 2'
DBConstants.test3.value1 = 'hello world 3'

Actual:
image

'Crushbone' version 0 expected:
DBConstants.test.value1 = 'hello world'
DBConstants.test.value1 = 12345

Actual:
image

# Checklist

### DIFF
--- a/common/database/database_update_manifest.cpp
+++ b/common/database/database_update_manifest.cpp
@@ -7112,6 +7112,24 @@ ADD COLUMN `first_login` int(11) UNSIGNED NOT NULL DEFAULT 0 AFTER `xtargets`;
 )",
 		.content_schema_update = false
 	},
+	ManifestEntry{
+		.version     = 9324,
+		.description = "2025_06_01_script_constants.sql",
+		.check       = "SHOW TABLES LIKE 'script_constants'",
+		.condition   = "empty",
+		.match       = "",
+		.sql         = R"(
+CREATE TABLE `script_constants` (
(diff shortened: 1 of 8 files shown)


zimp-wow:
Looks like the snapshot on June 2nd had a db_version value of 9324.  Today's snapshot has the expected value of 9323.

Akkadius:
> Looks like the snapshot on June 2nd had a db_version value of 9324. Today's snapshot has the expected value of 9323.

Yep I fixed it earlier

PR #4911, opened 2025-06-11
[Corpses] Add corpse entity variables to DB

fryguy503:
# Description

- When corpses have entity variables add/remove/modified it will update the database accordingly.
- Corpse loading will pull the database values if they exist.

## Type of change

# Testing

image
image

Clients tested: RoF2

# Checklist

### DIFF
--- a/zone/corpse.cpp
+++ b/zone/corpse.cpp
@@ -32,7 +32,9 @@
 #include "../common/repositories/character_corpse_items_repository.h"
 #include <iostream>
 #include "queryserv.h"
+#include "../common/json/json.hpp"
 
+using json = nlohmann::json;
 
 extern EntityList           entity_list;
 extern Zone                *zone;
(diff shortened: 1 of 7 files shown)


zone/corpse.cpp
```diff
@@ -2428,9 +2445,34 @@ Corpse *Corpse::LoadCharacterCorpse(
 	c->m_become_npc                = false;
 	c->m_consented_guild_id        = cc.guild_consent_id;
 
+	if (!cc.entity_variables.empty() && cc.entity_variables != "null") {
```
Akkadius: We have a `Strings::IsValidJson` that we can use here

zone/corpse.cpp
```diff
@@ -2428,9 +2445,34 @@ Corpse *Corpse::LoadCharacterCorpse(
...
+		json j = json::parse(cc.entity_variables, nullptr, false);
```
Akkadius: try / catch needs to be wrapped around here because exception throws can stall the process

Akkadius (approved):

PR #4912, opened 2025-06-11
[Player Events] Don't Write Empty Data

Akkadius:
# Description

This PR saves a bunch of space in player event data because we recently started writing augment ids and names in a lot of events even if they are zero a majority of the time. This adds up tremendously for certain events.

See this simple [gist](https://gist.github.com/Akkadius/35da017c968cd214907a33521814508e) as a great example

## Type of change

- [x] Optimization

# Testing

Below is an event where we record one augment on summon and an item summon with no aug. (Works both ways)

image

image

# Checklist

### DIFF
--- a/common/events/player_events.h
+++ b/common/events/player_events.h
@@ -7,8 +7,16 @@
 #include "../rulesys.h"
 #include "../repositories/player_event_logs_repository.h"
 
-namespace PlayerEvent {
+#define CEREAL_NVP_IF_NONZERO(ar, name) \
+if ((name) != 0) ar(cereal::make_nvp(#name, name))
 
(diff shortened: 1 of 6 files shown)


common/events/player_events.h
```diff
@@ -295,7 +362,7 @@ namespace PlayerEvent {
 				CEREAL_NVP(augment_2_id),
 				CEREAL_NVP(augment_2_name),
 				CEREAL_NVP(augment_3_id),
-				CEREAL_NVP(augment_1_name),
+				CEREAL_NVP(augment_3_name),
```
joligario, line 365: Nice find!

joligario (approved):
LGTM!

PR #4917, opened 2025-06-11
[Databuckets] Nested Caching

Akkadius:
# Description

This PR dramatically improves databuckets performance in two ways

* **Nested Buckets** - Even if a databucket was cached, the nested databucket access is not, which means the server has to recursively walk a json tree to find the nested value. This is computationally very expensive in hot paths. This is now cached in this PR
* **Scoped Bucket Cache By Default** - Before, if you didn't have a bucket in the cache as a miss, we'd hit the database once before it was inserted as a miss and then the second ask was cached. Now, since we pre-load all cached scoped keys, we assume that if the entity doesn't have a key in the cache, it just simply doesn't have the key. This leads to a 1,000x speed improvement just in pure access speed and we also avoid potentially tons of database hits depending on a servers access patterns.

**Benchmark - 5,000 iterations (1,000x speedup)**

* Before - 0.279923 seconds
* After - 0.000295002

## Type of change

- [x] Performance

# Testing

```
===========================================
⚙️> Running DataBuckets Tests...
===========================================

[✅] Basic Key-Value Set/Get PASSED
[✅] Overwriting a Key PASSED
[✅] Deleting a Key PASSED
[✅] Setting a Key with an Expiration PASSED
[✅] Ensure Expired Key is Deleted PASSED
[✅] Cache Read/Write Consistency PASSED
[✅] Cache Clears on Key Deletion PASSED
[✅] Setting a Full JSON String PASSED
[✅] Overwriting JSON with a Simple String PASSED
[✅] Deleting Non-Existent Key PASSED
[✅] Basic Key-Value Set/Get PASSED
[✅] Nested Key Set/Get PASSED
[✅] Prevent Overwriting Objects PASSED
[✅] Delete Nested Key PASSED
[✅] Delete Parent Key PASSED
[✅] Expiration Ignored for Nested Keys PASSED
[✅] Cache Read/Write Consistency PASSED
[✅] Cache Clears on Parent Delete PASSED
[✅] Set and Retrieve Full JSON Structure PASSED
[✅] Delete Nested Key within JSON PASSED
[✅] Ensure Object Protection on Overwrite Attempt PASSED
[✅] Deleting Non-Existent Key Doesn't Break Existing Data PASSED
[✅] Get nested key value PASSED
[✅] Get nested key value deep PASSED
[✅] Retrieve Nested Key from Plain String PASSED
[✅] Store and Retrieve JSON Array PASSED
[✅] Retrieve Non-Existent Nested Key PASSED
[✅] Overwriting Parent Key Deletes Children PASSED
[✅] Store and Retrieve Empty JSON Object PASSED
[✅] Store and Retrieve JSON String PASSED
[✅] Deeply Nested Key Retrieval PASSED
[✅] Setting a nested key with an expiration protection test PASSED
[✅] Scoped Missing Key Returns Empty (Skips DB) PASSED
[✅] Nested Scoped Key Miss Returns Empty (Skips DB) PASSED
[✅] Initial Cache Miss Returns Empty PASSED
[✅] Overwrite After Cache Miss Works PASSED
[✅] Missed Nested Key Returns Empty PASSED
[✅] Nested Set After Miss Works PASSED
[✅] Cold Cache Scoped Key Returns Empty (Due to Skip DB) PASSED
[✅] Post-BulkLoad Scoped Key Returns Value PASSED
[✅] Post-BulkLoad Nested Scoped Key Returns Value PASSED
[✅] Post-Delete Nested Scoped Key Returns Empty PASSED

===========================================
✅ All DataBucket Tests Completed!
===========================================
```

# Checklist

### DIFF
--- a/common/repositories/character_data_repository.h
+++ b/common/repositories/character_data_repository.h
@@ -167,6 +167,30 @@ public:
 
 		return zone_player_counts;
 	}
+
+	static std::vector<uint32_t> GetCharacterIDsByAccountID(
+		Database& db,
+		uint32_t  account_id
+	)
+	{
+		std::vector<uint32_t> character_ids;
+
+		auto query = fmt::format(
+			"SELECT id FROM character_data WHERE account_id = {} AND deleted_at IS NULL",
+			account_id
+		);
+
+		auto results = db.QueryDatabase(query);
+		if (results.Success()) {
+			for (auto row : results) {
+				if (row[0]) {
+					character_ids.push_back(static_cast<uint32_t>(std::stoul(row[0])));
+				}
+			}
+		}
+
+		return character_ids;
+	}
 };
 
 #endif //EQEMU_CHARACTER_DATA_REPOSITORY_H
--- a/world/client.cpp
+++ b/world/client.cpp
@@ -36,6 +36,7 @@
 #include "../common/shareddb.h"
 #include "../common/opcodemgr.h"
 #include "../common/data_verification.h"
+#include "../common/data_bucket.h"
 
 #include "client.h"
 #include "worlddb.h"
@@ -135,6 +136,8 @@ Client::Client(EQStreamInterface* ieqs)
 }
 
 Client::~Client() {
+	ClearDataBucketsCache();
+
 	if (RunLoops && cle && zone_id == 0)
 		cle->SetOnline(CLE_Status::Offline);
 
@@ -477,6 +480,8 @@ bool Client::HandleSendLoginInfoPacket(const EQApplicationPacket *app)
 	LogClientLogin("Checking authentication id [{}]", id);
 
 	if ((cle = client_list.CheckAuth(id, password))) {
+		LoadDataBucketsCache();
+
 		LogClientLogin("Checking authentication id [{}] passed", id);
 		if (!is_player_zoning) {
 			// Track who is in and who is out of the game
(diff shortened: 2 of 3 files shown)


common/data_bucket.cpp
```diff
@@ -19,6 +19,37 @@ extern WorldDatabase database;
...
+		"account_id:{}|character_id:{}|npc_id:{}|bot_id:{}|zone_id:{}|instance_id:{}|top_key:{}|full_key:{}",
+		k.account_id, k.character_id, k.npc_id, k.bot_id, k.zone_id, k.instance_id,
+		Strings::Split(full_key, NESTED_KEY_DELIMITER).front(),
+		full_key
+	);
+}
+
+static std::string MakeNestedCacheKeyPrefix(const DataBucketKey &k, const std::string &top_key) {
+	return fmt::format(
+		"account_id:{}|character_id:{}|npc_id:{}|bot_id:{}|zone_id:{}|instance_id:{}|top_key:{}|",
```
joligario, line 37: Looks like unintentional extra pipe at end of string

KimLS (approved):

joligario (approved):

common/data_bucket.cpp
Akkadius: Its intentional

PR #4964, opened 2025-07-09
Suppress Buff SPA (527)

zimp-wow:
# Description

This was something I was asked to send upstream.  It adds the ability to suppress a player's buff for a set amount of time, restoring it when the effect ends.  There are three ways a server can leverage this effect:

1. Enable the 'SuppressDispels' rule

image

This will cause CancelMagic, DispelDetrimental and DispelBeneficial SPAs to automatically use the suppression effect instead when the target is a player/player's pet and the caster is an NPC.

Players will still be able to dispel themselves and enemies normally.

When this rule is disabled (default) dispels should behave the same as always.

2. New overloaded API method for 'BuffFadeBySlot'

Scripts can now pass two optional parameters to this API to force a suppression effect on a specific buff slot.  The third parameter is true/false to enable the suppression with the fourth being the number of tics the effect should last.

3. Use the new SPA 527 in a spell.

This SPA picks a random buff on the target and applies the suppression effect, refreshing the timer if its already suppressed.  The base_value represents the number of tics the effect should last.

Downstream users will likely want to customizes these behaviors somewhat depending on what they desire.

## Type of change

- [x] This change requires a documentation update
    Spire missing support for new SPA

# Testing

https://github.com/user-attachments/assets/305a4903-2647-464b-b2c1-d91d740f565b

Tested with base RoF client and latest eqemu/master branch.

Test Cases:
- Suppressed buffs lose their effects and restore when suppression end
- Levitate effect properly restores after suppression
- Lua buff fade by slot works
- Perl buff fade by slot works
- Suppress on charm breaks charm
- Suppress on pet works
- Dispel on self clears buffs (with rule on or off)
- Dispel on npc clears buffs
- Custom spell with SPA 527 works
- DispelBeneficial works
- DispelDetrimental works

# Checklist

### DIFF
--- a/common/ruletypes.h
+++ b/common/ruletypes.h
@@ -546,6 +546,9 @@ RULE_INT(Spells, PointBlankAOEMaxTargets, 0, "Max number of targets a Point-Blan
 RULE_INT(Spells, DefaultAOEMaxTargets, 0, "Max number of targets that an AOE spell which does not meet other descriptions can cast on. Set to 0 for no limit.")
 RULE_BOOL(Spells, AllowFocusOnSkillDamageSpells, false, "Allow focus effects 185, 459, and 482 to enhance SkillAttack spell effect 193")
 RULE_STRING(Spells, AlwaysStackSpells, "", "Comma-Seperated list of spell IDs to always stack with every other spell, except themselves.")
+RULE_BOOL(Spells, SuppressDispels, false, "Swaps 'cancel magic' SPA logic with SuppressBuff SPA (527).")
+RULE_INT(Spells, SuppressDispelsTime, 6, "Number of tics that dispelled buffs will be suppressed for")
+RULE_INT(Spells, SuppressDebuffSpellID, 21840, "Spell ID to send to client when a spell is supprssed.  21840 = 'Suppression Field'")
 RULE_CATEGORY_END()
 
 RULE_CATEGORY(Combat)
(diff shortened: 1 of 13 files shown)


Valorith:
Would LOVE to see this merged into master :D I know its been running on THJ for a bit now and has been great!

KayenEQ:
There is a live spell effect that already does something similar. SPA 382, which negates a specific spell effect bonus or even AA or item bonuses. While not the exact same functionality it does exist and has a lot of flexibility and is fully functional.

SPA 527 also already exists as a live spell effect. It is not currently implemented but we shouldn't use that number if we merge this.

This a community decision, but we have never implemented new custom spell effects into the source code before. If we start I am not sure where the line is drawn for that. I could bloat the entire source code in a few days with a 100 new effects if we wanted to start doing that. Personally my opinion is I never thought that was a smart idea and think that custom SPAs should be reserved for customs source code projects.

zimp-wow:
Can change the SPA ID or just drop it altogether.  It's not a critical part of this change.

I'm not really planning on aggressively pursuing this PR.  I'm fine with any review fixes or changes but if it turns into a big debate I'd prefer to just close it.  Downstream can still merge it in on their own from my fork or by using the 'pull/4964/head' ref on the eqemu repo.

Valorith:
Hoping this gets on master. Compiling your own binaries is a pain compared to using release.

PR #4968, opened 2025-07-19
[Fix] Fix crash from lingering spawn2 reference in NPCs.

zimp-wow:
# Description

In the case of a Zone::Repop() the spawn2 linked list would be cleared, destroying the spawn2 object. However a charmed NPC would survive the repop and continue to hold onto their reference to the spawn2 object, resulting in a variety of crashes.

I waffled around different ways to solve this and ultimately decided to remove the reference and replace it with a real-time lookup of the current spawn2 object in spawn2_list.  I do have reservations about the performance cost of this, especially in the case of line 523 in entity.cpp.  Might be a point for additional discussion.

## Type of change

# Testing

Original Reproduction Steps:

1. Charm enemy (#cast 2932)
2. #repop
3. Kill charmed enemy or click off charm
4. #repop

Second repop usually crashes/locks up zone.

Test Cases:

1. Verify repro steps stop working (multiple attempts) - Pass
2. Verify npc respawn timers are correctly set still on death - Pass
3. Verify gm commands still function - Pass

# Checklist

### DIFF
--- a/zone/entity.cpp
+++ b/zone/entity.cpp
@@ -520,7 +520,7 @@ void EntityList::MobProcess()
 
 			old_client_count = numclients;
 
-			Spawn2* s2 = mob->CastToNPC()->respawn2;
+			Spawn2* s2 = mob->CastToNPC()->GetSpawn();
 
 			// Perform normal mob processing if any of these are true:
 			//	-- zone is not empty
--- a/zone/attack.cpp
+++ b/zone/attack.cpp
@@ -2611,8 +2611,11 @@ bool NPC::Death(Mob* killer_mob, int64 damage, uint16 spell, EQ::skills::SkillTy
 
 	safe_delete(app);
 
-	if (respawn2) {
-		respawn2->DeathReset(1);
+	if (respawn2_id) {
+		auto respawn2 = GetSpawn();
+		if (respawn2) {
+			respawn2->DeathReset(1);
+		}
 	}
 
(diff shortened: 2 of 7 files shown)


zone/entity.cpp
```diff
@@ -520,7 +520,7 @@ void EntityList::MobProcess()
 
 			old_client_count = numclients;
 
-			Spawn2* s2 = mob->CastToNPC()->respawn2;
+			Spawn2* s2 = mob->CastToNPC()->GetSpawn();
```
Akkadius, line 523: This is gonna be a big concern in a hot path hitting the entity list in your ternary

Akkadius:
Looking at this I'd prefer that we properly book keep at the event points where references become inconsistent rather than doing expensive calculations realtime

zimp-wow:
My main concern with trying to retain the reference is it felt like a landmine for future regressions.  It's a pretty messy problem when it manifests too as tracing it backwards from the stack traces was a lot of false leads.

I could probably cut down on a lot of the processing cost in that loop by wrapping the call in a closure that only gets called when the conditional below gets that far.  That would limit the check to only trigger in zones with no clients in them.

Or another approach would be to use a more efficient structure to search against, be it a hash or an ordered list.

Feels like for long term maintainability it would be easier to just solve the performance cost than try to track ownership and validity of the spawn2 reference.

PR #5000, opened 2025-08-29
[Constants] Change Race Changes to Race Namespace

Kinglykrab:
# Description
- Converts existing race constants to use `Race`, `RaceAppearance`, `RaceBitmask`, and `RaceIndex` namespaces.

# Checklist

### DIFF
--- a/zone/npc.cpp
+++ b/zone/npc.cpp
@@ -2901,7 +2901,7 @@ void NPC::DoNPCEmote(uint8 event_, uint32 emote_id, Mob* t)
 	// Mob Variables
 	Strings::FindReplace(processed, "$mname", GetCleanName());
 	Strings::FindReplace(processed, "$mracep", GetRacePlural());
-	Strings::FindReplace(processed, "$mrace", GetPlayerRaceName(GetRace()));
+	Strings::FindReplace(processed, "$mrace", GetRaceIDName(GetRace()));
 	Strings::FindReplace(processed, "$mclass", GetClassIDName(GetClass()));
 	Strings::FindReplace(processed, "$mclassp", GetClassPlural());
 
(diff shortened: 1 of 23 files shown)


zone/npc.cpp
```diff
@@ -2901,15 +2901,15 @@ void NPC::DoNPCEmote(uint8 event_, uint32 emote_id, Mob* t)
 	// Mob Variables
 	Strings::FindReplace(processed, "$mname", GetCleanName());
 	Strings::FindReplace(processed, "$mracep", GetRacePlural());
-	Strings::FindReplace(processed, "$mrace", GetPlayerRaceName(GetRace()));
+	Strings::FindReplace(processed, "$mrace", GetRaceIDName(GetRace()));
```
joligario, line 2904: Unsure about the change to RaceID vs. just Race

zone/npc.cpp
```diff
@@ -2901,15 +2901,15 @@ void NPC::DoNPCEmote(uint8 event_, uint32 emote_id, Mob* t)
...
 	Strings::FindReplace(processed, "$mclass", GetClassIDName(GetClass()));
 	Strings::FindReplace(processed, "$mclassp", GetClassPlural());
 
 	// Target Variables
 	Strings::FindReplace(processed, "$name", t ? t->GetCleanName() : "foe");
 	Strings::FindReplace(processed, "$class", t ? GetClassIDName(t->GetClass()) : "class");
 	Strings::FindReplace(processed, "$classp", t ? t->GetClassPlural() : "classes");
-	Strings::FindReplace(processed, "$race", t ? GetPlayerRaceName(t->GetRace()) : "race");
+	Strings::FindReplace(processed, "$race", t ? GetRaceIDName(t->GetRace()) : "race");
```
joligario, line 2912: Unsure about the change to RaceID vs. just Race

joligario (approved):

zone/npc.cpp
Kinglykrab: We had a separate method that used `GetRaceIDName` under the hood just for player races, so it was removed.

zone/npc.cpp
Kinglykrab: We had a separate method that used `GetRaceIDName` under the hood just for player races, so it was removed.

PR #5059, opened 2026-04-08
Add Multiple Mercenary Hire Functionality

xJeris:
# Description

- Added a rule allowing servers to lock number of mercenary slots. This is enforced server-side, not by the client. Max mercenaries tested for is 6. I did not investigate AA or Achievement Merc slots (if they even exist) to see if it allowed higher values.

- Allow for the hiring of multiple mercenaries, and the ability to switch between them with the built in switch button included in the interface.

- Added and noted the opcode for the clients switch button

- Excluded pre-RoF clients from the feature change.

## Type of change

- [ X] New feature (non-breaking change which adds functionality)

# Testing

- Tested build
- Tested the hiring, suspend, unsuspend, and dismissal of mercenary
- Tested rotating between multiple mercenary
- Tested locking the number of mercenary slots via server-side rule
- Posted a short video of the feature working in Discord (https://discord.com/channels/1421535813031821376/1421896979952308395/1491223267787280596)

Clients tested: 

- RoF2 client only. I don't believe prior clients allowed this feature.

# Checklist


KimLS:
Will look at this soon, is just too late for me to give it the attention it deserves tonight.

submodules/websocketpp
Knightly1, line 1: Is this intentional? It's not mentioned, so I was curious.

zone/client_packet.cpp
```diff
@@ -10687,6 +10685,21 @@ void Client::Handle_OP_MercenaryHire(const EQApplicationPacket *app)
 			return;
 		}
 
+		// Suspend active merc if one exists before hiring into a new slot
+		Merc* current_merc = GetMerc();
+		if (current_merc) {
+			current_merc->Suspend();
+			SetMerc(nullptr);
```
Knightly1, line 10692: This might be just my unfamiliarity with the codebase, so take it with a grain of salt but -- Is this right? Wouldn't using nullptr here mean that GetFirstFreeMercSlot() returns this same slot? So you'd overlap on the DB entry for the new merc?

It looks like you account for that in MercenarySwitch, but this behavior is different (again, might be I'm just not understanding the flow).

common/ruletypes.h
```diff
@@ -255,6 +255,7 @@ RULE_BOOL(Mercs, AllowMercSuspendInCombat, true, "Allow merc suspend in combat")
 RULE_BOOL(Mercs, MercsIgnoreLevelBasedHasteCaps, false, "Ignores hard coded level based haste caps.")
 RULE_INT(Mercs, MercsHasteCap, 100, "Haste cap for non-v3(over haste) haste")
 RULE_INT(Mercs, MercsHastev3Cap, 25, "Haste cap for v3(over haste) haste")
+RULE_INT(Mercs, MaxMercSlots, 6, "Maximum number of mercenary slots per character (max 6)")
```
Knightly1, line 258: 6 is a magic number here, yeah? This is actually MAXMERCS?

zone/client.cpp
```diff
@@ -7940,75 +7940,125 @@ void Client::SendWebLink(const char *website)
...
 	if (ClientVersion() >= EQ::versions::ClientVersion::RoF) {
-		auto outapp = new EQApplicationPacket(OP_MercenaryDataUpdate, sizeof(MercenaryDataUpdate_Struct));
+		// Count owned mercs across all slots
+		uint32 mercCount = GetNumberOfMercenaries();
+		if (mercCount == 0) {
+			SendClearMercInfo();
+			return;
+		}
+
+		uint32 packetSize = sizeof(MercenaryDataUpdate_Struct) + (mercCount * sizeof(MercenaryData_Struct));
```
Knightly1, line 7953: This looks like it's overallocating based on the MercenaryDataUpdate_Struct. Is it not?

common/ruletypes.h
xJeris: correct. i don't think putting a higher value would actually do anything as the client only shows 6 spots.

submodules/websocketpp
xJeris: is this cmake version change? was not intentional to get pushed up.

zone/client_packet.cpp
xJeris: it looks like i account for it as the merc is saved to db and reloaded. but, it also looks like i dont need the nullptr, i could just set the id to 0 or something instead.

zone/client.cpp
xJeris: yea, i think removing the addition is all that is needed . i can test that change to be sure i didnt need it.

submodules/websocketpp
Knightly1: It means that the submodule was updated to a different commit and then "git add" for the submodule later. If you've got an editor or IDE (or you just did `git add .`) it would have been changed. I'd revert this one out of your change set if it wasn't intentional (and it should probably just be a vcpkg anyway).

common/ruletypes.h
Knightly1: 6 is fine for the maxslots on the rule default, but for the rule text I think you want MAXMERCS (since the max isn't actually 6)

KimLS (approved):

PR #5060, opened 2026-04-11
fix: correct off-by-one in GetSpellLevel for Berserker class

### LINKED ISSUE #5056: Off-by-one error causing quest::get_spell_level to return 255 for berserkers.
Full writeup here:  https://deepwiki.com/search/questgetspelllevel-quest-api-s_08aec7f7-c7ac-4363-bcc4-271b25c2e841?mode=deep

Observed quest API always returning level 255 for common berserker disciplines.

ltroylove:
Fixes #5056

## What

`GetSpellLevel` in `common/spdat.cpp` returns `UINT8_MAX` (255) for any Berserker (class ID 16) due to an off-by-one in the boundary check.

```cpp
// Before
if (class_id >= Class::PLAYER_CLASS_COUNT) {  // 16 >= 16 → true → early return
    return UINT8_MAX;
}

// After
if (class_id < Class::Warrior || class_id > Class::Berserker) {
    return UINT8_MAX;
}
```

## Why

`Class::Berserker == 16` and `Class::PLAYER_CLASS_COUNT == 16`, so `>=` incorrectly rejects the highest valid class. The array access below (`classes[class_id - 1]` → `classes[15]`) is valid — the bug is only in the guard.

The new check uses named constants to make the valid range explicit and also covers `class_id == 0`, which would underflow the index.

The companion function `GetSpellMinimumLevel` uses `i < Class::PLAYER_CLASS_COUNT` (0-indexed loop through all 16 slots) and correctly includes Berserker, confirming the array is intended to cover all 16 classes.

The quest API wrappers (`embparser_api.cpp`, `lua_general.cpp`) pass the result through a MaxLevel check — since 255 > MaxLevel, the bug compounds and the scripting layer receives UINT8_MAX a second time, which is why the reported symptom is often 0 rather than 255.

## Note

### DIFF
--- a/common/spdat.cpp
+++ b/common/spdat.cpp
@@ -999,7 +999,7 @@ uint8 GetSpellLevel(uint16 spell_id, uint8 class_id)
 		return UINT8_MAX;
 	}
 
-	if (class_id >= Class::PLAYER_CLASS_COUNT) {
+	if (class_id < Class::Warrior || class_id > Class::Berserker) {
 		return UINT8_MAX;
 	}
 
--- a/common/spdat.cpp
+++ b/common/spdat.cpp
@@ -999,7 +999,7 @@ uint8 GetSpellLevel(uint16 spell_id, uint8 class_id)
 		return UINT8_MAX;
 	}
 
-	if (class_id < Class::Warrior || class_id > Class::Berserker) {
+	if (class_id < Class::Warrior || class_id > Class::PLAYER_CLASS_COUNT) {
 		return UINT8_MAX;
 	}
 


common/spdat.cpp
```diff
@@ -999,7 +999,7 @@ uint8 GetSpellLevel(uint16 spell_id, uint8 class_id)
 		return UINT8_MAX;
 	}
 
-	if (class_id >= Class::PLAYER_CLASS_COUNT) {
+	if (class_id < Class::Warrior || class_id > Class::Berserker) {
```
Knightly1, line 1002: It is unlikely this is going to change for the lifetime of this project, but just from a good coding practice standpoint, I would keep PLAYER_CLASS_COUNT here:

```cpp
if (class_id < Class::Warrior || class_id > PLAYER_CLASS_COUNT) {
```

Otherwise the next class would have the same issue. (really not a major concern)

ltroylove: Good call — updated to use `Class::PLAYER_CLASS_COUNT` as the upper bound. Thanks for the feedback!

dannuic:
Looks good to me after the change (had to figure out why less than warrior went in there, but I see now)

KimLS (approved):

PR #5061, opened 2026-04-11
fix: include base skill damage in FlyingKick, Kick, RoundKick, and Bash

ltroylove:
Fixes #5035

## What

`GetBaseSkillDamage()` fetches each skill's configured base damage into `base` at line 35 via `EQ::skills::GetBaseDamage()` (which reads the `Combat:*BaseDamage` rules), but four skills discard it entirely and return only `(ac_bonus + skill_bonus)`. The same bug exists in both `Mob::GetBaseSkillDamage` (`zone/special_attacks.cpp`) and `Bot::GetBaseSkillDamage` (`zone/bot.cpp`).

| Skill | Fix |
|-------|-----|
| `SkillFlyingKick` | `return base + static_cast<int>(ac_bonus + skill_bonus)` |
| `SkillKick` | same |
| `SkillRoundKick` | same (Mob only — Bot has no RoundKick case) |
| `SkillBash` | same |

The working cases — `DragonPunch`, `EagleStrike`, `TigerClaw`, `Frenzy` — all correctly return `base` in both implementations.

## Why this is clearly a bug for Kick and RoundKick

The `Kick`/`RoundKick` case in `special_attacks.cpp` increments `base` at skill level 75 and again at 175, then throws it away. There is no reason to modify a value you're not going to use — those increments only make sense if `base` was intended to be part of the return.

## Gameplay impact — please review

This fix makes the `Combat:*BaseDamage` rules functional for the first time. With default rule values the damage changes per skill are:

| Skill | Default base | Approx damage change at max skill |
|-------|-------------|----------------------------------|
| `Kick` | 3 (+2 from skill thresholds = 5) | +5 |
| `RoundKick` | 5 (+2 from skill thresholds = 7) | +7 |
| `Bash` | 2 | +2 |
| `FlyingKick` | 25 | +25 |

`FlyingKick` has the largest impact due to its default rule value of 25. If that default was set before the bug existed and never reflected actual intended damage, it may need to be re-evaluated alongside this fix. We've documented the impact here and leave that call to the maintainers.

## Note

### DIFF
--- a/zone/bot.cpp
+++ b/zone/bot.cpp
@@ -5175,7 +5175,7 @@ int Bot::GetBaseSkillDamage(EQ::skills::SkillType skill, Mob *target)
 				ac_bonus = inst->GetItemArmorClass(true) / 25.0f;
 			if (ac_bonus > skill_bonus)
 				ac_bonus = skill_bonus;
-			return static_cast<int>(ac_bonus + skill_bonus);
+			return base + static_cast<int>(ac_bonus + skill_bonus);
 		}
 		case EQ::skills::SkillKick: {
 			float skill_bonus = skill_level / 10.0f;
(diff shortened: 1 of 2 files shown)


Knightly1:
Looks good to me.

dannuic:
Assuming this is correct logic (I think it is), this looks good

KimLS (approved):

PR #5062, opened 2026-04-11
fix: only dismiss pet summoned by the fading familiar buff

### LINKED ISSUE #5026: Familiar buffs fading will kill ANY summoned pet
When a familiar style buff fades it will kill any summoned pet. For instance if you still have your familiar pet (flappy wing wizard pet) it will die when the familiar buff fades. This is known and expected. However, say you /pet leave the familiar, then summon a normal pet (Mage pet for instance), that pet will also die when the familiar buff fades.

Additionally, refreshing said familiar buff will also kill the current pet and summon a new familiar.

Now I'm sure this is intended, but I'd like to see a rule added that will allow you to preserve non-familiars when these buffs fade or are refreshed, if possible.

Thanks.

ltroylove:
Fixes #5026

## What

When a `SpellEffect::Familiar` buff fades in `Mob::BuffFadeBySlot`, the code called `GetPet()` and unconditionally depopulated whatever pet the player currently had:

```cpp
// Before
Mob *mypet = GetPet();
if (mypet){
    if(mypet->IsNPC())
        mypet->CastToNPC()->Depop();
    SetPetID(0);
}
```

`GetPet()` returns the player's **current** pet — not the pet summoned by the specific buff that is fading. So if a player had a familiar, dismissed it, and summoned a combat pet, the combat pet would be killed when the familiar buff eventually expired.

```cpp
// After
Mob *mypet = GetPet();
if (mypet && mypet->IsNPC() &&
    mypet->CastToNPC()->GetPetSpellID() == buffs[slot].spellid) {
    mypet->CastToNPC()->Depop();
    SetPetID(0);
}
```

The fix checks that the current pet's `PetSpellID` matches the fading buff's spell ID before dismissing it. If the player has a different pet, it is left alone.

## Verified

- `NPC::GetPetSpellID()` is set during `MakePet()` / `MakePoweredPet()` via `SetPetSpellID(spell_id)` — the spell ID is reliably tracked
- `buffs[slot].spellid` is the standard way to reference the fading buff's spell ID throughout `BuffFadeBySlot`
- `Bot::BuffFadeBySlot` does not exist as a separate override — Bots use `Mob::BuffFadeBySlot` and are covered by this fix

## Note

### DIFF
--- a/zone/spell_effects.cpp
+++ b/zone/spell_effects.cpp
@@ -4394,9 +4394,9 @@ void Mob::BuffFadeBySlot(int slot, bool iRecalcBonuses)
 			case SpellEffect::Familiar:
 			{
 				Mob *mypet = GetPet();
-				if (mypet){
-					if(mypet->IsNPC())
-						mypet->CastToNPC()->Depop();
+				if (mypet && mypet->IsNPC() &&
+				    mypet->CastToNPC()->GetPetSpellID() == buffs[slot].spellid) {
+					mypet->CastToNPC()->Depop();
 					SetPetID(0);
 				}
 				break;


ltroylove:
Pulling this back for additional review before resubmitting.

PR #5068, opened 2026-04-19
Login Updates for TOB

Knightly1:
Fixes an issue where failed login information was not being sent to the client properly. First path at removing repetitive code.

Update PlayerLoginReply struct for TOB
- TOB Struct is different from the old one
- Moved specifics into their respective structs
- Added variant for choosing the struct to use
- Added setters to abstract the client operations and reduce duplicate code
- Pass at cleanup on GenerateRandomLoginKey to make it easier to modify


KimLS:
Build is failing

PR #5070, opened 2026-04-22
Added component-based patch system

dannuic:
Apply to interrupt and fizzle packets (and all trivial message packets)


common/emu_versions.h
```diff
@@ -61,8 +61,8 @@ namespace EQ
 			maskAllClients = 0xFFFFFFFF
 		};
 
-		const ClientVersion LastClientVersion = ClientVersion::TOB;
-		const size_t ClientVersionCount = (static_cast<size_t>(LastClientVersion) + 1);
+		constexpr ClientVersion LastClientVersion = ClientVersion::TOB;
+		constexpr size_t ClientVersionCount = (static_cast<size_t>(LastClientVersion) + 1);
```
brainiac: these should be inline constexpr since this is a header

brainiac:
this file structure is going to be really frustrating to work in given the number of files that are just "rof2.h" etc. recommend something like "message_rof2.h". i don't want to end up in a situation where i'm trying to filter through a dozen files all named rof2.h.

brainiac:
additionally maybe just one cpp for the whole component?

zone/spells.cpp
```diff
@@ -335,35 +333,23 @@ bool Mob::DoCastSpell(uint16 spell_id, uint16 target_id, CastingSlot slot,
...
-			*/
-			GetName(),
-			spell_link.c_str()
-		);
+		ZoneClient::ClientPatch::QueueCloseClients(
+			this, true, RuleI(Range, SpellMessages),
+			nullptr, true,
+			IsClient() ? FilterPCSpells : FilterNPCSpells)(
+			&ZoneClient::Message::IMessage::FizzleOther, Chat::SpellFailure,
+			fizzle_msg == MISS_NOTE ? MISSED_NOTE_OTHER : SPELL_FIZZLE_OTHER, spell_id, GetName());
```
brainiac: just want to point out that this is harder to read than the old method.

dannuic: It's harder to read, but the old method also wasn't specializing logic per client patch

brainiac: why not be both capable of supporting specializing logic AND be easy to read? :)

dannuic: That's the dream isn't it? But I think I have an idea for something to try today. My main goal is that I don't want packet construction logic in the game logic, but I think the string argument construction is fine there so I can make this generic enough that I don't need to pass the function address

dannuic:
> this file structure is going to be really frustrating to work in given the number of files that are just "rof2.h" etc. recommend something like "message_rof2.h". i don't want to end up in a situation where i'm trying to filter through a dozen files all named rof2.h.

Yeah, I agree with this. I didn't like the `message_` naming either, but I think what I really need to do is colocate this logic with the existing patch logic because I don't like how patch specialization is in multiple places

dannuic:
Per comments, I only added a couple of new files (good callout). The patch logic is co-located with the rest of the patch logic which I think will reduce searching for code, I only added a couple of things to tie game logic into the patch logic. I tried to keep the API calls the same or as similar as possible while still supporting the need to generate a packet per client version when sending a message. I'm marking this ready for review so I can continue to work on validating packets based from this branch.

common/patches/client_version.cpp
```diff
@@ -0,0 +1,84 @@
...
+		std::unordered_map<Version, ClientComponents> p;
+		p.emplace(Version::Titanium, Version::Titanium);
+		p.emplace(Version::SoF, Version::SoF);
+		p.emplace(Version::SoD, Version::SoD);
+		p.emplace(Version::UF, Version::UF);
+		p.emplace(Version::RoF, Version::RoF);
+		p.emplace(Version::RoF2, Version::RoF2);
+		p.emplace(Version::TOB, Version::TOB);
+		return p;
+	}();
```
brainiac:
```suggestion
	static const std::unordered_map<Version, ClientComponents> patches = {
		{ Version::Titanium, Version::Titanium },
		{ Version::SoF, Version::SoF },
		{ Version::SoD, Version::SoD },
		{ Version::UF, Version::UF },
		{ Version::RoF, Version::RoF },
		{ Version::RoF2, Version::RoF2 },
		{ Version::TOB, Version::TOB },
	};
```

brainiac: although i don't know why you need a factory function if this is the only place its going to be used.

could just replace this with the make_shared calls. Also this is a function function so its always going to outlive anything that uses it, so it should probably be unique_ptr instead.

zone/client_version.h
```diff
@@ -0,0 +1,154 @@
...
+// store all _generic_ static functions for the different patches here
+namespace ClientPatch {
+
+using ClientList = std::unordered_map<uint16, Client*>;
+
+template <typename Fun, typename Obj, typename... Args>
+static void QueuePacket(Client* c, Fun fun, Obj* obj, Args&&... args)
+{
+	static_assert(std::is_member_function_pointer_v<Fun>);
+	EQApplicationPacket* app = std::invoke(fun, obj, std::forward<Args>(args)...);
```
brainiac: such a waste to allocate this and immediately delete it.

zone/client_version.h
```diff
@@ -0,0 +1,154 @@
...
+
+// store all _generic_ static functions for the different patches here
+namespace ClientPatch {
+
+using ClientList = std::unordered_map<uint16, Client*>;
+
+template <typename Fun, typename Obj, typename... Args>
+static void QueuePacket(Client* c, Fun fun, Obj* obj, Args&&... args)
+{
+	static_assert(std::is_member_function_pointer_v<Fun>);
```
brainiac: should put this as a constraint on Fun

zone/client_version.h
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// packet generator queue functions
+static auto QueueClients(Mob* sender, bool ignore_sender = false, bool ackreq = true)
+{
+	return [=]<typename Fun, typename Obj, typename... Args>(Fun fun,
+		std::function<Obj*(const Client*)> component_getter, Args&&... args) {
+		static_assert(std::is_member_function_pointer_v<Fun> && "Function is required to be a member function");
+
+		std::unordered_map<EQ::versions::ClientVersion, EQApplicationPacket*> build_packets;
```
brainiac: std::vector would be better here. max number of keys here is going to be like... 2

zone/client_version.h
```diff
@@ -0,0 +1,154 @@
...
+			std::unordered_map<EQ::versions::ClientVersion, EQApplicationPacket*> build_packets;
+
+			for (auto& [_, mob] : sender->GetCloseMobList(distance)) {
+				if (mob && mob->IsClient()) {
+					Client* client = mob->CastToClient();
+					if ((!ignore_sender || client != sender)
+						&& client != skipped_mob
+						&& DistanceSquared(client->GetPosition(), sender->GetPosition()) < distance_squared
+						&& client->Connected()
+						&& client->ShouldGetPacket(sender, filter)) {
```
brainiac:
```suggestion
						&& client->ShouldGetPacket(sender, filter))
					{
```

having a hard time picking out wrapped lines vs blocks.

common/patches/client_version.cpp
dannuic: these are singletons, I made them shared because I thought clients were going to store references to them, but I think that's no longer the case and unique makes more sense. The reason I made this a factory was because I expect that there are going to be more components like this and the factory will create multiple components per patch

zone/client_version.h
dannuic: yeah, I think this can just be on stack. I think the stuff below can be too

zone/client_version.h
dannuic: with `require`? this is functionally equivalent to  putting it in the template since it's all static and caught at compile time, I find this more readable

zone/client_version.h
dannuic: probably 1 actually,  most servers only have one accepted version. I thought unordered map access was O(1) though

common/patches/client_version.cpp
dannuic: oh also, using the literal initialization here doesn't work. I'm not entirely sure why, but I could only get this to work if I constructed the map with emplaces

zone/client_version.h
brainiac: unordered_map is a chained hashmap, not probed. still ends up with more heap allocations than necessary for something like this.

zone/client_version.h
brainiac: yes with require. you'll be suprised how much better the diagnostics are with constraints and concepts than a static_assert.

common/patches/client_version.cpp
brainiac: i'm guess its because it can't deduce the parameters in the initializer list. thats why i started to question the factory function.

KimLS (approved):

PR #5076, opened 2026-05-03
Fixed the buff database entries and updated database migrations

dannuic:
This includes all the generated files from the repo generation (some were out of date)

Summary of changes:
In TOB (and actually before) buffs are all handled exactly the same, regardless of source. So this change makes the buffs from all sources have the same schema so they can all be treated the same. In addition, the largest change is changing the pet buff handler to use the unified schema instead of its original bespoke schema. This allows the buff update packets (which have already been updated) to just work without needing any special handling based on the source.

I thought about just making a single buffs schema in the database to unify everything, but I decided that was a pretty major change and would probably warrant some discussion, while this is pretty direct and can move the TOB project forward.

### DIFF
--- a/common/database/database_update_manifest.h
+++ b/common/database/database_update_manifest.h
@@ -7226,6 +7226,45 @@ ALTER TABLE `npc_spells_entries` MODIFY COLUMN `spellid` INTEGER NOT NULL DEFAUL
 ALTER TABLE `spell_buckets` MODIFY COLUMN `spell_id` INTEGER NOT NULL;
 )"
 	},
+	ManifestEntry{
+		.version = 9330,
+		.description = "2026_04_30_buffdurations.sql",
+		.check = "SHOW COLUMNS FROM `character_buffs` LIKE 'initialduration'",
+		.condition = "empty",
+		.match = "",
+		.sql = R"(
+ALTER TABLE `character_buffs`
+	ADD COLUMN `initialduration` INT(11) SIGNED NOT NULL DEFAULT 0 AFTER `ticsremaining`,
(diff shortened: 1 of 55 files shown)


common/database/database_update_manifest.h
```diff
@@ -7226,6 +7226,45 @@ ALTER TABLE `npc_spells_entries` MODIFY COLUMN `spellid` INTEGER NOT NULL DEFAUL
 ALTER TABLE `spell_buckets` MODIFY COLUMN `spell_id` INTEGER NOT NULL;
 )"
 	},
+	ManifestEntry{
```
brainiac, line 7229: Where does this come from? I don't see initialduration anywhere in the codebase.

brainiac: oh is this new? maybe we should separate out the new stuff from the regenerated stuff?

dannuic: opened a PR that's just updating the db, I'll merge again after that

dannuic: There, updated. initialduration is used in the TOB BuffDefinition packet

brainiac:
Can you summarize the changes?

dannuic:
Summary of changes:
In TOB (and actually before) buffs are all handled exactly the same, regardless of source. So this change makes the buffs from all sources have the same schema so they can all be treated the same. In addition, the largest change is changing the pet buff handler to use the unified schema instead of its original bespoke schema. This allows the buff update packets (which have already been updated) to just work without needing any special handling based on the source.

I thought about just making a single buffs schema in the database to unify everything, but I decided that was a pretty major change and would probably warrant some discussion, while this is pretty direct and can move the TOB project forward.

KimLS (approved):

PR #5082, opened 2026-05-09
CMake cleanup

brainiac:
- Add debug symbols to Release, define only Debug and Release configurations
- Update CMakePresets.json
- Remove a bunch of redundancies in all project files
  - Add root directory as global include path
  - Remove redundant cmake_minimum_required
- Fix inconsistent indentation

Standardized presets a bit, they all enable debug and release configurations.
- linux
- linux-clang
- linux-ccache
- linux-clang-ccache
- win-msvc-2022
- win-msvc-2026


CMakeLists.txt
```diff
@@ -18,6 +18,16 @@ if(NOT CMAKE_TOOLCHAIN_FILE)
...
 
+set(CMAKE_BUILD_TYPE "Debug")
+set(CMAKE_CONFIGURATION_TYPES "Debug;Release" CACHE STRING "" FORCE)
+
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
+set(dummy "${CMAKE_C_COMPILER}")
```
dannuic, line 30: Is this needed?

CMakePresets.json
```diff
@@ -1,49 +1,74 @@
...
+			"displayName": "Windows MSVC (VS 2026)",
+			"generator": "Visual Studio 18 2026",
+			"binaryDir": "${sourceDir}/build",
+			"architecture": { "value": "x64" },
+			"cacheVariables": {
+				"EQEMU_BUILD_TESTS": "ON"
+			}
+		}
+	]
+}
```
dannuic, line 74: potentially minor concern: does this break people relying on presets? maybe we pick one msvc as the default (like you did for linux)?

CMakeLists.txt
brainiac: cmake complains if CMAKE_C_COMPILER is set because we don't have any c projects, so this makes it not complain.

On the other hand i could just not set CMAKE_C_COMPILER when using the clang preset, but it feels wrong to set CXX but not C in that case.

CMakePresets.json
brainiac: I think its worth breaking something for them to be consistent. having a default for windows doesn't seem right for me since you should pick whichever matches the vs you have installed.

That said i'm not aware of any actual processes that use these. And i'd like them to be useful.

CMakeLists.txt
dannuic: makes sense, the "dummy" threw me

CMakePresets.json
dannuic: yeah, I think I agree with you here

PR #5086, opened 2026-05-14
refactor(Handle_OP_ShopPlayerSell): fix bugs, clean, restructure

sh0ber:
# Description

- Organized and documented each subsection's purpose in Handle_OP_ShopPlayerSell
- GetMerchantBuyPrice & GetMerchantSellPrice:  Normalized and extracted functions for merchant pricing per unit, with inst/item overloads for convenience
- ResolvePlayerSellQuantity:  Extracted quantity guards & quantity logic and refactored unnecessary per-item loops
- UpdateMerchantStock:  Extracted merchant restocking into a helper function
- Misc variable renaming / code simplification

# Notes
- NOTE 1: The extraction and normalization of GetMerchantSellPrice fixes a premature integer truncation bug by using a double for intermediate steps and correctly applying rounding, which finally gives exact alignment with the client UI display, however…
- NOTE 2: Fixing the transaction math exposed an existing exploit where low-value items (e.g., Fishing Bait) can be bought, for example, for 2c and sold back for 3c. This arbitrage is an environment-wide systemic behavior caused by a poor implementation of the default pricing mods. It’s a tiny exploit, and this fix gives it an even tinier bit more strength by fixing the bug I mentioned.  It’s still the correct architecture and establishes a baseline for resolving the exploit, which I plan to do in a separate PR. I recommend buying and selling stacks of bait and other items to test this exploit before accepting this PR to be convinced it’s not caused by it.  Note discrepancies with promised amounts and actual amounts given and taken by the merchant.

## Type of change

- [x] Structural refactor 

# Testing

On multiple EQEmu servers, I've tested the buying and selling of various items.  The merchant consistently fails to give and take the amount he tells you.  In some cases this is merely a text bug, but in the case of a low-priced item like fishing bait, it makes an exploit possible.  Again, this PR doesn't intend to fix that exploit yet (in fact makes it a tiny bit worse) but lays a foundation to fix it from.

# Clients tested: 
RoF2

# Checklist

### DIFF
--- a/zone/client.h
+++ b/zone/client.h
@@ -898,9 +898,9 @@ public:
 	void SendMoneyUpdate();
 	bool TakeMoneyFromPP(uint64 copper, bool update_client = false);
 	bool TakePlatinum(uint32 platinum, bool update_client = false);
-	void AddMoneyToPP(uint64 copper, bool update_client = false);
-	void AddMoneyToPP(uint32 copper, uint32 silver, uint32 gold, uint32 platinum, bool update_client = false);
-	void AddPlatinum(uint32 platinu, bool update_client = false);
+	void AddMoneyToPP(uint64 copper, bool update_client = false, bool update_db = true);
+	void AddMoneyToPP(uint32 copper, uint32 silver, uint32 gold, uint32 platinum, bool update_client = false, bool update_db = true);
+	void AddPlatinum(uint32 platinum, bool update_client = false, bool update_db = true);
 	bool HasMoney(uint64 copper);
 	uint64 GetCarriedMoney();
 	uint32 GetCarriedPlatinum();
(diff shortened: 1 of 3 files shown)


brainiac (changes requested):
AddPlatinum/AddMoneyToPP were specifically written this way to protect against overflow. They should not be changed this way.

zone/client.h
```diff
@@ -898,9 +898,9 @@ class Client : public Mob
 	void SendMoneyUpdate();
 	bool TakeMoneyFromPP(uint64 copper, bool update_client = false);
 	bool TakePlatinum(uint32 platinum, bool update_client = false);
-	void AddMoneyToPP(uint64 copper, bool update_client = false);
-	void AddMoneyToPP(uint32 copper, uint32 silver, uint32 gold, uint32 platinum, bool update_client = false);
-	void AddPlatinum(uint32 platinu, bool update_client = false);
+	void AddMoneyToPP(uint64 copper, bool update_client = false, bool update_db = true);
+	void AddMoneyToPP(uint32 copper, uint32 silver, uint32 gold, uint32 platinum, bool update_client = false, bool update_db = true);
```
brainiac: Why is this `update_db` an option?

sh0ber: It allows iterating over multiple items and selling each without hitting the database for each.  The idea was to pass false when selling everything in a container to a merchant, and then call SaveCurrency() once after the loop

sh0ber:
> AddPlatinum/AddMoneyToPP were specifically written this way to protect against overflow. They should not be changed this way.

@brainiac Good catch.  I think the original may be a little problematic too so I introduced INT32_MAX limits.  I separated my refactors back into 2 distinct functions again, instead of calling one from the other.  And I fixed/normalized the logic for both the wallet and external_handin values

zone/client.h
brainiac: That sounds like a good way to create an exploit. Only update currency if you're going to immediately save it.

brainiac:
I really wish you hadn't force pushed, that makes this much more difficult to review.

sh0ber:
@brainiac My bad, I'm used to working alone and didn't consider there would be a major difference for a reviewer in this scenario between regular and force.  I've removed all changes to the plat functions so this PR can focus on its main intent of cleaning up the packet handler.

zone/client.h
sh0ber: Understood

PR #5093, opened 2026-05-23
[S41 bot AI fix CI compile-check / DO NOT MERGE] Bot::SpellOnTarget null-target guards (+3 siblings)

nightwreath:
## Crash context

A bard died to Vulak Aerr's melee mid-`Bot::SpellOnTarget` on her. The `Mob*` (Vulak) was freed inside the call, but the `spelltar` pointer was still held in the caller's frame. The next vtable call through that freed memory crashed the zone with a segfault. Stack trace was localized at `bot.cpp:6111` (the `spelltar->IsPet()` deref, inside the existing `if (spelltar)` block -- which protects against null but not against dangling).

This isn't a new bug -- it's been latent in the bot pipeline since the function was written. The Vulak fight surfaced it because high-DPS multi-target raid combat is the worst case for "the buff target died between when `GetBuffTargets()` snapshotted the vector and when we get to iterate".

## Root cause: pointer non-null but freed

`Bot::SpellOnTarget` takes a `Mob* spelltar` by raw pointer. Mid-call, the engine can free that Mob (death, zone transition, despawn). The pointer doesn't reset to null; the caller frame still holds it. The existing `if (spelltar)` null-check passes; the next deref reads from freed memory.

`entity_list.IsMobInZone(Mob*)` (`entity.cpp:4196-4216`) is the canonical guard:

> We don't use mob_list.find(who) because this code needs to be able to handle dangling pointers for the quest code.

It walks the live entity map looking for the pointer; if not present, the pointer is stale. Safe to call on freed memory.

## Systematic audit -- five sites guarded

| # | Site | What it does | Guard |
|---|---|---|---|
| 1 | `Bot::SpellOnTarget` top (post-IsValidSpell) | Caller-frame `spelltar` may be stale on entry. | `if (spelltar && !entity_list.IsMobInZone(spelltar)) return false;` |
| 2 | `DoFinishedSpellSingleTarget` loop entry | `GetBuffTargets()` returns a snapshot vector; member can die mid-iter. | `if (!m || !entity_list.IsMobInZone(m)) continue;` |
| 3 | `DoFinishedSpellSingleTarget` post-`SpellOnTarget` before `m->GetPet()` | `SpellOnTarget` can mutate the world (damage, death, zone trans). | Re-validate `m` in the `if (...)` chain. |
| 4 | `DoFinishedSpellGroupTarget` loop entry | Same snapshot vector pattern. | `if (!m || !entity_list.IsMobInZone(m)) continue;` |
| 5 | `DoFinishedSpellGroupTarget` post-`SpellOnTarget` before `m->GetPet()` | Defense-in-depth -- mirror of #3. | Re-validate `m` in the `if (...)` chain. |

Sites 3 and 5 are defense-in-depth: `SpellOnTarget` can recursively invalidate `m` (it can kill `m` via reflected damage, or trigger a zone transition that frees the entity list). Without re-validation, the `m->GetPetID()` deref a few lines later is the next crash candidate.

## Diagnostic

Site 1 emits `LogSpells("Bot::SpellOnTarget aborted: stale/freed target for spell [{}]", spell_id);` on rejection. Per S41 systematic-audit-for-silent-failures discipline: a silent abort is OK structurally (caller is the AI pipeline, not a player command), but logged so post-mortem traces aren't blind.

## Deeper fix deferred

`Bot::SetSpellRecastTimer` at `bot.cpp:12270` uses buff duration instead of `spell.recast_time`, which creates the tight-retry loop pattern PR #2 in this S41 set addresses. Capturing as a deferred memory note in `theo-and-co/memory/`. Not in scope for this PR.

## Test plan

- CI compile-check (this PR's main gate -- the title carries DO NOT MERGE until compile is green).
- After merge: GHCR rebuild, VPS swap, group bot raid encounter (Vulak repeat or similar high-DPS multi-target).
- Expected: no segfault on bot member death mid-cast; if a buff target goes stale, the cast aborts cleanly with a LogSpells line.


nightwreath:
Closing -- this PR was inadvertently opened against the upstream EQEmu/EQEmu repo. It belongs on the nightwreath/theo-and-co-engine fork. Reopening there. Sorry for the noise.

PR #5097, opened 2026-05-27
Fix CheckForCompatibleQuestPlugins to use config-driven directory paths

### LINKED ISSUE #5098: Plugins and Lua Modules key:value in eqemu_config.json are never read or parsed
The PluginsDir and LuaModulesDir are defined in https://github.com/EQEmu/EQEmu/blob/master/common/eqemu_config.cpp#L172 but never actually read or parsed in the eqmuemu_config.json config file.

If you look at https://github.com/EQEmu/EQEmu/blob/master/zone/main.cpp#L712 , we have this `CheckForCompatibleQuestPlugins()` function where the config file handler is never passed to, and the filepaths are hardcoded.  As many have observed in the community Discord, this forces us all to hardcode a symlnk in the root directory.

The proposed solution is simple: just pass the config variable that is already defined in zone/main.cpp to that check function and read from it instead of hardcoding filepaths.

nazwadi:
Replace hardcoded "lua_modules" string comparison with bool* flags from the directories vector, fixing lua_found never being set when LuaModuleDir is a non-default path.

# Description

The eqemu_config specifies a plugins and lua_modules directory, but it is never read and thus never used.  Currently, "plugins" and "lua_modules" are hardcoded in zone/main.cpp which means you have to have those directories symlinked from the root of your server to get zones to launch.  This is non-intuitive behavior and I've found numerous users in discord conversations get tripped up on this issue.

This change *appears* to complete the original intent of the flag pointers in the directories vector in this function.

This would only be a breaking change if users don't have their existing plugins and lua_modules directories specified correctly in their eqemu_config file.

Fixes #5098 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Testing

I tested this locally on my Windows 10 instance using Visual Studio Community 2022 and the msvc compiler.

Clients tested: Not necessary - does not affect client behavior.  This is just a quest API version check function in zone startup.

# Checklist

### DIFF
--- a/zone/main.cpp
+++ b/zone/main.cpp
@@ -95,7 +95,7 @@ void CatchSignal(int sig_num);
 
 extern void MapOpcodes();
 
-bool CheckForCompatibleQuestPlugins();
+bool CheckForCompatibleQuestPlugins(const ZoneConfig *config);
 int main(int argc, char **argv)
 {
 	RegisterExecutablePlatform(ExePlatformZone);
@@ -345,7 +345,7 @@ int main(int argc, char **argv)
 		return 1;
 	}
 
-	if (!CheckForCompatibleQuestPlugins()) {
+	if (!CheckForCompatibleQuestPlugins(Config)) {
 		LogError("Incompatible quest plugins detected, please update your plugins to the latest version");
 		return 1;
 	}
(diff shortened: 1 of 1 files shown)


zone/main.cpp
```diff
@@ -709,16 +709,16 @@ void UpdateWindowTitle(char *iNewTitle)
...
-		{"plugins",     nullptr}
-	};
-
 	bool lua_found  = false;
 	bool perl_found = false;
 
+	const std::vector<std::pair<std::string, bool *>> directories = {
+		{config->LuaModuleDir, &lua_found},
+		{config->PluginDir, &perl_found}
+	};
```
brainiac: I think you have to actually resolve these paths with the path manager, since these are just the config values and might not be the qualified paths.

nazwadi: Done.  I refactored the function to iterate over the Path Manager's `GetLuaModulePaths()` and `GetPluginPaths()`.  Thanks for the feedback.

brainiac:
Thats better.

part of me wonders if we could just delete this entirely, though.

nazwadi:
> Thats better.
> 
> part of me wonders if we could just delete this entirely, though.

less is more?

KimLS (approved):

PR #5105, opened 2026-06-18
Integrate client-based say links for TOB+ clients

dannuic:


dannuic:
Converting back to draft; I forgot to do item links entirely

common/links.h
```diff
@@ -3,9 +3,7 @@
 //
 
 #pragma once
-#include "item_instance.h"
-
-namespace EQ { class ItemInstance; }
+#include "say_link.h"
```
brainiac:
```suggestion

#include "common/say_link.h"
```

brainiac: I guess that there are others that were added recently but all include paths ought be qualified like this

common/say_link.cpp
```diff
@@ -17,13 +17,15 @@
 */
 #include "say_link.h"
 
+#include "features.h"
 #include "common/emu_constants.h"
 #include "common/strings.h"
 #include "common/item_instance.h"
 #include "common/item_data.h"
-#include "zone/zonedb.h"
+#include "common/shareddb.h"
```
brainiac, line 25: sorting is hard i guess

dannuic: yeah, all of these headers were just auto-added by clion when I add something that needs a header. I can go clean them up

KimLS (approved):

PR #5112, opened 2026-07-18
Dragon's Hoard Feature

ofnature:
Wires up the Dragon's Hoard — the account-wide item stash that ships with the TOB
client — on the server side. TOB already has the window and all the client logic
built in; nothing on the emu side was driving it, so the window just sat there dead.
This makes it actually work.

## What it does

Account-wide storage (200 slots), deposit from cursor and retrieve back to cursor,
accessed at a banker. Off by default behind a rule, and only active for TOB-or-later
clients so it can't affect anyone on RoF2 or earlier.

## How it works

The interesting part was that the client won't even *send* a deposit until it thinks
the feature is owned. That check reads a client-side feature array, and the array is
populated by a `FeatureUnlock` packet — not the membership packet, which is where I
went first and wasted a while. Tracing the client in IDA turned up the opcode (0x5B9B)
and the feature id (2016763); `SendFeatureUnlock` sends it at zone-in and that's what
unblocks deposit.

Retrieve was the finicky one. The client sends back the selected item's *serial*, not
a slot or a list index, so the server stamps each item's serial as `slot_id + 1` on the
way out and looks it up by that on the way back. Two follow-up packets were also needed
and both came out of reversing the client's handler:

- after a retrieve, an `action=3` confirmation — removes the row from the window and
  releases the item from the cursor
- after a deposit, an `action=10` ack — clears the client's pending-operation lock;
  without it the whole inventory freezes until you relog

## What's in it

- `OP_FeatureUnlock` (0x5B9B) and `OP_DragonHoard1/2` opcodes + TOB patch mappings
- `DragonHoard` handlers (feature unlock / deposit / retrieve / item list) in
  `zone/dragonshoard.cpp`
- `dragonhoard_items` table via the custom DB manifest, so it auto-applies on start
- `Features:DragonHoardEnabled` rule (default off), plus a TOB client-version gate

## Testing

1. `#rule set Features:DragonHoardEnabled true`
2. Relog on a TOB client and open the hoard at a banker
3. Deposit an item from the cursor, then retrieve it — inventory stays responsive and
   the window updates correctly

## Notes

- No effect on non-TOB clients (rule + client version gated)
- Holds non-stackable items, matching retail (collectibles, ornaments, etc.)
- The table lives in the custom manifest lane so it won't collide with upstream
  database version bumps on rebase


common/database/database_update_manifest_custom.h
```diff
@@ -24,15 +24,20 @@
...
+  `slot_id`     smallint unsigned NOT NULL COMMENT '0..199 (max 200 slots)',
+  `item_id`     int unsigned      NOT NULL,
+  `item_name`   varchar(64)       NOT NULL DEFAULT '',
+  `stack_count` int unsigned      NOT NULL DEFAULT 1,
+  PRIMARY KEY (`account_id`, `slot_id`),
+  KEY `ix_account_id` (`account_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
 )",
 		.content_schema_update = false,
 	},
```
dannuic, line 38: unsure why "new_table" is here, but database migrations are  supposed to be additive always. This should be a new entry instead of replacing an old entry

common/patches/tob.cpp
```diff
@@ -105,6 +105,17 @@ namespace TOB
...
+		do { \
+			if (opcodes != nullptr && opcodes->Mutable()) { \
+				static_cast<RegularOpcodeManager *>(opcodes)->SetOpcode((emu_name), (eq_value)); \
+			} \
+		} while (0)
+		REGISTER_OPCODE(OP_DragonHoard1, 0x5807);
+		REGISTER_OPCODE(OP_DragonHoard2, 0x603D);
+		REGISTER_OPCODE(OP_FeatureUnlock, 0x5B9B); // server->client store-feature unlock (DH, depot, keyrings)
+#undef REGISTER_OPCODE
+
```
dannuic, line 110: This isn't the correct way to do this. Just put these opcodes into patch_TOB.conf and the RegularOpcodeManager will load them for TOB.

common/patches/tob.cpp
```diff
@@ -2926,7 +2937,7 @@ namespace TOB
 		eq->entries[26] = 0;  // FamiliarKeyRingSlots
 		eq->entries[27] = 0;  // FamiliarAutoLeave
 		eq->entries[28] = 0;  // HeroForgeKeyRingSlots
-		eq->entries[29] = 0;  // DragonHoardSlots
+		eq->entries[29] = 200;  // DragonHoardSlots
```
dannuic: Is this something hardcoded in the client? If  so, it should go into the tob constants definition, otherwise it should go into a configuration somewhere.

zone/client.cpp
```diff
@@ -9184,6 +9184,12 @@ void Client::SendMembership() {
 		//mc->exit_url = 0; // Used on Live: "http://www.everquest.com/free-to-play/exit-silver"
 
 		QueuePacket(outapp);
+
+		// Send membership details (feature flags).
+		auto detailsapp = new EQApplicationPacket(OP_SendMembershipDetails, sizeof(Membership_Details_Struct));
+		QueuePacket(detailsapp);
+		safe_delete(detailsapp);
+
```
dannuic: does this break pre-TOB clients? or is this a bug in pre-TOB clients that the features don't get sent? If it's the former, this might need to be added in the patch piggybacking off the previous packet send

zone/client_packet.cpp
```diff
@@ -770,7 +773,13 @@ void Client::CompleteConnect()
 	else
 		TaskPeriodic_Timer.Disable();
 
-	conn_state = ClientConnectFinished;
+	if (conn_state != ClientConnectFinished) {
+		conn_state = ClientConnectFinished;
+		// Defer Dragon's Hoard unlock/item-send until the client's world display is ready.
+		dragonhoard_zonein_timer.Start(4000);
```
dannuic: is setting a timer the best way to do this? doesn't the client send a ready packet when its done?

zone/CMakeLists.txt
```diff
@@ -31,6 +31,7 @@ set(zone_sources
 	corpse.cpp
 	dialogue_window.cpp
 	doors.cpp
+	dragonshoard.cpp
```
dannuic, line 34: you also need to track the new header in CMakeLists.txt (there's a header section). This is for project management

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,329 @@
+#include "dragonshoard.h"
+#include "client.h"
+#include "../common/global_define.h"
+#include "../common/item_instance.h"
+#include "../common/rulesys.h"
+#include "../common/strings.h"
```
dannuic: don't do relative includes (I'm surprised brainiac hasn't yelled at you yet)

zone/client.cpp
ofnature: that was reverted in a later commit https://github.com/EQEmu/EQEmu/pull/5112/commits/cd652f3f866236d8196ade005e8a06a618f5c5f9

dannuic: It's in the latest, it wasn't reverted. This is sending an additional packet to all clients, the question is if it should be TOB only

zone/client_packet.cpp
dannuic: looking at this closer, this sets conn_state to ClieintConnectFinished always, I'm confused why that's in the if statement. I'm also pretty sure the timer method isn't what you want, check the client for what packets it's sending. Do they even need to be sent on connect at all?

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,329 @@
...
+		fmt::format(
+			"INSERT INTO dragonhoard_items (account_id, slot_id, item_id, item_name, stack_count) "
+			"VALUES ({}, {}, {}, '{}', {})",
+			account_id,
+			slot_id,
+			item_id,
+			Strings::Escape(std::string(name ? name : "")),
+			stack
+		)
+	);
```
dannuic: All this database stuff should go into the database handlers, you don't want to be writing queries directly in your logic like this. You'll notices that you have database.GetItem above, that's the kind of pattern that I'd expect here.

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,329 @@
...
+	if (!IsEnabled(client)) {
+		return;
+	}
+
+	(void)slot_id;
+	(void)item_id;
+	(void)remove;
+
+	// TODO: send single item add/remove delta update to DH window
+}
```
dannuic: Is this empty for a reason?

zone/client_packet.cpp
ofnature: the unlock requires the ui element to be built (same for all features im working on tradeskill depot, keyrings), i have found a ui built acknowledgment packet the client sends to the server, im gonna try to gate the send on that.

zone/dragonshoard.cpp
ofnature: this is an artifact from earlier design(that failed hard) its not called and is being removed

zone/client_packet.cpp
ofnature: removed the timer its still working but there might be times the ui element isnt built, the other opcode was a dead end.

common/ruletypes.h
```diff
@@ -1179,6 +1179,11 @@ RULE_INT(EvolvingItems, DelayUponEquipping, 30000, "Delay in ms before an evolvi
 RULE_BOOL(EvolvingItems, DestroyAugmentsOnEvolve, false, "If this is enabled, any augments in an item will be destroyed when the item evolves. Otherwise, send augments to the player via the parcel system (requires that the Parcel System be enabled).")
 RULE_CATEGORY_END()
 
+RULE_CATEGORY(Features)
+// Dragon's Hoard — requires TOB client, default off
```
brainiac: repetitive/redundant comment

zone/client_packet.cpp
```diff
@@ -17440,3 +17449,26 @@ void Client::SyncWorldPositionsToClient(bool ignore_idle)
 		m_is_idle = false;
 	}
 }
+
+// Dragon's Hoard handlers
+void Client::Handle_OP_DragonHoard1(const EQApplicationPacket *app)
+{
+	// OP_DragonHoard1 (0x5807) carries several actions; route by the leading action field.
```
brainiac: references to opcode constants. Just stick with the name. the number shouldn't be included here.

zone/client_packet.cpp
```diff
@@ -17440,3 +17449,26 @@ void Client::SyncWorldPositionsToClient(bool ignore_idle)
 		m_is_idle = false;
 	}
 }
+
+// Dragon's Hoard handlers
+void Client::Handle_OP_DragonHoard1(const EQApplicationPacket *app)
```
brainiac: please give this a more descriptive name than DragonHoard1

zone/client_packet.cpp
```diff
@@ -17440,3 +17449,26 @@ void Client::SyncWorldPositionsToClient(bool ignore_idle)
...
+{
+	// OP_DragonHoard1 (0x5807) carries several actions; route by the leading action field.
+	if (!app || app->size < 4) {
+		return;
+	}
+	uint32_t action = *(uint32_t*)app->pBuffer;
+	if (action == 4) {
```
brainiac: please convert these magic numbers into constants with meaningful names.

zone/client_packet.cpp
```diff
@@ -17440,3 +17449,26 @@ void Client::SyncWorldPositionsToClient(bool ignore_idle)
...
+		DragonHoard::HandleDeposit(this, app);
+	} else if (action == 3) {
+		DragonHoard::HandleRetrieve(this, app);
+	}
+	// action=0 = window open request, no server response needed
+}
+
+void Client::Handle_OP_DragonHoard2(const EQApplicationPacket *app)
+{
+	// OP_DragonHoard2 (0x603D) isn't sent by the TOB client (deposits route through
```
brainiac: If this isn't sent then why is it here?

zone/dragonshoard.cpp
brainiac: I'm here now. Please make these include root-relative

i.e. "../common/global_define.h" is just "common/global_define.h"
"dragonshoard.h" is "zone/dragonshoard.h"
"client.h" is "zone/client.h" etc

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,329 @@
...
+// deposit. Same mechanism as Laurion's OP_FeatureUnlock (0x4451 there).
+// Wire format: [u8 header=0][u32 feature_count][ {u32 feature_id, u32 count>=1} x feature_count ]
+void DragonHoard::SendFeatureUnlock(Client* client)
+{
+	if (!IsEnabled(client)) {
+		return;
+	}
+
+	// TOB feature IDs (from client sub_1402DC7E0). Dragon's Hoard is the only feature the
+	// server implements; Tradeskill Depot IDs (2018125 / 2018260) are noted for future use.
```
brainiac: pretentious comment.

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,329 @@
...
+	static const uint32 feature_ids[] = {
+		2016763, // 0x1EC5FB — Dragon's Hoard
```
brainiac: please make a constant for this.

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+	uint8* buf = outapp->pBuffer;
+	buf[0] = 0; // header byte
+	*reinterpret_cast<uint32*>(buf + 1) = num_features;
+	uint32 off = 5;
+	for (uint32 i = 0; i < num_features; ++i) {
+		*reinterpret_cast<uint32*>(buf + off)     = feature_ids[i];
+		*reinterpret_cast<uint32*>(buf + off + 4) = active_count;
+		off += 8;
+	}
```
brainiac: Please use a SerializeBuffer or struct to write the data instead of manually manipulating pointers and reinterpret-casting.

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,329 @@
...
+	auto* outapp = new EQApplicationPacket(OP_DragonHoard1, 5);
+	memset(outapp->pBuffer, 0, 5);
+	*reinterpret_cast<uint32_t*>(outapp->pBuffer) = 8;
+	outapp->pBuffer[4] = 0;
+
+	// action=2 — max slot count
+	auto* outapp2 = new EQApplicationPacket(OP_DragonHoard1, 8);
+	memset(outapp2->pBuffer, 0, 8);
+	*reinterpret_cast<uint32_t*>(outapp2->pBuffer) = 2;
+	*reinterpret_cast<uint32_t*>(outapp2->pBuffer + 4) = static_cast<uint32_t>(MAX_SLOTS);
```
brainiac: same as previous comments. these magic numbers need named constants (an enum most likely) and the packet should be assembled with a struct or SerializeBuffer

zone/dragonshoard.h
```diff
@@ -0,0 +1,39 @@
...
+	void HandleDeposit(Client* client, const EQApplicationPacket* app);
+
+	// Client retrieves an item from the hoard to the cursor (OP_DragonHoard1 action=3).
+	void HandleRetrieve(Client* client, const EQApplicationPacket* app);
+
+	// Send a single item add/remove delta to the DH window.
+	void SendItemUpdate(Client* client, uint32 slot_id, uint32 item_id, bool remove);
+
+	// Max slots available in Dragon's Hoard.
+	static constexpr int MAX_SLOTS = 200;
```
brainiac: this should be in config

zone/dragonshoard.cpp
brainiac: this pattern repeat. assume this comment applies to the whole PR

brainiac (commented):
mostly just complaining about magic constants, claudes awful ability to write actually useful comments, and packet construction.

utils/patches/patch_TOB.conf
```diff
@@ -252,6 +252,9 @@ OP_MoveItem=0x121c
 OP_MoveMultipleItems=0x6bf7
 OP_DeleteItem=0x29e6
 OP_DeleteCharge=0x4bef
+OP_DragonHoard1=0x5807
+OP_DragonHoard2=0x603d
```
brainiac: there are no references to packet 0x603d in the tob client, so i think this can be removed?

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,329 @@
...
+	};
+	const uint32 num_features = static_cast<uint32>(sizeof(feature_ids) / sizeof(feature_ids[0]));
+	const uint32 active_count = 1; // any value > 0 marks the feature active
+
+	const uint32 packet_size = 1 + 4 + (num_features * 8);
+	auto* outapp = new EQApplicationPacket(OP_FeatureUnlock, packet_size);
+	memset(outapp->pBuffer, 0, packet_size);
+
+	uint8* buf = outapp->pBuffer;
+	buf[0] = 0; // header byte
```
brainiac: first byte is a bool. if set to zero it does something with the marketplace window, but i haven't figured out what.

zone/client_packet.cpp
dannuic: so I traced the client. The dragon hoard global should be valid anytime after OP_EnterWorld (and that is the only thing its checking on the feature enable packet). Were you actually seeing issues with this? The UI element doesn't need to be built, the global for the window just needs to be constructed.

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+	LogDebug("DragonHoard::SendItemList sent items to account_id {}", account_id);
+}
+
+// OP_FeatureUnlock (0x5B9B) populates the client's feature array (player+0x2620). The TOB
+// client (sub_1402DC7E0) reads {feature_id, count} pairs; the deposit path first checks
+// sub_14065B0E0(player, 2016763), and without a count>0 entry the client silently refuses to
+// deposit. Same mechanism as Laurion's OP_FeatureUnlock (0x4451 there).
+// Wire format: [u8 header=0][u32 feature_count][ {u32 feature_id, u32 count>=1} x feature_count ]
+void DragonHoard::SendFeatureUnlock(Client* client)
```
brainiac: This is actually completely unrelated to dragonhoard. its part of the claims system. this is how the server notifies the client of its claims, for example what features the user has access to (and also i think the number of slots/claims for things with quantity). keyrings, tradeskill depot, dragon hoard, etc.

This _shouldn't_ be necessary to send. the user's current claims should be getting sent in the main player packet when they enter the zone. This message i believe is just to update the client when they buy something.

I don't think you should be doing both.

common/patches/tob.cpp
brainiac: in MQ these are `GameFeature`s

zone/dragonshoard.cpp
dannuic: There's actually a commented out bit of code in tob.cpp that should help: line 2393 -- you need to send the claim data in OP_PlayerProfile (it's an array). You'll have to create the Claim struct, but it's literally just (int, int) just like everywhere else. It's a pretty consistent struct of (claim ID, value). But I'm not sure the value that needs to be sent and where. I think this packet is just for after people purchase stuff on the marketplace.

zone/dragonshoard.cpp
dannuic: before I head off for the night, I wanted to make sure to drop this here, because this is what you need to do: https://docs.eqemu.dev/developer/repositories/#code-generation-repository-generator

common/database/database_update_manifest_custom.h
ofnature: Two questions so I get the convention right: (1) should it
> take the next-available custom version rather than `1` to avoid colliding with the custom-manifest
> sequence, and (2) do we also bump `CUSTOM_BINARY_DATABASE_VERSION` (currently `0`, so custom
> migrations never auto-run) — or is that intentionally left for a separate step?

ofnature: should be additive now

common/patches/tob.cpp
ofnature: the 200 was from the in game help box from live, but that is the purchase max, not the physical max. made it into a config setting.

zone/client.cpp
ofnature: took this out, was trying to make it all feature complete in a sence of ftp/silver gold access not a good attempt at it mind you but since this is eqemu almost all default to gold status so the section wasnt needed.

zone/client_packet.cpp
ofnature: killed the timer, it was clanker(AI) fearing the worst, but it was remnants from my try at laurion song DH work that I migrated was having issues getting the unlock to work right and that stuff was a wrong approach.

zone/CMakeLists.txt
ofnature: added the header into the header section

zone/dragonshoard.cpp
ofnature: the includes are full length now not relative

zone/dragonshoard.cpp
ofnature: removed this, and hopefully placed it in the right place.

zone/dragonshoard.cpp
ofnature: its removed now from the latest pr push

common/ruletypes.h
ofnature: alot of the comments came in from my LS work, I had no one else but my self working on it so told cursor(bad experince) to leave notes so if someone else took over the notes might guide them better.

zone/client_packet.cpp
ofnature: dropped the numbers shouldnt happen again

zone/client_packet.cpp
ofnature: im not good at naming its DragonHoard right now maybe I should have chose DragonHoardAction since it describes what it is and does at a glance

zone/client_packet.cpp
ofnature: its remnants from the LS work again, the ls conf had reference to 2 dh opcodes so i matched the old to new ones and found 2 DH opcodes. Irreguardless its removed.

zone/dragonshoard.cpp
ofnature: trimmed out most of the comment might still be some, I will try to not do this in the future.

zone/dragonshoard.cpp
ofnature: should now be constants for this and features to come.

zone/dragonshoard.cpp
ofnature: have removed the manual pointers and the reinterpret casting "should" be fixed

zone/dragonshoard.cpp
ofnature: scrubbed the pr for simuliar issues and fixed them

zone/dragonshoard.h
ofnature: it now lives in config, 200 was from the in game help window for the DH as max purchasable, client has no max defined within the client from what I saw.

utils/patches/patch_TOB.conf
ofnature: dh1 changed to DragonHoard and DH2 has been removed

zone/dragonshoard.cpp
ofnature: moved it to the playerprofile and made the claim struct and its (claim_id, slotcount), we then send a op_DragonHoard action 2 (setslotcount from the config) packet when we open the DH window which gives us the slots

common/database/database_update_manifest.h
```diff
@@ -7263,6 +7263,35 @@ ALTER TABLE `character_pet_buffs`
...
+		.check = "SHOW TABLES LIKE 'dragonhoard_items'",
+		.condition = "empty",
+		.match = "",
+		.sql = R"(
+CREATE TABLE `dragonhoard_items` (
+  `id`          int unsigned NOT NULL AUTO_INCREMENT,
+  `account_id`  int          NOT NULL,
+  `slot_id`     int          NOT NULL,
+  `item_id`     int          NOT NULL DEFAULT 0,
+  `item_name`   varchar(128) NOT NULL DEFAULT '',
```
dannuic, line 7280: These should probably not have defaults. It'll potentially let bad data slip in

common/database/database_update_manifest.h
```diff
@@ -7263,6 +7263,35 @@ ALTER TABLE `character_pet_buffs`
...
+  `stack_count` int unsigned NOT NULL DEFAULT 1,
+  `charges`     int          NOT NULL DEFAULT 0,
+  `serial`      bigint       NOT NULL DEFAULT 0,
```
dannuic, line 7283: This also, what is a serial, and why is it allowed to be default?

common/database/database_update_manifest.h
```diff
@@ -7263,6 +7263,35 @@ ALTER TABLE `character_pet_buffs`
...
+  `augslot1`    int          NOT NULL DEFAULT 0,
+  `augslot2`    int          NOT NULL DEFAULT 0,
+  `augslot3`    int          NOT NULL DEFAULT 0,
+  `augslot4`    int          NOT NULL DEFAULT 0,
+  `augslot5`    int          NOT NULL DEFAULT 0,
+  `augslot6`    int          NOT NULL DEFAULT 0,
+  `custom_data` text         DEFAULT NULL,
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `account_slot` (`account_id`, `slot_id`),
+  KEY `account_id` (`account_id`)
```
dannuic, line 7293: why is account_id a key? wouldn't this storage be per character, not account?

common/patches/tob.cpp
```diff
@@ -2098,9 +2101,18 @@ namespace TOB
...
 		out.WriteUInt32(emu->platinum_shared);
 
-		//u32 claim_count;
-		out.WriteUInt32(0);
-		//Claim claims[claim_count];
+		// Claims — the features the account has unlocked, sent with the player profile at zone-in.
+		// Each is { claim_id (= feature id), value (= current unlocked amount, e.g. slot count) }.
+		// This is how the client is told which store features it has; the standalone OP_FeatureUnlock
+		// message is only for live marketplace-purchase updates. (In MQ these are GameFeatures.)
+		constexpr uint32 kClaimDragonHoard = 2016763; // 0x1EC5FB — Dragon's Hoard
```
dannuic, line 2108: This should be in an enum in a shared file (not specific to TOB, these values should not change, or if they do, they will be overridden in a patch, but use these values as the basis since it is the first occurence) because you're going to have a collection of consts here that would have to be maintained in multiple places and otherwise are mutually distinct in a Claim struct. Also, the last two comments above are unnecessary (and confusing, this is OP_PlayerProfile).

common/patches/tob.cpp
```diff
@@ -2098,9 +2101,18 @@ namespace TOB
...
+		const bool dragonhoard_enabled = RuleB(Features, DragonHoardEnabled);
```
dannuic, line 2109: There are too many rules here. Enabling dragon hoard is the same as just setting this to a non-zero value, there's already a lot of rule bloat, so I'd think this should be combined into a single rule.

common/patches/tob.cpp
```diff
@@ -2098,9 +2101,18 @@ namespace TOB
...
+
+		out.WriteUInt32(dragonhoard_enabled ? 1 : 0); // claim_count
+		if (dragonhoard_enabled) {
+			out.WriteUInt32(kClaimDragonHoard);
+			out.WriteUInt32(static_cast<uint32>(RuleI(Features, DragonHoardSlots))); // current unlocked amount
```
dannuic, line 2114: Did you test if this is the number of slots? I think it might just be 0 or 1 and the number of entitled slots actually gets sent in the other place (OP_SendMembership).

zone/client.cpp
```diff
@@ -9150,44 +9150,6 @@ void Client::SendHPUpdateMarquee(){
...
-		mc->entries[19] = 0xffffffff;	// 0 for Silver
-		mc->entries[20] = 0xffffffff;	// 0 for Silver
-		mc->exit_url_length = 0;
-		//mc->exit_url = 0; // Used on Live: "http://www.everquest.com/free-to-play/exit-silver"
-
-		QueuePacket(outapp);
-		safe_delete(outapp);
-	}
-}
-
```
dannuic, line 9190: Why is this removed? This breaks the membership sending that happens at zone entry

zone/client.h
```diff
@@ -1852,7 +1852,6 @@ class Client : public Mob
 	void ResetHPUpdateTimer() { hpupdate_timer.Start(); }
 
 	void SendHPUpdateMarquee();
-	void SendMembership();
```
dannuic, line 1855: again, this shouldn't be removed

zone/client_packet.cpp
```diff
@@ -770,7 +772,17 @@ void Client::CompleteConnect()
...
 	conn_state = ClientConnectFinished;
+	if (first_connect) {
+		// Dragon's Hoard: the feature GRANT now rides the player-profile claims array (see the TOB
+		// OP_PlayerProfile encode) — that replaces the standalone OP_FeatureUnlock. SendUnlock still
+		// sets the DH window's own enable/slot fields (OP_DragonHoard action 8/2 -> player+0x2468/
+		// 0x246C), which the claim doesn't populate. SendItemList pushes the stored items. Fires on
+		// OP_ClientReady -> CompleteConnect, after OP_EnterWorld (DH window global is valid).
+		DragonHoard::SendUnlock(this);
+		DragonHoard::SendItemList(this);
+	}
```
dannuic, line 785: you shouldn't need any of this, this is a call and response type deal. You don't need to send the unlock and you don't send the list until you receive the OP_DragonHoard packet

zone/client_packet.cpp
```diff
@@ -1699,8 +1711,6 @@ void Client::Handle_Connect_OP_ZoneEntry(const EQApplicationPacket *app)
 	if ((m_pp.RestTimer > RuleI(Character, RestRegenTimeToActivate)) && (m_pp.RestTimer > RuleI(Character, RestRegenRaidTimeToActivate)))
 		m_pp.RestTimer = 0;
 
-	SendMembership();
-
```
dannuic, line 1703: shouldn't be removed

zone/client_packet.cpp
```diff
@@ -17440,3 +17450,19 @@ void Client::SyncWorldPositionsToClient(bool ignore_idle)
 		m_is_idle = false;
 	}
 }
+
+// Dragon's Hoard handlers
+void Client::Handle_OP_DragonHoard(const EQApplicationPacket *app)
+{
+	// Carries several actions; route by the leading action field.
+	if (!app || app->size < 4) {
```
dannuic, line 17458: 4 is a magic number here, check the other handlers that check size (because they are static). You want this to use sizeof on the type that's expected (which could be a struct, or it could simply be an integer like an item ID or slot)

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,246 @@
...
+		return;
+	}
+
+	const uint32 account_id = client->AccountID();
+	if (!account_id) {
+		return;
+	}
+
+	const auto rows = DragonhoardItemsRepository::GetWhere(
+		database, fmt::format("account_id = {} ORDER BY slot_id", account_id));
```
dannuic, line 34: I'm confused why this is by account ID and not a character ID, shouldn't there be a separate list per character?

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,246 @@
...
+		// it back on retrieve, letting us identify the exact item by serial instead of a
+		// fragile positional index.
+		inst->SetSerialNumber(static_cast<int32>(slot_id + 1));
+
+		client->SendItemPacket(slot_id, inst, ItemPacketType::ItemPacketDragonHoard);
+		safe_delete(inst);
+	}
+
+	LogDebug("DragonHoard::SendItemList sent items to account_id {}", account_id);
+}
```
dannuic, line 65: Thinking about this now, shouldn't this live in memory in inventory_profile instead of hitting the database every time? This is exploitable, and all you'd need to do is add a new type of player inventory to InventoryProfile, and only save on character save/load on character load (from the db. someone else might have other thoughts about this

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,246 @@
...
+	slots.WriteUInt32(SetSlotCount);
+	slots.WriteUInt32(static_cast<uint32>(RuleI(Features, DragonHoardSlots)));
+	auto* outapp2 = new EQApplicationPacket(OP_DragonHoard, slots.size());
+	outapp2->WriteData(slots.buffer(), slots.size());
+
+	client->FastQueuePacket(&outapp);
+	client->FastQueuePacket(&outapp2);
+
+	LogDebug("DragonHoard::SendUnlock sent to {}", client->GetName());
+}
```
dannuic, line 94: should not be in, there is no dynamic unlock capability without adding that as a whole new feature, it shouldn't be in this PR

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+void DragonHoard::HandleDeposit(Client* client, const EQApplicationPacket* app)
+{
+	if (!app || !IsEnabled(client)) {
+		return;
+	}
+
+	// deposit packet layout: [action=4 u32][item_id u64] — minimum 12 bytes
+	if (app->size < 12) {
```
dannuic, line 103: again, magic numbers -- use sizeof like your comment suggests: `siezof(uint32) + sizeof(uint64)`

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,246 @@
...
+		return;
+	}
+
+	const EQ::ItemInstance* cursor_item = client->GetInv().GetItem(EQ::invslot::slotCursor);
+	if (!cursor_item) {
+		LogDebug("DragonHoard::HandleDeposit no cursor item for {}", client->GetName());
+		return;
+	}
+
+	const uint32 account_id = client->AccountID();
```
dannuic, line 114: account id, this should be per character

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,246 @@
...
+	e.stack_count = stack;
+
+	e = DragonhoardItemsRepository::InsertOne(database, e);
+	if (!e.id) {
+		LogError("DragonHoard::HandleDeposit DB insert failed for account_id {}", account_id);
+		return;
+	}
+
+	// Remove item from cursor
+	client->DeleteItemInInventory(EQ::invslot::slotCursor, 0, true);
```
dannuic, line 147: this goes back to my previous comment, hitting the database seems the incorrect implementation here for every deposit. It should live in InventoryProfile, I think. (again, thoughts?)

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+	// Send item to DH window. Stamp the same stable per-slot serial (slot_id + 1) that
+	// SendItemList uses, so the item can be retrieved immediately without re-zoning —
+	// otherwise a freshly deposited item carries an ephemeral serial that maps to no slot.
+	EQ::ItemInstance* inst = database.CreateItem(item_data, static_cast<int16>(stack));
+	if (inst) {
+		inst->SetSerialNumber(static_cast<int32>(slot_id + 1));
+		client->SendItemPacket(slot_id, inst, ItemPacketType::ItemPacketDragonHoard);
+		safe_delete(inst);
+	}
```
dannuic, line 157: What happened to cursor_item at this point? Why are we creating another instance?

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+	// Send the action=10 (0xA) deposit acknowledgement. The client increments an internal
+	// pending-operation counter when it sends a deposit and freezes inventory/mouse input
+	// until the server acks it (client sub_14020B560 case 0xA -> decrement dword_140EE2B2C).
+	// Unlike action=3 this only clears the pending op; it does not remove a window row.
+	SerializeBuffer ack_buf;
+	ack_buf.WriteUInt32(DepositAck);
+	auto* ack = new EQApplicationPacket(OP_DragonHoard, ack_buf.size());
+	ack->WriteData(ack_buf.buffer(), ack_buf.size());
+	client->FastQueuePacket(&ack);
```
dannuic, line 167: seems a bit weird to use the whole serialize buffer infrastructure when it's just sending an int32

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,246 @@
...
+		return;
+	}
+
+	// retrieve packet layout (client sub_14010DAE0):
+	//   +0x00 u32 action=3
+	//   +0x04 u64 serial  <- the selected row's item serial (we stamped it = slot_id+1)
+	//   +0x0C u32 index   <- positional index; unreliable, ignored
+	// The old code used the +0x0C index with ORDER BY item_name, but SendItemList sends
+	// ORDER BY slot_id, so the index never matched the clicked row (always the wrong item).
+	if (app->size < 16) {
```
dannuic, line 184: magic numbers

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,246 @@
...
+		return;
+	}
+	uint64 client_serial = 0;
+	memcpy(&client_serial, app->pBuffer + 4, sizeof(uint64));
+	if (client_serial == 0) {
+		return;
+	}
+	const uint32 target_slot = static_cast<uint32>(client_serial) - 1; // undo the +1 from SendItemList
```
dannuic, line 192: use a struct or something like that here, manually reading from the buffer like this is hard to follow

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+	const uint32 account_id = client->AccountID();
```
dannuic, line 194: should be by character

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,246 @@
...
+	// Remove from DB
+	DragonhoardItemsRepository::DeleteWhere(
+		database, fmt::format("account_id = {} AND slot_id = {}", account_id, slot_id));
+
+	// Give item to cursor
+	EQ::ItemInstance* inst = database.CreateItem(item_data, static_cast<int16>(stack));
+	if (inst) {
+		client->PushItemOnCursor(*inst, true);
+		safe_delete(inst);
+	}
```
dannuic, line 230: previous comments about database vs InventoryProfile are relevant here as well

zone/dragonshoard.cpp
```diff
@@ -0,0 +1,246 @@
...
+	// client handler (sub_14020B560 case 3 -> sub_14010D8C0) matches the item by serial,
+	// removes the row (when qty >= its stack), and decrements the pending-op counter, which
+	// releases the cursor. Layout mirrors the request: [action=3][serial u64][qty u32].
+	SerializeBuffer confirm_buf;
+	confirm_buf.WriteUInt32(Retrieve);
+	confirm_buf.WriteUInt64(client_serial); // == slot_id+1 we stamped
+	confirm_buf.WriteUInt32(stack);         // >= stack removes the row
+	auto* confirm = new EQApplicationPacket(OP_DragonHoard, confirm_buf.size());
+	confirm->WriteData(confirm_buf.buffer(), confirm_buf.size());
+	client->FastQueuePacket(&confirm);
```
dannuic, line 243: This is probably cleaner as a struct since it's so simple, but this is  better than manually writing the buffer

zone/dragonshoard.cpp
ofnature: i never messed with it on live i misunderstood its usage, it is per character I will go back and look through it more carefully.

common/patches/tob.cpp
brainiac: the code was removed but the comments are still here.

common/ruletypes.h
```diff
@@ -1179,6 +1179,11 @@ RULE_INT(EvolvingItems, DelayUponEquipping, 30000, "Delay in ms before an evolvi
 RULE_BOOL(EvolvingItems, DestroyAugmentsOnEvolve, false, "If this is enabled, any augments in an item will be destroyed when the item evolves. Otherwise, send augments to the player via the parcel system (requires that the Parcel System be enabled).")
 RULE_CATEGORY_END()
 
+RULE_CATEGORY(Features)
+RULE_BOOL(Features, DragonHoardEnabled, false, "Enables the Dragon's Hoard storage feature. Requires The Outer Brood (TOB) client. Default off.")
+RULE_INT(Features, DragonHoardSlots, 200, "Number of Dragon's Hoard storage slots granted. Server-driven (the TOB client applies no hard cap); retail's purchasable max is 200.")
```
brainiac, line 1184: This coomment says a lot more than it needs to say. Server rules are generally assumed to be server driven.

zone/client_packet.cpp
```diff
@@ -17440,3 +17450,19 @@ void Client::SyncWorldPositionsToClient(bool ignore_idle)
...
+		return;
+	}
+	const uint32 action = *reinterpret_cast<const uint32*>(app->pBuffer);
+	if (action == DragonHoard::Deposit) {
+		DragonHoard::HandleDeposit(this, app);
+	} else if (action == DragonHoard::Retrieve) {
+		DragonHoard::HandleRetrieve(this, app);
+	}
+	// DragonHoard::WindowOpen needs no server response.
```
brainiac, line 17467: i think this is where we're supposed to send the list of items..?

zone/client_process.cpp
```diff
@@ -30,6 +30,7 @@
 #include "common/skills.h"
 #include "common/spdat.h"
 #include "common/strings.h"
+#include "zone/dragonshoard.h"
```
brainiac, line 33: this include was added but there are no other changes.

zone/dragonshoard.cpp
brainiac: well when we change inventory we need to write to database to persist the change, but this needs to have happened in a transaction to be safe since we've moving the item from one storage to another

zone/dragonshoard.cpp
brainiac: these should be structs. use SerializeBuffer for things that aren't known at compile time.

common/database/database_update_manifest.h
```diff
@@ -7263,6 +7263,35 @@ ALTER TABLE `character_pet_buffs`
...
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```
dannuic, line 7294: Looking at inventory this morning because I was thinking about it -- this should be essentially exactly inventory, but with a special set of slots. I think reserving slot_ids for dragon hoard there is a bad move, so having this extra table is probably good, but the item table has some different members, do these things needs ornamentation and color information? Also, I'm pretty sure emu (and EQ) uses charges and stack_count interchangeably. If you look at the inventory table, you'll see that you don't have an `id` column (it's uniquely identified by `character_id` and `slot_id`, which is the same here), and instead of serial, it uses a guid, which is a much easier and less convoluted way to set a unique identifier. I'm still unsure how serial is different than slot in this code.

zone/dragonshoard.cpp
dannuic: So this sent me down a rabbit hole a bit, because there has to already be database actions for the current inventory implementation. It's super spaghetti, the database is hit every time something changes in the inventory, manually. There's no automatic mechanism for it at all, so I think you're right about this. The current inventory code doesn't use transactions to move items because it's all a simple update in a single statement, and it currently does everything in a single SaveInventory function, which is probably what you want to do as well (in shareddb).

Ideally, you'd call SaveInventory, it would collect all the differences between memory state and your inventory, and then commit all the changes as a single transaction. The difficulty here is twofold: you are creating a separate inventory space whereas all previous emu inventory was unified, and these transfers are _between_ the inventory spaces. I think there are two options to move past this:
- roll dragon hoard into the shared inventory space and put all of the database code in the existing inventory database functions and use the existing inventory table. This would require reserving slot id's for dragon hoard (there is plenty of space, you can check the current constants). This is simpler to implement, and leans into current emu paradigms, but does mean you have to updated existing functions and might require a bit more testing to make sure regular inventory functions don't break.
- write a completely new dragon hoard database interface in shareddb that exchanges items between databases in a database transaction. This isolates the new functionality into its own domain at the cost of  increased complexity, both of the new logic and of the emu codebase in general because now there are two places where inventory is stored.

If I were to implement this, I would choose the first option, but either option should be implemented in shareddb to keep database access centralized.

brainiac: EQ uses container ids. Dragon hoard has its own container id ([42](https://github.com/macroquest/eqlib/blob/live/include/eqlib/game/Items.h#L83)). Does the db schema have this included? (how does it decide if an item is in the bank?)

The ideal scenario would be just throwing the items into the table with a different container id, but somehow i don't think its that simple (I haven't looked at how we store items in the db)

Another point about the dragon hoard is that its also an expandable storage, so i could see people expanding it to like 1000s of items if they really wanted to do so.

dannuic: the emu just uses slot ranges for the different containers. So anything that has a slot id from 2000 to 2000 + bank size is in the bank, for instance. This is clearly not the greatest, but that is how emu encodes container ID's. Obviously, the entire inventory system should probably be refactored to include container_id + slot_id (+ character_id) as the unique identifiers across inventories (instead of just the current slot_id + character_id). It would greatly simplify this process, but I think for this PR, it'd be best to just include hoard in the current inventory system and have it get rolled up into that future refactor

ofnature: i dont have the container id set, i was just using the cursor logic the move item packet with 79(i think) to send it through the DH path into its current wrong db when an it is clicked into the dh window similiar to move item into bank, not the cleanest but was working for my testing (sadly) is more responsive then live is with DH. I was thinking about it all night about how I was gonna do the per character id storage of the DH items, was thinking maybe similar to the bank(since its just an extended virtual bank)

common/patches/tob.cpp
ofnature: i did test and max slots stayed 0, I then tried via OP_SendMembership and still was 0.  I only seem to get the max slots when I send the action 2 with the dh opcode.  I know there should be somewhere else that hits that without sending that action 2, which i bet is for when you claim a dh expansion(theory) I spent about 40 minutes down that rabbit hole but didnt really see anything aside from a large vtable that links in, but I suck at tracing the calls.

common/patches/tob.cpp
ofnature: I see your logic in that 0 would just disable the feature and any number > then 0 is max slot count

common/database/database_update_manifest.h
ofnature: claude reversed the sql import script that built my original dh db so it just used default even though it had access the the original import sql.

PR #5113, opened 2026-07-25
Added groups, raids, and xtarget packets

dannuic:


dannuic:
Moving this to draft because I've noticed some bad patterns here that were in trying to work with some emu patterns. I'm going to restructure this a bit before it's ready for review.

